PR #3, opened 2020-02-21
Switch workspaces in slack

hbk619:
We could make the modifier_number function more genericly named as it should work with strings

Followed https://slack.com/intl/en-gb/help/articles/201374536-Slack-keyboard-shortcuts#mac-3
for mac shortcuts (although the linux ones are wrong there so someone will need to test a mac)


knausj85:
you mind re-opening this once you've grabbed the latest? I royally borked the move, sorry, hehe.

PR #5, opened 2020-02-22
Firefox linux support

hbk619:
Updated a couple of shortcuts as per https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly I think they apply to Windows as well

### DIFF
--- a/code/win/apps/firefox.talon
+++ b/code/win/apps/firefox.talon
@@ -1,4 +1,5 @@
 os: windows
+os: linux
 app: Firefox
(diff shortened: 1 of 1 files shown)


knausj85:
These few shortcuts actually appear to be different. :'(

I think we'd probably be better offer creating a linux-specific version.

hbk619:
The one time I don't test it on Windows! No problem, I'll close this and make a new branch so the fit history looks nice

PR #11, opened 2020-02-28
Terminal navigation/control

hbk619:
Add page up/down to allow scrolling in editors
Windows doesn't have tabs so nothing for those
Followed https://support.apple.com/en-gb/guide/terminal/trmlshtcts/mac for shortcuts for mac

It'd be better if "go tab" was part of the generic_browser but I'm not sure how to pass along the number the user says so did it this way for now

### DIFF
--- a/apps/linux/terminal.talon
+++ b/apps/linux/terminal.talon
@@ -1,0 +1,26 @@
+os: linux
+app: /.*terminal/
+-
+action(app.tab_open):
+  key(ctrl-shift-t)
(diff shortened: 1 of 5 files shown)


knausj85:
go tab would be worth asking aegis to add to the app actions, if it’s not there.

Note that I relocated 
user.knausj_talon.code.formatters.modifier_number
to 
user.knausj.code.keys.modifier_key on the tip, so that’ll break when/if you update it.

PR #38, opened 2020-03-27
Add a few useful git commands

hunner:
I'm not sure about always having `key(enter)` since it means commands
can't be built up with arguments, but we can address that later. a `dash
dash <phrase>` command would be useful for adding ` --<phrase> ` for
example.


knausj85:
Sean beat you to the punch a little bit. Would you mind re-evaluating this now that

https://github.com/knausj85/knausj_talon/pull/35

has been merged? Sorry for the inconvenience.

hunner:
Looks like `word git` and `git fetch all` are the two commands that I have here extra. I'll fix that up sometime.

PR #44, opened 2020-03-30
List of running applications

Chreschtf:
Should the intuitive name of a program not open the window you want, simply list all the available names - makes it easier to select the one you want.
also a small addition to git.talon


code/switcher.py
```diff
@@ -55,6 +55,21 @@ def switcher_launch(path: str):
         """Launch a new application by path"""
         ui.launch(path=path)
 
+    def list_running():
```
knausj85, line 58: I'd like to try to maintain the convention of prefixing these with the file title.

e.g. switcher_show_running()

code/switcher.py
```diff
@@ -55,6 +55,21 @@ def switcher_launch(path: str):
...
+
+    def hide_list_running():
+        """Hides list of running applications"""
+        gui.hide()
+
+@imgui.open(y=0)
+def gui(gui: imgui.GUI):
+    gui.text("Names of running applications")
+    gui.line()
+    for line in  ctx.lists['self.running']:
```
knausj85, line 70: Should this account for/display any overrides?

knausj85:
I'm going to go ahead and take this as-is and do some minor cleanup. Thanks for the contribution!

PR #46, opened 2020-03-30
Assorted changes

mkandalf:
Still not using Talon all the time, but as I try to use it a bit I've made some changes to my scripts. Here are some potential changes for merging, let me know which if any you want to accept. My guess is the Vim and 1PasswordX bindings are the most controversial.

- [x] Allow `refresh it` as alternative for `reload it`
- [x] Add some DataGrip bindings
- [ ] Add some vim bindings for common text actions
- [ ] Add a few VSCode bindings
- [x] Update password manager bindings for 1PasswordX defaults
- [x] Add some common switcher overrides
- [x] Add SQL lang bindings
- [x] Add bindings for the 1 password actions
- [x] Add `git add -p` interactive support (char + enter)
- [x] Remove some perhaps less frequently used bindings in `standard.talon`
- [x] Allow `new tab` as alternative for `open tab`
- [x] Allow `spamma` as alternative for `comma and`
- [x] Add some more python keywords
- [x] Add `<word>` binding
- [x] Fix `slap` to match common usage (end of line and enter, not beginning of line)
- [x] Add suspend and resume (fg and ctrl-z) to terminal


code/mac/apps/vscode.talon
```diff
@@ -0,0 +1,19 @@
+app: Code
+os: mac
+
+-
+# Navigation and commands
+run command: key(cmd-shift-p)
+
+action(user.ide_find_file):
```
knausj85, line 8: For things like this, where the implementation of a particular action is OS-specific, i think i'd prefer to break those out into separate files from the voice commands.

Similar to like this
https://github.com/knausj85/knausj_talon/blob/master/code/win/apps/chrome.talon

This might be a good time to consider creating a generic_ide.talon (similar to generic_broswer), and consolidate the voice commands between this and

https://github.com/knausj85/knausj_talon/blob/master/apps/jetbrains.talon


Thoughts?

misc/git_add_patch.talon
```diff
@@ -0,0 +1,20 @@
+app: /.*terminal/
+app: /.*term/
+app: Cmd.exe
+title: /git add .*\-p/
+-
+yank:
```
knausj85, line 6: Will re-using the alphabet here cause any issues if people need to actually spell something? I don't really use the commandline git.

misc/standard.talon
```diff
@@ -1,6 +1,8 @@
 dragon words: "<dgnwords>"
 dragon dictation: "<phrase>"
-slap: key(home enter)
+slap:
+	key(cmd-right)
```
knausj85, line 4: Please use the
edit.line_end()
action since these are defined per-OS.

misc/git_add_patch.talon
mkandalf: The only options during a `git add -p` are single letters or question mark.

mkandalf:
There is a vim plugin for vscode, good point. So maybe a use-case for the tags system? Make a vim tag, and then apply it to ide, browser (vimium), vim, as needed?

knausj85:
I'm just going to go ahead and take this as-is and clean it up a bit here and there.

PR #51, opened 2020-04-01
add paste match

2shea:


misc/standard.talon
```diff
@@ -89,6 +89,7 @@ zoom out: edit.zoom_out()
 copy: edit.copy()
 cut: edit.cut()
 paste: edit.paste()
+paste match: key(cmd-alt-shift-v)
```
knausj85, line 92: There's a Talon-defined action in the edit thing for this edit.paste_match_style()

For cross-platform support, would you mind using that action here, and implementing the OS X action in 
https://github.com/knausj85/knausj_talon/blob/976990dfced6a4c030ab30cf3807b753f4cf3d5f/code/mac/edit.talon
?

PR #53, opened 2020-04-05
Updates to 1Password: make it more consistent with their published keyboard shortcuts

2shea:
A few changes to bring 1Password more consistent with their published keyboard shortcut documentation:
* 1Password has a keyboard shortcut to fill the password. We can use that rather than implementing ourselves.
* Use shortcut provided by the docs for "password show". I'm not sure where `cmd-shift-x` came from... it doesn't work for me, but maybe @mkandalf can comment on the intent if I'm missing something? IMO, seems best to use the shortcuts provided in the docs unless there's a good reason.
* Use app.bundle() for the context scope as 1Password7 didn't seem to work (I checked "help context" with 1Password in focus and the context wasn't active).
* Only global commands should be in global context and move others to app scoped context. These are grouped based on the 1Password keyboard shortcut documentation that lives here: https://support.1password.com/keyboard-shortcuts/

### DIFF
--- a/apps/mac/password_manager.talon
+++ b/apps/mac/password_manager.talon
@@ -20,5 +20,5 @@ action(user.password_edit):
 	key(cmd-e)
 	
 action(user.password_delete):
-	key(cmd-delete)
+	key(cmd-backspace)
 
--- a/apps/mac/password_manager.talon
+++ b/apps/mac/password_manager.talon
@@ -5,14 +5,10 @@ os: mac
 #the below are for 1password, redefine as needed
 -
 action(user.password_fill):
-	user.password_show()
-	sleep(300ms)
-	key(enter)
-	sleep(300ms)
-	key(enter)
+	key(cmd-\)
 
 action(user.password_show):
-	key(cmd-shift-x)
+	key(cmd-alt-\)
 	
 action(user.password_new):
 	key(cmd-i)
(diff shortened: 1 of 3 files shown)


mkandalf:
It appears there are two browser options for filling passwords, 1Password (mac w/ optional paired browser extension) or 1PasswordX (separate standalone extension). The shortcuts are [here](https://support.1password.com/getting-started-1password-x/#use-keyboard-shortcuts). The default shortcut for opening that extension might be `cmd-shift-x` (I don't recall changing it, and that is what it is set to in chrome).

That said, my guess is using the desktop app may end up being more talon friendly, so is reasonable to be the default approach included. We can split out a 1passwordX file if necessary.

knausj85:
Sounds good. Sorry for the inconvenience here, and thanks both of you for these contributions.

2shea:
> It appears there are two browser options for filling passwords, 1Password (mac w/ optional paired browser extension) or 1PasswordX (separate standalone extension). The shortcuts are [here](https://support.1password.com/getting-started-1password-x/#use-keyboard-shortcuts). The default shortcut for opening that extension might be `cmd-shift-x` (I don't recall changing it, and that is what it is set to in chrome).

I see! Thanks for explaining! I actually hadn't seen 1Password X.

> That said, my guess is using the desktop app may end up being more talon friendly, so makes sense to default to that. We can split out an optional 1PasswordX file if necessary.

Wonder if there's a way for Talon to know which one is being used?

2shea:
I made an issue for 1Password X support: https://github.com/knausj85/knausj_talon/issues/58

PR #54, opened 2020-04-06
Update README.md

ma-anwar:
Maybe adding a tip for how to easily access ~/.talon/user might be useful? Not sure if this is in-line with the style of the README.


knausj85:
Thanks for taking the time!

PR #69, opened 2020-04-14
change code.language matching so it also works on Linux terminals

fidgetingbits:
Test on Linux in termite running VIM where the titles are something like `code.py (~/.talon/user/fidget/code) - VIM` which causes the old code not to match correctly.

Confirmed that the `.talon` `code.language` context won't load when editing `code.py` despite `.talon` being in the path, but editing an actual `.talon` file will load it correctly.

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -12,27 +12,25 @@ extension_lang_map = {
 "talon": "talon",
 }
 
-regex_ext = re.compile("\.(\S*)\s*")
- 
+# The [^\\/] is specifically to avoid matching something like a .talon folder
+# that would otherwise cause the .talon file action to load
+regex_ext = re.compile("[^\\/]\.(\S*)\s*")
+
 @ctx.action_class('code')
 class CodeActions:
     def language(): 
         title = ui.active_window().title
         #print(str(ui.active_app()))
-        #workaround for VS Code on Mac. The title is "", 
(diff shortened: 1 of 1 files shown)


code/code.py
```diff
@@ -12,27 +12,25 @@
 "talon": "talon",
 }
 
-regex_ext = re.compile("\.(\S*)\s*")
- 
+# The [^\\/] is specifically to avoid matching something like a .talon folder
+# that would otherwise cause the .talon file action to load
+regex_ext = re.compile("[^\\/]\.(\S*)\s*")
```
knausj85, line 17: I think this needs a minor tweak to ensure the correct platform-specific path separator is used, particularly for Windows

`regex_ext = re.compile("[^\\{}]\.(\S*)\s*".format(os.path.sep))`

could you give that a whirl on Linux?

knausj85: I was able to poke at this a bit more today. A windows path such as

C:\...\.talontest\test.talon

will result in this match:

talontest\test.talon

when running talon. This is in spite of the proposed regex doing the correct thing on the online regex testers I typically use.

Escaping the '/' character looks like it fixes it, so i'll merge this in and make that minor tweak.

regex_ext = re.compile("[^\\\/]\.(\S*)\s*")

PR #73, opened 2020-04-28
switch windows by opening a UI with a letter combo for every window

timo:


knausj85:
This is neat. Let me think about this a bit, as I've been kinda standardizing on using numbers for this sort of selections out of the box (e.g., filemanager, homophones, etc).

knausj85:
This turns out to be a bit unusable on Windows as-is, and I'm not exactly sure how to fix it.

There's like a billion windows that are just "Windows Explorer" - from my experience, this is things like the taskbar/clock which have no title etc.

example

Ideas?

code/windowswitcher.py
```diff
@@ -0,0 +1,65 @@
...
+
+        shown_windows = ui.windows()
+
+        window_spelling = list(map(lambda kv: " ".join(map(lambda d: default_alphabet[int(d)], str(kv[0]))), enumerate(shown_windows)))
+        print(window_spelling)
+        ctx.lists['self.window_selection_words'] = window_spelling
+
+        window_codes = list(map(lambda kv: "".join(map(lambda d: letters_string[int(d)], str(kv[0]))), enumerate(shown_windows)))
+
+        gui.show()
```
knausj85, line 68: I'd suggest using gui.freeze() here for performance.

misc/windowswitcher.talon
```diff
@@ -0,0 +1,2 @@
+switch window:  user.show_window_switcher()
+switch <user.window_selection_word>: user.switch_to_window(window_selection_word)
```
knausj85: How would you feel about using <numbers> here instead?

With an upcoming PR, you'd be able to say either the whole number, or the digits, to switch windows. Similar to file manager.

pokey:
@timo think you'd be able to address some of the issues raised by @knausj85 here?  lmk if you wanna brainstorm / discuss

knausj85:
I think the time has come to close this one. :'(

We can revisit in the future & reopen.

PR #77, opened 2020-05-09
add an abbreviate command

fidgetingbits:
this is a command i used to use with aenea, and i find it useful if i am not thinking of the alphabet fast enough


knausj85:
This is very handy, thanks!

PR #79, opened 2020-05-09
support microsoft teams messenger

fidgetingbits:
only tested on linux but i suspect the windows shortcuts are the same

### DIFF
--- a/apps/linux/teams.talon
+++ b/apps/linux/teams.talon
@@ -17,9 +17,13 @@ open (apps|applications): key(ctrl-`)
 open settings: key(ctrl-,)
 open help: key(f1)
 close: key(escape)
-zoom in: key(ctrl-=)
-zoom out: key(ctrl--)
-reset zoom: key(ctrl-0)
+
+#zoom in: key(ctrl-=)
+#zoom out: key(ctrl--)
+#reset zoom: key(ctrl-0)
+action(edit.zoom_in): key(ctrl-=)
+action(edit.zoom_out): key(ctrl--)
+action(edit.zoom_reset): key(ctrl-0)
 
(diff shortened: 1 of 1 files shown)


apps/linux/teams.talon
```diff
@@ -0,0 +1,68 @@
...
+[go] [to] search: key(ctrl-e)
+show commands: key(ctrl-/)
+open filter: key(ctrl-shift-f)
+go to: key(ctrl-g)
+open (apps|applications): key(ctrl-`)
+[start] new chat: key(ctrl-n)
+open settings: key(ctrl-,)
+open help: key(f1)
+close: key(escape)
+zoom in: key(ctrl-=)
```
knausj85, line 20: This looks great - thanks!

Minor thing-
- Zoom in/out commands are currently part of standard.talon, so we might be better off overriding the edit.zoom_in(), edit.zoom_out() actions instead, and leaving the commands in standard.talon? 
https://github.com/knausj85/knausj_talon/blob/ab6462b0a7443a91d1dfb26386c39449841cad19/misc/standard.talon#L55

knausj85: I'm open to suggestions if there's a better approach than just making these global commands.

fidgetingbits: i don't have a better idea the moment. the only thing i don't like about them being globals, which is really unrelated to this problem, is that when i forget commands i often use the help active context command as a quick way to see currently enabled commands, but with this method i only see the action itself and not the corresponding voice command. but i feel like there's a different way to solve this with help menu eventually

knausj85:
Thanks for these contributions, and your patience!

PR #80, opened 2020-05-09
add support for protonmail web application

fidgetingbits:
i originally had this in its own web/ directory on my fork rather than apps since its technically a website and so the short cuts should roughly match for every system, but i guess the problem is that on mac the control keys will end up being command key. but i think for stuff like this it would be better to have some sort of setting where we say on mac use command key and on everything else use control key and then just have talon figure it out on the fly?

### DIFF
--- a/apps/linux/protonmail.talon
+++ b/apps/linux/protonmail.talon
@@ -1,0 +1,82 @@
+app: firefox
+app: Firefox
+win.title: /ProtonMail/
+-
+# General
+## Application
+open help: key(?)
+[focus] search: key(/)
+confirm active: key(enter)
+close active: key(escape)
+open command [palette]: key(shift-space)
+
+## Composer
+new message: key(c)
+send message: key(ctrl-enter)
+save message: key(ctrl-s)
+
+# Mail
+## Jumping
+(go|jump) [to] inbox:
+    key(g)
+    key(i)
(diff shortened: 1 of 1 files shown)


knausj85:
I think a /web directory makes sense, i can see a need for both e.g. /linux/web and just /web/, depending on how things are implemented for a given app. 

Just for reference, there a few ways to handle cmd vs ctrl
- defining actions in python. the platform can be detected, and the appropriate shortcut used. 
- just declaring actions in python, and defining per-OS web app actions in a .talon file. the command file can be defined in a separate .talon file. This is what i've done for e.g. the edit commands
- just separate .talon files for each OS

Few minor notes:
- Until someone vets these on Windows or Mac, I'd vote to add 
  os: linux 
- i do wonder if the win.title is specific enough? Should it be some regex of mail.protonmail.com ?

fidgetingbits:
unfortunately win.title doesn't include the url (nor do any other window attributes on linux afaict). Instead you get something like this which comes from the tab name (this output is from running `xprop`):

```
WM_ICON_NAME(STRING) = "Inbox | blah@blah.com | ProtonMail - Mozilla Firefox"
```

fidgetingbits:
I tested it with tags and it works fine, and I also added the linux os bit.

knausj85:
Thanks!

PR #81, opened 2020-05-09
add support for outlook web application

fidgetingbits:
this should also be the same on windows but again have not tested it. this is also another web application that could eventually go into a web folder.


knausj85:
Few minor notes, similar to proton:

- Until someone vets these on Windows or Mac, I'd vote to add
os: linux
- i do wonder if the win.title is specific enough? Should it be some regex of mail.outlook.com or whatever?
- if you've time to try tags on v1276, i'd say use tag: browser instead of listing the apps.

(This is beyond the scope of this, but i think we may be able to usefully consolidate the grammars between messengers/rich text/email clients some down the road. It's been on my list for a bit, but I don't have access to my work stuff currently)

fidgetingbits:
i've confirmed tags work well, i also added the linux filter. similar to the protonmail problem the window title doesn't come from the actual url, so what i used is about as granular we can get at the moment

knausj85:
Sounds good, thanks for this!

PR #82, opened 2020-05-09
add some useful symbols for when programming

fidgetingbits:
also switch ellipses to actually be three dots, i left the old ellipses as a commented out version because i assume whoever originally added it uses the real ellipses, but i suspect the majority of people will use just three dots.


knausj85:
Thanks!

PR #83, opened 2020-05-09
mark 1password_global.talon with an app tag that matches the other 1password.talon file

fidgetingbits:
prevent it from globally triggering

### DIFF
--- a/misc/1password_global.talon
+++ b/misc/1password_global.talon
@@ -1,4 +1,5 @@
 #todo: tags
+app: 1Password.exe
 -
 password fill: user.password_fill()
 password show: user.password_show()


knausj85:
I think this breaks the desired functionality: these commands should be available in e.g. safari, Firefox, etc. I might be missing something though?

This might be a better use case for tags, but not 100% sure.

knausj85:
Ah, I see. Wondering if there’s room here to potentially have e.g. generic password manager global or something instead.

fidgetingbits:
i think a generic password manager probably makes more sense ya. if this is actually meant for use in the browser, i guess we can just add the browser tag now instead of tying to an app.

knausj85:
I'll think about this one a bit, and leave this open for now. Does your password manager have plugins for the browser(s) like 1Password and such?

fidgetingbits:
> I'll think about this one a bit, and leave this open for now. Does your password manager have plugins for the browser(s) like 1Password and such?

No keepassx (a fork of keepass) doesn't support plugins. Though https://keepass.info/plugins.html implies there are some browser plugins for the original version of keepass.

knausj85:
Gotcha, sounds like this one is probably best left as a branch for those using then since keepassx requries focus?

I can add that detail to the readme.

fidgetingbits:
sounds good

PR #84, opened 2020-05-09
add support for the keepassx program on linux

fidgetingbits:


apps/linux/keepassx.talon
```diff
@@ -0,0 +1,27 @@
+os: linux
+#app: /KeePassX/
+#app: /eepassx/
+# XXX for some reason the app: field doesn't match atm
```
knausj85, line 4: Just curious. Is the issue with app a bug with talon, or is keepassx one of those apps with a bad name (many things on Windows are 'ApplicationHost.dll' or something).

fidgetingbits: I believe it's a bug in talon, as xprop shows the correct name, but I need to retest on the latest version

fidgetingbits: fWIW it still doesn't match on the latest version talon for some reason, but i'm not sure what/where to print from a ui.register() callback to see what talon is actually seeing the app string as.

knausj85: The easiest way is to have a command that dumps the focused app info to the console, or perhaps the clipboard. (clip.set_text())

debug app:
 print(app.name())
 print(app.bundle())
 print(app.executable())

fidgetingbits: thanks

```
2020-05-14 09:23:13    IO <pid 2181004>
2020-05-14 09:23:13    IO 
2020-05-14 09:23:13    IO 
```

Definitely a talon bug I think. So in the meantime we can just use the `win.title` if you were fine with that

fidgetingbits: Filed as https://github.com/talonvoice/beta/issues/62

fidgetingbits: bug is fixed in 1280, so will update the rules soon

PR #85, opened 2020-05-09
add support for some common gdb commands

fidgetingbits:
currently only matches on terminal windows running gdb. this causes a problem if you're running gdb through a script, but right now i solve this by having the script set the terminal window name so that it matches with gdb. i don't particularly like that, but don't have a better solution at the moment


knausj85:
I'm wondering if this might be a good use case for a custom mode that we could explicitly enable via a command. Hmm.

fidgetingbits:
yeah that would be a decent solution, in that case I think I would still have it automatically trigger if it knows for sure that the program is running, but also having the option to just turn it on if I know I need it without having to hacking some other thing like changing the terminal name is probably ideal

apps/linux/gdb.talon
```diff
@@ -0,0 +1,108 @@
+os: linux
+app: /term/
```
knausj85, line 2: maybe use tag: terminal instead of app?

fidgetingbits: will do. also been playing around with a 'debug mode' that triggers `mode: gdb` to get around the script problem and some new problems from me switching to using vim as my terminal. so will add that too.

knausj85: Oki doki, I'll leave this open and wait for these tweaks.

fidgetingbits:
added the mode changes and switched to terminal tag. in the mode changes include a couple other tweaks, like saying 'go lang' (which is what most of my friends call it), forcing markdown, etc.

apps/linux/gdb.talon
```diff
@@ -1,7 +1,7 @@
 os: linux
-app: /term/
 # XXX - this matches .gdb files atm
 #win.title: /gdb/
+tag: terminal
 mode: user.gdb
 -
 tag(): gdb
```
knausj85, line 7: Doesn't this need to be `user.gdb`?

knausj85: Is there a need for both a tag and a mode? Just curious.

knausj85, line 7: Doesn't this need to be user.gdb?

PR #87, opened 2020-05-11
added vocabulary

jonaskindler93:
Hey I added the functionality to define your own vocabulary, based on Artemis Everfrees version.
I thought this might be something of general use? If so, but somethings wrong with the code I'd love to get input.I'm still new to talon.


knausj85:
Would you be willing to try this version from aegis out on the latest beta release? It’s a bit simpler, and it sounds like the @ bug has been fixed,

https://talonvoice.slack.com/archives/G9YTMSZ2T/p1589477423054800

code/formatters.py
```diff
@@ -141,16 +141,16 @@ def format_text(m) -> str:
...
         
 @ctx.capture(rule='{self.formatters}+')
 def formatters(m):
     return ','.join(m.formatters_list)
  
-@ctx.capture(rule='<self.formatters> <phrase>')
+@ctx.capture(rule='<self.formatters> <user.textObject>')
 def format_text(m):
-    return FormatText(m.phrase, m.formatters)
+    return FormatText(m.textObject, m.formatters)
```
knausj85, line 154: With aegis version, we could modify this to perhaps take either a string or a phrase via a Union. It would potentially be a bit cleaner (gets rid of textobject etc)? This has the added benefit we could use this to reformat selected text down the line.

Thoughts?

Thoughts?

jonaskindler93: I'm not sure what you mean by union? 
If I remove textObject that would mean we have to check in FormatText wether its a str or a phrase and we cant use the actions.dictate stuff anymore. I found this a less intrusive change, but I get that its a bit intransparent. I'm happy to change it, but I'm not sure yet, if I understand how (at least such that we don't have to completely rewrite FormatText)

knausj85: I'll think about this one a bit, not sure either tbh.

knausj85: I had some to take a peek/tinker, and took this in a slightly different direction after some conversations with aegis. It sounds like the intention is for actions.dictate.* stuff to support stuff beyond just phrase down the line.

In the interim, 
- the formatters have been updated to take a string, or a phrase. This will also allow easy reformatting of selected text in the future too, which is nice.
- Some dragon-specific parsing nonsense was added to the user.text capture

https://github.com/knausj85/knausj_talon/commit/275496bbd309485a955c80784231508bcbed63b6
https://github.com/knausj85/knausj_talon/commit/16022342883966122d76b2df6288c358808c2bc3

If I overlooked anything, please let me know. Leaving this open for now, just in case(tm)

jonaskindler93: I have one question: why do we lowercase all of the words? I have a few words in my vocabulary list, where I included custom capitalization rules. shouldn't this be an option for the user?

jonaskindler93: also: I kind of dislike the fact that the actual vocabulary has to be defined inside of the knausj repository. I get that there are some standard words that we maybe want to include here, but the point of this is that there might be custom words that will cause merge conflicts and don't make sense to put in a pull request. 
if you agree with this I would adapt your version such that it defines a custom  user list and merges it with the global one in vocabulary.py. 
I guess this I feeds into a bigger discussion about how monolithic this whole thing should be.

knausj85: I'll test around the lower case stuff, might be for dragon since w2l doesn't do any capitalization. 

Eventually, Talon will support lists in a .talon file, so the vocab list could be redefined there. I might wait for that, and maybe branch for now? What do you think about that?

jonaskindler93: do you mean,we branch out at the completed to vocabulary functionality? I say we just merge it. it's cool functionality and I'm just talking about details.

knausj85: I tested around the lower casing stuff, and removed it from formatters.talon. Good catch. https://github.com/knausj85/knausj_talon/commit/606421ea22a321a22c44065de6133f396b81c1c4

I'd say go ahead and make your proposed change and open a new PR. It sounds reasonable to me.

PR #91, opened 2020-05-14
Add kubenetes support

khornlund:


apps/kubectl.talon
```diff
@@ -0,0 +1,52 @@
+tag: terminal
+-
+cube: insert(user.kubectl())
```
trapier, line 3: :joy_cat: someone is else done with fighting `cube`.

Thanks for the PR!

knausj85: Main thing I'm wondering: should we throw all the 'generic' terminal command stuff into a single folder so people can remove things easily (git, kubere) etc? May not be within scope of this PR, but figure it's worth discussing now.

Other than that, what was the advantage of implementing the actions in .py vs just e.g.
`cube: "kubectl "` in .talon? This is just for my own education.

fidgetingbits: i agree we should start putting them into their own directory. apps folders will get chaotic fast otherwise (part of why i already added `vim_plugins`, etc on my branch)

khornlund: I think it's tempting to nest things early on when there are clear separations, but later on you end up with these ambiguities and regret it. In general I think flat is better. 

There isn't much of an advantage, other than it being more flexible going forward. I could imagine people adding more complex functionality, in which case you need a .py

knausj85: Thanks for this contribution, going to merge as-is. I tend to agree that flatter's probably a bit better, but I probably need to rethink the folder structure in general anyway. A problem for another day, I think.

PR #93, opened 2020-05-15
don't use -1 for ordinals by default

fidgetingbits:
this changes the ordinal value returned so it's not -1 by default. instead I change the use in the repeater itself do do the -1. this will allow us to use ordinals another places where it doesn't need to be -1 by default, which means with them we don't have to hack in +1s in those places. an example would be in an editor like vim you can do 'delete 5th word' and in this case you don't have the explicit -1 in the count because the action hasn't actually happened yet at the time of parsing.

I also included two convenience repeaters i use

i left in the comment and previous command for why i made the change in case anyone is using repeaters/ordinals elsewhere in a fork i don't know about. but also happy to remove it

Finally, I am also of the opinion that what is currently repeaters.py should be either integrated into numbers.py or changed into ordinals.py because eventually it won't be explicitly tied to repeaters.talon as it gets integrated elsewhere. i didn't want to change that quite yet, as i figured we can discuss first


misc/repeater.talon
```diff
@@ -1,5 +1,5 @@
-<user.ordinals>: core.repeat_command(ordinals)
-
-
-
-
+# -1 because we are repeating, so the initial command counts as one
+<user.ordinals>: core.repeat_command(ordinals-1)
+repeat that: core.repeat_command(1)
+twice: core.repeat_command(1)
+# add N times
```
knausj85, line 5: maybe add this for n-times?
repeat that <number>: core.repeat_command(number)

also fyi you can do (in case you didn't know)
(repeat that | twice): core.repeat_command(1) 
to alias commands, though i *might* remove twice for my repro to avoid too many aliases for newbies

knausj85:
Yeah, I think ordinals.py is a better name. I’d kinda prefer to keep this separate from numbers since that one implements a bunch of talon-declared stuff.

fidgetingbits:
should be done now but github says it there's a conflicting file but i'm not sure how to fix

PR #97, opened 2020-05-17
vim support

fidgetingbits:
this is my first big pass at adding vim support. right now it doesn't support automatic detection and/or switching of modes for the most part, but it's still very much usable and I've been heavily using this for the last almost month, so I think it's good for other people to start testing it. I plan to keep adding to it and adding in the mode switching hopefully in the next week or two, but I don't think it needs to wait to get merged into upstream until that functionality is supported or anything

I do include a tutorial file because there is a few quirks related to getting it working. its still a work in progress but it should be enough to get people up and running.

there's a few things I'm not sure how to do deal with in the best way. for instance a lot of people won't have all of the plugins that are supported, so in that case you might not want them to actually load when you the vim tag, but I'm not sure if there's like a sane way to configure this at the moment. I would guess that one possibilities is to have a settings section inside of the vim.talon file and then you can manually enable specific plugins and only then are they loaded? but I'm not sure exactly how we would cleanly force load them only when the setting is enabled.


fidgetingbits:
i'm slowly switching all this over to the slightly smarter mode-aware version already, so can likely just wait until i update the commits now

knausj85:
This is a big one! I'll dedicate some time for this next week. Sorry for the delay.

fidgetingbits:
I have way more changes for this coming down the pipe btw. Preliminary mode switching, special terminal support, is all supported. I've just been testing and tweaking the last couple weeks, but will update the PR with the latest in the next couple days

fidgetingbits:
this is updated now with the latest changes. I think it's pretty stable, as I've been using it daily I guess almost a month now, so I've fixed a lot of the little bugs that I run into and most of the mode switching stuff works at least how I want it to.

I do support a lot of settings for disabling certain features, and I have an extensively tested then being disabled because most of them his match the functionality I want by default. So we definitely will run into people having problems I suspect if they start tweaking things, but I'm happy to try my best supporting people using it for now

doc/vim.md
```diff
@@ -0,0 +1,341 @@
...
+## Talon Change - The key `word`
+
+By default talon will use the command `word` as a command for saying a single
+word. See `misc/formatters.talon`. In vim "word" is a natural movement motion
+who it is included by default. If you decide to use this you might want to
+change the talon command to be a separate word.
+
+## The `generic_editor.talon` commands
+
+The default actions defined in `generic_editor.talon` are too simple for more
```
knausj85, line 58: Would it make sense to disable generic_editor.talon instead?

`not tag: vim`

apps/linux/vim_plugins/snippets/markdown_snippets.talon
```diff
@@ -0,0 +1,7 @@
+# these commands are based off the official them markdown vim-snippets
+code.language: markdown
```
knausj85, line 2: since this is vim-specific, i would say include `tag: vim`

apps/linux/vim.talon
```diff
@@ -0,0 +1,799 @@
...
+# XXX - needs to be adjusted to handle the actions associated with key presses
+<user.vim_normal_counted_action>:
+    insert("{vim_normal_counted_action}")
+
+###
+# File editing and management
+###
+# NOTE: using `save` alone conflicts too much with the `say`
+save file:
+    user.vim_command_mode(":w\n")
```
knausj85, line 174: Instead of creating a new 'save file' or 'save all' command, I'd suggest maybe implementing the existing actions
https://github.com/knausj85/knausj_talon/blob/87efebdcae0ad23db7a29dee899b7177d8391e40/apps/win/edit.talon#L157

and sticking with the global commands where possible?
https://github.com/knausj85/knausj_talon/blob/87efebdcae0ad23db7a29dee899b7177d8391e40/misc/standard.talon#L61
?

apps/linux/vim.talon
```diff
@@ -0,0 +1,799 @@
...
+# assumes visual mode
+deleted selected empty lines:
+    insert(":")
+    # leave time for vim to populate '<,'>
+    sleep(50ms)
+    insert("g/^$/d\j")
+
+swap global:
+    user.vim_normal_mode(":%s///g")
+    key(left)
```
knausj85, line 385: Just FYI, you can do `key(left:3)`.

apps/linux/vim.talon
```diff
@@ -0,0 +1,799 @@
...
+    user.vim_set_normal_mode_exterm(":bufdo diffthis\n")
+
+buffer end diff:
+    user.vim_set_normal_mode_exterm(":bufdo diffthis\n")
+
+###
+# Tab
+###
+(list|show) tabs: user.vim_normal_mode(":tabs\n")
+(close this tab|tab close): user.vim_normal_mode_exterm(":tabclose\n")
```
knausj85, line 577: I would suggest implementing the existing tab-related actions in `app` where possible, so the common tab grammar can be used as much as possible 

https://github.com/knausj85/knausj_talon/blob/master/misc/tabs.talon

apps/linux/vim.talon
```diff
@@ -0,0 +1,799 @@
...
+    user.vim_normal_mode(":vertical wincmd f\n")
+
+(show|list) current directory: user.vim_normal_mode(":pwd\n")
+change buffer directory: user.vim_normal_mode(":lcd %:p:h\n")
+
+###
+# Standard commands
+###
+# XXX - technically should be handled by vim.py, but need to add "ctrl" key support
+redo:
```
knausj85, line 218: I'd suggest implementing the `edit.redo()` action instead, so we can stick with the global command 

`redo that: edit.redo()`

knausj85:
> there's a few things I'm not sure how to do deal with in the best way. for instance a lot of people won't have all of the plugins that are supported, so in that case you might not want them to actually load when you the vim tag, but I'm not sure if there's like a sane way to configure this at the moment. I would guess that one possibilities is to have a settings section inside of the vim.talon file and then you can manually enable specific plugins and only then are they loaded? but I'm not sure exactly how we would cleanly force load them only when the setting is enabled.

I would say this is up to the user. we can document what plugins are supported, and they can simply remove any plugins they don't use?

doc/vim.md
fidgetingbits: ah ya, I will try that out. originally I just wanted to get rid of errors in a rush so i just nuked it, but ya.. perhaps there's some other way to also integrate with it even better but I just haven't had time to look yet

apps/linux/vim.talon
fidgetingbits: ya that works

apps/linux/vim.talon
fidgetingbits: I hadn't noticed those existed, will do

apps/linux/vim.talon
fidgetingbits: this is actually change in the latest unpushed to this PR version so its already part of `vim.py`. I think will just support both. 

The main reason I want it part of `vim.py` as that it's one of the main action commands that is countable in `vim.py`, and the counting works differently in talon itself (count prefix, not ordinal suffix) but as a vim user it feels very natural to do it the vim way. it also allows vim to repeat counted commands with the `repeat` command, or record into macros . strictly speaking i suspect redo and undo are unlikely to get used in macros so they don't have to be counted this way, but because when you're heavily using vim all the other commands you end up counting the same way, suddenly not being able to count redo and undo I found quite unintuitive which is also why I added them to `vim.py` eventually. 

all that said having the `redo that` and `undo that` commands  supported in general via `actions` makes sense, but still also supporting the way I'm currently doing it will be the best of both worlds.

apps/linux/vim.talon
fidgetingbits: should note that `save all` is fundamentally different though, in that `save file` saves the current buffer, and `save all` saves all open buffers. so will still end up having a `file save all` or something that isn't an `edit.talon` action probably

apps/linux/vim.talon
knausj85: Makes sense

apps/linux/vim.talon
fidgetingbits: I was thinking about this while I was going to add it, and there is an inconsistency in the repo with regards to how generic command order is approached across the repo, specifically with the order of [subject][action] vs [action][subject]. The most obvious example is earlier in this review you recommended using the `file save`, which follows the `[subject][action]` order. However, with the generic tab commands they all support the `[action][subject]` order. I also mix up the order situationally within the vim grammars, but it's making me wonder if we should try to standardize on one over the other, since I know you don't want to have duplicate ways of saying commands in your own repo.

I personally prefer [subject][action] in general because I can speak the subject that I want to interact with and then say what I want to do with it and it helps me think about what commands I actually going to give. I also think it correlates well to command-line stuff, where we say the number of the command first, and then the sub command. That said I guess it will rely depend on the user as to what order seems most intuitive though. I can move this discussion into a separate issue if you prefer.

knausj85: I hadn't really consciously considered this, but it's a great point. 

The tab commends were based on how it's spoken in English... "open a new tab and go here..." whereas file save was probably based on the the menu order.

That said, I think this should be a separate issue -
https://github.com/knausj85/knausj_talon/issues/148

apps/linux/vim.talon
knausj85: Make sense

fidgetingbits:
this now has more of the latest updates just in case anyone wants to try it that isn't aware of my branch, but I still haven't the tweak things to follow all the latest generic talon actions serves still not actually suitable fur merging I think :|

knausj85:
Sorry for the moving target on this one. Hopefully, the generic actions prove worth the effort.... so far, I’ve been very happy with them with the apps I use.

fidgetingbits:
I guess I'm going to close this for now as it's been almost a year, end there's been significant changes to all of this. there is stuff now that I think probably shouldn't be merged in its current state, and the docs are out of date and stuff. I may end up opening another pull request at some point, with the updated stuff

PR #99, opened 2020-05-17
click on pop fix

jonaskindler93:
Literally only a 2 letter PR :) 
Not sure if this is a bug or intended behaviour:
Right now if the user specifies that pop should click and it should not end the gaze_wheel, it does not click on pop when gaze_wheel is active.
The behaviour that I think is logical is that it does click under these circumstances. Am I the only one?

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -190,7 +190,7 @@ def on_pop(active):
     if (gaze_job or scroll_job):
         if settings.get("user.mouse_enable_pop_stops_scroll") >= 1:
             stop_scroll()
-    elif not eye_zoom_mouse.zoom_mouse.enabled and eye_mouse.mouse.attached_tracker is not None:
+    if not eye_zoom_mouse.zoom_mouse.enabled and eye_mouse.mouse.attached_tracker is not None:
         if settings.get("user.mouse_enable_pop_click") >= 1:
             ctrl.mouse_click(button=0, hold=16000)
 


code/mouse.py
```diff
@@ -190,7 +190,7 @@ def on_pop(active):
     if (gaze_job or scroll_job):
         if settings.get("user.mouse_enable_pop_stops_scroll") >= 1:
```
knausj85, line 191: What if we flip these checks instead? 
`if settings.get("user.mouse_enable_pop_stops_scroll")
   if (gaze_job or scroll_job)
elif
   ...`
?

jonaskindler93: You mean like this:
```
if settings.get("user.mouse_enable_pop_stops_scroll") >= 1:
        if (gaze_job or scroll_job):
            stop_scroll()
elif not eye_zoom_mouse.zoom_mouse.enabled and eye_mouse.mouse.attached_tracker is not None:
        if settings.get("user.mouse_enable_pop_click") >= 1:
            ctrl.mouse_click(button=0, hold=16000)
``` 

That won't work if the user wants to stop a scroll by pop, but then also wants to use pop to click while no scroll is running (which I don't do, but would assume is a common scenario?)

knausj85: You're correct, of course. *facepalm*

This particular behavior was intentional for my personal preference, but perhaps we'd be best off with a setting (mouse_pop_stop_scoll_and_click) here?

jonaskindler93: To me that sounds a bit hard to understand. how about this:
```
    if (gaze_job or scroll_job):
        if settings.get("user.mouse_enable_pop_stops_scroll") >= 1:
            stop_scroll()
        elif settings.get("user.mouse_enable_pop_click") >= 1:
            ctrl.mouse_click(button=0, hold=16000)
    elif not eye_zoom_mouse.zoom_mouse.enabled and eye_mouse.mouse.attached_tracker is not None:
        if settings.get("user.mouse_enable_pop_click") >= 1:
            ctrl.mouse_click(button=0, hold=16000)
```
that should get us the same behavior.

knausj85: I'm fine with the above, that makes sense to me.

knausj85: I just realized that you'll want to add the check for zoom mouse/attached tracker check to the click in the scroll case too. Prevents double clicking with zoom mouse. Might break that common stuff into a function.

PR #102, opened 2020-05-18
Fix vscode windows

auphofBSF:
Thank you @khornlund for the initial base work implementing vscode The following commits enable Visual Studio Code functionality to work in windows.
I am unable at the moment  to prove this does not break for linux. I imagine the same functionality would also be suitable for MAC but also unable test for MAC so have not added

### DIFF
--- a/apps/vscode.talon
+++ b/apps/vscode.talon
@@ -1,4 +1,7 @@
+# see app/vscode_generic.talon for linux and windows action implementation 
+
 app: Code
+app: Code.exe
 -
 
 # editing
--- a/apps/vscode.talon
+++ b/apps/vscode.talon
@@ -1,6 +1,12 @@
 # see app/vscode_generic.talon for linux and windows action implementation 
 
+os: linux
 app: Code
+
+os: mac
+app: Code
+
+os: windows
 app: Visual Studio Code
 app: Code.exe
 -
--- a/apps/vscode.talon
+++ b/apps/vscode.talon
@@ -1,6 +1,7 @@
 # see app/vscode_generic.talon for linux and windows action implementation 
 
 app: Code
+app: Visual Studio Code
 app: Code.exe
 -
 
--- a/apps/vscode.talon
+++ b/apps/vscode.talon
@@ -1,4 +1,3 @@
-# see app/vscode_generic.talon for linux and windows action implementation 
 
 os: linux
 app: Code
--- a/apps/linux/vscode.talon
+++ b/apps/linux/vscode.talon
@@ -1,0 +1,81 @@
+# Microsoft - Visual Studio Code
+# see app/vscode.talon for voice commands
+
+os: linux
+app: Code
+-
+
+action(user.ide_refactor):
+  key(ctrl-a)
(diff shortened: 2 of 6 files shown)


apps/vscode.talon
```diff
@@ -1,4 +1,6 @@
-app: Code
+# see app/vscode_generic.talon for linux and windows action implementation 
+
+app: /Code.*/
```
knausj85: The best practice here is to add another entry, and for Windows include both the executable and friendly name if different. (On occasion, the MUICache can break, so the friendly name may not work for everyone.)

`app: code
app: Code.exe
app: whatever-app.name() for VSCode`

Note that separate app entries get ORed
FWIW, I use this command to get this info for Window
`debug app:
    print(app.name())
    print(app.executable())`

khornlund:
I should add that windows and linux have different shortcuts, most of them are the same but a few are different so maybe we should just have two files.

auphofBSF:
> I should add that windows and linux have different shortcuts, most of them are the same but a few are different so maybe we should just have two files.

How about common commands all kept at this app level and platform specific variations in the relevant platform folders?

auphofBSF:
> > I should add that windows and linux have different shortcuts, most of them are the same but a few are different so maybe we should just have two files.
> 
> How about common commands all kept at this app level and platform specific variations in the relevant platform folders?

The changes in this PR with regards the app definition for windows are not working .  I was so sure it worked before committing but can't make it work today.  You mentioned the MUICache becoming corrupt.  I don't know about this cache,  could you assist me please.

The debug print shows `Visual Studio Code` for **app.name**  and **app.executable** is `C:\Program Files\Microsoft VS Code\Code.exe`

The following is interpreted correctly and works `app: /Code[.exe]/`

The following what I understood was better practice does not work 
```
app: Code
app: Code.exe
```
However the following his interpreted correctly and works 
```
app: Code
app: Visual Studio Code
app: Code.exe
```
Appreciative you could please assist with some deeper understanding

knausj85:
The MUICache is a registry key that Talon uses to pull the app.name() from on Windows. If it’s not present, it’ll use the exe name for Windows. 

I don’t endorse/haven’t used this app, but they have an accurate description of the MUICache here http://www.nirsoft.net/utils/muicache_view.html.

Aegis also recommends avoiding regex where possible, but it’s unavoidable for some use cases.

As for khornlund’s note, I agree that it’s easiest just to maintain separate action definition files for each platform, and a separate command file that works for all. This isn’t necessarily fully reflected in the depot since I’ve come to this conclusion the hard way. 

Eventually, my goal is to define a common grammar for the ides. Maybe.

auphofBSF:
> MUICache

Thank you for the structuring advice I'm just trying to understand how to best squash the commits !

auphofBSF:
I don't believe it is possible for me to the squash the commits in this PR branch, but I believe you will be able to squash and merge this PR once it is acceptable to merge

knausj85:
Oh, I didn't know about squash and merge. Thanks for your patience, and the contribution.

PR #104, opened 2020-05-19
linux firefox vimvixen support

fidgetingbits:
Some of this in theory overlaps with commented out stuff in `apps/generic_browser.talon`, which I have never tried. 

Only tested on Linux. Also worth noting the setting for enabling a little hack to use vimvixen commands in annoying scenarios like when the URL bar is highlighted, etc. I chose to enable this by default, but we can change that if needed.

### DIFF
--- a/apps/linux/vimvixen.talon
+++ b/apps/linux/vimvixen.talon
@@ -1,0 +1,126 @@
+os: linux
+tag: firefox
+-
+settings():
+    user.vimvixen_auto_focus = 1
+
+(page|tab) (previous|left):
+    user.vimvixen_key("K")
+(page|tab) (next|right):
+    user.vimvixen_key("J")
+(page|tab) (home|first):
+    user.vimvixen_key("g")
+    key(0)
+(page|tab) (end|last):
+    user.vimvixen_key("g")
+    key($)
+(page|tab) new:
+    user.vimvixen_key("ctrl-t")
+(page|tab) reopen:
+    user.vimvixen_key("ctrl-shift-t")
+(page|tab) close:
+    user.vimvixen_key("d")
+(page|tab) select:
+    user.vimvixen_key("b")
+last (page|tab):
+    user.vimvixen_key("ctrl-6")
+(page|tab) find:
+    user.vimvixen_key("b")
+    key(tab)
+(page|tab) find <phrase>:
+    user.vimvixen_key("b")
+    insert("{phrase}")
+(page|tab) open:
+    user.vimvixen_key("escape")
+    insert(":open ")
+    key(tab)
+(page|tab) open <phrase>:
+    user.vimvixen_focus()
+    insert(":open {phrase}")
+    key(tab)
+[(page|tab)] back:
+    user.vimvixen_key("H")
+[(page|tab)] forward:
+    user.vimvixen_key("L")
+[(page|tab)] refresh:
+    user.vimvixen_key("r")
+link:
+    user.vimvixen_key("f")
+link new:
+    user.vimvixen_key("shift-f")
+(address|Earl|link) bar:
+    user.vimvixen_key("ctrl-l")
+(address|Earl|link) copy:
+    user.vimvixen_key("y")
+copy (address|Earl|link):
+    user.vimvixen_key("y")
+add bookmark:
+    user.vimvixen_key("a")
+focus input:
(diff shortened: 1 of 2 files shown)


fidgetingbits:
one thing I guess this raises is that we will only trigger this rule if firefox is loaded, but it will contextually replace the generic browser actions that are defined by firefox.talon rule. I didn't add those yet for vim vixen because  I was thinking they are already defined by firefox, but I guess it makes sense to possibly override.

I see this causing at least one quirk though, one we would need to be able to flip  between a mode where firefox actions are enabled versus vimvixens, and have a way to flip back quickly. one reason is that a lot of websites have built in keyboard short cuts or other features that conflict with the vim vixen so you often disable it for a specific site and so in those cases you want to flip back over to firefox defaults. 

so then one solution might be to have a vimvixen mode and those rules for the generic browser actions are in separate firefox rule files with a `not mode: vimvixen` rule whereas the rest of firefox is always loaded no matter the mode

fidgetingbits:
my 'focus' trick is pretty buggy atm, so will definitely disable it by default, but will spend a bit of time trying to find a usable method.

knausj85:
I've been kinda avoiding browser plugins by default so things work consistently across all browsers. Maybe this is a mistake, particularly if a lot of folks are indeed navigating entirely by voice. hmmm... let me think about this one a bit.

fidgetingbits:
I find the tobii eyetracker mostly unusable so far unfortunately, even with the zoom to click feature, so 99% is voice based browsing so far. Can't speak for the others. I'll love it if I can get eye tracking accurate enough to use my eyes comfortably though.

knausj85:
Gotcha, thanks for the insight. 

With that in mind, I'm kinda thinking vimium might be a better fit for this repo since it supports Chrome, Firefox, Brave, and there's something vimium-based for Safari. Annoyingly, it looks like the shortcuts are ever so slightly different between vimvixen and vimium. I'm not sure there's a great way to support both. 

thoughts?

fidgetingbits:
kk I have never used it but I can give it a whirl and see

knausj85:
thinking a bit about the website shortcut conflict a bit, we could possibly have a tag to disable the vimium/vimvixen commands.

```
tag: firefox 
and not tag: custom_web_app_shorts
```

or something. Then the web app-specific things could assert the custom_web_app_shorts

fidgetingbits:
so I have a different `vimium.talon` file working and it seems to be a good enough for equivalent, so I'll send a separate pr request for that the next time I get a chance to do a pr storm. I didn't bother  adding support for all of the focus trickery yet, since it wasn't working and vim vixen the well anyways

having the `and not` tag make sense, though I did notice that vimium supports the idea of having partial black lists of command for certain websites, so I suppose of some people will want to disable that `and not` tag from theirs.

knausj85:
Thanks for taking the time to take a look.

Does it make sense to support both vimvixen and vimium here? I suppose this could potentially be a use case for disabling vimvixen by default?

fidgetingbits:
I think we may as well keep both in case other people happen to use it over vimium for some particular reason, but I agree we can have vimium as the default of one of them and a being enabled by default since it's more cross-platform. I still need to try to fix the auto focus stuff in vim vixen or just switch back to not using it by default at all before this gets merged, so still have to wait for me to have time to edit it.. hopefully next week.

one thing I didn't do yet is implement the vimvixen/vimium command as actual browser actions to make sure they match the commands, so I suppose I could do that as well. when problem I see though is since the browser and the plugging will be enabled at the same time they'll technically be overlap for the commands, but maybe that's actually ok because in theory they should both do the same thing no matter which one fires? but I guess minimizing grammar overlap in general is better.

so along that line of thinking it makes me wonder if vimvixen and vimium talon files should only have support for commands that don't already have support in the browser-specific files? so it's basically just augmenting what already exists. probably makes the most sense...

knausj85:
Generally, i prefer to keep the grammar smaller where possible. i tend to lean towards the augmentation approach, but that'll vary with the browser I suspect.  implementing duplicate things as browser actions might cause issues - not sure which one talon would choose if e.g. we enable/disable vimium via a tag or mode.

fidgetingbits:
since you merged vimium and I stopped using vimvixen, as well as the focus trick being broken and me not planning to fix it any time soon (if ever), I will close this PR

PR #107, opened 2020-05-22
initial port of window_snap.py to new api

timo:
couldn't test on linux (assigning a window's rect isn't
implemented there yet, could work-around with wmctrl)

also moving windows by name is NYI

### DIFF
--- a/code/window_snap.py
+++ b/code/window_snap.py
@@ -1,0 +1,181 @@
+import time
+
+from talon import Module, Context, ui, config
+
+"""Provides a voice-driven window management application implemented in Talon.
+
+You can use this to replace applications like Spectacle, BetterSnapTool, and Divvy.
+
+Todo:
+    - Provide for mapping keyboard shortcuts as a fallback when the new API is launched.
+"""
+
+mod = Module()
+
+change_screen_mode = mod.setting("window_manager_change_screen_mode", type=str, default="same")
+
+def sorted_screens():
+    """
+    return screens sorted by their left most edge, from left to right
+    """
+    return sorted(ui.screens(), key=lambda screen: screen.visible_rect.left)
+
+
+def move_screen(off=None, screen_number=None, win=None):
+    if win is None:
+        win = ui.active_window()
+
+    src_screen = win.screen
+    screens = sorted_screens()
+    if screen_number is None:
+        screen_number = (screens.index(src_screen) + off) % len(screens)
+    else:
+        screen_number -= 1
+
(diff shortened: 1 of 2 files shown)


knausj85:
I get a bunch of NYI errors on Windows too, so I think it might be a bit too early to merge this one.

Is this a bug for this on the beta github?

```
2020-05-30 23:07:15 ERROR    12: talon\scripting\talon_script.py:431|
   11: talon\scripting\talon_script.py:210|
   10: talon\scripting\talon_script.py:398|
    9:      talon\scripting\actions.py:56 |
    8: -----------------------------------# cron thread
    7:        talon\scripting\types.py:249|
    6:  user\knausj_talon\code\snap.py:163| grid_actions_by_name[name]()
    5: -----------------------------------# 'cron' talon_plugins.engines.dragon:<lambda>()
    4:  user\knausj_talon\code\snap.py:106| return lambda: resize_to_grid(column, ..
    3:  user\knausj_talon\code\snap.py:93 | resize_window(x, y, w, h)
    2:  user\knausj_talon\code\snap.py:68 | win.rect = rect
    1:             talon\windows\ui.py:200|
NotImplementedError
```

dwiel:
given the conversation in #137 what is needed to merge this? Is it just renaming the files? Which did we decide on:

knausj_talon/desktop_management/

1. `window_snap.talon` `window_snap.mac.py`
2. `mac/window_snap.talon` `mac/window_snap.py`
3. `window_snap.mac.talon` `window_snap.py`

In this case #1 seems the most reasonable since the enumeration of snap commands will likely be the same across platforms but the implementations will differ significantly based on the os. I'm open to any of them and am happy to update this PR accordingly

code/window_snap.py
```diff
@@ -0,0 +1,181 @@
+import time
+
+from talon import Module, Context, ui, config
```
dwiel, line 3: `from talon import config` gives me an error on talon v0.0.8.29-1482

knausj85:
I don't necessarily see a lot here that needs to be OS-specific, particularly once the Rect is writiable on windows and Linux.

May make sense just to stick with

```
window_management\window_snap.py
window_management\window_snap.talon
```

and do whatever temporary workaround for a particular OS? 

For windows, i believe we could use the method I specify here for now
https://github.com/talonvoice/beta/issues/106

dwiel:
Good point. That makes sense

knausj85:
A version of this has been merged.

PR #109, opened 2020-05-25
vscode: New window, command palette, editor split

auphofBSF:
Further voice commands, for vscode under windows

### DIFF
--- a/apps/vscode.talon
+++ b/apps/vscode.talon
@@ -10,6 +10,9 @@ app: Visual Studio Code
 app: Code.exe
 -
 
+# general
+command palette: user.ide_command_palette()
+
 # editing
 code refactor all: user.ide_refactor()
 code refactor: user.ide_refactor_in_line()
--- a/apps/vscode.talon
+++ b/apps/vscode.talon
@@ -42,6 +42,9 @@ console last: user.ide_terminal_focus_previous()
 console up: user.ide_terminal_scroll_up()
 console down: user.ide_terminal_scroll_down()
 
+#app window
+new code window: user.ide_window_new()
+
 # focus
 show explorer: user.ide_show_explorer()
 show search: user.ide_show_search()
--- a/apps/vscode.talon
+++ b/apps/vscode.talon
@@ -46,9 +46,9 @@ console up: user.ide_terminal_scroll_up()
 console down: user.ide_terminal_scroll_down()
 
 #app window
-new code window: user.ide_window_new()
+new instance: user.ide_window_new()
 
-# focus
+# focus | (see alternate commands in ide.talon)
 show explorer: user.ide_show_explorer()
 show search: user.ide_show_search()
 show source control: user.ide_show_source_control()
--- a/apps/win/vscode.talon
+++ b/apps/win/vscode.talon
@@ -6,6 +6,21 @@ app: Visual Studio Code
 app: Code.exe
 -
 
+# General
+action(user.ide_command_palette):
+  key(ctrl-shift-p)
+
+# Splits
+action(user.ide_split_right):
+  key(ctrl-shift-p)
+  insert("workbench.action.splitEditorRight")
+  key(enter)
+
+action(user.ide_split_left):
+  key(ctrl-shift-p)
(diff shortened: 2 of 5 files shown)


code/ide.py
```diff
@@ -266,6 +275,9 @@ def ide_git_annotate():
     def ide_git_menu():
         """Show the git menu for IDE"""
 
+    def ide_toggle_explorer():
```
knausj85, line 278: Any reason not to just use ide_toggle_project?

code/ide.py
```diff
@@ -499,3 +511,6 @@ def ide_show_debug():
 
     def ide_show_extensions():
         """Show Extensions"""
+
+    def ide_window_new():
```
knausj85, line 515: I don't think this new action is necessary, should be able to override the talon-declared one

https://github.com/knausj85/knausj_talon/blob/master/apps/win/app.talon#L37

and we could stick with the global 'new window' command with VSCode
https://github.com/knausj85/knausj_talon/blob/606421ea22a321a22c44065de6133f396b81c1c4/misc/window_management.talon#L1

apps/win/vscode.talon
```diff
@@ -6,6 +6,31 @@ app: Visual Studio Code
 app: Code.exe
 -
 
+# General
+action(user.ide_command_palette):
+  key(ctrl-shift-p)
+
+# Splits
+action(user.ide_split_right):
+  key(ctrl-shift-p)
```
knausj85, line 15: I'd suggest calling the user.ide_command_palette() here, so if the shortcut's just in one place.

apps/vscode.talon
```diff
@@ -46,9 +46,9 @@ console up: user.ide_terminal_scroll_up()
 console down: user.ide_terminal_scroll_down()
```
knausj85, line 46: Do we still need any of these commands now that we're using ide.talon? Seems to be a lot of overlap, this should probably be reserved for anything that's specific to VSCode.

apps/win/vscode.talon
auphofBSF: Agreed

code/ide.py
auphofBSF: Thanks that works

code/ide.py
auphofBSF: I felt there was enough Symantec difference to create this, but on checking a few IDE's I now have a different opinion.   
Yes in vscode this Explorer view maps to ide_toggle_project. However there is the ability to open an OS file explorer/manager and maybe this `ide_toggle_explorer`  should be that command `vscode> File:Reveal in file explorer` however it is not able to be toggled so it should finally be `ide_show_explorer`

additionally there are a number of other `ide_show_*` commands 
```
    def ide_show_explorer():
        """Show Explorer"""

    def ide_show_search():
        """Show Search"""

    def ide_show_source_control():
        """Show Source Control"""

    def ide_show_debug():
        """Show Debug"""

    def ide_show_extensions():
        """Show Extensions"""
```
I believe are hangovers from the merge of vscode.talon and ide.talon.   only the `ide_show_explorer` should stay for reason above.

I see you have done some great refactoring over weekend. I suggest I pull back this PR , cherry pick my updates on a new PR, I have also overweekend got my Linux Dev environment up and running ,so can test simultaneous against both platforms

knausj85: Yeah, I agree with the ide_show_* stuff, good catch.

i can see the value in that ide_toggle_explorer/ide_show_explorer action, but would suggest something like `ide_reveal_in_file_manager`. Arguably a bit more descriptive/generic. 

Pulling this PR back sounds reasonable to me. Sorry for the inconvenience due to the refactor, been sitting on it for too long.

apps/vscode.talon
auphofBSF: Agreed I think these are also legacy, I will attempt review  each and make changes via a  new PR as discussed

auphofBSF:
As discussed, in code review, will be best with all vscode refactor introduced by bba1b56156eb264477ef73344751f91804f38957 to close this PR and resubmit on the new structure after with necessary changes discussed here

PR #111, opened 2020-05-31
Add standard programming commands

brxck:
Hi, I wonder if there's any interest in incorporating this pattern of a standard programming command interface.

This idea is taken from [Caster](https://github.com/dictation-toolbox/Caster/blob/master/castervoice/rules/ccr/standard.py). It defines a common set of programming commands for languages to implement. I think it makes it more straightforward to start adding a new language, and makes it easier to switch between languages as a user.

I added a JavaScript implementation and adapted the existing Python one. When adapting the Python file, I attempted to keep any commands that weren't covered by the standard ones. I did not change Go because it makes use of dictation.

More commands, (or even dictation commands, like the Go implementation?), could be added.

Thanks, I appreciate the amount of work that goes into this repository.

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -5,11 +5,13 @@ ctx = Context()
 key = actions.key
 
 extension_lang_map = {
-"py"   : "python",
-"cs"   : "csharp",
-"cpp"  : "cplusplus",
-"h"    : "cplusplus",
-"talon": "talon",
+    "vue": "javascript",
+    "js": "javascript",
+    "py": "python",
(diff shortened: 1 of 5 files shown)


knausj85:
Thanks, this is the direction I intend to move. I've got something very similar in a different branch (https://github.com/knausj85/knausj_talon/tree/generic-language-grammar). I'll take a look, and take the best of both worlds.

knausj85:
Closing this out since I've merged our approaches and committed

PR #114, opened 2020-06-01
Add help search

mkandalf:
Adds help search.

The search matches against literal worlds (not captures) in rules. The semantics are AND for all terms, ignoring order.

This also refactors pagination to produce a consistent number of lines for showing commands (either in a context or search view).

Screen Shot 2020-06-01 at 5 12 47 PM

### DIFF
--- a/misc/help.talon
+++ b/misc/help.talon
@@ -1,6 +1,7 @@
 help alphabet: user.help_alphabet(user.get_alphabet())
 help context$: user.help_context()
 help active$: user.help_context_enabled()
+help search <phrase>$: user.help_search(phrase)
 help refresh$: user.help_refresh()
 help next$: user.help_next()
 help previous$: user.help_previous()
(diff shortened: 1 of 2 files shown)


misc/help.talon
```diff
@@ -1,6 +1,7 @@
 help alphabet: user.help_alphabet(user.get_alphabet())
 help context$: user.help_context()
 help active$: user.help_context_enabled()
+help search <phrase>$: user.help_search(phrase)
```
knausj85, line 4: I'd vote to user user.text here instead of phrase.

mkandalf: @knausj85 Updated

mkandalf: I think the ideal thing is to actually use a matcher consisting of all words in your commands, but I was getting some weird parses when I tried that.

knausj85: Looks like the help.talon thingy didn't get updated? 

`help search <user.text>$: user.help_search(user.text)`

or am I missing something?

mkandalf: My bad, didn't add the file. Check again.

PR #115, opened 2020-06-03
add browser/firefox tag for plugins and an extra app match

fidgetingbits:
also select all which is nice for the url bar


apps/linux/firefox.talon
```diff
@@ -1,9 +1,9 @@
 os: linux
 app: Firefox
+app: firefox
 -
-#todo tags
-# settings():
-	# tags = browser
+tag(): browser
+tag(): firefox
```
knausj85, line 6: I think this needs to be `user.firefox`, no?

apps/linux/firefox.talon
```diff
@@ -72,3 +72,5 @@ action(browser.submit_form):
 
 action(browser.toggle_dev_tools):
 	key(ctrl-shift-i)
+
+select all: key(ctrl-a)
```
knausj85, line 76: I can't believe I missed this command. There's an existing action that's implemented in edit.talon, `edit.select_all`

I think 'select all' should be moved here - 
https://github.com/knausj85/knausj_talon/blob/master/text/generic_editor.talon

apps/linux/firefox.talon
fidgetingbits: not for tags afaik. my vimvixen talon file triggers based on `tag: firefox` with this np.

i agree it's kinda weird though, as it differs from modes, settings, etc namespace

apps/linux/firefox.talon
fidgetingbits: sounds good, moved it and just pushed it for this PR

PR #117, opened 2020-06-03
github shortcut support

fidgetingbits:
this is stuff I added for github. unfortunately they don't include the actual name github in their title, so currently the window matching has to be tweaked for whatever repo you are using or just enable it across the board. in light of that I just said it as github for now so people can change it if they want. I did contact github and  asked them to change it on their website, but I haven't heard anything back


knausj85:
maybe try something like https://addons.mozilla.org/en-US/firefox/addon/domain-in-title/ ?

fidgetingbits:
I tried out that out and it works perfectly so I guess in the meantime I'll just to add that url to the title match and add a  comment that  people will need to use it

knausj85:
Thanks for giving it a whirl. I think this is a sensible approach for web app stuff - m ost browsers have some sort of plugin for adding the domain to the title

knausj85:
FYI I re-arranged the app folder as discussed for https://github.com/knausj85/knausj_talon/issues/129, and added a 'web' subfolder. Just a note to perhaps move it when we circle back to this one.

e.g.,
https://github.com/knausj85/knausj_talon/blob/master/apps/web/outlook.talon

fidgetingbits:
cool will do

fidgetingbits:
has been moved to web now and using the URL

knausj85:
Thank you!

PR #119, opened 2020-06-03
initial i3wm support

fidgetingbits:
there should be enough to get people going, but there's a shortcuts in there that are specific to my configuration I think. also it currently relies on `{vim_arrow}`, which is part of `vim.py` from the vim pull request


dwiel:
Is i3wm detectable with something like this? https://github.com/knausj85/knausj_talon/blob/master/code/app_running.py then we can leave it "enabled" and just detect when someone is using i3wm automatically.

fidgetingbits:
no it doesn't get picked up in `ui.apps()`. if there was an acceptable way to just parse the system process list than we could do it that way for sure though. 

the other problem with this as is is that it still uses m non-default key bindings, so the real solution is for me to write an i3wm.py that just parse the users i3 config end auto configures the rules to match whatever keys they have configured there, but I just haven't found the time yet.

dwiel:
Bummer it's so complex

jspam:
Instead of parsing the config file, couldn't you use `i3-msg` to send the commands directly? E.g. I have the following in my .talon file:
```
screen five: user.system_command("i3-msg workspace 5")
```

fidgetingbits:
ya that should work actually, good idea. will do that soon

fidgetingbits:
I was just thinking about this and at least for workspace switching it will cause problems, because people don't necessarily use 1-10 for their workspace names, a lot of people will use special characters (including myself), it may be at the very least these need to be pulled out of the config, or we have settings to declare the name on the talon side. I think for anything is not using the workspace names though it will be fine.

initially the easiest option is probably to have people just not us special work space names though, and I will add support for pulling them out later. 

ex of what i mean about special characters:

```
set $workspace1 "1:一"
set $workspace2 "2:二"
set $workspace3  "3:三"
set $workspace4  "4:四"
set $workspace5  "5:五"
set $workspace6  "6:六"
set $workspace7  "7:七"
set $workspace8  "8:八"
set $workspace9  "9:九"
set $workspace10 "10:十"
```

if I want to switch to work space 9 my command would be `i3-msg workspace 9:九`

fidgetingbits:
this should be good to go now. I didn't add in config parsing for the work based stuff I mentioned, and there's a couple of things where it's up to the user to set their mod key in the settings, but otherwise its generic using i3-msg now. I've been using it today with these changes so far and haven't run into any problems. I still add in config paring eventually for the remaining stuff but I think for now it's not a priority

i know @ma-anwar  uses this, so may be good for him to test. also @jspam since it sounds like you're using i3 you may want to give it a whirl before we merge

jspam:
You could also use `i3-msg -t get_config` to get the currently active configuration instead of specifying the file path. (It's available since i3 v4.14, which was released in September 2017.)

fidgetingbits:
ok thanks, for now i dont actually do anything with the config so it doesn't matter, but I'll do that once I add in the parsing

fidgetingbits:
good to go in as is  for now, I'll do the config parsing at another time

PR #123, opened 2020-06-03
basic twitter shortcut support

fidgetingbits:
twitter website shortcuts


dwiel:
it looks useful to me, any reason not to merge this?

fidgetingbits:
I think the main problem is relying on `win.title` with a fuzzy match for twitter, even though it works in practice it's probably prone to false positive. so the option suggested in the github issue #117 is to use a browser extension that exposes the actual in the tab, so I just need to add a common about that (will do shortly)

the other thing I need to do is move it to the new `web` folder

fidgetingbits:
this is updated now, with a markdown file talking about matching using the plugin (not sure the best place to put that, so freedom move)

knausj85:
Thank you!

PR #124, opened 2020-06-03
slack.talon: switch to user.text

fidgetingbits:


apps/win/slack.talon
```diff
@@ -10,9 +10,9 @@ previous workspace: key(ctrl-shift-tab)
 next workspace: key(ctrl-tab)
 # Channel
 channel: key(ctrl-k)
-channel <phrase>: 
+channel <user.text>:
     key(ctrl-k)
-    dictate.lower(phrase)
+    dictate.lower(text)
```
knausj85, line 15: Does this actually work? Unless something's changed, I think you need to user the formatters currently since dictate.* expects a Phrase object.

`insert(user.formatted_text(user.text, "ALL_LOWERCASE"))`

fidgetingbits: no it didn't work. I fixed now and tested it, must have been using the wrong branch when I thought i tested it or something

PR #128, opened 2020-06-03
VSCODE further support for win and linux based on ide.talon

auphofBSF:
Replacing #109 , this PR extends ide.talon for vscode in windows and linux,   some comments are inline . 
A key issue is I have reverted `misc/git.talon`  as cannot get tag:terminal working in vscode, cmd.exe or pwsh.exe  I just don't fully understand the use of tag: and its got late in day, maybe you can identify a quick fix.
A second issue is the Mac support, I want to just copy the Linux for start but need a tester, I can probably reactivate a MAC virtual machine but may take me a  couple of days


knausj85:
probably preferable to alias the commands in the generic vscode. i think there's a lot of value in keeping the grammar consistent across OS', unless something's really, really specific

`open containing folder | show in explorer | show in finder: dostuff()`

auphofBSF:
Further reduction of duplication with ide.talon,  added some more common debug support, 
Outstanding:  ( should do's before merge)
1.     create a `vscode.mac.talon` based on `vscode.linux.talon`
2.     some code folding support
3.     editor find/replace support
1.

auphofBSF:
@knausj85 can you please look at and comment on in particular
a594837 ide - New Commands and Actions to support find & replace
97172ef git.talon   - I think we should change in `misc/git.talon`  the `tag:terminal` to (or add if multiple tags are supported) the `tag:git` 

also in 97172ef is there any reason we can't use `tag: ide` in `vscode.*.talon`   it appears that you cannot use `app:` and `tag:` simultaneous in a `.talon`

- [ ] I will now see if I can do a quick port of `vscode.linux.talon` to mac

knausj85:
I can test out the Mac stuff tomorrow. I'll poke around the PR at the same time.

apps/vscode/vscode.talon
```diff
@@ -11,41 +11,8 @@ app: Visual Studio Code
...
-# selection
-cursor up: user.ide_up_cursor()
-cursor down: user.ide_down_cursor()
+# general
 
-# debugging
-step over: user.ide_step_over()
-step into: user.ide_step_into()
-step continue: user.ide_continue()
+open containing folder | show in explorer | show in finder: user.ide_reveal_in_file_manager()
```
knausj85, line 16: I don't see a reason not to include this in the generic IDE.talon, to be honest. That way, if you forget which OS you're on, it'll still work.

apps/vscode/vscode.win.talon
```diff
@@ -0,0 +1,235 @@
...
+  user.ide_command_palette()
+  insert("workbench.action.splitEditorDown")
+  key(enter)
+
+#Refactor
+
+action(user.ide_refactor):
+  key(ctrl-a)
+  # Format Document
+  key(alt-shift-f)
```
knausj85, line 66: This is looks like the equivalent of the `ide_fix_format.` I think this should triggering the refactor command ('ctrl-shift-r'), no?.

apps/vscode/vscode.win.talon
```diff
@@ -0,0 +1,235 @@
...
+  key(ctrl-s)
+
+action(user.ide_refactor_in_line):
+  # Format Document
+  key(alt-shift-f)
```
knausj85, line 71: I think in vscode, this action is the same as ide_refactor.

apps/vscode/vscode.win.talon
```diff
@@ -0,0 +1,235 @@
...
+  # Go to Definition
+  key(f12)
+
+action(user.ide_go_back):
+  key(alt-left)
+
+action(user.ide_go_forward):
+  key(alt-right)
+
+action(user.ide_up_cursor):
```
knausj85, line 90: The ide_*_cursor actions are intended for adding multi-select cursors, not for selections. 

See this: https://code.visualstudio.com/docs/editor/codebasics

e.g., from jetbrains
```
action(user.ide_multi_cursor): key(shift-alt-insert)
action(user.ide_multi_cursor_stop): key(escape)
action(user.ide_up_cursor): user.idea("action EditorCloneCaretAbove")
action(user.ide_down_cursor): user.idea("action EditorCloneCaretBelow")
```

misc/ide.talon
```diff
@@ -3,12 +3,21 @@ app: /jetbrains/
 app: IntelliJ IDEA
 app: PyCharm
 app: PyCharm64.exe
+app: Visual Studio Code
+app: Code - OSS
+app: Code
+app: Code.exe
+app: Atom
```
knausj85, line 10: I don't think we have any Atom support yet, so I'd probably leave that out for the moment.

misc/git.talon
```diff
@@ -1,4 +1,5 @@
 tag: terminal
+tag: git
```
knausj85, line 2: Where is this tag defined?

knausj85: To activate a tag, you do
```
-
tag(): terminal
```

though I'm not sure we'd want to bulk enable terminal commands in VSCode, might cause issues.

Note that custom tags need to be declared in .py, e.g.
https://github.com/knausj85/knausj_talon/blob/41760365bafcb9672e0c570de774d8aba4192e16/code/gdb.py#L5

knausj85:
> also in 97172ef is there any reason we can't use tag: ide in vscode.*.talon it appears that you cannot use app: and tag: simultaneous in a .talon

Yeah, this seems like a good time to do so. You'd just to declare the tag in ide.py, and update the jetbrains/vscode.talons to assert the tag

auphofBSF:
Thanks for the review, looks all sensible, I will get into it tomorrow

knausj85:
I'm going to take this as-is and take care of these minor things tonight - I'm switching to VSCode myself.

I really appreciate this contribution, and your patience.

knausj85:
Most of these are resolved on my branch, plus some extra support added fyi. Not sure about bulk enabling terminal commands.

PR #130, opened 2020-06-03
Update jetbrains app names + allow nicer app names

hbk619:
Outlook doesn't end with .exe so misses the filter
Also map lock to Slack as surely no one wants the lockapp to have focus (it's the lockscreen, but focusing it doesn't trigger the lockscreen)

### DIFF
--- a/code/switcher.py
+++ b/code/switcher.py
@@ -7,8 +7,19 @@ import platform
 
 app_cache = {}
 overrides = {
-    'grip': 'DataGrip', 
-    'term': 'iTerm2'
+    'grip': 'DataGrip',
+    'term': 'iTerm2',
+    'lock': 'Slack'
+}
+
+# If you don't like an app name you can remap it here
+# key: name OS thinks app is
+# value: what you want to say
+friendly_names = {
+    'pycharm64': 'pycharm',
+    'idea64': 'idea',
+    'webstorm64': 'webstorm',
+    'outlook.exe': 'outlook'
 }
 
 mod = Module()
@@ -18,7 +29,7 @@ mod.list('launch', desc='all launchable applications')
 @mod.capture
 def running_applications(m) -> str:
     "Returns a single application name"
-    
+
 @mod.capture
 def launch_applications(m) -> Capture:
     "Returns a single application name"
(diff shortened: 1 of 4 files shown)


code/switcher.py
```diff
@@ -7,8 +7,19 @@
...
-    'term': 'iTerm2'
+    'grip': 'DataGrip',
+    'term': 'iTerm2',
+    'lock': 'Slack'
+}
+
+# If you don't like an app name you can remap it here
+# key: name OS thinks app is
+# value: what you want to say
+friendly_names = {
```
knausj85, line 18: Aren’t friendly name and overrides doing the same thing? Or is there a need for both that I’m missing?

hbk619: Hmm yes, sorry. I think I was thinking overrides is aimed at changing the app that it's mapped to, rather than what you say to trigger it. I guess the advantage of friendly names is it removes the "unfriendly" one rather than adding to the map. Having thought about it, it's not worth the extra code though. I'll remove that when I'm at my PC :-)

knausj85: I'm fine with either, so I leave it up to you which method to keep.

hbk619: Thinking about it some more, this would break Linux as pycharm64 is specific to Windows. What do you think of making overrides platform specific? I saw https://github.com/knausj85/knausj_talon/pull/136/files maybe we could use talon's built in settings for users to define overrides for each platform and use the app.platform to determine which ones to use.

knausj85: I don't quite follow why this would break Linux? Wouldn't there just be another linux-specific entry if needed?

Not opposed to platform-specific, if it's necessary. Note I did merge 136.

hbk619: If I set pycharm to focus pycharm64 then on Linux it'd try to focus pycharm64 which doesnt exist. Should the part that breaks the app name down also remove numbers? Or have I misunderstood that?

knausj85: Oh, you're correct of course.

knausj85: I don't see a need for a setting with the csv stuff, separate platform-specific csvs loaded by the switcher seems reasonable

hbk619:
Closing this as I'll do the separate platform csvs instead

PR #135, opened 2020-06-06
Language additions for Talon Files, key chord voice commands

auphofBSF:
Just a consistency on quotes in insert and some useful voice commands to make key chord composition easier

### DIFF
--- a/lang/talon.talon
+++ b/lang/talon.talon
@@ -20,6 +20,17 @@ insert:
 key:
 	insert('key()')
 	edit.left()
+control key:
+	insert('ctrl')
+super key:
+	insert('super')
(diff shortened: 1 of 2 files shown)


lang/talon.talon
```diff
@@ -20,6 +20,17 @@ insert:
 key:
 	insert('key()')
 	edit.left()
+control key:
```
knausj85, line 24: wondering if we should use the modifiers capture here, so the same words are used?

`<user.modifiers> key: "{user.modifiers}"`

auphofBSF: Yes, thanks for showing some more good Talon

PR #137, opened 2020-06-08
Windows window/app management additions

auphofBSF:
Here are a number of generic Windows window management actions.

I think this should follow the structure you started for vscode and under apps have a folder and sub files   `generic_windows_management` etc and `generic_app` 

I will be willing to refactor this PR on the feedback


knausj85:
I'm going to add some extra brains for discussion, as I've got (3) window management-y PRs, lol.

@fidgetingbits has some i3-specific window management things
https://github.com/knausj85/knausj_talon/pull/119

@timo has ported window_snap.py, which presently works only on Mac OS, I believe
https://github.com/knausj85/knausj_talon/pull/107

I like the generic actions @auphofBSF defined, as we can implement them per-OS. We can probably do something similar for workspaces/virtual desktops that @fidgetingbits added in his scripts. Can't cover everything in a generic fashion, but probably worth extracting out what we can.

Thoughts?

auphofBSF:
Certainly looks like #119 and  #107 can extend what I have here in #137. I propose 
```
knausj_talon\platform\generic_desktop_management
      \generic_desktop_management.talon   #voice commands
      \generic_desktop_management.linux.talon  #implementation linux
      \generic_desktop_management.mac.talon #implementation mac
      \generic_desktop_management.win.talon #implementation windows
      \generic_desktop_management.py # code support for all os implementation

```
We fit #137, #119 and #107 into this. 
Does this work for all ?

> I'm going to add some extra brains for discussion, as I've got (3) window management-y PRs, lol.
> 
> @fidgetingbits has some i3-specific window management things
> #119
> 
> @timo has ported window_snap.py, which presently works only on Mac OS, I believe. This might cover a lot of the desired functionality once setting the rect is supported.
> #107
> 
> I like the generic actions @auphofBSF defined, as we can implement them per-OS. We can probably do something similar for workspaces/virtual desktops that @fidgetingbits added in his scripts. Can't cover everything in a generic fashion, but probably worth extracting out what we can.
> 
> Thoughts?

fidgetingbits:
yeah I'm fine moving in this direction, I agree on the suggested file layout in general. what else do you envision this `platform/` folder holding?

I think for i3 we don't want the implementation to actually go into the `generic_desktop_management.linux.talon` file, since there's a lot of different but popular window managers (kde, gnome, i3, openbox, dwm, etc) on linux, all with different short cuts, so it's not really feasible to just have one file that implement all of the `.linux.talon` actions. This differs from Windows (assuming people use 10 these days I guess) and Mac, where it's mostly generic. 

In general having some separate folder, perhaps under `platform/`, like `platform/desktop_managers/` with `i3wm.talon`, `gnome.talon`, etc probably makes sense? One confusing thing here though, is we start to move towards `<appname>.<platform>.talon` naming for a lot of stuff, but when we run into single OS only things we may still want to have an OS specific folder like  `platform/desktop_managers/linux/` and `platform/desktop_managers/windows/`. For instance, some people might use [bug.n](https://github.com/fuhsjr00/bug.n) tiling wm on Windows, etc. Dunno if that gets too confusing though, just throwing the problem out there.

As for apps, I'm not sure there's actually any sane defaults short of if we did it for gtk or qt specific app shortcuts, but it may be for linux we still need to have mostly separate per-app files. For instance the `app.talon` in this PR for Linux doesn't make sense to me... some of those commands work on Firefox/Chrome, but afaik isn't actually generic set of commands for most Linux apps in general? Linux is hard in this regard... so I don't think on Linux we necessarily want a generically enabled talon file randomly pressing buttons. 

In general how do you @knausj85 settle on the best command name, for instance `maximize window` in this PR is good in so far as it obvious, but as someone that flips in and out of full screen windows a good 100+ times a day, saying a 5 syllable word (`maximize window`) vs 2 syllable word (`full screen` in my i3wm.talon) is annoying. I guess we will basically favor the most obvious/intuitive command and if people find it too long they simply customize? I'm fine with whatever, as I'll customize mine I'm sure, but just mostly curious on what is favored.

knausj85:
I don't necessarily see a need for a platform folder, and my use of "generic" early on strikes me as an unenforced error. 

Something like this perhaps?
`knausj_talon\desktop_management
      \desktop_management.talon   #voice commands
      \desktop_management.i3.talon  #implementation for i3
      \desktop_management.gnome.talon  #implementation for gnome
      \desktop_management.mac.talon #implementation mac
      \desktop_management.win.talon #implementation windows
      \desktop_management.py # code support for all os implementation`

If there's a need for .py stuff for implementing things, could also have e.g. a /gnome/ subdirectory.

Extra linux desktop managers could be disabled by default - or maybe all - to force users to pick since it's a bit of a mess?

For Windows and Mac, I'm thinking to keep the repo as close to default Mac OS/Windows as possible, as that should hopefully get more people up and running with Talon faster.

> In general how do you @knausj85 settle on the best command name, for instance maximize window in this PR is good in so far as it obvious, but as someone that flips in and out of full screen windows a good 100+ times a day, saying a 5 syllable word (maximize window) vs 2 syllable word (full screen in my i3wm.talon) is annoying. I guess we will basically favor the most obvious/intuitive command and if people find it too long they simply customize? I'm fine with whatever, as I'll customize mine I'm sure, but just mostly curious on what is favored.

@fidgetingbits ^^ (1) it works consistently without conflicts (2)  shorter is better. "full screen" is a reasonable default, IMO

dwiel:
So what's the plan here? It seems like this a case of perfect is the enemy of good enough. Window managers have always been a diverse ecosystem and I dont see that changing. I mean some commands are common enough, but many aren't. Many tiling window managers will have commands like rotate windows left which dont make sense in non-tiling window managers. Instead of trying to define a window manager taxonomy and all possible commands of all possible window managers in one place, maybe we should have a window management folder as proposed, a few super generic commands in a general .talon file, and then each platform/wm gets it's own .talon/.py combo? If patterns emerge, we can refactor and build up a taxonomy of tiling vs non-tiling or whatever becomes clear.

knausj85:
I haven’t had a chance to poke at this.

Maybe we just take the window_snap PR (mac only for now) and a modified version of this PR to start for windows and Linux?

knausj85:
I might be able to make window_snap work for Windows too.

dwiel:
that seems reasonable to me. I'd be happy to help flesh things out, but also don't want to duplicate effort

knausj85:
It'll be a week or so before I can get to this, so if you've the bandwidth to do something first, please feel free.

knausj85:
Closing this since the bulk of the functionality was added via https://github.com/knausj85/knausj_talon/pull/246

PR #139, opened 2020-06-12
Add repeat command

ma-anwar:


lang/talon.talon
```diff
@@ -42,4 +42,6 @@ tag require:
 	insert("tag: ")
 user:
 	insert("user.")
-
+repeat:
+	insert("repeat()")
+	edit.left()
```
knausj85, line 47: fwiw, probably don’t need the edit.left here since repeat() doesn’t take any parameters.

ma-anwar: repeat() can take integers to indicate the number of times for the previous line to be repeated. Perhaps that is not used as much so may not be worth the edit.left. Your call.

knausj85: Oops, you're right! My bad.

PR #142, opened 2020-06-13
Add tmux commands to use in terminal

ma-anwar:
Not sure if prefixing every command with mux is a good idea?

### DIFF
--- a/apps/linux/tmux.talon
+++ b/apps/linux/tmux.talon
@@ -49,6 +49,7 @@ mux move <user.arrow>:
 mux close pane:
     key(ctrl-b)
     key(x)
+#Say a number right after this command, to switch to pane
 mux pane numbers:
     key(ctrl-b)
     key(q)
--- a/apps/linux/tmux.talon
+++ b/apps/linux/tmux.talon
@@ -1,5 +1,5 @@
 os: linux
-tag: terminal 
+tag: tmux
 -
 mux: "tmux "
 
--- a/apps/linux/tmux.talon
+++ b/apps/linux/tmux.talon
@@ -1,0 +1,54 @@
+os: linux
+tag: terminal 
+-
+mux: "tmux "
+
+#session management
+mux new session:
+    insert('tmux new ')
+mux sessions:
+    key(ctrl-b)
+    key(s)
+mux name session:
+    key(ctrl-b)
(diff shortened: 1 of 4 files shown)


fidgetingbits:
I think using `mux` is good

knausj85:
Pardon my ignorance, is this a standard feature of linux terminals, or would a tag make sense here?

ma-anwar:
No, this is not a standard feature. Did I use the tag incorrectly? Sorry I still get mixed up, should it be,
`tag(): terminal`?

knausj85:
No, you used tag correctly - I meant: does a custom, separate tag from terminal (tmux or something), make sense here that could be asserted as-needed by terminal programs? Particularly since some apps might opt into some ‘standard’ terminal stuff (e.g. vscode).

I really know nothing about this one, just asking dumb questions. 🤔

ma-anwar:
Yes definitely a custom tag makes sense in this case. I was not aware of vscode opting into terminal commands as well but that makes sense.

knausj85:
Thanks. This tag needs to be declared... somewhere, no? Maybe you missed a file? :)

e.g., https://github.com/knausj85/knausj_talon/blob/master/code/gdb.py

It probably makes sense to have a /code/tags.py for something like this, where we can just toss tags that need no other code support

ma-anwar:
Thank you for teaching me the perils of testing on my own fork as opposed to the patch branch! The tags file is a useful idea. I have added a tags file that declares tags with generic descriptions. I think I managed to collect the custom tags defined so far. 

I also added a line to declare modes, but I'm not sure if that should be in the scope of this file. Also not sure if mode description should be generic as well,

apps/linux/tmux.talon
```diff
@@ -0,0 +1,55 @@
...
+tag: tmux
+-
+mux: "tmux "
+
+#session management
+mux new session:
+    insert('tmux new ')
+mux sessions:
+    key(ctrl-b)
+    key(s)
```
knausj85, line 11: This is just an FYI as I'll be merging this as-is, but you should be able to combine some of these into one command, unless you need arbitrary delay between
e.g.,
`key(ctrl-b s)`
`key("ctrl-b {number}")`

PR #144, opened 2020-06-14
add "push to talk" actions for press/release of a button

timo:
delays turning speech off until after the next phrase is parsed,
so that we don't cut stuff off too soon

### DIFF
--- a/code/delayed_speech_off.py
+++ b/code/delayed_speech_off.py
@@ -1,0 +1,28 @@
+from talon import Context, Module, actions, app, speech_system
+
+delay_mod = Module()
+
+delayed_enabled = False
+
+def do_disable(e):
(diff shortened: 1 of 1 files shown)


knausj85:
could you describe the use case for this? I'm not quite sure i follow, but I'm pretty tired

timo:
i've recently gotten an azeron keypad and i've bound buttons to speech on and speech off, but if's not convenient for when you just want to say one command. naively i could have bound "speech on" to "button pushed" and "speech off" to "button released", but when i tried that i released too early basically 99% of the time.

hence this feature. you can "turn off" while still speaking or directly after speaking and it'll wait until a phrase has been recognized and only then actually turns speech off.

additionally, it won't do anything if speech is already on

PR #146, opened 2020-06-18
Add JavaScript language implementation

brxck:
Converted this over to the standard programming actions. Let me know if anything should be adjusted!


knausj85:
Thanks, looks good.

Did this approach to language support work well for you? Just curious if anything should be tweaked.

brxck:
It did work well, I didn't run into any issues really.

PR #147, opened 2020-06-18
Add commands to delete or reformat the last formatted phrase

brxck:
This is another thing inspired by Caster. 

- The `nope` command will to delete the last phrase formatted by `formatters.py`. 
- The `nope that was <user.formatters>` command will delete the last phrase and then reformat it with the spoken formatter

Both are anchored to prevent accidental activations.

I wonder if there is a way to better track state; this is just how I remember it being implemented in Caster. Let me know what you think!


misc/formatters.talon
```diff
@@ -8,3 +8,4 @@ phrase <user.text> over: insert(user.text)
 word <user.word>: insert(user.word)
 list formatters: user.list_formatters()
 hide formatters: user.hide_formatters()
+nope: user.clear_last_phrase()
```
knausj85: is the nope command necessary? seems like we could just use the global `undo that` command instead

brxck: In my experience undoing can clear more or less than you want it to, especially if you are chaining commands together.

fidgetingbits: Agreed, in vim `undo` command will often undo an entire line of added text rather than the most recent added word, etc depending on the mode switching that has happened and stuff. So I could still see something like `nope` being useful still, but I do think multiple undo style commands will get confusing for people.

`nope that was <format>` will be super useful for me though

I guess if the `nope` command is used, something longer than one word `nope` will be necessary to avoid accidental misfires though, perhaps just `nope that` and `nope that was <formatter>`?

knausj85: Thanks, makes sense @brxck. 

I tend to agree with fidget, *probably* best to avoid single word commands - particularly global ones - in this instance. `nope that` seems like a reasonable compromise

brxck: That sounds good to me as well!

PR #149, opened 2020-06-20
Select with no arguments will select a single line by default.

barrcodes:
This fixes an issue where the command `select` would map to `select all` by default. Selecting a single line is far less destructive, as selecting all will move the cursor to the end of the document.


knausj85:
I actually haven't seen this - are you using wav2letter?

I'm a little hesitant to add many global one-word commands variants.

barrcodes:
Yes w2l, this is using 0.0.8.29 and the current stable version of w2l.

I am able to reproduce it consistently if i revert this line. I am open to other solutions!

PR #153, opened 2020-06-20
Fixed default "select" functionality.

barrcodes:
This fixes an issue where the command `select` would map to `select all` by default. Selecting a single line is far less destructive, as selecting all will move the cursor to the end of the document.

This is using 0.0.8.29 and the current stable version of w2l.

I am able to reproduce it consistently if i revert this line. I am open to other solutions!z

Duplicate of #149 because I had to fix my git history.

### DIFF
--- a/text/generic_editor.talon
+++ b/text/generic_editor.talon
@@ -48,7 +48,7 @@ go page up:
     edit.page_up()
 
 # selecting
-select line:
+select [line]:
     edit.line_start()
     edit.extend_line_end()
 


knausj85:
I'll think about this one a bit - might be one of those situations where it's just best to branch the command(s) as-needed for you.

barrcodes:
Yeah I considered that. No problem if that's the decision you make.

The new voice model may make this not a big issue, but the old voice model sometimes catches false positives. So I was often selecting all, and losing my position in editing code documents.

knausj85:
Just curious, have you tried this with the b4 model? 

I’m thinking to leave this as-is for now, and re-evaluate with the new model

Issue #155, opened 2020-06-20
Request: Block completion for class, method, struct


barrcodes:
I've been reading breadcrumbs left in the code, so I know that talon scripts can't take parameters yet. There's a note left to put "()" inside a language-specific function at some point.

This ticket is just a reminder for one of us to consider adding language specific block completion to whatever method is created. example:
```
insert(" {}")
# the following is just an example, not sure if it's ideal
key(left enter up end left left left)
```

knausj85:
We could also implement these in python for now, if it's getting in the way.

barrcodes:
I'm not blocked or anything like that. If it's something that you want to put a band aid on, i'm ok with that. I don't want to go to get difficult to maintain, whatever you think is best.

PR #157, opened 2020-06-23
[draft] some convenient scripts for typescript

barrcodes:
@knausj85 I stubbed out some quick scripts for what we were discussing in #155 to discuss your opinion of the ideal implementation. It works almost perfectly, the caveat that `state class` does not take a type.

In a previous version of this code, I had the following. It worked, but this would fail in languages that are not typed. I see now why aegis prefers not abstracting code.

_programming.talon_
```
^state class <user.text>$:
    user.code_type_class()
    user.code_type_formatter(user.text)
```

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -9,6 +9,8 @@ setting_public_function_formatter    = mod.setting('code_public_function_formatt
 setting_private_variable_formatter   = mod.setting('code_private_variable_formatter', str)
 setting_protected_variable_formatter = mod.setting('code_protected_variable_formatter', str)
 setting_public_variable_formatter    = mod.setting('code_public_variable_formatter', str)
+setting_type_formatter               = mod.setting('code_type_formatter', str)
+setting_interface_formatter          = mod.setting('code_interface_formatter', str)
 
 mod.tag("code_comment", desc='Tag for enabling generic comment commands')
 mod.tag("code_operators", desc='Tag for enabling generic operator commands')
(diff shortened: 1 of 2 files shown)


code/code.py
```diff
@@ -9,6 +9,8 @@
 setting_private_variable_formatter   = mod.setting('code_private_variable_formatter', str)
 setting_protected_variable_formatter = mod.setting('code_protected_variable_formatter', str)
 setting_public_variable_formatter    = mod.setting('code_public_variable_formatter', str)
+setting_type_formatter               = mod.setting('code_type_formatter', str)
```
knausj85, line 12: I'm not very familiar with typescript, but isn't this covered by the _*_variable_formatters?

knausj85: Oops, nevermind. Derp.

knausj85:
There might be value in a slightly separate grammar for typed vs untyped languages via tag?

```
tag: typed_language
-
^state class <user.text> of type <user.text>$:
    ... do typed stuff...
```

```
tag: untyped_language
-
^state class <user.text>$:
    ... do untyped stuff...
```

lang/typescript.talon
```diff
@@ -153,3 +220,39 @@ state reduce:
...
+
+state return default:
+  edit.right()
+  insert(": ")
+  insert("void")
+
+^state return type <user.text>$:
+  edit.right()
+  insert(": ")
+  user.code_type_formatter(user.text)
```
knausj85, line 258: Are types in typescript uniformly public camel case? I'm completely unfamiliar with it, just a dumb question.

This one is beyond this PR, just picking your brain. What do you think about adding a capture for some types in code.py, similar to what manwar's done for C? 
https://github.com/ma-anwar/knausj_talon/blob/bb893df3ede9f77944903c6c98139ecead1a5d70/code/c.py#L17

barrcodes: Both js and ts *should* use PascalCase for modules / types and camelCase for members. The exceptions are aliases for primitives (as is common in many languages) like `string`, `boolean`, `number`, etc. So yes, some combination of these may work really well together.

I was going to solve that with two commands, but that may be ugly. Something like `state return type` and `state return primitive`

knausj85: Thanks. A combo seems like a reasonable approach here, I'll look into the captures sometime this week hopefully.

FYI, I did some prelim work on snippet support, which you can likely use easily with your rewrite fwiw
https://github.com/knausj85/knausj_talon/commit/1e4eea598e795c77da960da4e0880baddd4fa2f7

PR #162, opened 2020-06-26
Update operators.talon

lexjacobs:
Fixes two transposed operator definitions.

Small edit to fix the transposed key/value pairs defining `<` and  `>=` .


knausj85:
Oops, thanks.

PR #167, opened 2020-07-12
Fix issue where some vocab words aren't transcribed correctly

jenmei:
Due to [https://github.com/talonvoice/beta/issues/88].

For example, saying "dee en ess" results in “@dns” instead of “DNS". Changing the keys to lowercase fixes the problem for now.

Also, made it so that the dictionary is created first with simple vocabulary so that the things defined later override the things defined earlier.

### DIFF
--- a/code/vocabulary.py
+++ b/code/vocabulary.py
@@ -10,15 +10,17 @@ simple_vocabulary = [
     "minecraft",
 ]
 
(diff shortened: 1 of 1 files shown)


lunixbochs:
don't lowercase the terms now, this was fixed in talon

PR #169, opened 2020-07-17
dictate python functions from list

RodrigoChT:
right now I only have a short list of python functions, once I get approval for the structure of this me and the rest of python users can continue adding to it


knausj85:
I’m thinking we could make these commands generic (programming.talon) with context-specific captures for supported languages.

This would be similar to the initial snippet support
https://github.com/knausj85/knausj_talon/blob/master/apps/vscode/snippets/csharp_snippets.py
https://github.com/knausj85/knausj_talon/blob/master/code/snippets.py

The capture could probably be declared in code.py

I’m happy to help with this, if it sounds reasonable. 

A step towards making the ideas here
https://github.com/ma-anwar/knausj_talon/blob/master/code/c.py
generic and cross-language

lang/python.talon
```diff
@@ -113,3 +113,11 @@ action(user.code_state_return):
 dunder in it: insert("__init__")
 state (def | deaf | deft): "def "
 pie test: "pytest"
+
+# show and print python functions
+list python: user.list_python_functions()
+hide python: user.hide_python_functions()
```
knausj85, line 119: One thing that might be worth adding is the ability to select the function by index when imgui is open.

Similar to what’s done with the hello system when open perhaps.
https://github.com/knausj85/knausj_talon/blob/eeea6454aa217aa5f24d0eb36885c639a439d87c/misc/help_open.talon#L5

I’ve found selecting via numbers VERY convenient fast and reliable. Could have a command like
`fun <number>: `

Could also be done later, but I’d be happy to help. What do you think?

RodrigoChT:
I would love to extend this to other programming languages, but I'm a little lost on how to do it so your help would be greatly appreciated. I also liked the idea of selecting the functions by index!

knausj85:
@RodrigoChT  - I took a first pass at making this something that might work reasonably across languages. Sorry for the formatting changes. I'm switching the depot to using `black` for formatting needs, so a lot of noise in this one 👎 
Hopefully you don't mind my use of the same PR!

 @fidgetingbits @barrcodes  - you two might be interested in taking a peek at this.

Meat of the idea is really in these two, with whatever supporting code in code.py:
https://github.com/knausj85/knausj_talon/blob/de6c69a22e3a2e9b8900f400a4bc4177733d3ada/lang/csharp/csharp.py
https://github.com/knausj85/knausj_talon/blob/de6c69a22e3a2e9b8900f400a4bc4177733d3ada/lang/python/python.py

I would be to potentially extend this to include common libraries and types for each language as appropriate. Haven't had enough time to think about those just yet.

```
toggle fun: user.code_toggle_functions()
fun <user.code_functions>: user.code_insert_function(code_functions)
```

fidgetingbits:
indeed black certainly adds a lot of noise at first, but totally worth it :D

I like this idea, and certainly goes well with the recent C stuff I hope to PR soon. I think the  selection window idea will be really useful too

Since you mentioned types, one the thing I have been wanting to do with types is have the ability to toggle between default types used by the C stuff, since for instance stdint (ex: `uint8_t`) vs default types (`unsigned char`) are two common used but different typing styles across many code bases (i constantly work on different codebases). originally I tried to do this with just updating a talon setting, but it seems like you can up to a setting on the fly, so I guess we need some way of maybe standardizing non talon setting 'settings' for different programming languages. may be suitable for a different issue if you want me to file it, but I thought I'd raise it since you mentioned types here if you plan on thinking about it. I suspect this might be something ran into other languages (cpp, windows (DWORD, etc)), so may also be worth having some quasi generic method of selecting the default type set

knausj85:
Is that changing the typeset something that could be done cleanly per-project? I have my doubts, but...

For example, the default type list could be the std_*, or whatever’s considered vanilla C. Then for projects using some other set, the list could be changed via a more specific python context.

So a python file with something like, plus a modified list.
```
ctx.matches = r"""
mode: user.c
mode: command 
and code.language: c
tag: ctypes_fancy
```

Then, elsewhere opt-in:

```
title: \some project\
title: \some other project\
-
tag: ctypes_fancy
```

RodrigoChT:
This looks great, what has to be done before it is ready to be merged?

knausj85:
I think with my latest changes, it's pretty reasonable and ready to go in. Types and libraries and stubbed in code.py once we're ready for those.

I switched to copy and paste for these, so we can hopefully reliably wrap selections without fighting the IDE(s).

```
toggle funk: user.code_toggle_functions()
funk <user.code_functions>: 
    old_clip = clip.text()
    user.code_insert_function(code_functions, "")
    clip.set_text(old_clip)
funk cell <number>: 
    old_clip = clip.text()
    user.code_select_function(number - 1, "")
    clip.set_text(old_clip)
funk wrap <user.code_functions>: 
    old_clip = clip.text()
    edit.copy()
    sleep(100ms)
    user.code_insert_function(code_functions, clip.text())
    clip.set_text(old_clip)
funk wrap <number>: 
    old_clip = clip.text()
    edit.copy()
    sleep(100ms)
    user.code_select_function(number - 1, clip.text())
    clip.set_text(old_clip)
```

PR #170, opened 2020-07-21
Allow app name overrides to be platform specific

hbk619:
This means pycharm etc can be remapped across os's
On linux Mint I found the updater is called MintUpdater.py which means "py" gets mapped to that, so trying to "focus pycharm" fails as talon picks up the "py" and focuses the updater. This will work for pycharm community edition on Linux but the paid version doesn't have -ce so a more intelligent solution will need to be implemented (or map it to another name, which ever is preferred!)


knausj85:
Looks good, thanks. Hopefully, this makes it easier for folks to branch as needed.

PR #176, opened 2020-07-27
fix a bunch of operator/lang stuff

fidgetingbits:
the main notable change here is typos in operators for dereference and exclusive, as well as fixing some bitwise operator typos unmixed up commands in various language files.

as a result of these changes touching files with some other things I hadn't scent pull requests for, I'm mixing a few other things together:

* `go` to use user.text
* block comments (used by c, but c stuff not PRed yet)
* more extension types
* whitespacing
* null addition for some commands (saying none is not intuitive when using C)

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -1,6 +1,7 @@
-from talon import Context, actions, ui, Module, settings, registry, imgui, fs
-import re
 import os
+import re
+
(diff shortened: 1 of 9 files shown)


knausj85:
Thanks!

PR #179, opened 2020-07-27
fuzzy focus switching

### LINKED ISSUE #145: Windows switching could use fuzzier matching
Right now when you want to switch window focus you have to say a word the matches a generated list of running applications, and this list has an override that lets you more easily say the names of complex application names, for instance `term` is `iTerm2` or whatever.

I propose we change the way the list works so that actually we just use `focus <user.text>` for the command, and then `switcher_focus()` does the lookup for whatever text it gets inside the running application list. Doing it this way allows us to leverage any additional vocabulary that's been added to `vocabulary.py` without necessarily having to hardcode it again inside of `switcher.py`. 

in addition I propose that we introduce slightly fuzzer matching. IFor example, my `vocabulary.py` as a rule for "key pass" which becomes "keepass", and this is helpful when I am just talking to other people in chat or whatever. But the actual name of the running process is `KeePassXC`. I don't necessarily want to always be typing that exact word into chat with other people when I say `key pass` though. I suspect this type of thing will be increasingly common the more people are using it with different applications 

As it stands if I wanted to match the app with current switcher I'd have to put "key pass":"KeePassXC" override in `switcher.py`, instead of just matching the result I already have in `vocabulary.py`. 

In addition to using `vocabulary.py`, I propose we check for a partial match using `startswith()`, so then I would be able to say something like "focus key pass" and it would actually come through as `focus keepass`, which would in turn match against `keepassxc`. 

We can of course keep the overrides for certain things that are a bit harder to say in `switcher.py`, for instance I think people aren't going to have `term` becoming `iTerm2` in `vocabulary.py`.

This can mostly all be done with a simple `switcher_focus()` change like:

```
        running = ctx.lists["self.running"]
        wanted_app = None
        for running_name in running.keys():
            if running_name == name or running_name.lower().startswith(name):
                wanted_app = running[running_name]
                break
        if wanted_app is None:
            return

        for app in ui.apps():
            if app.name == wanted_app and not app.background:
                #os.system("i3-msg '[class=\"(?)%s\"] focus'" % app.name)
                app.focus()
                break
```

Not sending a PR for now as I figured I'd run it past others first.

One possible change could be that it does two passes, the first pass matches on explicit name matches so you don't accidentally match against something that has a more explicit match, and then if there are no explicit matches, it tries a second pass with fuzzy matching.

fidgetingbits:
fix #145, including your suggestion


code/switcher.py
```diff
@@ -68,10 +68,19 @@ def get_words(name):
...
+                break
+        if wanted_app is None:
+            return
+
         for app in ui.apps():
-            # print(f"--------- app.name:{app.name}  app.bundler:{app.bundle}")
-            if name in app.name and not app.background:
-                app.focus()
+            if app.name == wanted_app and not app.background:
+                os.system("i3-msg '[class=\"(?)%s\"] focus'" % app.name)
```
knausj85, line 82: This looks like a i3-specific call - i assume there an issue with app.focus() for i3?

fidgetingbits: oh right good catch, app.focus can be used in general . I started to use monkey patching for that kind of stuff so these don't show up in diffs, but forgot to switch this part back

fidgetingbits: I switched this back and also fixed a couple other warnings I noticed

PR #185, opened 2020-07-29
Update switcher.py for reals

knausj85:
- Since the <user.running_applications> capture provides the actual app.name for switching purpose, we only need to do the fuzzier matching if the name isn't already in the list of running apps. In that case, the input is coming from <user.text> in practice. Ugly code, but working.
- For <user.text> input, ignore input < 3 characters
- Add watcher to update overrides when the override file changes. Old overrides appear to stick around for 1 or 2 focus command sometimes, but take affect thereafter.
- Add length check to the CSV processing to prevent silly crashes.

### DIFF
--- a/code/switcher.py
+++ b/code/switcher.py
@@ -2,32 +2,28 @@ import os
 import re
 import time
 
-from talon import Context, Module, app, imgui, ui
-from talon.voice import Capture
+from talon import Context, Module, app, imgui, ui, fs
 
 # Construct at startup a list of overides for application names (similar to how homophone list is managed)
 # ie for a given talon recognition word set  `one note`, recognized this in these switcher functions as `ONENOTE`
 # the list is a comma seperated `<Recognized Words>, <Overide>`
 # TODO: Consider put list csv's (homophones.csv, app_name_overrides.csv) files together in a seperate directory,`knausj_talon/lists`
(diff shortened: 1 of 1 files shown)


ym-han:
I'm weirdly still having issues with the focus switcher on the mac for some reason, though I'm also using a tweaked version of the code (I didn't bother with a csv for the overrides). Is anyone having issues as well?

knausj85:
@ym-han Can you be more specific - what sort of issues?

#179 also added some “fuzzier” switching, which may not be preferred by everyone. 

https://github.com/knausj85/knausj_talon/blob/844739820e81db74cb92c4df722ed64f25610048/code/switcher.py#L39 

^<user.text> In the capture will allow eg “focus fire” for Firefox,

ym-han:
Sorry about my previous question; I hadn't done enough testing and asked impulsively right before heading to bed... 

I've tested more now, and I think the issue was my inept merging of the edits. It's good to know about the fuzzier switching, though; I might end up disabling that. 

Thanks and sorry once again!

knausj85:
@ym-han  No apology necessary! If the fuzzy thing is causing grief, we can make it opt-in (separate command in window_management.talon that’s disabled by default) or something. 

Let me know

PR #186, opened 2020-07-30
Implement splits actions for tmux.talon

ma-anwar:
I implemented some of the actions in splits.talon that were applicable to tmux, also cleaned up the file a bit in terms of moving one line inserts to the same line as the command.


fidgetingbits:
You might want to consider switching to `number_small` instead of `number`  since the pain numbers should reasonably never hit over a hundred ya? it will help reduce the longer term rule explosion. otherwise I can do it as part of https://github.com/knausj85/knausj_talon/pull/180 as I plan to change bunch of stuff, but since I don't use tmux it's better for you to confirm I guess

ma-anwar:
Great point, that should be fine. I'll change that.

knausj85:
I'm not sure I see the advantage of enabling the splits tag here if you keep the "mux <split command>" stuff too. Now, you have two commands enabled for most things.

e.g.,
tmux.talon:
`mux split horizontal: user.split_window_horizontally()`

splits.talon:
https://github.com/knausj85/knausj_talon/blob/844739820e81db74cb92c4df722ed64f25610048/misc/splits.talon#L8

The advantage here, if you forego the additional split-specific commands, is that the same commands will work if you switch to e.g. VSCode or jetbrains down the line. Thoughts?

Also, this guy
```
mux move <user.arrow>:
    key(ctrl-b)
    key(arrow)
```

is the equivalent of these

```
split right: user.split_window_right()
split left: user.split_window_left()
split down: user.split_window_down()
split up: user.split_window_up()
```

no?

I think I'll refactor these to user.split_window(direction), as I think your use of <user.arrow> is better. Only tradeoff is we'll have to implement in .py for now.

fidgetingbits:
this is an interesting problem,  because say my terminal is inside vim ( which is how I do it) and vim itself has splits  which leverage the generic talon split grammars,  and I want the generic split grammars  to be used for vim and not tmux,   then having the 'redundant' mux specific commands are still actually good  so i can explicit.  the problem is I'm not sure the cleanest way for us to specify a generic grammar app priority when you have multiple apps that would otherwise overlap? right now the main way would just be to remove  the tag from the given talon files  where you don't want it to trigger.  so I'd remove it from tmux.talon  in my branch.  maybe there's a better way though...

knausj85:
There’s no separate priority at the moment. Talon will use the most specific implementation of an action. 

eg, os, tag, and app alone each have the same specificity. os + app is more specific. os + app + title more so.

Is splitting a terminal via tmux hosted in vim a valid thing? I don’t think we can use the generic actions right now for both in such a case. vim could implement the generic actions, and tmux would need to do its own, entirely separate thing

fidgetingbits:
it's valid (for me) ya, but possibly edge case-ish. i use vim as my terminal locally for convenient terminal interaction via talon, but if i often have to ssh into a remote box i admin and then i will often have a screen (similar to tmux) session to save my work state, etc and that screen session will thus be running embedded within the local vim terminal, often in a vim split.

knausj85:
Oh my, hehe

Yeah, I'm not sure entirely sure what the best approach is for such a case. Maybe pick what we think is the "most common" use case, and use the generic grammar for that one. My initial thought is tmux might be the "better" choice for the generic grammar.

vim -> non-generic grammar/don't use the actions 
tmux -> generic splits actions/grammar 

?

ma-anwar:
I was kind of "split" about how to do this as well. would it be reasonable if I left the `mux <split command>` in but commented out so that if someone wanted all of the mux specific commands then they could just uncomment that? Otherwise, we can just take out the mux specific commands as, I think, the most common use case would be to have the generic split actions?


Also I will wait for `user.split_window(direction)` to be changed or make a PR myself and then implement it with the generic grammar for tmux.

fidgetingbits:
yeah I don't know what is "better" out of the two honestly,  at the moment I'm just open to whatever's easiest.  I guess some people will just end up manually changing things on their own branch to use what they want,  so it that vein we could keep two copies in each talon file,  and just have them commented out by default?  maybe that's too annoying  to maintain and explain though

dwiel:
As another data point, I use iterm splits but even though I do use tmux and vi occasionally but not their splits. Implementing all the splits and then me commenting out vi/tmux splits would be pretty easy for my use case.

dwiel:
Another option would be to somehow allow me to prefix the vi/tmux splits with so I could still do "tmux split left" even if I have it disabled as the "primary" split left implementation.

apps/linux/tmux/tmux.talon
```diff
@@ -1,26 +1,25 @@
 os: linux
```
dwiel, line 1: Maybe also os: mac? Though once you are in a terminal you could be on any OS, might not make sense for most terminal apps to be OS bound

dwiel:
This would be verbose, but flexible:
```
mux split left: user.tmux_split_left()
action(user.tmux_split_left): user.split_left()
```

If you want tmux as "primary" leave how it is. Otherwise comment out the 2nd line. This means you only have to comment out a block of code to disable "primary" status. Since actions with arguments have to go in python, those would have to be commented out there too. Alternatively the action mapping could happen entirely in python and enabled/disabled as a setting or one big block comment.

apps/linux/tmux/tmux.talon
knausj85: I think based on the discussion here on #210, we should probably just ditch the OS entirely and let folks opt-in to tmux via the tag.

knausj85:
> This would be verbose, but flexible:
> 
> ```
> mux split left: user.tmux_split_left()
> action(user.tmux_split_left): user.split_left()
> ```
> 
> If you want tmux as "primary" leave how it is. Otherwise comment out the 2nd line. This means you only have to comment out a block of code to disable "primary" status. Since actions with arguments have to go in python, those would have to be commented out there too. Alternatively the action mapping could happen entirely in python and enabled/disabled as a setting or one big block comment.

I think this is a reasonable approach.  

We may want to consider adding some custom tmux actions, so it can be customized per-OS if needed? 

I haven't used tmux in ages, so pardon my ignorance. Does the shortcut for 

```
mux window <number_small>:
    key(ctrl-b )
    key('{number}')
    key('{number_small}')
```

or anything vary per-OS (per terminal?)?

knausj85:
After some back-and-forth with aegis on using captures for certain things, I think I'd prefer to forgo this refactor:

> I think I'll refactor these to user.split_window(direction), as I think your use of <user.arrow> is better. Only tradeoff is we'll have to implement in .py for now.

While I like the idea, it does obfuscate things a bit for newbies.

dwiel:
> or anything vary per-OS (per terminal?)?

Not unless your terminal also uses ctrl-b for something.

splondike:
Closing this pull request since knausj indicated he'd like to keep things the way they are and there's been no response in a while.

PR #189, opened 2020-08-06
remove redundant "replace this" command

rntz:
"replace this" was defined twice (the same way) by accident.


knausj85:
Thanks, looks like I submitted a fix this just before you submitted this PR hehe. 🥇 

I went with this to consolidate them:
`replace this [<user.text>]: user.replace(text or "")`

https://github.com/knausj85/knausj_talon/commit/ca20e96c65fb3644c65e90617d85eb0510c7226d

PR #192, opened 2020-08-09
move further towards OV/subject-action word order

rntz:
update command history, running, and window commands to follow object-verb order


knausj85:
Thanks!

PR #196, opened 2020-08-10
keep more history, but only show the most recent ten lines by default

rntz:
This is especially useful if you want to have the history always be displayed, in which case you may need to reduce the number of lines displayed to avoid colliding with other windows, which is what I do; this patch lets you reduce the number of lines displayed by default, while retaining a lot of history for debugging.


knausj85:
Good idea, thanks.

PR #198, opened 2020-08-19
allow users to launch commands with arguments

rntz:
Modify switcher_launch to split its string into a command and arguments, so that launch_applications can be used for commands with arguments. For example, I use it for launching an emacs window via `emacsclient -nc`.

### DIFF
--- a/code/switcher.py
+++ b/code/switcher.py
@@ -174,7 +174,7 @@ class Actions:
 
     def switcher_launch(command: str):
         """Launch a new application by path and arguments"""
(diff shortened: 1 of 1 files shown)


knausj85:
Yeah, `shlex` does strange things on windows i.e. doesn't work. 

I wouldn't be opposed to making the `shlex` stuff Mac/Linux-only behavior for the moment.

rntz:
Eh, I'd rather keep core stuff working on all platforms. If someone needs or wants this we can reopen.

PR #203, opened 2020-08-24
generic debugger commands

fidgetingbits:
This is some generic debugger stuff I've been using lately for windbg and gdb.  it certainly not perfect and there's stuff missing I want to add but it's in a usable enough state that it should be okay to add what's there.. 

I guess there's some overlap with basic debugging stuff inside ide.talon, but I don't use an ide and I don't have any tooling set up to test it I didn't want to mess with it,  but I guess the debug stuff from the ide files could probably use this new generic debug stuff?

This also includes moving out mode setup into modes/modes.py mostly because I found it being in tags quite confusing every time I went looking for it.  also there are still tags in there which will eventually need to get replaced with the new apps stuff  but I haven't fix that in my branch yet.


knausj85:
Note ide.talon is dead anyhow, moved to using tags for common stuff (line_commands, multiple_cursors, find_and_replace) that I found reasonable.

Wasn’t entirely sure if it would work well for debuggers, but will give this a whirl. Thanks!

PR #205, opened 2020-08-24
more abbreviations

fidgetingbits:
the one contentious one might be 'funny' to 'lol', but I am tempted to start adding acronyms another generic shortening into this as well rather than only strictly abbreviations, because it seems like the natural place to do it...


knausj85:
Acronyms probably make sense here too. Good idea.

PR #206, opened 2020-08-24
basic c lang support

fidgetingbits:
this is a merging of my and @ma-anwar C stuff I've been using. is not perfect, as with everything I push it seems lol, but should be enough to get people going


knausj85:
Thanks!

PR #208, opened 2020-08-24
make screenshots work on linux

fidgetingbits:
I don't know if we want to start going the route of using settings like this, but I found it pretty convenient with other things I've been doing on my own branch. I can confirm this work so can linux, but because I don't have easy access to windows or mac at the moment I can't guarantee nothing broke

this also changes it so that the screen shots don't clobber previous screenshotted files by default

### DIFF
--- a/code/screenshot.py
+++ b/code/screenshot.py
@@ -1,9 +1,31 @@
-from talon import Module, screen, ui, actions, clip, app
-import os
+from talon import Module, screen, ui, actions, clip, app, settings
+from datetime import datetime
+import os, subprocess
 
 active_platform = app.platform
 
 mod = Module()
+mod.setting(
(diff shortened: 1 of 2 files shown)


knausj85:
Thanks, I've "merged" this, but I ran into a silly onedrive-related bug, and was on the wrong branch when I committed some minor refactoring.

The one thing I didn't keep is the default setting for the folder.
`user.screenshot_folder = "~/img/screenshots"`
since I think this setting should probably be done in a user-specific Talon file by default? Maybe.

https://github.com/knausj85/knausj_talon/commit/f5f470283d8f4c46ab004807968783941ea002ec

fidgetingbits:
Sounds good. Having a default/standard user specific talon config file make sense, I don't know why I started adding them to each file really.. as it will get pretty confusing hehe

PR #210, opened 2020-08-24
bunch of terminal stuff

fidgetingbits:
This is a bunch of terminal commands and updates that I've been using.  for some of it I hope to eventually switch to something abstract,  like a generic package manager on top of `apt` and `yay`, etc  that would work for any  command line. but at the moment I just don't have time  to do everything I want so for now  I figure I'l just send a PR  in case it's useful for anyone  in the meantime :D

summary of stuff:
* buku is a command line  bookmark manager
* apt is debian/ubuntu  package manager
* yay is an arch  package manager
* ghidra server  is a collaboration server  for reverse engineering
* nmcli  is a linux  network management  front end ( also possibly a good abstraction candidate)
* time warrior  is a time management tool that works in tandem with task warrior
* make is a development tool

### DIFF
--- a/apps/buku/buku.talon
+++ b/apps/buku/buku.talon
@@ -1,0 +1,22 @@
+os: linux
+tag: terminal
+-
+
+# general options
+bookoo help: "buku -h\n"
+bookoo version: "buku -v\n"
+bookoo auto add:
+    insert("buku -a ")
+    edit.paste()
+    key(enter)
+    insert("buku -w -1\n")
+
+bookoo update: "buku -u "
+bookoo edit: "buku -w "
+bookoo edit last:
+    insert("buku -w -1\n")
+
+# edit options
+
+# search options
+bookoo search: "buku -s "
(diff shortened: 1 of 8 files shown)


knausj85:
Hmm, my only thought here is maybe we should consider using tags for these sorts of things to make them more opt-in friendly?

`tag: boku`


... then in whatever terminal.talon

```
app: someTerminal
- 
tag: boku
#tag: kuberctrl
tag: yay
```

Or maybe that's overkill?

fidgetingbits:
I think that makes sense, also it lets you have something like a terminal `admin` mode that lets you temporarily enable a bunch of stuff that you don't normally need and so don't want boing ta down with too many commands

knausj85:
I made the existing terminal command sets opt-in friendly [here](https://github.com/knausj85/knausj_talon/commit/6493a5f75caec8ea0b686b52ac3af1d1dd782b0e), fwiw. Seems like a low-risk change, so I just committed it.

fidgetingbits:
I've since abstracted the package management stuff, like yay/apt, and done other changes that will conflict with this pull in a way that it doesn't make sense to keep it open, will eventually just do new pull requests

PR #220, opened 2020-09-04
Support for the R language

RodrigoChT:
Support for the R language

- Added R implementations for operator and programming actions.
- Added list of R functions and libraries.
- Libraries can now also be seen in a GUIi window and entered with commands the same way as functions.
- Added general actions for break, next, true, false, and NA.
- Added action for the operator "is in". In R this checks if the element is in a vector, it could appear in other languages.
- Showing function list will close library list and viceversa.

The R functions and libraries were taken from mrob95 (https://github.com/mrob95/MR-talon.git).


RodrigoChT:
I left the `user.code_private_function` action commented out because I don't know how to modify the `^funky <user.text>` command so that it works for both Python and R. Functions in R are created like this: `foo <- function(arg1, arg2) {}`. Maybe the `funky` command should be defined by each programming language separately?

knausj85:
I was thinking to re-work the funky stuff to be a single action implemented in the language-specific .py (https://github.com/knausj85/knausj_talon/issues/193). I think I'll do this today, if it's not too big a merge for ya.

It would end up looking like this:

```
^funky <user.text>$:  user.code_private_function(user.text)
```

and that would probably call the `code_private_function_formatter` which just formats the user.text appropriately.

I know there's a variety of funky commands though - for like static functions and such. 

We got a few options for those, as you mentioned:
- move to lang-specific files
- live with a few things that aren't implemented in a particular language that'll do nothing
- make the unneeded variants call down a supported one, in case you try something silly. e.g. if you don't have public/protected/private variants to deal with, these could all just call `user.code_private_function(user.text)`

What do you think?

lang/programming.talon
```diff
@@ -34,6 +34,12 @@ state include local: user.code_include_local()
 state type deaf: user.code_type_definition()
 state type deaf struct: user.code_typedef_struct()
 state (no | nil | null): user.code_null()
+state break: user.code_break()
+state next: user.code_next()
+state true: user.code_true()
+state false: user.code_false()
+state na: user.code_na()
```
knausj85, line 41: `state na` I'm a little on the fence on - might be better as an R-specific think? From a cursory search, I don't think there's an analog in thing in many languages (python, C#, C++), though I could be very wrong!

RodrigoChT: I wasn't sure either, I will change it to an R specific command.

knausj85:
I took a crack at the discussed tweaks + a first pass of R functions. I don't know R at all, so bear that it mind!

Looking at the R snippet format though
https://support.rstudio.com/hc/en-us/articles/204463668-Code-Snippets

...
Somewhat aside:
you might be well-served by using snippets similar to what I'm doing for VSCode for functions and such -- for tab stops. I watch the user snippet folder + parse on load. It's disabled by default, but snippet_watcher.py and csharp.py might serve as inspiration.

code/code.py
```diff
@@ -279,27 +283,37 @@ def code_state_while():
...
+    def code_break():
+        """Inserts break statement"""
+
+    def code_next():
+        """Inserts next statement"""
+
+    def code_true():
+        """Insert True value"""
+
+    def code_false():
```
knausj85, line 295: I didn't see these implemented anywhere?

RodrigoChT: My bad, I've now implemented them for R and Pyhon (https://github.com/knausj85/knausj_talon/pull/220/commits/485d775296ede11cc93fca056d325fb3fa54be91).

PR #222, opened 2020-09-06
adjust vocabulary.py to separate word_map from user.vocabulary

rntz:
Per some discussion in the slack (https://talonvoice.slack.com/archives/G9YTMSZ2T/p1599345924139300), it would be nice if rewriting words Talon already recognizes ("i'm" -> "I'm") didn't change recognition behavior - apparently, adding new words to user.vocabulary makes Talon prefer recognizing them over ordinary words. This PR adjusts vocabulary.py to separate "changing spellings of things Talon recognizes" from "adding new words Talon doesn't recognize".

However, I've only tested this on w2l. Since this affects a pretty core part of knausj_talon, probably it should get tested by someone using Dragon!

I also added capitalization for days of the week & months to the word map.

### DIFF
--- a/code/vocabulary.py
+++ b/code/vocabulary.py
@@ -1,30 +1,67 @@
 from talon import Context, Module, actions, grammar
 
-
-simple_vocabulary = [
-    "nmap",
-    "admin",
-    "Cisco",
-    "Citrix",
-    "VPN",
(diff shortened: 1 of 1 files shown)


knausj85:
Thanks!

PR #225, opened 2020-09-09
fix bug in homophones.py that was forgetting all the homophones

rntz:
Previously homophones.py was conditionally reading homophones.csv, but unconditionally updating all_homophones and ctx.lists["self.homophones_canonical"], which meant that updates that didn't change homophones.csv would wipe the homophones list.


knausj85:
🤦 , thanks for fixing this

PR #228, opened 2020-09-16
Updates to R language files

seananderson:
- Switch 'funky' to 'function define' (noun verb) for R language since 'funky' clashes with 'funk' often for me
- Add common R libraries
- Add many common R functions, especially from ggplot2 and the tidyverse
- Divide package-specific functions into sections
- Complete the function alphabetization within groups
- Re-tab (turn one tab to spaces) in lang/r/r.talon
- Add 'pipe that' for %>%


lang/r/r.talon
```diff
@@ -52,19 +52,19 @@ action(user.code_import):
...
-chain:
+(chain|pipe that):
     key(end)
     " %>%"
     key(enter)
 state na:
     insert("NA")
 
-^funky <user.text>$: user.code_private_function(text)
+^function define <user.text>$: user.code_private_function(text)
```
dwiel, line 70: I agree this is better than existing since funky is already defined in `programming.talon`. As a separate issue, we should probably move this line into `programming.talon` as well. No need for that to gate this PR though.

dwiel (approved):
LGTM

PR #234, opened 2020-09-18
many additions

dwiel:
I have a bunch of changes I've made in the last week or two and was driving myself crazy trying to split them up. This is about half of my changes, but it is the least contentious half so hopefully easy to merge. Afterwards, I can start merging in / creating pull request in smaller chunks.


apps/amethyst/amethyst.talon
```diff
@@ -0,0 +1,11 @@
+
```
knausj85, line 1: I'm wondering about the best way to handle window managers.

Maybe we can make stuff like i3 and amethyst opt-in by default, kinda like I did for the mouse grid?
https://github.com/knausj85/knausj_talon/blob/master/mouse_grid/grid_always.talon#L1
https://github.com/knausj85/knausj_talon/blob/a2830d844f505c37200a8f5dfdf7415cb2e3250b/settings.talon#L25

lang/python/python.talon
```diff
@@ -106,3 +106,5 @@ state past: "pass"
 #^static funky <user.text>$: user.code_private_static_function(text)
 #^pro static funky <user.text>$: user.code_protected_static_function(text)
 #^pub static funky <user.text>$: user.code_public_static_function(text)
+raise {user.python_exception}: user.insert_cursor("raise {python_exception}([|])")
```
knausj85, line 109: probably something that could be moved to programming.talon / add the list to code.py? Not a deal breaker, just a thought.

code/word_selection.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def word_neck(word_index, valid_characters=alphanumeric):
+    old = clip.get()
+
+    stop_selection("right")
+
+    actions.key("shift-end")
+    time.sleep(0.25)
+    actions.key("cmd-c")
```
knausj85, line 75: This should probably use the platform/edit-agnostic actions

`edit.copy()`
`edit.left()`
`edit.right()`

What does "shift-end" do? I don't know if there's an action for this.

code/word_selection.py
```diff
@@ -0,0 +1,154 @@
...
+
+def big_word_prev(index):
+    return word_prev(index, valid_characters=set(alphanumeric) | set("/\\-_.>=<"))
+
+
+def stop_selection(cursor_position):
+    assert cursor_position in ("left", "right")
+
+    with clip.capture() as s:
+        actions.key("shift-right")
```
knausj85, line 58: `edit.extend_right()`

code/word_selection.py
```diff
@@ -0,0 +1,154 @@
...
+        time.sleep(0.25)
+        actions.edit.copy()
+    current_highlight = s.get()
+    actions.key("shift-left")
```
knausj85, line 62: `actions.edit.extend_left()` ?

code/word_selection.py
```diff
@@ -0,0 +1,154 @@
...
+    for i in range(0, end_position - start_position):
+        actions.key("shift-right")
+
+
+def word_prev(word_index, valid_characters=alphanumeric):
+    old = clip.get()
+
+    stop_selection("left")
+
+    actions.key("shift-home")
```
knausj85, line 121: `edit.extend_line_start()` instead?

code/word_selection.py
```diff
@@ -0,0 +1,154 @@
...
+    # cursor over to the found word
+    for i in range(0, start_position):
+        actions.key("right")
+    # now select the word
+    for i in range(0, end_position - start_position):
+        actions.key("shift-right")
+
+
+def word_prev(word_index, valid_characters=alphanumeric):
+    old = clip.get()
```
knausj85, line 117: Should probably use the `with clip.revert()` method?

code/switcher.py
```diff
@@ -169,6 +169,10 @@ def switcher_focus(name: str):
         for cur_app in ui.apps():
             if cur_app.name == wanted_app and not cur_app.background:
                 cur_app.focus()
+
+                while cur_app != ui.active_app():
+                    time.sleep(0.1)
```
knausj85, line 174: Hmm, I'm not how reliable this will be on linux/Windows right now. Might end up an infinite loop sometimes - at least I've seen indications that apps can hang around on Linux recently? 

Might need a failsafe.

apps/amethyst/amethyst.talon
dwiel: Ah yeah, in public talon, I gate the commands on is_amethyst_running which worked well. Not sure how to do that now

lang/python/python.talon
dwiel: that's a reasonable thought, I can create an issue to record the desire

apps/amethyst/amethyst.talon
dwiel: amethyst context is now only active if an app called amethyst is running (not necessarily focused)

fidgetingbits: until we have a better way of detecting window manager, and in cases where the app running approach doesn't work,  I think having opt-in the same way that you did for mouse grid makes sense. I'll do it this way for i3wm when I revisit it

code/word_selection.py
rntz: shift-end selects until the end of the line; edit.extend_line_end()

misc/media.talon
```diff
@@ -1,6 +1,7 @@
 volume up: key(volup)
 volume down: key(voldown)
+set volume <number>: user.media_set_volume(number)
```
rntz, line 3: This action is not defined anywhere; did you forget to include it in the push?

rntz (commented):
we seem to be missing the action user.media_set_volume

misc/media.talon
dwiel: I did, thank you. I also need to modify the implementation to only apply to mac

PR #240, opened 2020-09-23
Rename {symbol,arrow,number,modifier,function,special} lists/captures

rntz:
The potential confusion between `<number>` (which matches a spoken number) and `<user.number>` (which matches a number **key**) was bugging me, so I made this PR to rename most of the lists and captures in keys.py to end in `_key`. In the process I uncovered some instances of this confusion in existing code:

- @dwiel's `word_selection.talon` used `<user.number>` (a number key) when it probably wants to use `<number_small>` (which matches a spoken number below a hundred).

- @fidgetingbits code in `lang/c/c.talon` used <user.function> and I don't understand why, but it seems unlikely it meant to be a function key? I've commented that out, but feel free to fix it to whatever it should be.

Anyway, for the sake of consistency & clarity I ended up renaming everything except `user.letter`, `user.letters`, and `user.modifiers`, because those seem commonly used and unambiguous. I'm happy to adjust this if people think some names ought to be shorter (maybe `user.symbol` is better than `user.symbol_key`?), but I think at least `user.number_key`, `user.function_key`, and `user.special_key` ought to be named as such, because `user.number` `user.function` `user.special` are easy to misunderstand.

This changes a lot of names and I might have missed something, so I'd appreciate a review & testing if any of y'all have the time <3.


dwiel:
I also noticed this today and fixed a few things in my repo, but hadnt made a PR yet. Looks good to me.

lang/c/c.talon
```diff
@@ -145,14 +145,15 @@ push brackets:
     insert("{variable} {letter} ")
 
 # Ex. int * testFunction
-fun <user.function> <phrase>:
-    insert("{function} ")
-    insert(user.formatted_text(phrase, "PRIVATE_CAMEL_CASE,NO_SPACES"))
-    insert("()")
-    edit.left()
+# TODO: these clearly don't want to use function_key, so what do they want to use?
+# fun <user.function> <phrase>:
```
knausj85, line 149: Looks like user.function here is a copy/paste error. Should probably just be fun <user.text>?

This might be out of scope, but we may want to take a moment to update this to use the newer .py actions e.g. `code_private_function` as in python, csharp and r.

lang/c/c.talon
```diff
@@ -145,14 +145,15 @@ push brackets:
...
+#     insert("{function} ")
+#     insert(user.formatted_text(phrase, "PRIVATE_CAMEL_CASE,NO_SPACES"))
+#     insert("()")
+#     edit.left()
 
-<user.function>:
-    insert("{function} ")
+# <user.function>:
```
knausj85, line 155: This was probably supposed to be `c_functions`. 

This is probably be out of scope for this, but I think this should probably be moved over to use the `code_functions` and `code_libraries`, as done for python and R. And maybe move to the new code_insert_function stuff, which I think can basically be copied from csharp while we're here, or open an item for it. This way, we benefit from the imgui and stuff in code.py. I guess we'd need to explore how to do that for the libraries though.

https://github.com/knausj85/knausj_talon/blob/0a9c40f9e979ba7b1bae24b94fd57635914ee613/lang/csharp

fidgetingbits:
i like this idea. looks good to me too. will file a separate issue for the c stuff

fidgetingbits (approved):

PR #241, opened 2020-09-25
Add guake.talon

khornlund:
Add talon implementation for [guake](http://guake-project.org) terminal

It's been a few months since I've worked on talon scripts so let me know if I've missed some part of best practice

### DIFF
--- a/apps/linux/guake.talon
+++ b/apps/linux/guake.talon
@@ -1,0 +1,17 @@
+os: linux
+app: Guake
+-
+tag(): terminal
+tag(): user.tabs
+
+quake show: user.system_command("guake --show")
+quake hide: user.system_command("guake --hide")
+
+action(app.tab_open):
+  key(ctrl-shift-t)
+action(app.tab_close):
+  key(ctrl-shift-w)
+action(app.tab_next):
+  key(ctrl-pagedown)
+action(app.tab_previous):
+  key(ctrl-pageup)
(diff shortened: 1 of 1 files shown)


knausj85:
Shouldn't their be an app definition in https://github.com/knausj85/knausj_talon/blob/master/code/application_matches.py for this?

khornlund:
> Shouldn't there be an app definition in https://github.com/knausj85/knausj_talon/blob/master/code/application_matches.py (or its own .py) for this?

How do they work? Do you need one for each app you match on?

knausj85:
Yeah, that's essentially correct. These are defined in .py now so the same match can be used in several places.

The old approach of matching the executable/bundle/etc is deprecated in .talon directly, and will be removed eventually.

dwiel:
When should applications_matches be used/not be used?

knausj85:
My 2c:
If an application has it's own .py, I think the app def should go there.
If there isn't a need for its own .py, slap it applications_matches.

?

dwiel:
So then it should never go at the top of a .talon file?

knausj85:
Oh, you were asking a different question. My bad.

This snippet: 
```
from talon import Context, Module
mod = Module()
apps = mod.apps
apps.windows_terminal = """
os: windows
and app.name: WindowsTerminal.exe 
"""
```

Defines the application match for window terminal that can be used in both .py and .talon. 

```
app: windows_terminal
-
stupid test: "Do something
```

The approach where talon would magically match against name/exe/bundle via `app: magic.bundle.something` behind the scenes is deprecated and will disappear someday.

dwiel:
Gotcha

PR #242, opened 2020-09-25
Add docstring command

khornlund:
Apologies if this already exists elsewhere!

### DIFF
--- a/lang/python/python.talon
+++ b/lang/python/python.talon
@@ -91,8 +91,7 @@ action(user.code_state_return):
 	insert("return ")
 action(user.code_true): "True"
 action(user.code_false): "False"
-
-
+action(user.code_document_string): user.insert_cursor("\"\"\"[|]\"\"\"")
 
 #python-specicic grammars
 dunder in it: insert("__init__")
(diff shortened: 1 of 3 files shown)


lang/python/python.talon
```diff
@@ -107,4 +107,4 @@ state past: "pass"
 #^pro static funky <user.text>$: user.code_protected_static_function(text)
 #^pub static funky <user.text>$: user.code_public_static_function(text)
 raise {user.python_exception}: user.insert_cursor("raise {python_exception}([|])")
-
+dock string: user.insert_cursor("\"\"\"[|]\"\"\"")
```
knausj85, line 110: I think this is a good idea, though I think it's worth taking a moment to make it language-neutral via code.py. 

Then we could move it into programming.talon and use it for all languages
`dock string: user.code_document_string()`

then python here would implement it
```
actions(user.code_document_string):
    user.insert_cursor("\"\"\"[|]\"\"\"")
```

as this does vary per language.
Thoughts?

khornlund: Perhaps it would need a different name as doc *strings* are a bit of a python thing, other languages typically use comments like
```
/* ... */
```

dwiel:
I hadnt used any other language with docstrings, but just looked it up and it looks like Julia does https://docs.julialang.org/en/v1/manual/documentation/

knausj85:
There's also xmldoc stuff for C#, java and others that i frequently use, and consider similar enough even though it isn't really a language feature.

khornlund:
How's this?2

knausj85:
Looks good, thanks!

PR #243, opened 2020-09-25
Better formatters

### LINKED ISSUE #238: feature request: `phrase hello slash world` -> `hello/world`
I'm not sure what the best way to go about this is, it might be simple.

some more examples/unit tests:

- `phrase hello slash world` -> `hello/world` (currently `hello /world`)
- `phrase hello slash snake name of file slash world` -> `hello/name_of_file/world` (currently `hello/name_of_file_/world`)

Similarly: `phrase file dot extension` -> `file.extension`, though this one might be complicated by the fact that I have a formatter called `dot` defined so that `dot hello world` -> `.hello_world` because I use python so much.

jcaw:
At dweil's request, a simple implementation of nested punctuation & chained (separate) formatters. 

(Details of each change in the commit messages)

Closes #238

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -283,7 +283,7 @@ def formatter_immune(m) -> ImmuneString:
     # Note that if the user speaks something like "snake dot", it will
     # insert "dot" - otherwise, they wouldn't be able to insert punctuation
     # words directly.
-    rule="<self.formatters> <user.text> [<user.text> | <user.formatter_immune>]+"
+    rule="<self.formatters> <user.text> (<user.text> | <user.formatter_immune>)*"
 )
 def format_text(m):
     out = ""
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -273,10 +273,14 @@ def formatters(m):
 @ctx.capture(
     # Add anything else into this that you want to be able to speak during a
     # formatter.
-    rule="(<user.symbol> | <user.letter> | <user.number>)"
+    rule="(<user.symbol_key> | <user.letter> | numb <number>)"
 )
 def formatter_immune(m) -> ImmuneString:
-    return ImmuneString(m[0])
+    if hasattr(m, "number"):
+        value = m.number
+    else:
+        value = m[0]
+    return ImmuneString(str(value))
 
 
 @ctx.capture(
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -259,6 +259,11 @@ class Actions:
         actions.insert(text)
         return text
 
+    def insert_many(strings: List[str]) -> None:
+        """Insert a list of strings, sequentially."""
+        for string in strings:
+            actions.insert(string)
+
 
 @ctx.capture(rule="{self.formatters}+")
 def formatters(m):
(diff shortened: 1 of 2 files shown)


dwiel:
Looks good to me. Any thoughts from anyone else?

knausj85:
I'm a little on the fence on including `user.numbers` in the capture, but I think it's fine to give whirl and tweak if needed.

dwiel:
Yeah, what is the effect of the number capture here? "phrase the best one" -> "the best one" vs "the best1"?

jcaw:
Number will be interpreted as a number - I can remove that. What about `"num <number>"` so the user can still enter numbers as e.g. part of a variable name?

rntz:
I'm against including either <user.number> or <letter> without some qualifying prefix; otherwise I think too many ordinary words would get misinterpreted as letters or numbers.  I'm going to try out merging this branch on my local config and see if I run into any issues.

rntz:
I have just tested this and it doesn't seem to work for me. For example, I tried to the phrase "snake object dot call method" and instead of `object.call_method` I get `object_dot_call_method`.

dwiel:
This is exactly what unit tests are perfect for. Sure would be nice to have a simple test harness

jcaw:
@rntz are you on Dragon or wav2lettter?

rntz:
wav2letter.

rntz:
@jcaw did you test this on Dragon, and did it work?

jcaw:
Just to mirror from Slack: I've seen similar behaviour happening on Dragon. Sometimes dictation takes precedence, sometimes the command. I'm not sure what's happening - I'm using a chunked system similar to this on my personal config and this doesn't happen. Commands always take priority.

rntz:
Further testing suggests that changing formatters does work but punctuation does not:

"hammer hello slash world" -> HelloSlashWorld :(
"snake hello world smash hello world" -> hello_worldhelloworld :)

rntz:
whoops this was a mistake on my part, that happened because #240 is merged in my local config and renames <user.symbol> to <user.symbol_key>.

rntz:
Planning to merge this as soon as I get around to it, probably in the next couple of days, unless someone beats me to it.

code/formatters.py
```diff
@@ -228,15 +244,41 @@ def formatters_reformat_last(formatters: str) -> str:
...
 @ctx.capture(rule="{self.formatters}+")
 def formatters(m):
     return ",".join(m.formatters_list)
 
 
-@ctx.capture(rule="<self.formatters> <user.text>")
+@ctx.capture(
+    # Add anything else into this that you want to be able to speak during a
+    # formatter.
+    rule="(<user.symbol> | <user.letter> | <user.number>)"
```
rntz, line 261: This needs to be changed to use <user.number_key> and <user.symbol_key> now that #240 has been merged.

rntz (changes requested):
The number/symbol vs number_key/symbol_key thing needs to be fixed before merging though. Otherwise everything looks good to me.

code/formatters.py
jcaw: Ok - I'll gate "number" with a prefix too, that seems to be the consensus. I'll leave letters in for now, you guys experiment with it and gate that too if it becomes a problem.

jcaw:
Should be compatible now.

PR #244, opened 2020-09-25
Add command to reformat selected text

### LINKED ISSUE #224: formatters should apply to selected text if they are spoken alone
for example `yeller` while some text is selected should pass the selected text through the formatter and replace the selected text with the formatted text.

jcaw:
Small command that reformats the currently selected text.

This implementation strips punctuation before formatting, but it doesn't separate out camel & studley case variables. I can add a heuristic for that if you'd like though.

I also replaced the dictation mode implementation - let me know if I've misinterpreted what that code was doing.

Closes #224

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -1,9 +1,11 @@
 from talon import Module, Context, actions, ui, imgui
 from talon.grammar import Phrase
 from typing import List, Union
+import re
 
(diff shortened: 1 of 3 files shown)


dwiel:
Breaking apart snake and camel case can be separate

PR #246, opened 2020-09-26
Port window snapping commands from Community

### LINKED ISSUE #217: window management: `snap left`, `snap bottom right`, etc
public talon implementation here: https://github.com/dwiel/talon_community/blob/master/misc/window_snap.py

### LINKED ISSUE #218: window management: `window move screen {number}`
for moving a window to a different screen: https://github.com/dwiel/talon_community/blob/master/misc/window_snap.py

jcaw:
Shortcuts to snap windows to predefined positions, as well as between
different screens.

Required extracting the "currently running at" logic from
`user.switcher_focus` so the current app is generally accessible. I also
added a timeout and some overt errors to make it a bit cleaner.

Also add the ability to snap named apps.

Closes #217 
Closes #218

### DIFF
--- a/code/window_snap.py
+++ b/code/window_snap.py
@@ -1,0 +1,215 @@
+"""Tools for voice-driven window management.
+
+Originally from dweil/talon_community - modified for newapi by jcaw.
+
+"""
+
+# TODO: Map keyboard shortcuts to this manager once Talon has key hooks on all
+#   platforms
+
+import time
+from operator import xor
+from typing import Optional
+
+from talon import ui, Module, Context, actions
+
+
+def sorted_screens():
+    """Return screens sorted by their topmost, then leftmost, edge.
+
+    Screens will be sorted left-to-right, then top-to-bottom as a tiebreak.
+
+    """
+
+    return sorted(
+        sorted(ui.screens(), key=lambda screen: screen.visible_rect.top),
+        key=lambda screen: screen.visible_rect.left,
+    )
+
+
+def _set_window_pos(window, x, y, width, height):
+    """Helper to set the window position."""
+    # TODO: Special case for full screen move - use os-native maximize, rather
+    #   than setting the position?
+    window.rect = ui.Rect(round(x), round(y), round(width), round(height))
+
+
+def _bring_forward(window):
+    current_window = ui.active_window()
+    try:
+        window.focus()
+        current_window.focus()
+    except Exception as e:
+        # We don't want to block if this fails.
+        print(f"Couldn't bring window to front: {e}")
+
+
+def _get_app_window(app_name: str) -> ui.Window:
+    return actions.self.get_running_app(app_name).active_window
+
+
+def _move_to_screen(
+    window, offset: Optional[int] = None, screen_number: Optional[int] = None
+):
+    """Move a window to a different screen.
+
+    Provide one of `offset` or `screen_number` to specify a target screen.
+
+    Provide `window` to move a specific window, otherwise the current window is
+    moved.
+
+    """
+    assert (
+        screen_number or offset and not (screen_number and offset)
+    ), "Provide exactly one of `screen_number` or `offset`."
+
+    src_screen = window.screen
+    screens = sorted_screens()
+    if offset:
+        screen_number = (screens.index(src_screen) + offset) % len(screens)
+    else:
+        # Human to array index
+        screen_number -= 1
+
+    dest_screen = screens[screen_number]
+    if src_screen == dest_screen:
+        return
+
+    # Retain the same proportional position on the new screen.
+    dest = dest_screen.visible_rect
+    src = src_screen.visible_rect
+    # TODO: Test this on different-sized screens
+    #
+    # TODO: Is this the best behaviour for moving to a vertical screen? Probably
+    #   not.
+    proportional_width = dest.width / src.width
+    proportional_height = dest.height / src.height
+    _set_window_pos(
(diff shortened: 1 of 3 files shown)


code/window_snap.py
```diff
@@ -0,0 +1,221 @@
...
+def _set_window_pos(window, x, y, width, height):
+    """Helper to set the window position."""
+    # TODO: Special case for full screen move - use os-native maximize, rather
+    #   than setting the position?
+    rect = ui.Rect(round(x), round(y), round(width), round(height))
+    window.rect = rect
+    # HACK: If on Windows and the window was maximised, this can leave the
+    #   window in a weird semi-maximised state. Some windows will be tiny or
+    #   invisible in this state. Repeating seems to solve this.
+    actions.sleep("250ms")
```
knausj85, line 39: Just curious,  is this long of a delay necessary? On windows 10 here, seems to work without a delay, but haven't tested under load or anything.

code/window_snap.py
```diff
@@ -0,0 +1,221 @@
...
+    window.rect = rect
+
+
+def _bring_forward(window):
+    current_window = ui.active_window()
+    try:
+        window.focus()
+        current_window.focus()
```
knausj85, line 47: Feels a little weird to maintain the focus of the current window to me. What's the reasoning for this, out of curiosity?

code/window_snap.py
jcaw: The original had a similar hack. 

https://github.com/dwiel/talon_community/blob/master/misc/window_snap.py#L58

You could probably get away with shorter on Windows, but I copied that in case it needed to be reintroduced for Mac. The original method set the rect three times - not sure if that would still be necessary.

code/window_snap.py
jcaw: Copying original behaviour - just improving the feel slightly. Use case is presumably something like, you're in the program you want and you want to put, say, Firefox on the left for reference:

`"snap left, snap firefox left"` vs. `"snap firefox left, snap emacs right"`

First is more concise, but the second seems more intuitive. You may be right, forcing focus for the moved window seems a little better.

code/window_snap.py
jcaw: (To be clear, testing on this end, it does need a small delay to be stable on Windows.)

code/window_snap.py
dwiel: I hadnt known this behavior was available, but I like the idea. I assumed first time I read it that it would maintain focus on whatever you are in. It could be an option to preserve or not.

I have a similar situation with a global command "shell rerun". It reruns whatever was in the shell history without losing focus on current app. Was thinking of something like "shell rerun stay" to keep focus on shell after. Many times though I just want to rerun the shell ad watch while keeping focus in my editor.

code/window_snap.py
dwiel: Maybe the delay should be gated with a if windows block similar to the mac specific hack?

jcaw:
Removed the double-move, Aegis wants to fix it first-party

knausj85:
> 
> 
> Removed the double-move, Aegis wants to fix it first-party

I guess we hold off on merging this one until then? Or did I miss an update?

code/window_snap.py
knausj85: I think that's reasonable.

code/window_snap.py
knausj85: Yeah, maybe consider adding a setting for this behavior before merging? i.e. while we wait for the official fix.

It really feels weird to me, so I'd probably vote for a default of changing focus to the app that is snapped in settings.talon? I think it'll be really confusing to newbies.

Reasonable compromise?

dwiel:
👍

jcaw:
>I guess we hold off on merging this one until then? Or did I miss an update?

It should work on Mac in the meantime and it will still work on Windows, it will just have janky behaviour if the window is maximized. Up to you, can leave it until it's fixed in Beta too.

code/window_snap.py
jcaw: I can add a setting. How are settings managed in knausj at the moment?

knausj85:
> 
> 
> > I guess we hold off on merging this one until then? Or did I miss an update?
> 
> It should work on Mac in the meantime and it will still work on Windows, it will just have janky behaviour if the window is maximized. Up to you, can leave it until it's fixed in Talon Beta.

I generally vote not to commit things that we know are broken, hehe. It's confusing for newbies, and may give a bad impression of Talon.

That said, we could make it Mac-only for now if anyone really wants it? 

I was also fine with the workaround, though looks like aegis wasn't.

dwiel:
can we just use a mac-only context with a comment about what is preventing it from working on windows? It seems unnecessary to require that every command work on every platform before including here.

jcaw:
Added a comment with a TODO to audit it later

jcaw:
Just in case anyone is subscribed by email - I was testing my local copy with the hack. Ignore that last comment, it's still broken.

code/window_snap.py
knausj85: Oops, missed this. the setting's declared/defaulted and used as appropriate .py, in this case switcher.py.

Then the user can set it here as desired.
https://github.com/knausj85/knausj_talon/blob/master/settings.talon 

Goal being one file to branch for most settings.

knausj85: This can certainly wait for another day too, honestly

PR #251, opened 2020-10-03
CSV-based Talon lists

### LINKED ISSUE #239: lists/dicts in vocabulary in separate json files for easier merging of changes to vocabulary.py
As it stands, many refactors to `vocabulary.py` result in merge conflicts due to custom vocabulary defined. If these were instead read from json files, this would no longer be a problem.

jcaw:
First attempt at pulling Talon lists out into user-defined CSV files.

The system I ended up implementing is a 1:1 mapping between CSV files and Talon dicts. You call a function to tie a Talon list to a CSV file, and it's mirrored from then on. When the CSV is changed, it's reloaded into a list in the `user.user_settings` context (which is active globally). All CSVs are stored in a separate `"settings"` folder, so the user can be track them independently. The API is designed to be simple, calls look like this:

```
bind_list_to_csv(
    "user.vocabulary",
    "additional_words.csv",
    csv_headers=("Word(s)", "Spoken Form (If Different)"),
    default_values={"the csv": "will be created", "with these": "values", "if it does not": "exist"},
)
```

The default values are only used when initially creating the CSV.

In the CSV files, values are stored on the left. Keys are optional, and stored on the right (if left blank, the key will be the same as the value). Each CSV takes a pair of headers to explain the columns:

```
| Word(s) | Spoken Form (If Different) |
| ------- | -------------------------- |
| def     |                            |
| nmap    |                            |
| nmap    | enn map                    |
| nmap    | N map                      |
```

This is a generic system - you can register any talon list and bind it to a CSV file, all from Python (the CSVs are created on-demand).

I initially implemented the approach discussed in #239 (four separate files) but the wiring was a lot more complicated than this system (and was pretty vocab-specific). We could go back to that for vocab settings, but this is a lot simpler on the backend it would muddy the 1:1 mapping I have set up here. Let me know what you think.

Closes #239

### DIFF
--- a/code/user_settings.py
+++ b/code/user_settings.py
@@ -22,7 +22,6 @@ def _load_csv_dict(
     file_name: str, headers=Tuple[str, str], default: Dict[str, str] = {}
 ) -> Dict[str, str]:
     """Load a word mapping from a CSV file. If it doesn't exist, create it."""
-    # TODO: utf-8
     assert file_name.endswith(".csv")
     path = SETTINGS_DIR / file_name
 
--- a/code/user_settings.py
+++ b/code/user_settings.py
@@ -1,0 +1,149 @@
+from talon import Module, fs, Context
+import os
+import csv
+from pathlib import Path
+from typing import Dict, List, Tuple
+import threading
+
+
+# NOTE: This method requires this module to be one folder below the top-level
+#   knausj folder.
+SETTINGS_DIR = Path(__file__).parents[1] / "settings"
+
+if not SETTINGS_DIR.is_dir():
+    os.mkdir(SETTINGS_DIR)
+
+
+mod = Module()
+ctx = Context()
+
+
+def _load_csv_dict(
+    file_name: str, headers=Tuple[str, str], default: Dict[str, str] = {}
+) -> Dict[str, str]:
+    """Load a word mapping from a CSV file. If it doesn't exist, create it."""
+    # TODO: utf-8
+    assert file_name.endswith(".csv")
+    path = SETTINGS_DIR / file_name
+
+    # Create the file if it doesn't exist
+    if not SETTINGS_DIR.is_dir():
+        os.mkdir(SETTINGS_DIR)
+    if not path.is_file():
+        with open(path, "w", encoding="utf-8") as file:
+            writer = csv.writer(file)
+            writer.writerow(headers)
+            for key, value in default.items():
+                writer.writerow([key] if key == value else [value, key])
+
+    # Now read from disk
+    with open(path, "r", encoding="utf-8") as file:
+        rows = list(csv.reader(file))
+
+    mapping = {}
+    if len(rows) >= 2:
+        actual_headers = rows[0]
+        if not actual_headers == list(headers):
+            print(
+                f'"{file_name}": Malformed headers - {actual_headers}.'
+                + f" Should be {list(headers)}. Ignoring row."
+            )
+        for row in rows[1:]:
+            if len(row) == 0:
+                # Windows newlines are sometimes read as empty rows. :champagne:
+                continue
+            if len(row) == 1:
+                mapping[row[0]] = row[0]
+            else:
+                mapping[row[1]] = row[0]
+                if len(row) > 2:
+                    print(
+                        f'"{file_name}": More than two values in row: {row}.'
+                        + " Ignoring the extras."
+                    )
+    return mapping
+
+
+_mapped_lists = {}
+_settings_lock = threading.Lock()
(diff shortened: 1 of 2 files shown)


code/user_settings.py
```diff
@@ -0,0 +1,148 @@
...
+) -> Dict[str, str]:
+    """Load a word mapping from a CSV file. If it doesn't exist, create it."""
+    assert file_name.endswith(".csv")
+    path = SETTINGS_DIR / file_name
+
+    # Create the file if it doesn't exist
+    if not SETTINGS_DIR.is_dir():
+        os.mkdir(SETTINGS_DIR)
+    if not path.is_file():
+        with open(path, "w", encoding="utf-8") as file:
```
dwiel, line 32: Maybe dont write defaults to disk. Otherwise to remove a default it needs to be removed from the default list in the .py file and the csv file. This gets even more confusing with derived values which end up in multiple csv files.

code/user_settings.py
```diff
@@ -0,0 +1,148 @@
...
+def _load_csv_dict(
+    file_name: str, headers=Tuple[str, str], default: Dict[str, str] = {}
+) -> Dict[str, str]:
+    """Load a word mapping from a CSV file. If it doesn't exist, create it."""
+    assert file_name.endswith(".csv")
+    path = SETTINGS_DIR / file_name
+
+    # Create the file if it doesn't exist
+    if not SETTINGS_DIR.is_dir():
+        os.mkdir(SETTINGS_DIR)
```
fidgetingbits, line 30: since you do this check here, you probably don't need it globally at line 13?

code/user_settings.py
jcaw: At the moment this is by design, but I can change that. I didn't want upstream defaults to affect the user's CSV files, so the defaults are only used once, when the CSV is initially created. If the user wants to delete them, `knausj_talon` won't override those changes. This creates a clean split between user CSVs and upstream knausj - you should never see a merge conflict and `settings` can be tracked independently.

Not sure what you mean by derived values - each CSV represents one Talon list in the global (thus top-level) context.

code/user_settings.py
jcaw: I made this trigger-happy to protect against the user deleting the settings dir, but looking at it now you're right, it doesn't make sense. I'll remove this.

code/user_settings.py
dwiel: Ah I see what you mean, and I read the code wrong the first time around. Thanks for the clarification!

dwiel: You dont need the fs watch at the end of the file if you instead use `talon.resource.open` here instead of open.

jcaw:
Removed the redundant directory creation. Is this OK to be merged now?

fidgetingbits (approved):

fidgetingbits:
I haven't had a chance to test it but it looks fine

code/user_settings.py
```diff
@@ -0,0 +1,148 @@
...
+import os
+import csv
+from pathlib import Path
+from typing import Dict, List, Tuple
+import threading
+
+
+# NOTE: This method requires this module to be one folder below the top-level
+#   knausj folder.
+SETTINGS_DIR = Path(__file__).parents[1] / "settings"
```
rntz, line 11: why is this parents[1] instead of parents[2]?

parents[1] puts it inside the knausj_talon git repository, which seems undesirable.

On the other hand, parents[2] puts it as a sibling. I guess for people who are using knausj_talon as their user/ directory, parents[2] would put it directly in the .talon/ directory instead of in .talon/user/. is this likely to be a problem?

is there some way of getting our hands on the .talon/user/ directory directly? that seems like the right thing to do here.

rntz (commented):
Sorry for only getting to this now, I've been busy with work. I have basically two concerns:

1. What directory should we be put the settings in? Currently it's put inside the knausj_talon directory (which may or may not be the same as `~/.talon/user`), which is a bit annoying for people like me who try to keep as much of their customization as possible outside the knausj_talon repo. It seems to me we really want to put it in `~/.talon/user/settings/` or similar. To do this we probably want to use `actions.path.talon_user() / "settings"`.

2. Why is the order value,key instead of key,value? It seems more consistent to do key,value, especially if we're going to be using this generically for many lists.

code/user_settings.py
rntz: EDIT: `actions.path.talon_user()` produces the user directory

jcaw: (Note for anyone reading, this was addressed on Slack)

PR #253, opened 2020-10-07
Abbreviate

fidgetingbits:
more abbreviations i've he started using. as often as the case some might be a bit questionable: ex hypertext -> http, web -> www, etc


dwiel:
LGTM after merge conflicts fixed

fidgetingbits:
I don't get a merge conflict warning?

dwiel:
Ah you are right. Merge is ok. Rebase is not. Haven't looked into why

fidgetingbits:
I have never used rebase on github so I'm not sure what kind of a problem it would have. should probably read more into rebasing in general..

PR #255, opened 2020-10-07
bunch of small language fixes

### LINKED ISSUE #248: lang/c files needs to be updated to use the new code_functions and code_libraries lists
see comments in https://github.com/knausj85/knausj_talon/pull/240, specifically https://github.com/knausj85/knausj_talon/pull/240#discussion_r493987137

> This was probably supposed to be `c_functions`.
> 
> This is probably be out of scope for this, but I think this should probably be moved over to use the `code_functions` and `code_libraries` lists, as done for python and R. And maybe move to the new code_insert_function stuff, which I think can basically be copied from csharp while we're here, or open an item for it. This way, we benefit from the imgui and stuff in code.py.
> 
> https://github.com/knausj85/knausj_talon/blob/master/lang/csharp/csharp.py

fidgetingbits:
this includes a fix #248, a couple automatic white space changes, a typo in csharp, some additional python stuff for adding docstring documentation as well as type hinting. 

also changed the way the generic library command worked in programmin.talon is because it seems specific to r? now I make it possible in programming.talon to just to add from a list of available includes or libraries or whatever, which makes it a little bit harder to generically do because the terminology differs for different programing languages, so it may be something we don't want generic?

### DIFF
--- a/lang/c/c.py
+++ b/lang/c/c.py
@@ -263,7 +263,5 @@ class user_actions:
         actions.user.code_insert_function(result, None)
 
     def code_insert_library(text: str, selection: str):
-        actions.insert("#include <>")
-        actions.edit.left()
-        actions.clip.set_text(text + "{}".format(selection))
-        actions.edit.paste()
+        actions.user.paste("include <{}>".format(selection))
+
--- a/lang/c/c.py
+++ b/lang/c/c.py
@@ -1,14 +1,19 @@
-# XXX - reorder some dicts for human readability so we see functions and libraries
-# first
-
 from talon import Context, Module, actions, settings
 
 mod = Module()
 mod.setting(
-    "use_stdint_datatypes ", type=int, default=1, desc="beep",
+    "use_stdint_datatypes ",
+    type=int,
+    default=1,
+    desc="Use the stdint datatype naming in commands by default",
 )
 
 ctx = Context()
+ctx.matches = r"""
+mode: user.c
+mode: command
+and code.language: c
+"""
 
 ctx.lists["self.c_pointers"] = {
     "pointer": "*",
@@ -64,7 +69,7 @@ ctx.lists["self.c_types"] = {
     "float": "float",
 }
 
-ctx.lists["self.c_libraries"] = {
+ctx.lists["user.code_libraries"] = {
     "assert": "assert.h",
     "type": "ctype.h",
     "error": "errno.h",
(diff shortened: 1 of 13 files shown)


lang/python/python.py
```diff
@@ -23,6 +22,48 @@
     "update": "update",
 }
 
+"""a set of fields used in python dog strings that will follow the
```
dwiel, line 25: Dog strings

lang/programming.talon
```diff
@@ -44,9 +44,7 @@ toggle funk: user.code_toggle_functions()
 toggle library: user.code_toggle_libraries()
 funk <user.code_functions>:
     user.code_insert_function(code_functions, "")
-library <user.code_libraries>:
-    insert("library()")
-    key(left)
+(include|import|library) <user.code_libraries>:
```
dwiel, line 47: I agree, seems like this should be language specific. Include doesnt make sense in python, import doesnt in c, etc.

lang/python/python.talon
```diff
@@ -93,16 +93,32 @@ action(user.code_true): "True"
...
+    insert("{python_type_list}")
+dock {user.python_docstring_fields}:
+    insert("{python_docstring_fields}")
+    edit.left()
+dock type {user.python_type_list}:
+    user.insert_cursor(":type [|]: {python_type_list}")
+dock returns type {user.python_type_list}:
+    user.insert_cursor(":rtype [|]: {python_type_list}")
+
+dunder in it: "__init__"
```
dwiel, line 121: Why move these, just makes mergers harder

lang/programming.talon
fidgetingbits: i moved it to per file

lang/python/python.talon
fidgetingbits: good call. was just effect of my python file being way more full of cruft and and trying to keep in the original stuff, but I think it should be fix now.

dwiel: Cool thanks

lang/c/c.py
```diff
@@ -227,25 +210,60 @@ def stdint_signed(m) -> str:
...
+            )
+        )
+
+        actions.user.code_insert_function(result, None)
+
+    def code_insert_library(text: str, selection: str):
+        actions.insert("#include <>")
+        actions.edit.left()
+        actions.clip.set_text(text + "{}".format(selection))
+        actions.edit.paste()
```
knausj85, line 269: I'd suggest using user.paste, which preserves the clipboard.

```
actions.user.paste("include <{}>".format(selection)
```

PR #257, opened 2020-10-07
initial vim script support

fidgetingbits:


fidgetingbits:
i'll put this in for now, in case anybody else will use it but I'll certainly make more improvement next time I have to write some them script

PR #261, opened 2020-10-08
personal info convenience commands

fidgetingbits:
is something I use pretty heavily for convenience. my personal circumstances that a lot of the entries need to have multiple options, like multiple cities, addresses, emails, devices, etc. this is why I ended up going with the json array route that I did, as well as the optional gui pop up.

I remember @dwiel you said you had something similar like this but I haven't had a chance to look yet

### DIFF
--- a/misc/personal_info/personal_info.py
+++ b/misc/personal_info/personal_info.py
@@ -1,0 +1,135 @@
+import json
+import os
+import sys
+from typing import Set
+
+from talon import Context, Module, actions, fs, imgui, settings, ui
+
+mod = Module()
+mod.mode("personal_info")
+mod.setting(
+    "personal_info_auto_select",
+    type=int,
+    default=1,
+    desc="Auto select specified entry in json list, or none if 0",
+)
+mod.list("personal_info", desc="List of personal info populated by json file")
+ctx = Context()
+
+
+@mod.capture
+def personal_info(m) -> list:
+    """Returns a personal_info name"""
+
+
+@ctx.capture(rule="{user.personal_info}")
+def personal_info(m):
+    return m.personal_info
+
+
+# ctx.matches = r"""
+# mode: command
+# """
+
+main_screen = ui.main_screen()
+
+personal_info_list = []
+
+
+def close_personal_info():
+    gui.hide()
+    actions.mode.disable("user.personal_info")
+
+
+@imgui.open(y=0, x=main_screen.width / 2.6, software=False)
+def gui(gui: imgui.GUI):
+    global personal_info_list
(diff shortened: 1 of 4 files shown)


misc/personal_info/personal_info.py
```diff
@@ -0,0 +1,135 @@
...
+mod.list("personal_info", desc="List of personal info populated by json file")
+ctx = Context()
+
+
+@mod.capture
+def personal_info(m) -> list:
+    """Returns a personal_info name"""
+
+
+@ctx.capture(rule="{user.personal_info}")
```
knausj85, line 25: fyi you don't need these captures if it's just a straight mapping

you can use the list in .talon like so
https://github.com/knausj85/knausj_talon/blob/master/lang/snippets.talon#L3

i learned this recently - can save some time:)

knausj85: oh, that might not work here, nvm:)

dwiel:
OK with me, though honestly not sure the benefit of putting this upstream and also in a json file. A json file is just an extra level of indirection. A .talon file seems actually more clear and simple:
```
my full name: john doe
my phone number: 555-555-5555
my work number: 777-777-7777
my address: ...
```

knausj85:
Yeah, I think I agree with dwiel on this one. I've got something similar in a straight .talon file that lives on its own.

Is there an advantage I'm missing? You should be able to hop to it via the help two "help context personal" or whatever  :-)

fidgetingbits:
so I guess two things: 

1) putting it up stream in general, json or not, is mostly to help people that have no idea what they're doing and feel like adding talon functionality on their own is a barrier to entry, which seems increasingly common based on the slack discussions. I think it's easy for all of us to think the stuff trivial, but I guess to a lot of people it probably not

2) the reason I went with json verses the talon/python files was so that it would avoid merge conflict in the event that lists in a .py file got updated frequently (unlikely), and ya basically 80% of the entries on mine are multi-option, because my circumstance is a bit weird I have multiple phones, addresses, countries, phones, mics, etc, so having the lists is really useful compared to just one big talon file .

that said if this seems like overkill I'm happy not to push it. probably having "help context personal" for people would be better less confusing.

dwiel:
Maybe for people less confident, we could include an example personal.talon? My guess is aegis will want something pretty specific for people who dont edit talon files.

The multi option thing is also probably overkill for most people. I guess I do a similar thing, but I just use a talon file with names. My work email, my google email, etc.

fidgetingbits:
sounds good. we can just have a simple personal.talon file for people then

knausj85:
Closing this one then. 👍

PR #262, opened 2020-10-11
Fix "play/pause" media command 

jpaddison3:
I believe this to be platform-independent, and have tested that it works on mac and windows.


rntz:
Works for me on linux.

PR #264, opened 2020-10-13
Various Additions for Jetbrains

jcaw:
Various additions for Jetbrains (and some changes) I've pulled from @dwiel's config at his request. I don't have the Jetbrains RPC stuff set up, so it may be worth someone who uses Jetbrains testing these first.


stevegeek:
Will test them out

stevegeek:
@knausj85 @jcaw tested with the rubymine and all work well for me!

PR #269, opened 2020-10-20
rewrite code/ordinals.py, fixing off by one error

rntz:
ordinals.py had a bug where it was generating numbers that were off by one for ordinals more than “tenth”. This fixes that bug and simplifies the implementation.

### DIFF
--- a/code/ordinals.py
+++ b/code/ordinals.py
@@ -1,54 +1,4 @@
-import os
-import re
-import time
-from math import floor
-
 from talon import Context, Module, actions, app, ui
-
-ordinal_words = {}
-ordinal_ones = [
-    "first",
-    "second",
-    "third",
-    "fourth",
-    "fifth",
-    "sixth",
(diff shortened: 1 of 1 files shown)


dwiel:
Looks good to me. Weird how complex the previous implementation was, I wonder why

knausj85:
This is kinda a separate thing, but should we add an <ordinal_small> capture that's used for repeats? 

Say, 2nd -> 20th or something to prevent silliness like 196th that seems not very likely to me.

rntz:
Currently only ordinals up to ninety ninth are recognized, so "one ninety sixth" would not be recognized, but it shouldn't be too hard to make the limit tunable. I'd like to make it configurable via a Talon setting, but I'm not sure how to do that, since it involves running code in response to a setting change, which can in theory happen on any context change, although I don't imagine many folks will want to change the maximum ordinal contextually...

PR #274, opened 2020-10-27
fix edit.selected_text to return empty string if there is no selection

rntz:
edit.selected_text has a bug where if there is no selection it returns the current clipboard value. this fixes that bug, per discussion with aegis in slack (https://talonvoice.slack.com/archives/G9YTMSZ2T/p1603817479056200?thread_ts=1603815852.050700&cid=G9YTMSZ2T). I'm a bit worried that we might have some code that depends on this behavior though, so I'd appreciate people testing this out before merging.

### DIFF
--- a/code/edit.py
+++ b/code/edit.py
@@ -8,19 +8,12 @@ mod = Module()
 @ctx.action_class("edit")
 class edit_actions:
     def selected_text() -> str:
-        # try:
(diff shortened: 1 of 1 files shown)


dwiel:
Looks good to me, and I can't think of any cases where I would want selected_text to default to whatever is already in the clipboard

dwiel (approved):

PR #277, opened 2020-10-29
Add iterm as terminal

stevegeek:
iTerm2 is a terminal on mac.


knausj85:
Would you mind creating an entry in https://github.com/knausj85/knausj_talon/blob/master/code/application_matches.py for iterm and using that in the .talon file? The old approach is deprecated now, as I recall.

The `talon copy context` when iterm is focused will attempt to create the necessary code for application_matches.py and add it your clipboard. May require some light editing.

knausj85:
I had a moment to take care of this since I was iterm-curious. 

The terminal stuff had evolved a bit, so I did it in separately: https://github.com/knausj85/knausj_talon/commit/28ad1bb8ce3fc58b7c6e97ef67544e54489389dd

PR #279, opened 2020-11-04
Generic messaging, discord app support

epilanthanomai:
Generalize slack app commands into a generic messaging tag. Implement that tag for the Discord app.

This PR makes some small changes to the slack grammar where the old commands were very slack specific:
 * `previous workspace` -> `previous (workspace | server)`: alternate discord terminology
 * `next workspace` -> `next (workspace | server)`: alternate discord terminology
 * `(slack | lack) (find | search)` -> `go (find | search)`: remove slack name from generic commands
 * `mark all read` -> `mark (all | workspace | server) read`: use slack and discord terminology
 * `(slack | lack) upload` -> `upload file`: remove slack name from generic commands


misc/messaging.talon
```diff
@@ -0,0 +1,17 @@
...
+previous (workspace | server): user.messaging_workspace_previous()
+next (workspace | server): user.messaging_workspace_next()
+channel: user.messaging_open_channel_picker()
+channel <user.text>:
+    user.messaging_open_channel_picker()
+    insert(user.formatted_text(user.text, "ALL_LOWERCASE"))
+channel up: user.messaging_channel_previous()
+channel down: user.messaging_channel_next()
+([channel] unread last | gopreev): user.messaging_unread_previous()
+([channel] unread next | goneck): user.messaging_unread_next()
```
fidgetingbits, line 13: I'm not sure if we have some sort of standard already in other abstract command sets for go previous and go next, but I don't think we use either of these. It may be worth considering.

apps/discord/discord.talon
```diff
@@ -0,0 +1,20 @@
+app: discord
+-
+# Navigation: Channels
+[channel] mentions last: action(user.discord_mentions_last)
+[channel] mentions next: action(user.discord_mentions_next)
+oldest unread: action(user.discord_oldest_unread)
+
+# UI
+toggle pins: action(user.discord_toggle_pins)
```
fidgetingbits, line 9: I'm also not sure if we decided that we would abstract OS differences this way, because I know aegis doesn't like that it becomes harder to tell what commands actually do. But I guess if we are going to accept this, we should consider making an issue to convert the others... but I want to say we originally decided against this and favored duplication cross OS-specific set of actions? @knausj85 do you remember what the preference is, if we made one?

fidgetingbits (commented):
as I think over all this is good, and it will make sense to add this type of abstraction for teams and signal messengers as well, which I can plan to do since I wrote those ones originally.

misc/messaging.talon
epilanthanomai: You mean like `([channel] unread next | go next | goneck)`? I'm fine with that. I was starting with the existing slack command wording wherever I could, but certainly happy to change it up if y'all have a clearer vision for the direction to take it.

apps/discord/discord.talon
knausj85: Doesn't this need to be 

```
oldest unread:user.discord_oldest_unread()
```

to work? The syntax above is for implementing actions, or did I just miss a memo?

Other than that, there are two ways to support multiple OSes, @fidgetingbits 
- As emily's done here
- implementing the actions in .py

We haven't established any preference, and the repo is a mixed bag. The tradeoff with implementing things in .py: it's kinda out of reach for non-programmers. I find myself firmly on the fence, so I'm fine with either approach generally. Perhaps something to discuss elsewhere.

epilanthanomai: > Doesn't this need to be

Yikes, sure does! Last-moment edit I failed to test thoroughly. Thanks, I'll fix it right up.

epilanthanomai: Pushed a fix to the action calls.

fidgetingbits:
lgtm

knausj85:
Looks good to me too, thanks!

Issue #249, opened 2020-10-01
number parsing is slow, eg "one two three four five six seven eight nine"


rntz:
Passing of numbers with many spoken components is slow for me. I'd like to know whether anybody else is having this problem. I have tested this on a vanilla checkout of knausj_talon master on the latest beta (v0.1.2-12-gdebaf4e) with w2l gen2 on Ubuntu 20.04. Example tests: say "one two three four five six seven eight nine" or "one million two hundred and thirty four thousand five hundred and sixty seven". For me it takes a noticeable amount of time to process those, eg see the end of the post for a trace.

One thing that seems to fix it is editing text/numbers.talon to require a 'number' prefix:

```
not tag: user.mouse_grid_showing
-
number <number>: insert("{number}")
```

but it would be nice not to have to do this.

Example trace:
```
2020-10-01 23:10:33    IO 
2020-10-01 23:10:35    IO EMIT ['one', 'two', 'thre', 'four', 'five', 'six', 'seven', 'eight', 'nine']
2020-10-01 23:10:35    IO DECODING
detecting in viterbi toks: |||||||||||||||||||||||||||||ooonee|||||twooo|||thrree|||||ffouurrrr||ffiivvve||||ssixxxxx||seevvveen||eighttt|nnniiinnee|||||||||||||
832.144 |||||||||||||||||||||||||||||ooonee|||||twooo|||thrree|||||ffouurrrr||ffiivvve||||ssixxxxx||seevvveen||eighttt|nnniiinnee|||||||||||||
823.666 |||||||||||||||||||||||||||||ooonee|||||twooo|||thrree|||||ffouurrrr||ffiivvve||||ssixxxxx||seevvveen||eighttt|nnniiinnth|||||||||||||
  result: @one @two @thre @four @five @six @seven @eight @nine

2020-10-01 23:10:35    IO DECODED ['@one', '@two', '@thre', '@four', '@five', '@six', '@seven', '@eight', '@nine']
2020-10-01 23:10:37 WARNING [watchdog] "talon.vad.<lambda>" @2.0s
   23:       lib/python3.7/threading.py:890| 
   22:       lib/python3.7/threading.py:926| 
   21:       lib/python3.7/threading.py:870| 
   20:                    talon/cron.py:112| 
   19: ------------------------------------# cron thread
   18:                    talon/cron.py:77 | 
   17:          talon/scripting/rctx.py:200| 
   16: ------------------------------------# 'cron' main:call()
[...]

Issue #252, opened 2020-10-06
Leading zeros being stripped off number being dictated a digit at a time


nriley:
Dictating "zero three" produces "3".

PR #280, opened 2020-11-10
overhaul numbers.py to use a simpler grammar

rntz:
This is a complete rewrite of numbers.py in order to use a simpler grammar as suggested in #249 by lunixbochs, namely a sequence of numbery words optionally separated by "and":

```
<number> ::= {number_word}+ (and {number_word}+)*
```

In the process I've also removed the need to separate the digit parser from the number parser; the number parser handles concatenating digits itself. However, **the behavior of number parsing has changed a bit**. Some slightly strange number phrases like "one thousand one one" (previously 10011, now 1011) or "one hundred thousand and five thousand" (previously 105000, now 100005000) give different results. Take a look at the tests in numbers.py for more examples. If these differences are problematic, let me know what behaviors you need in practice and I can try to fix it.

On the other hand, some rather ordinary phrases are now parsed correctly, such as "ten sixty six" for 1066 or "twenty oh nine" for 2009.

Other changes:

- Implements `<digit_string>` and `<user.number_string>`, which return a string and avoid dropping leading zeros.

- Allows teens and tens in digit strings, eg. `<digits>` parses "three forty five six twelve" into 345612.

- `<number_signed>` allows a leading "minus" for negation, not just "negative". In my dialect I almost always say "minus 4" instead of "negative 4".

### DIFF
--- a/code/numbers.py
+++ b/code/numbers.py
@@ -13,8 +13,8 @@ digits_map = {n: i for i, n in enumerate(digits)}
 digits_map["oh"] = 0
 teens_map = {n: i + 11 for i, n in enumerate(teens)}
 tens_map = {n: 10 * (i + 1) for i, n in enumerate(tens)}
-scales_map = {"hundred": 100}
-scales_map.update({n: 10 ** (3 * (i+1)) for i, n in enumerate(scales[1:])})
+scales_map = {n: 10 ** (3 * (i+1)) for i, n in enumerate(scales[1:])}
+scales_map["hundred"] = 100
 
 numbers_map = digits_map.copy()
 numbers_map.update(teens_map)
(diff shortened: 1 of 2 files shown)


rntz:
this would hopefully fix #249 and make it easy to fix #252

knausj85:
I generally dictate numbers digit-by-digit, so this change is fine by me. Adding @dwiel too for any concerns.

knausj85 (approved):

fidgetingbits (approved):
based off a look and the sanity of the tests that are included I think this will be good for me too. I don't have time at the moment to merge it in and test, but I don't see any problems with it. I don't tend to use digits only but the examples given seem to match what I do when I imagine most people do anyways. the test cases that no longer work from the original don't sense to me either, so I don't think losing them matters

dwiel:
Looks good to me. I will likely end up modifying it anyway, as I forget which, but I think some of the teen numbers caused false positives, so I just get rid of them.

PR #281, opened 2020-11-12
Add user.symbol_keys to user.text capture

knausj85:
With Dragon and recent changes to the beta, commands such as

`say this is a test period` ->`this is a test period` 

whereas it used to be `this is a test.`

This change would allow similar behavior on both wav2letter or Dragon for special symbols, if we want it. 

Thoughts?

### DIFF
--- a/code/vocabulary.py
+++ b/code/vocabulary.py
@@ -79,7 +79,7 @@ def word(m) -> str:
 punctuation = set(".,-!?;:")
 
 
-@mod.capture(rule="(<user.vocabulary> | <phrase>)+")
+@mod.capture(rule="(<user.vocabulary> | <phrase> | <user.symbol_key>)+")
 def text(m) -> str:
     words = []
     for item in m:


rntz:
I like the idea in principle, but after trying it I think this may include some words that people would rather have as words than as punctuation. For example: plus, dollar, star, dot, bang, hash, caret (similar to "carrot"), pipe, etc. Also, some people customize symbol_keys to use short, common words for some pieces of punctuation, eg. "lack"/"rack"/"lace"/"race" for []{}, which wouldn't play well with this.

It might be better to have a separate list of punctuation words, although there would be a lot of overlap. On the plus side, we could reuse this punctuation word list for dictation mode.

knausj85:
Good points!

rntz:
I've made a branch that adds a list for punctuation: https://github.com/rntz/knausj_talon/tree/punctuation, see https://github.com/knausj85/knausj_talon/compare/master...rntz:punctuation?expand=1 for comparison with master.

I'm going to try it out and might turn it into a PR sometime in the next few days depending on how I like it. I definitely like having punctuation in `say`, but I'm not sure whether I like having it everywhere `user.text` is used.

knausj85:
Maybe say should have a separate capture.

Issue #271, opened 2020-10-24
Re-work file_manager.py


knausj85:
This is a bit of a monolithic, fragile mess.

A better design would might be:
(a) Simply declare the necessary actions & lists
(b) Implement the few actions that are reasonably generic (e.g., imgui stuff)
(c) Re-implement things per-application, similar to programming language support (e.g., https://github.com/knausj85/knausj_talon/blob/master/lang/csharp/csharp.py#L9)

PR #282, opened 2020-11-15
File manager rewrite

knausj85:
#271 

I don't know if anyone uses the file manger things other than me, but it was a rather cumbersome implementation. I've tested these changes on Mac and Windows, seems to work well.


rntz:
I don't use the file manager stuff, and I'm on linux, so I'm not sure I have much useful feedback to give here. Shall I tag in @dwiel to review instead?

knausj85:
I would be curious to learn what you guys use to navigate folders and such in terminal/file managers efficiently if there's an alternative?

rntz:
I just use a terminal. I use tab completion a lot, have a command "parent" for going up a directory, and my shell is set up so that I don't need to type "cd directory-name <enter>", just "directory-name <enter>" works. But I have to admit it's not the most efficient thing in the world. If I didn't keep most of the files I need open in my editor all the time, I think it would be more annoying – I use emacs, usually have a hundred plus buffers open, and emacs remembers what files I have open on shutdown and restores them on boot up. I'll probably take a look at trying to make file navigation more efficient eventually, maybe using this file manager code, but for now it works ok and I have more pressing needs.

dwiel:
I rarely use the finder app. For navigating directories in a terminal, I also use tab completion a lot. I also have saved directory names so I can quickly navigate to directories I commonly use: `cd {user.directories}: "cd {directories}"`.

fidgetingbits:
I generally only use the terminal (tab completion, fuzzy finding), or for vim I use NERDTree/Fern or fzf to fuzzy search for anything under the current directory I'm working in. I have some generic global code where I can say "path" in some key word that print out my commonly used paths. I use the path thing for all sorts of stuff, and also for the terminal (I have it to if I just press enter like rntz i can enter a dir) and if I end up getting a prompt from Qt or Gtk it's a little bit more annoying to navigate I can sti.l just say the path in roughly get places. That said, I really don't like the way my navigation experience is in general and it's something I want to revisit, mostly because when what I do use fails and I have to manually construct paths or navigate around, is quite awful. I've just been swamped with work, so I haven't really been doing anything with talon that actually takes any real time investment lately

knausj85:
Thanks for the insights. You folks might appreciate this, then.

This is how I use it with both finder and terminal:
image

Got an imgui for the folders and files, can say "follow 1" or "follow <some rudimentary spoken version>" to navigate folders and such. With small folders + nothing hidden, usually don't even need the imgui.

based on Mike Robert's idea, just trying to make it workable for more apps.

rntz:
shall we go ahead and merge this? Nobody seems to have any objections.

knausj85:
Seems reasonable.

PR #283, opened 2020-11-16
Additions to generic editor functionality

RodrigoChT:
These actions are very common in everyday use and they currently take
2 commands to perform.

 - "**(select/clear/copy/cut) this**": apply action to word on which the
                                   cursor is on (not in borders).

 - "**(copy/cut) line**": apply action to whole line.


dwiel:
I put my `cut line` and similar into `line_commands.talon`. Seems like that might be a better place

RodrigoChT:
@dwiel I didn't realize there was a `line_commands.talon` file, specially because the commands `select line` and `clear line` were already in `generic_editor.talon`.

The copy and cut commands in `line_commands.talon` seem to require the line number to work. I think it's useful to have a few **basic** line commands in `generic_editor.talon` to help people that don't delve too deeply into Talon. 

@knausj85 what do you think?

rntz:
I think "cut/select/etc *this*" should probably be renamed "cut/select/etc *word*". We tend to use *this* or *that* for commands that operate on the currently selected region.

I also think there might be some built in actions that do, or are supposed to do, some of these things? I think edit.delete_word exists, for example. Take a look through the results of `actions.list()` in the reply. We should use those in the command implementations where appropriate.

RodrigoChT:
@rntz, I agree that changing the commands from `(action) this` to `(action) word` is a good idea so I did so.

Thanks for letting me know that these type of functionality already existed somewhere else, I implemented the existing actions for the commands in `generic_editor.talon`. There was a minor problem for the `edit.delete_word` action in both windows and mac so I changed that. Also the command `edit.select_word` in mac sometimes selected the previous word even when the cursor was located after the first letter of the word. I made it so that it always selects the current word unless the cursor is before the first letter. I don't know how to make that "edge" case work.

knausj85:
The line between line_commands.talon and generic_editor is pretty blurry, but my thought was anything with robust support for number-based line selection should opt-in into line_commands, and anything that doesn't require a line # and works more-or-less globally should be in generic editor.

I think this change is fine as-is.

dwiel:
I know I'm late to the party, but I really think non line number specific line commands should still go in line_commands, this is how I have it in mine. One rationale is that I change my line command names, and now I have to do this same change basically in two places.

knausj85:
Yeah, it's a trade-off for sure. At the time, my primary thought was to avoid having too many commands in the global space that didn't work in most apps.

Maybe open a PR when you can so can better evaluate?

rntz:
I think it's important that cut/copy/clear line be available globally, and not only in editors that support "jump to line", which seems to be the current condition for line_commands to be enabled. The other alternative would be to enable line_commands globally. I think that might be okay too, worth trying.

PR #285, opened 2020-11-20
make a user.prose capture that supports punctuation

rntz:
This patch adds a <user.prose> capture that supports punctuation words like "comma" and "period" and uses it for the `say|speak` command. Things worth noting:

1. My eventual goal is for `<user.prose>` to act like dictation mode, handling auto-capitalization as well. However, for now it only handles punctuation.

2. I previously tried adding punctuation to `user.text`, but decided against it; `<user.text>` is unchanged by this PR. This is for a few reasons. First, sometimes it's handy to be able to speak text without punctuation. Second, adding punctuation to `<user.text>` without also adding auto-capitalization was breaking auto-capitalization in dictation mode (because it was getting parsed as part of the `user.text` capture instead of the dictation mode commands).

3. `phrase` and `say|speak` now **behave differently**. We have `phrase <user.text>` and `(say|speak) <user.prose>`. This is so that we have commands for both inserting with & without punctuation/auto-capitalization. However, this is just my personal preference; perfectly happy to change this if folks prefer otherwise, e.g. having all three act the same.

Things I'd like your feedback on:

1. Are you happy with `phrase` behaving differently than `say|speak`?

2. What punctuation words should we have? The PR makes a settings CSV file for them, like for additional vocabulary. The default list that I came up with is pretty bare-bones, happy to have more.

### DIFF
--- a/code/vocabulary.py
+++ b/code/vocabulary.py
@@ -1,9 +1,75 @@
 from talon import Context, Module, actions, grammar
 from .user_settings import bind_list_to_csv, bind_word_map_to_csv
 
+mod = Module()
+ctx = Context()
+
+mod.list("vocabulary", desc="additional vocabulary words")
+mod.list("punctuation", desc="words for inserting punctuation into text")
+
+
+@mod.capture(rule="{user.vocabulary}")
+def vocabulary(m) -> str:
+    return m.vocabulary
+
+
+@mod.capture(rule="({user.vocabulary} | <word>)")
+def word(m) -> str:
+    try:
+        return m.vocabulary
+    except AttributeError:
+        # TODO: if the word is both a regular word AND user.vocabulary, then in
+        # principle it may parse as <word> instead; we ought to pass it through
+        # mapping_vocabulary to be sure. But we should be doing that in
+        # user.text, below, too.
+        words = actions.dictate.replace_words(actions.dictate.parse_words(m.word))
+        assert len(words) == 1
+        return words[0]
+
+
+punctuation = set(".,-!?;:")
+
+
+@mod.capture(rule="({user.vocabulary} | {user.punctuation} | <phrase>)+")
+def text(m) -> str:
+    words = []
+    for item in m:
+        if isinstance(item, grammar.vm.Phrase):
+            words.extend(
+                actions.dictate.replace_words(actions.dictate.parse_words(item))
+            )
+        else:
+            words.extend(item.split(" "))
+
+    result = ""
+    for i, word in enumerate(words):
+        if i > 0 and word not in punctuation and words[i - 1][-1] not in ("/-("):
+            result += " "
+        result += word
+    return result
+
+
+
+# ---------- LISTS ----------
+# Default punctuation words.
+default_punctuation = {
+    "period": ".",
+    "comma": ",",
+    "colon": ":",
+    "semicolon": ";", "semi colon": ";",
+    "question mark": "?",
+    "exclamation mark": "!",
+}
+
+bind_list_to_csv(
+    "user.punctuation",
+    "punctuation.csv",
+    csv_headers=("Punctuation Mark", "Word or Phrase"),
+    default_values=default_punctuation,
+)
 
 # Default words that will need to be capitalized (particularly under w2l).
-_capitalize_defaults = [
+capitalize = [
     "I",
     "I'm",
     "I've",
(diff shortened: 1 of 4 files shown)


knausj85:
I think this model seems pretty reasonable on the surface:

`phrase/word` => always delivers the words
`say/speak` =>  lazy dictation

Other thoughts:
1. Perhaps the punctuation stuff should live in keys so we're not duplicating lists?
2. Alternatively, is there a compelling reason not to just use the self.symbol_key capture in prose? We could take the opportunity to remove the duplicates too.

code/vocabulary.py
```diff
@@ -38,92 +95,40 @@
...
-    default_values=_default_word_map,
+    default_values=default_word_map,
 )
 
-_default_vocabulary = {}
-_default_vocabulary.update({word: word for word in _simple_vocab_default})
-_default_vocabulary.update(_mapping_vocab_default)
+
+# Default words that should be added to Talon's vocabulary.
+simple_vocabulary = ["admin", "LCD", "VPN", "DNS", "USB", "FAQ", "PhD", "Minecraft"]
```
dwiel, line 116: Shouldnt change these defaults. Really not sure why they are in python at all instead of only in the csv files.

code/vocabulary.py
```diff
@@ -38,92 +95,40 @@
 ]
 
 # Default words that need to be remapped.
-_word_map_defaults = {
+default_word_map = {
     # E.g:
     # "cash": "cache",
+    # "centre": "center",
```
dwiel, line 101: Unnecessary change

code/vocabulary.py
```diff
@@ -1,9 +1,66 @@
...
+        "semicolon": ";", "semi colon": ";",
+        "question mark": "?",
+        "exclamation mark": "!",
+    }
+)
+
 
 # Default words that will need to be capitalized (particularly under w2l).
-_capitalize_defaults = [
+capitalize = [
```
dwiel, line 63: There are lots of variable name changes that aren't obviously an improvement and make this change harder to understand. Would be nice separate functional changes and stylistic changes, though not a show stopper.

code/vocabulary.py
```diff
@@ -1,9 +1,66 @@
 from talon import Context, Module, actions, grammar
 from .user_settings import bind_list_to_csv, bind_word_map_to_csv
 
+mod = Module()
```
dwiel, line 4: Reordering this file makes it harder to verify your claim that it doesnt change user.text, which is made extra difficult by our lack of ability to write unit tests. I do tend to prefer the reorganization that has been done, just I'd rather smaller, easier to manually verify changes.

dwiel:
I agree punctuation should probably go elsewhere to avoid duplication

rntz:
I would rather not unify symbol_keys and punctuation, as there are a lot of symbol key names that I would rather have as words than as punctuation when dictating: plus, dollar, star, dot, bang, hash, pipe, etc. I also personally use other common english words for symbol keys: lack `[`, rack `]`, lace `{`, race `}`, tick `'`, ditto `"`, stroke `/`, percy `%`.

What we could do is factor things so that punctuation words are included in symbol_key, but you can also have symbol key names that aren't active during dictation. This has a few downsides:

1. It means we can't use the csv file machinery (I think?), so users will have to customize keys.py to change the dictation mode punctuation words.

2. Because keys must be a single character, it won't cleanly handle multi-character punctuation marks, or characters that don't correspond to a key. For example, ‘--’ or ‘—’, or ‘ "’ vs ‘" ’ (which might be useful to distinguish between opening & closing quotes for prose formatting).

I don't think either of these are total blockers though (if we figure out how to do it better, we can always change it later), so if nobody objects or comes up with a better idea, I think I'll try to rewrite the PR this way. This week is rather busy for me, so I probably won't be able to get to it until the weekend or next week.

code/vocabulary.py
rntz: Good point, in future I will try to make the changes smaller and separate reorganizations from changes. Sorry about that.

code/vocabulary.py
rntz: I've reverted the variable name changes.

code/vocabulary.py
rntz: Reverted.

code/vocabulary.py
rntz: The defaults are only used when someone doesn't have a corresponding csv file, and they are used to create the initial csv file, so I think it might be sensible to have defaults that are more likely to be useful for most people. I've reverted these changes, however.

dwiel: that makes sense

code/vocabulary.py
dwiel: no worries. It sure would be nice if there is a standard way for us to integrate unit tests here ...

rntz:
Ok, I've refactored this PR so that the punctuation words live in keys.py and are included in symbol_keys. I have tried not to introduce any new symbol_key names, I've merely included some existing symbol_key names into punctuation, mostly the longer, unambiguous, more dictation-friendly ones. I would like to add some more for dictation purposes, but I'll do that in a separate PR if this is accepted.

Comments? @knausj85 @dwiel @fidgetingbits

knausj85:
Hmm, doesn't quite work reliably with Dragon. I'll have to poke around later.

Something like
`say this is a test comma so far it is working period` -> `this is a test, so far it is working.` 👍 

whereas 
`say so far comma it is working well period` -> `so far comma it is working well period` 👎 

*blinks slowly* Presumably because it's caught by phrase in the second case for reasons? Might be a Talon bug

rntz:
Hm, that's concerning. Does the same problem happen in dictation mode?

knausj85:
Yeah, same result in dictation mode.

knausj85:
For some reason, Dragon captures the entire thing via `<phrase>` sometimes. I guess we could manually map things in prose too. I'll see if I can convince aegis to give us a variant of parse_words with the old behavior, else we'll have to map it here I guess.

rntz:
I think the solution here is that we should take whatever user.prose recognizes and do a pass over it where we replace punctuation words with the equivalent punctuation. this means that we will replace punctuation words with punctuation whether or not they are recognized as commands. At that point we could even experiment with not binding them as commands, although that might make them less likely to be recognized.

I'm a bit confused by this behavior, though, since it seems it works the same way as additional vocabulary. Does the vocabulary from additional_words.csv also sometimes not get recognized by Dragon? And if not, what is the difference that's making that code work, while this code doesn't?

knausj85:
The difference: nothing in vocabulary overlaps with any existing Dragon silliness. "comma" and "period" are built-in Dragon symbol commands, so we're at the mercy of whatever Dragon decides to do -- and it's sometimes inconsistent.

Opened this issue after dicussing with aegis, which has more detail
https://github.com/talonvoice/talon/issues/165

rntz:
Ok, I think I have pushed a fix that should make this work on dragon, but I can't test it because I don't have dragon. can somebody who does have Dragon test this for me? @knausj85?

I'm also not sure whether this is the *right* way to do it, as it involves hooking into `registry.lists` in a way that aegis has suggested is not appropriate in a user script.

rntz:
It looks like recent changes to the talon beta mean that my fix to make this work with dragon isn't necessary on beta anymore, but I would still appreciate somebody testing it on public Talon with dragon.

knausj85:
It sounds like a new public release is pretty imminent, so I'd vote just to wait for that change.

knausj85:
Before I forget, I was thinking it might make sense to switch the stuff in comment.talon/block_comment.talon to use prose too. What do you guys think?

Also, thanks for your patience on this one.

PR #287, opened 2020-11-22
Clickless mouse

knausj85:
Might be useful to others. Quick and dirty initial pass, but seems to work well enough on Mac and Windows. 

Allows one to use a mouse without clicking. When the mouse stops moving, various options appear for some time; dwell to select left/right/scroll etc

image


knausj85:
Closing until I've time to work on this again.

Issue #290, opened 2020-11-30
Index out of bound in eclipse.py


daanvdh:
When using eclipse my logs are clogged with the errors below. You can reproduce it by opening quicksearch (ctrl-alt-shift-l) or another pop-up window. The window name then does not contain a dash. 

2020-11-30 21:02:41 ERROR cb error topic="scope_update:win" cb=<function scope at 0x0000000007F54DC0>
   30:                              threading.py:890| 
   29:                              threading.py:932| 
   28:                              threading.py:870| 
   27:                talon\windows\winevents.py:180| 
   26:               talon\scripting\dispatch.py:98 | 
   25:               talon\scripting\dispatch.py:133| 
   24:               talon\scripting\dispatch.py:124| 
   23:                   talon\scripting\rctx.py:200| 
   22: ---------------------------------------------# 'win_focus' main:_on_event()
   21:                       talon\windows\ui.py:86 | 
   20:               talon\scripting\dispatch.py:98 | 
   19:               talon\scripting\dispatch.py:133| 
   18:               talon\scripting\dispatch.py:124| 
   17:                   talon\scripting\rctx.py:200| 
   16: ---------------------------------------------# 'win_focus' win:update()
   15:                  talon\scripting\scope.py:28 | 
   14:               talon\scripting\dispatch.py:98 | 
   13:               talon\scripting\dispatch.py:133| 
   12:               talon\scripting\dispatch.py:126| 
   11:                   talon\scripting\rctx.py:200| 
   10: ---------------------------------------------# 'update' main:update_one()
    9:                  talon\scripting\scope.py:104| 
    8:                   talon\scripting\rctx.py:202| 
    7: ---------------------------------------------# 'scope_update:win' win:scope()
    6: ---------------------------------------------# stack splice
    5:                   talon\scripting\rctx.py:200| 
    4:                   talon\scripting\core.py:712| 
    3:                talon\scripting\actions.py:56 | 
    2:                  talon\scripting\types.py:313| 
    1: user\knausj_talon\apps\eclipse\eclipse.py:27 | result = title.split(" - ")[1]
IndexError: list index out of range

PR #291, opened 2020-11-30
Fix: #290

daanvdh:
I'm not sure yet what  the code I touched does,  my fix at least removes the errors for my logs.


knausj85:
Thanks. The code attempts to extract the active filename from the eclipse title. It's used to activate the correct language when one isn't forced.

At least on Windows, the title looks something like the below when a code file is open
`workspace - somefile.java`

daanvdh:
Thnx for the explanation. I'm also on windows and for the default editor there was no issue. But pop-up windows in Eclipse do not follow the format with dashes in between. Thnx for the quick merge.

PR #294, opened 2020-12-09
remove print statement

daanvdh:


daanvdh:
Every time you open a different window this print statements will print the name of that window about 70 times making the talon log unreadable.

PR #296, opened 2020-12-09
implement keyboard right click

daanvdh:


rntz:
Uh, could you explain the motivation for this?

rntz:
Oh, i just saw the conversation on slack: https://talonvoice.slack.com/archives/C9MHQ4AGP/p1607528069090500. Apparently shift-f10 acts like a right click at the text cursor location (instead of the mouse cursor location), at least on Windows and Linux in some applications. It also might have some relation to the menu key?

That seems useful but I'm not sure here is the right place for it. We should probably make sure it works on Mac, too.

daanvdh:
The keyboard  menu button and right-click as text cursor are the same thing, so maybe calling this menu instead makes sense. My argument for putting this in this file is that after importing this file I expect  a command to exist for al buttons of my keyboard. I expect this to work for Mac as well but we can also check before merging this.

rntz:
So I've tested this on my linux machine and it seems like shift-f10 and menu occasionally have different behavior, eg. in Emacs (where it behaves just like f10 unshifted) - not entirely unexpected since Emacs often has strange keybindings. Presumably we really want the menu key and not shift-f10?

I think the right place to put this is in `code/keys.py` as a `special_key`. Looks like `'menu'` is its name, on linux at least. So the remaining questions are:

- Confirm that `key(menu)` works on mac & windows.
- What voice command to give it? I vote for "menu key", since that's pretty unambiguous & users can always customize.

daanvdh:
I vote for "menu". I have trouble figuring out what happens after the keys are added to the context insidde keys.py. How does adding keys to the context end up in that I can for instance say "insert", is That a talon keyword? I found this on the menu key wiki: "Programmers using the Windows API can intercept this key by looking for a WM_KEYDOWN message with wParam VK_APPS (defined as 0x5D in winuser.h). It has key code 93 (VK_APPS 0x5D.)" Is this something we can use?

knausj85:
The "menu" key isn't a thing on OS X, so this might need to be Linux/windows specific in some fashion.

@daanvdh - The captures defined in keys.py return a string based on the Talon keyword that action.key(...) supports

e.g. simple_keys maps each entry to themselves. so saying "end" will provide "end"
```
simple_keys = [
    "end",
    ...
]
```

whereas symbol_key is mapped to the actual symbol (e.g., ctrl-shift-`) action.key(...) expects

```
ctx.lists["self.symbol_key"] = {
    "comma": ",",
```

so you could say "control shift comma end" -> `ctrl-shift-,-end`

daanvdh:
Closing this as it does not have any meaning for OS X.

knausj85:
Just in case it wasn't clear, I for one am fine with making platform-specific keys here via `app.platform` checks.

rntz:
Now #310 has been merged you can say "menu key" to press the menu key in Windows or Linux. If you want to customize this (eg. to just "menu") you can edit keys.py.

PR #302, opened 2020-12-21
add code navigation interface and Eclipse implementation

daanvdh:


text/line_commands.talon
```diff
@@ -1,4 +1,5 @@
 tag: user.line_commands
+tag: user.navigation_commands
```
rntz, line 2: We don't want to make all of line_commands depend on navigation_commands, since many commands in it work even without navigation. Could you split the commands that require navigation out into a separate file?

code/navigation.py
```diff
@@ -0,0 +1,24 @@
+# This file contains interfaces for navigation actions that can typically be executed within an editor. 
+# These interfaces can then be implemented per editor and called directly from .talon files by 
+# prefixing the method name with "user." and adding "tag: user.navigation_commands".
+
+from talon import ctrl, ui, Module, Context, actions, clip
+mod = Module()
+mod.tag(
+    "navigation_commands",
+    desc="Tag for enabling generic editor commands",
```
rntz, line 9: we probably want a more specific description, maybe something as simple as "Tag for enabling generic editor *cursor navigation* commands"?

code/navigation.py
```diff
@@ -0,0 +1,24 @@
...
+)
+#Do we need to at" user"To the annotation?
```
rntz, line 11: no, just `@mod.action_class` is fine for declaring stuff, it's only on `@ctx.action_class("user")` that you need the namespace :)

rntz (changes requested):
The line commands that also need navigation should be split out into a separate file rather than making all of line_commands depend on the `user.navigation_commands` tag, but otherwise this looks good to me.

text/line_commands.talon
daanvdh: I disagree, All commands inside this file are navigating somewhere. It's only that the "go line" Method is implemented in Talon by default. In my opinion we should group functionality in files and not let what it depends on determined grouping. 

With that reasoning I actually think this whole file should not exist, but that's another story.

code/navigation.py
daanvdh: copy paste mistake, fixed

code/navigation.py
daanvdh: Removed comment

text/line_commands.talon
daanvdh: I figured we don't need the module and tag at all, soo removed it. Looks cleaner now.

rntz (changes requested):
Looks good to me except the one out of date comment about the no longer existing navigation_commands tag.

rntz:
This adds the following commands to the generic_editor context: "go back" & "go forward" for cursor history movement, "go inside" for jumping inside a method/class/etc. I think this is ok but if anyone else wants to weigh in (since this is a pretty common/important context), now's the time, otherwise I'll merge this as soon as the comment is fixed. Opinions, @knausj85 @dwiel @fidgetingbits ?

knausj85:
I'm not sure "generic_editor" is the best place for these, as it will introduce a few commands that won't do anything in many applications. There's also a bit of overlap with generic_browser, command-wise, which may cause problems when a supported brwoser has focus.

Maybe an editor_navigation.talon or something that can be opted into as support is added? Just a half-baked thought.

```
tag: user.editor_navigation
-
go back: user.yadayada()
```

daanvdh:
Moving into another file I agree. But as far as I understand, adding a tag will result in log warnings, if a specific editor does not implement every method specified. I will be adding much more to this navigation interface, potentially methods that are not common for other editors. I don't want to pollute the log with necessary warnings. Is there any way around this?

apps/editor/navigation.talon
```diff
@@ -0,0 +1,6 @@
+
```
pokey, line 1: Need a tag here as [suggested](https://github.com/knausj85/knausj_talon/pull/302#issuecomment-752359463) by @knausj85

apps/eclipse/eclipse.talon
```diff
@@ -139,10 +139,10 @@ refactor this: key(alt-shift-i)
...
 #code navigation
 (go declaration | follow): key(f3)
-go back: key(alt-left)
-go forward:  key(alt-right) 
-# go implementation: 
-# go recent: 
+action(user.jump_back): key(alt-left)
+action(user.jump_forward): key(alt-right)
+action(user.go_inside): key(f3)
+action(user.go_bracket): key(ctrl-shift-p)
```
pokey, line 145: These need to be implemented in a python file as talon action definitinos are deprecated

apps/win/notepad++/notepad++.talon
```diff
@@ -16,4 +16,5 @@ action(edit.line_swap_up):
 action(edit.line_swap_down):
 	key(ctrl-shift-down)
 action(edit.indent_more): key(tab)
-action(edit.indent_less): key(shift-tab)
\ No newline at end of file
+action(edit.indent_less): key(shift-tab)
+action(user.go_bracket): key(ctrl-b)
```
pokey, line 20: These also need to go into a python file

code/navigation.py
```diff
@@ -0,0 +1,21 @@
+# This file contains interfaces for navigation actions that can typically be executed within an editor. 
+# These interfaces can then be implemented per editor and called directly from .talon files by 
+# prefixing the method name with "user.".
+
+from talon import ctrl, ui, Module, Context, actions, clip
+
+mod = Module()
```
pokey, line 7: tag would prob be defined here

apps/editor/navigation.talon
```diff
@@ -0,0 +1,6 @@
...
+-
+
+go back: user.jump_back()
+go forward: user.jump_forward()
+go inside: user.go_inside()
```
pokey, line 6: Add "go bracket" below here

pokey (commented):
This is looking really good; just a few minor tweaks in comments below, and looks like there are a couple merge conflicts to resolve

pokey:
> Moving into another file I agree. But as far as I understand, adding a tag will result in log warnings, if a specific editor does not implement every method specified. I will be adding much more to this navigation interface, potentially methods that are not common for other editors. I don't want to pollute the log with unnecessary warnings. Is there any way around this?

So it's actually fine if a specific editor doesn't implement every method specified; there will only be an error if user tries to say the given commands.  The reason a tag is important here is that otherwise these commands will be active everywhere, not just when you're in an editor where these actions make sense

rntz:
This PR has been inactive for basically a year, so I'm going to close it. If anyone wants to revisit it, make a command and we'll reopen.

PR #303, opened 2020-12-23
Fix slack commands not working

daanvdh:


knausj85:
Can you try with this fix instead? 
https://github.com/knausj85/knausj_talon/commit/fa809f3d6e8b752683b4d7f6411fbb762de403eb
This just lets us define "app: slack" in a single place

daanvdh:
It was fixed by your commit indeed. Thanks for fixing  it! Can you explain your commit? What are thee triple quotes for? Adding "Windows" sounds like that the statement is now valid for only Windows, but apparently that's not the case.

knausj85:
Triple quotes in python just allow a string to span multiple lines.

There are two ways to define application matches:
- directly in .talon files. Benefit: quick. Tradeoff: you have to maintain it potentially in multiple places
- python. Benefit: Define the match once, use everywhere (either in .py or .talon). Tradeoff: Lives separately from .talon, requries a bit more technical understanding.


The matches really work the same in either place
```
# linux specific apps
apps.keepass = """
os: linux
and app.name: KeePassX2
os: linux
and app.name: KeePassXC
os: linux
and app.name: KeepassX2
os: linux
and app.name: keepassx2
os: linux
and app.name: keepassxc
os: linux
and app.name: Keepassxc"""
```
^ each of the above is OR'd

```
apps.slack = "app.name: Slack"
apps.slack = """
os: windows
and app.name: slack.exe
"""
```
^ each .slack call is a separate entry as well

PR #304, opened 2020-12-24
Java language support

daanvdh:


lang/java/java.talon
```diff
@@ -0,0 +1,87 @@
...
+
+-
+tag(): user.code_operators
+tag(): user.code_comment
+tag(): user.code_generic
+
+action(user.code_operator_indirection): ""
+action(user.code_operator_address_of): ""
+action(user.code_operator_structure_deference): ""
+action(user.code_operator_lambda): ""
```
knausj85, line 13: for things that aren't supported. I'd recommend using `skip()`

lang/java/java.talon
```diff
@@ -0,0 +1,87 @@
...
+    insert('for () ')
+    key(left)
+    key(left)
+action(user.code_state_while): 
+    insert("while () ")
+    edit.left()
+	edit.left()
+action(user.code_type_class): "class "
+action(user.code_comment): "//"
+action(user.code_private_function):
```
knausj85, line 79: These `code_*_function` are usually implemented in .py. It isn't perfect for typed languages, though. See https://github.com/knausj85/knausj_talon/blob/master/lang/csharp/csharp.py

(Side note: this + some ability to extract snippets from eclipse would work great though. Or even just a hardcoded list. I haven't had time yet.)

lang/java/java.talon
daanvdh: Done.

lang/java/java.talon
daanvdh: As an initial Implementation I think this is fine. Maybe I'll extend this in future iterattions.

lang/java/java.talon
```diff
@@ -35,11 +33,8 @@ action(user.code_operator_less_than_or_equal_to): " <= "
 action(user.code_operator_and): " && "
 action(user.code_operator_or): " || "
 action(user.code_operator_bitwise_and): " & "
-action(user.code_operator_bitwise_and_equals): " &= "
```
knausj85: Is there a reason you removed the code_comment and the bitwise operators? I believe Java supports all of these?

lang/operators.talon
```diff
@@ -42,9 +42,7 @@ op mod equals: user.code_operator_modulo_assignment()
 
 #bitwise operators
 [op] bitwise and: user.code_operator_bitwise_and()
-(op | logical | bitwise) and equals: user.code_operator_bitwise_and_equals()
```
knausj85, line 45: Generally, our approach is to leave these commands here, even if not all languages support them. Not perfect.

daanvdh: This was dead code, "code_operator_bitwise_and_equals" Was never implemented anywhere.

lang/java/java.talon
daanvdh: It is supported Indeed. But I only want to add a small set of Java commands that are Widely used. I don't consider this widely used. This can always be added later.

knausj85:
This is a great start, thanks for your patience.

PR #306, opened 2020-12-25
dunst: use system cmds via dunstctl

bhipple:
The keybindings ctrl-space and ctrl-shift-space for notifications are not setup
by default on all window managers / desktop environments, so we should use
`dunstctl` for portability.

### DIFF
--- a/apps/linux/dunst.talon
+++ b/apps/linux/dunst.talon
@@ -2,8 +2,8 @@ os: linux
 -
 
 show notifications: key(ctrl-`)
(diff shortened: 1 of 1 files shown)


knausj85:
I'll have to defer this to one of the Linux gurus. @rntz or @fidgetingbits?

rntz:
Seems like a good idea, but I don't actually use dunst so I can't test this. I think fidget does though.

fidgetingbits (approved):

PR #308, opened 2020-12-26
Generic terminal

knausj85:
Initial generic terminal stuff. I find myself bouncing between various things for reasons (power shell, wsl, etc).

### DIFF
--- a/apps/eclipse/eclipse.py
+++ b/apps/eclipse/eclipse.py
@@ -32,7 +32,6 @@ class win_actions:
         if "." in result:
             return result
 
-        print(result)
         return ""
 
     def file_ext():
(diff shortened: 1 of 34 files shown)


knausj85:
This is based on some initial work @fidgetingbits did in his branch.

fidgetingbits (approved):
seems sane to me as a first pass..one other thing I had in the works before I took my break was a lot of predefined paths. my idea was to eventually pass them into certain commands like kate/lisa by name, as speaking paths is quite annoying. https://github.com/fidgetingbits/knausj_talon/blob/master/misc/paths.py for now I just use the path command I added.

I haven't actually tested your commits, but they look about the same as what I was using regularly cross windows and linux so I think it should be fine

PR #317, opened 2021-01-07
Change phones so that first word is used instead of longest word

admiralbumblebee:
Currently phones appears to be using the longest word from the row. This seems sub-optimal since it's possible to order homophones.csv so that the most popular/common word is first

Separate PR coming in a few minutes with homophones.csv re-ordered with most popular word first.


lunixbochs:
Note that you pushed both commits to this PR, so it should be sufficient to just merge this one.

beret:
Thank you!

PR #325, opened 2021-01-10
Added functionality for mac finder: preferences, options, focus search, sort by ..., icon/col/list/gallery view, copy path, trash, hide

ym-han:


ym-han:
Quick aside: `window_hide_others` in `app.talon` is currently using cmd-option-h, but that doesn't work for me. It does however work when it's cmd-alt-h. I don't know whether this is something idiosyncratic to my mac set up.

PR #328, opened 2021-01-12
add outlook.talon

daanvdh:


nriley:
This should be limited to Windows as Mac Outlook commands are different; you might also look at some of the other email app implementations (e.g. the Web version of Outlook) to try to use the same phrasing (e.g., "new message").  Not sure if it's worth abstracting email commands like generic_browser.talon?

Some email implementations I've done (feel free to steal/merge).

Windows Outlook: https://github.com/nriley/knausj_talon/blob/nriley/apps/win/outlook.talon

Mac Outlook (not "new Outlook" currently in beta): https://github.com/nriley/knausj_talon/blob/nriley/apps/mac/outlook.talon

Apple Mail: https://github.com/nriley/knausj_talon/blob/nriley/apps/mac/mail.talon

daanvdh:
I made it window specific. For me "new e-mail" sounds more logical than "new message", When I am in Outlook. I think it is worth abstracting this indeed. However, I will not Abstrract it In this PR, since I'm still waiting for feedback on the previous abstraction I did in: https://github.com/knausj85/knausj_talon/pull/302.

fidgetingbits:
I don't have outlook to test this, but it looks fine to me as is for the sake of having anything for outlook users in the interim, but agree that eventually abstraction should be added which I think make sense to implement for all mail clients.

PR #329, opened 2021-01-13
Use run-command-by-id in vscode

pokey:
I noticed that some commands in vscode can't be accessed via cmd+p. I created an extension that allows the user to type in the id of any command to run, and switched the talon vscode to use this functionality.


knausj85:
It looks like you've got a few extra, unrelated changes in this PR. :)

FWIW, these are the extensions we discussed today
https://github.com/brxck/commander
https://github.com/serenadeai/code

Brxck's a talon user too ( https://github.com/brxck/talon-config ), so there's probably an opportunity to collaborate on this.

pokey:
Oh haha whoops looks like I forgot to switch back to my master branch after filing this PR 😂. Had no idea I'd been pushing here the whole time

I'll close this PR tho, as I think we're both agreed that @brxck 's approach is better. Maybe it would make sense to switch to his approach at least feature-flagged to MacOS in the short term? Happy to work on a PR for that one

PR #332, opened 2021-01-14
add settings for command history size

rntz:
Making these settings makes them easier for users to change.


rntz:
I'm getting some weird errors that might have to do with this code, please don't merge it just yet.

rntz:
fixed it

PR #333, opened 2021-01-14
media.talon: avoid single word mute/play/pause commands

rntz:
We've had some instances of people activating `play` accidentally; aegis suggested not using single-word commands for them.


dwiel:
Doesnt `that` have a symbolic meaning in the talon grammar that doesnt apply here? What about `media play`?

rntz:
That's fair. I just used `play/pause that` because it's what I use myself. If we use `media play` instead we should probably rename `play next` to either `media play next` or `[media] play next` for consistency (ditto for `play previous`). And I guess `mute` becomes `volume mute`. 

Does that sound good? Alternative suggestions?

dwiel:
Yeah that all seems reasonable to me. Open to other ideas

fidgetingbits:
the media prefix versions seems fine to me as well. 

I suppose if you're interacting with media some people might be inclined to also expect media mute command rather than volume mute (I certainly make this type of mistake all the time what I'm in the middle of doing something else), so maybe we would want to consider having both, though I know there's generally a feeling of not wanting duplicate commands though, as if we have volume up and down commands then I suppose to adding a media mute only possibly adds confusion lol

rntz:
Ok, I've updated these to use a media or volume prefix instead.

daanvdh:
Looks good to me.

PR #337, opened 2021-01-21
Change backspace and delete keys

chtenb:
Whenever I say 'delete', the backspace key seems to be pressed. I find this unexpected since delete is a dedicated key and I would expect that to be pressed in almost every circumstance. For instance, when in the file explorer and wanting to delete a file, I have to say "forward delete" to delete the file, which is awkward. Even when text editing delete should remove the character in front of the cursor, not behind. 

And I know I can change this myself, but I'd like to stay close to the base to make it easier to merge upstream changes. Moreover, if I find this unexpected I'm sure many other people will run into the same thing. I believe this change would make talon more newbie-friendly.

### DIFF
--- a/code/keys.py
+++ b/code/keys.py
@@ -210,13 +210,11 @@ simple_keys = [
     "pageup",
     "space",
     "tab",
+    "delete",
+    "backspace",
 ]
 
-alternate_keys = {
-    "delete": "backspace",
-    "forward delete": "delete",
-    #'junk': 'backspace',
-}
+alternate_keys = { }
 # mac apparently doesn't have the menu key.
 if app.platform in ("windows", "linux"):
     alternate_keys["menu key"] = "menu"


rntz:
aegis had some comments in slack that I've copied over here:

> It's an awkward situation because the right answer is per platform
> I would vote against changing the binding for all platforms at once.
> I think we need a clear distinction between "the correct command to delete one letter" vs the key commands, and try to make the key commands os specific

lunixbochs:
another consideration is whether talon behaves the same across operating systems, which changing this would break as far as text goes

Chiel92:
I'm only familiar with linux and windows. Is delete a fundamentally different key on Mac?

lunixbochs:
https://store.storeimages.cdn-apple.com/4982/as-images.apple.com/is/MLA22LL?wid=2000&hei=2000&fmt=jpeg&qlt=95&op_usm=0.5%2C0.5&.v=1496944005839

Chiel92:
Ok that is awkward. For reference all my (and windows/linux users) keyboards look like this 
image

Chiel92:
It makes sense to make this mapping differ based on the OS then. Would you all agree? If yes I will try to modify the PR to reflect that.

daanvdh:
I agree

lunixbochs:
My biggest issue with an os specific change is if a user is using both Mac and Windows, and installs knausj on both:

On one computer, saying delete will delete a character to the left. On the other, saying delete will delete a character to the right.

Chiel92:
I agree that is suboptimal in that case. But it is a niche case and right now for windows and linux users the mapping makes no sense at all, which is arguably the biggest chunk of the users. Can you think of a solution that would solve all cases in a satisfactory way? What about:

- Map backspace always to backspace
- Map 'forward delete' always to delete
- Map 'delete' to delete on windows and linux, and to 'backspace' on Mac

This way all users get their native expected behavior, and crossplatform users could use the explicit backspace/forward delete combination.

lunixbochs:
I just checked and Talon has slightly more Mac users than both Windows and Linux users combined, FYI. 

This proposed change also changes the behavior of delete for existing Windows/Linux users who have been using it this way for a while. I'm still leaning strongly against this.

"muscle memory" is a big thing. I completely understand the motivation behind some new users wanting something different, but this is not a common complaint and it's really unfortunate if a few hundred existing Windows/Linux users `git pull` and a common command changes without warning. Even if there's a solid argument that the keys are in the wrong order for them, they've learned the wrong order and been dealing with it, so making the decision to change the mapping for existing users would be unfortunate.

I searched the Slack and only found two obvious complaints about the delete key order in the last three years, yours included. Someone also actually tried to make this change in 2019 with the talon_community repo, and there was a user complaint immediately:
> oof. who switched us to the Windows style of deletion? delete became backspace and forward delete became delete

For backwards compatibility, I would recommend adding a setting for this, which could cover more behavior than just delete, would be easy for any user to change, and would solve both backwards compatibility and consistency across operating systems.

Like `user.os_feel = "windows"`

Chiel92:
Backwards compatibility makes sense here I guess. An os_feel variable sounds like a reasonable option to me. I'm not very fluent with the talon API yet, so I will try to make some time in the near future to figure out how that would work.

daanvdh:
I would not be in favor of setting for this, I rather have MAC style by default. Users of this Repository have to become accustomed to changing keywords to their likings anyway. But I'm more in favor of making this OS specific, I value "lower learning curve" Over "backwards compatibiility". The steep learning curve will scare people off and not having a backspace is part of that, but existing users will simply change the key binding back (after complaining a bit).

knausj85:
I personally think there's an immense amount of value in consistent behavior across OSes. I do understand this perspective/frustration, though.

I personally use a combination of OS X and Windows. Once you develop some muscle memory, platform-specific behavior on common commands become a substantial liability, error-wise. I've put in some effort to keep things platform-agnostic for this reason.

That said, I'm not really opposed to a setting for this.

lunixbochs:
The benefit of a setting is you can set it to windows on your Mac too, if you went the other direction

Chiel92:
If the `os_feel` setting would not be set to the native OS by default, then we'd lose benefit of giving new windows/linux users a sane mapping by default.

lunixbochs:
knausj_talon can upgrade new users gradually without disrupting any existing users by doing two complementary updates at different times.

1. Automatically create a knausj_settings.talon file when knausj_talon runs, if it does not already exist. Start by putting os_feel = "mac" in it.
2. A few months later, update knausj_talon to create the file (for new users only, if the file does not already exist) using the current OS value instead.

This means as long as existing users update within the few month grace period, the behavior won't ever unexpectedly change. And new users will eventually get the current OS behavior out of the box.

Chiel92:
That sounds like a viable strategy. Are there any other mappings we need to consider for os_feel, apart from backspace/delete?

Chiel92:
@lunixbochs  I wanted to take a crack at this today, but I realized that this scenario is likely to occur more than once.
It would be really beneficial if we could generalize the update-scheme laid out above to work more than once. And for more plugins. Checking if a settings file exists will only work once, since for any future breaking tweaks it will already exist.
So maybe we shouldn't check if a file exists, but rather if a setting exists.

The general idea is then
1. Add an explicit setting that preserves the current behavior and ship.
2. Later, remove this explicit setting in a way that existing users will keep it but new users will not.

Some thoughts:
- When these settings are simply tracked in the git repository of the plugin it won't work, since a git pull to update the plugin will apply the breaking change for *all* users.
- When the setting created in step 1 is not tracked in the git repository of the plugin there needs to be an untracked file somewhere in which they are kept. Should this exist on a plugin level, or on a user level? The latter sounds better since you won't end up with untracked files all over your plugin infrastructure, and you keep all compatibility settings in once place which improves the visibility.

codecat555:
Some thoughts (FWIW):

1. When a user switches between systems, they have to adjust to differences in the platforms - like the delete key behavior or using forward-/back-slashes for paths. This is completely independent of any voice considerations. I'm not saying that talon shouldn't make things easier when possible, but this cognitive dissonance is already a familiar fact of life and it's not really talon's job to decide which variations should take precedence. That is, I think it's enough to provide people with a way to customize things but perhaps the default behavior should simply mirror the native style for each platform.
2. Two new settings might be called for, one for os feel and another for selecting between new and legacy behavior.
3. The legacy behavior can remain the default for as long as desired, as long as people have a way to switch away from it. However, it's probably less disruptive to make the change now than it would be later when the community has grown. That is, if 10% of the people are going to ask for help due to confusion over the change in behavior, then it's better to do it when there are 100 users than 1000 (or 10,000). Also, the current user community is likely more sophisticated and flexible than future users might be.

lunixbochs:
You only need one setting for this as long as you roll it out in the way I specified. Legacy is roughly mac feel.

Chiel92:
@lunixbochs Suppose we implement you roll out scheme and after a few months everyone has automatically defined an `os_feel` setting reflecting their original behavior. So far so good. Then some plugin or script decides to take `os_feel` into account for some OS-sensitive mappings. Suddenly a bunch of people who pull that script have breaking behavior, because the mapping they got used to are not the ones matching their defined `os_feel` setting. Do you feel what I'm getting at?

lunixbochs:
Call it knausj_os_feel if you're that worried, and document what it's for. There's no issue as long as only the knausj repo is using it and the folks maintaining it know what it's for.

lunixbochs:
Call it knausj_os_feel if you're that worried about third parties messing it up, and document what it's for. There's no issue as long as only the knausj repo is using it and the folks maintaining it know what it's for. I don't think we have enough cross repo dependency to worry about that right now anyway. It's fine to only guarantee how it affects knausj.

After the switch over, code shouldn't use the setting unless it uses the setting from the first time the code is written. That means it's fine to write new code at any point in knausj using os_feel, but it's incorrect to modify existing knausj code down the line to use the setting in a way that breaks user workflows.

bhipple:
As a Linux/Windows user who initially found this behavior confusing but has now gotten used to it, I'd vote we maintain the status quo.  It's nice to keep a consistent cross-platform experience, since many of us use one OS at work and a different OS at home.

astropuffin:
As a new user of talon who develops on both windows and Mac daily, I didn't find the mapping of `delete` to `backspace` on windows at all challenging. IIRC I did have to make a mental note when I was first learning that it wasn't directly mapped to the named key, but tbh I prefer this mapping. I quickly learned to use `clear right` when I want to delete the character to the right of the cursor. The `clear` ligatures are generally more useful/easier for me to pronounce clearly anyway. 100% agree commands should be platform agnostic as much as possible, even if that results in a few quirks. When I switch between windows/Mac keyboards, I actually use the same physical keyboard and press a button to change the key mapping. The physical movement of my fingers is _almost_ identical between OSs, with the vast majority of the difference being key combos that only exist on one OS.

knausj85:
I'm probably going to close this one since I generally prefer the commands to be platform-agnostic as much as is reasonable.

That said, I'm not entirely opposed to the proposed "choose you preference" approach, if someone really wants to take the time and re-open. It's not a priority for me, though.

PR #340, opened 2021-01-24
remove "software=" option from all calls to imgui.open

### LINKED ISSUE #335: Revert using software=True on linux
Now that https://github.com/talonvoice/talon/issues/190 is fixed, we should be able to use imgui.open(software=False) by default everywhere again. I haven't tested this on my box yet, though.

rntz:
Fixes #335.

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -472,7 +472,7 @@ def update_function_list_and_freeze():
     gui_functions.show()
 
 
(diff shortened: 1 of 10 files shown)


rntz:
already resolved by https://github.com/knausj85/knausj_talon/commit/93241369e3723f9ba2378d46956fa042708000e8

PR #343, opened 2021-02-01
Search tab by title for firefox

timo:


rntz:
This looks very useful! Do any other browsers have this feature? As far as I can tell chrome does not have it yet except on ChromeOS, but might soon: https://www.howtogeek.com/704212/how-to-search-open-tabs-on-google-chrome/.

1. It would be nice if the `<user.text>` could be made optional, so "tab search" on its own just brought up the tab search. That would require not hitting the `down` key afterwards, though, since that just selects the first match, eg add the following:

   ```
   tab search:
     browser.focus_address()
     insert("% ")
   ```

2. In fact, I'm not sure whether hitting the down key is a good idea at all here, since it means you can't add further text to the search without hitting up. (Also if there are no matches it ends up selecting a search engine, which is a bit weird although though not really a problem.) I guess it makes it easy to go to the first match, though, which is handy.

What do you think?

apps/firefox/firefox.win.talon
```diff
@@ -27,7 +27,12 @@ action(browser.bookmarks_bar):
 
 action(browser.focus_address): 
 	key(ctrl-l)
-	
+
+tab search <user.text>:
```
knausj85, line 31: for Firefox-specific commands, I'd recommend creating a new firefox.talon file that matches on
`app: firefox`
so we don't have to maintain the commands in multiple places

apps/firefox/firefox.win.talon
```diff
@@ -27,7 +27,12 @@ action(browser.bookmarks_bar):
...
+	browser.focus_address()
+	"% {text}"
```
knausj85, line 33: this is pretty nitpicky, but for multiline commands I prefer to use explicit `insert` calls

fidgetingbits (changes requested):
on my branch I actually implemented a tab search action in tabs.py and then implemented this that way:

```
action(user.tab_search):
	browser.focus_address()
    key(%)
```

I hadn't got around to merging into main line yet but I feel like it's probably better to go that route, since we already have tab commands?

I use this command a lot though I agree it's useful

rntz:
Bumping this, because it looks really useful and seems to have stalled — timo, are you planning on working on this PR? If not I can try and finish it up this weekend.

PR #345, opened 2021-02-01
Recognize jsx and tsx extensions

nbudin:
This adds simple support for the .jsx and .tsx extensions. There's no special support for react here, it just uses the javascript and typescript modes respectively.


rntz:
Looks good to me, at least until/if we grow some more specific solution for react.

PR #349, opened 2021-02-07
generic_editor.talon: use "all" consistently, add "clear all" command

rntz:
EDIT: PR now uses option (2), using *all* consistently.

I noticed that generic_editor was inconsistent between "select _all_", "copy _all_" vs "cut _everything_". This PR makes them all use "everything" for consistency (easier to memorise) and adds a "clear everything" command. I chose "everything" over "all" because `cut everything` and `clear everything` are slightly dangerous commands, and I figure "everything" is less likely to be accidentally recognized than "all". But I have no strong preference if folks prefer otherwise. Options:

1. Use 'everything' consistently: `select everything`, `copy everything`, `cut everything`, `clear everything`. 

2. Use 'all' consistently: `select all`, `copy all`, `cut all`, `clear all`.

3. Use `all|everything` for the common commands, `everything` for the dangerous ones: `select (all|everything)`, `copy (all|everything)`, `cut everything`, `clear everything`.

4. Leave it as is: `select all`, `copy all`, `cut everything`, `clear everything`.

Preferences?

### DIFF
--- a/text/generic_editor.talon
+++ b/text/generic_editor.talon
@@ -51,7 +51,7 @@ go page up:
 select line:
     edit.select_line()
 
-select all:
+select everything:
     edit.select_all()
 
 select left:
@@ -140,7 +140,7 @@ clear way down:
     edit.delete()
 
 #copy commands
-copy all:
+copy everything:
     edit.select_all()
     edit.copy()
 #to do: do we want these variants, seem to conflict
--- a/text/generic_editor.talon
+++ b/text/generic_editor.talon
@@ -139,6 +139,10 @@ clear way down:
     edit.extend_file_end()
     edit.delete()
 
+clear everything:
+    edit.select_all()
+    edit.delete()
+
 #copy commands
 copy everything:
     edit.select_all()
--- a/text/generic_editor.talon
+++ b/text/generic_editor.talon
@@ -51,7 +51,7 @@ go page up:
 select line:
     edit.select_line()
 
-select everything:
+select all:
     edit.select_all()
 
 select left:
@@ -139,12 +139,12 @@ clear way down:
     edit.extend_file_end()
     edit.delete()
 
-clear everything:
+clear all:
     edit.select_all()
     edit.delete()
 
 #copy commands
-copy everything:
+copy all:
     edit.select_all()
     edit.copy()
 #to do: do we want these variants, seem to conflict
@@ -178,7 +178,7 @@ copy line:
     edit.copy()
 
 #cut commands
-cut everything:
+cut all:
     edit.select_all()
     edit.cut()
 #to do: do we want these variants


nriley:
Are you seeing misrecognitions of 'all'? My preference would be 2, as the menu item is called "Select All" pretty consistently on Mac/Windows. It should all be undoable in any case, so doubt the actual destructive potential is that much.

rntz:
@nriley How do you feel about option (3)?

I'll try option (2) for a while on my computer and see if I get any accidental recognitions of "cut all" or "clear all".

nriley:
Sure, I would be fine with 3 too. To be honest, I'd never actually been aware of these other commands and would have done "select all copy that" prior to a few minutes ago…

knausj85:
I would vote to keep the aliases/synonyms to a minimum going forward. plus, doesn't having both all and everything kind of defeat the purpose?

I have been [slowly reworking my grammar](https://github.com/knausj85/knausj_talon/tree/knausj-personal-experiments) based  [on these principles](https://handsfreecoding.org/2018/09/04/utter-command-why-i-rewrote-my-entire-grammar/)

and I think it does make a substantive difference in terms of improving accuracy

mankoff:
Seems like **principles** is a good idea over whim, which is my current method for adding vocabulary. I was wondering why a recent change had the word `exchange` rather than `swap` or something. I know `knausj` is a personal repository, but at the same time I would be happy to see a review process or voting system or review system for new vocab...

knausj85:
for better or worse, it's not really a personal repository anymore :-p. if I've been aware of these principles back when I was a newbie, things would be different. 

it's difficult to agree upon specific words when what's accurate from person to person varies; worse, it's difficult to support both Dragon and wav2letter as their strengths and weaknesses re: recognition are quite different.

mankoff:
Do you use Dragon? I think it is Windows-only? Also, I have to assume that on the ~1 year timeframe (or less?) Talon or Talon + webspeech beats Dragon. These suggest supporting Dragon should not be a priority.

knausj85:
Yes I mostly use Dragon on both Mac and Windows. I feel the sconv models are getting competitive with Dragon's dictation quality, but I think dragons command recognition is still quite a bit better. After all, they've got about a thirty year head start. That said, I'll be glad to get rid of Dragon sometime in the future.

fidgetingbits:
I already use `all` for most stuff, and I haven't had any problems with it although I haven't used cut all before, so I don't know how dangerous it actually is. that said I'm good with 2 or 3. if 1, for the sake of reducing syllables ill almost definitely just change it back in mine anyway. I'd rather not have the inconsistency of 4, but I also know that generally the solution in 3 it is frowned upon in the repo.

rntz:
Most people seem to prefer "all", so I've switched the PR to use that. I'm still testing this locally. Haven't run into any accidental recognitions so far. Will merge this in the next week if nobody objects.

PR #352, opened 2021-02-10
Mouse grid improvements

lunixbochs:


lunixbochs:
please don't merge this without testing on all platforms

PR #353, opened 2021-02-10
Unify formatter history and dictation history

rntz:
Currently, formatters.py keeps track of a list of formatted things, and formatters.py and dictation mode each keep track of the last phrase emitted so that it can be undone (by "nope that" or "scratch that"). This unifies all of that into one phrase history. This PR is partly in preparation for my context-sensitive dictation stuff, which uses this code. But if we decide against this PR I can rewrite the context-sensitive dictation stuff not to use it.

Advantages:

- "scratch that" will work consistently across dictation & command mode, deleting the last thing. And if you use it again, it deletes the thing you said before that, etc.
- "select that" to select the last phrase will work in command mode.

Changes / possible disadvantages:

- The list of recent phrases will now include dictated text as well as formatted text. This might clutter it up.
- `format recent`, `format copy`, `format repeat` have become `recent list`, `recent copy`, `recent repeat`.

Please test this code before merging. I've been using this for a long time but I had to carefully extract it from other changes I made to knausj and while I did some testing I may have made a mistake.


knausj85:
we should probably make "scratch that" work with
`word <user.word>: insert(user.word)`
as well, then?

beyond that, I'm fine with this

dwiel (approved):

PR #356, opened 2021-02-15
Context sensitive dictation

rntz:
This implements context sensitive dictation, which determines how to space and capitalize dictated text by examining the text around the cursor before inserting.

I suggest reading this diff in the following order:

1. `modes/dictation_mode.talon`
2. `code/vocabulary.py`
3. `code/dictation.py`
4. `settings.talon`

The `word`, `text`, and `prose` captures got moved from vocabulary.py to dictation.py. This is because `text` & `prose` both need `format_text` and `prose` needs `auto_capitalize` and dictation needs `auto_capitalize`. We could split this by making something (eg. `auto_capitalize`) a talon action, but it seems like all this text-formatting code wants to be in the same place anyways. Open to other ways to organize this.

The capitalization code all became character based instead of word based, to avoid mangling whitespace. This makes it slightly more complex.

The PR leaves context sensitive dictation off by default, since, though useful, it is flaky and invasive; in some apps the copy-to-clipboard dance is unreliable or it doesn't work. I think it merits more testing before/if we decide to turn it on by default. Since it uses a setting, we can enable it on a per-application basis too.

It is possible to have a command that does context sensitive dictation in command mode, `dictate <user.prose>$: user.dictation_insert(prose)`. Should we add a command for this? I've been using `say` for it, but that is a big change in behavior from current knausj `say` (eg. if context-sensitivity is off, it will often make wrong guesses about capitalization & spacing; if context-sensitivity is on it does the clipboard dance a lot).

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -95,7 +95,7 @@ class DictationFormat:
     def __init__(self):
         self.reset()
 
-    def reset(self, before=None):
+    def reset(self):
         self.before = ""
         self.state = "sentence start"
 
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -1,102 +1,229 @@
-from talon import Module, Context, ui, actions
+# Descended from https://github.com/dwiel/talon_community/blob/master/misc/dictation.py
+from talon import Module, Context, ui, actions, clip, app, grammar
+from typing import Optional, Tuple, Literal
 
 mod = Module()
-ctx = Context()
-# Courtesy of https://github.com/dwiel/talon_community/blob/master/misc/dictation.py
-# Port for Talon's new api + wav2letter
 
-# dictionary of sentence ends. No space should appear before these.
-sentence_ends = {
-    ".": ".",
-    "?": "?",
-    "!": "!",
-    # these are mapped with names since passing "\n" didn't work for reasons
-    "new-paragraph": "\n\n",
-    "new-line": "\n",
-}
+setting_dictation_context_sensitive = mod.setting(
+    "dictation_context_sensitive",
+    type=bool,
+    default=False,
+    desc="Look at surrounding text to improve auto-capitalization/spacing in dictation mode. By default, this works by selecting that text & copying it to the clipboard, so it may be slow or fail in some applications.",
+)
 
-# dictionary of punctuation. no space before these.
-punctuation = {
-    ",": ",",
-    ":": ":",
-    ";": ";",
-    "-": "-",
-    "/": "/",
-    "-": "-",
-    ")": ")",
-}
+@mod.capture(rule="({user.vocabulary} | <word>)")
+def word(m) -> str:
+    """A single word, including user-defined vocabulary."""
+    try:
+        return m.vocabulary
+    except AttributeError:
+        return " ".join(actions.dictate.replace_words(actions.dictate.parse_words(m.word)))
 
-no_space_after_these = set("-/(")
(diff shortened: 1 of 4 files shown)


knausj85:
Yeah, I also hesitate to globally enable this. Are there any well-behaved applications that should opt-in by default?

I personally use say as a poor-mans dictation mode, so I'm not opposed to changing its behavior. Presumably we would respect the setting too.

knausj85 (approved):

rntz:
I think there are really two issues. One is application support and the other is clipboard flakiness. 

For application support, most text editors and web browsers will work fine, but with exceptions. Some web pages override default text editing behavior in strange ways, e.g. notion.so makes ctrl-c copy the surrounding block if nothing is selected, which screws us over; some text editors do a similar thing. Emacs doesn't work out-of-the-box because cancelling selections is different in emacs. Etc.

For flakiness, I know the clipboard is a bit flaky on Linux right now but I don't know what the state is on Mac/Windows, hopefully solid?

What I'd like to do is merge it disabled-by-default (which is the current default in the PR) and ask folks to test enabling it. Then we can selectively enable it by default in apps/contexts it works well in based on feedback. Once the draft window/text area stuff becomes more solid we can enable it there by default as well, I think that will be really nice.

knausj85:
Makes sense - sounds reasonable.

knausj85:
Oops, I haven't really encountered any issues with the clipboard stuff on Windows/Mac.

knausj85:
I'm thinking to merge this alongside the draft ui -- either as a submodule or directly -- once 0.1.5 lands. The combo works very well.

rntz:
I just had a video chat with someone on Mac where I had them try to use this in Gmail in a browser (not sure which one) on Mac, and it looked like they experienced some clipboard issue. Didn't have time to look into it but more reason to be cautious.

code/dictation.py
```diff
@@ -1,102 +1,229 @@
...
+            # this will go right over the newline. Argh.
+            actions.edit.right()
+        return text
 
-    def auto_format_resume():
-        """Resumes the autoformatter"""
-        return auto_formatter.pause(False)
+    def clobber_selection_if_exists():
+        """Deletes the currently selected text if it exists; otherwise does nothing."""
+        actions.key("space backspace")
```
knausj85, line 189: A lot of the flakiness I'm observing in various applications is not actually the clipboard, but the fact that talon presses keys so fast. 

peek left is basically equivalent to
`actions.key("space backspace shift-alt-left shift-alt-left")` on Mac. This reliably fails in applications like Messages.

Potential fixes:
1.  check whether there's a selection before clobber, which is slow and may fail under load
2. Add a delay here. Fragile.
3. Go nuclear: we could consider making this context sensitive dictation only available in the draft window.

knausj85: 4. Tinker with the talon's key_wait stuff per-app

code/dictation.py
```diff
@@ -1,102 +1,229 @@
...
-            if self.caps:
-                word = word.capitalize()
-
-            result += word
-            self.space = "\n" not in word and word[-1] not in no_space_after_these
-            self.caps = is_sentence_end
 
-        return result
+    def reset(self, before=None):
+        self.before = ""
```
rntz, line 99: Why is there a `before` parameter if it's not being used? Looks like there are no call sites that pass it in, so might as well remove the parameter.

knausj85:
@dwiel @fidgetingbits - last call for concerns before I merge this on Sunday evening

PR #357, opened 2021-02-15
1-aedison: bug fix for switching less than three characters

adam-edison:
Hi, my first bug fix for a public repo here. Let me know if you need me to do anything more.

### DIFF
--- a/code/switcher.py
+++ b/code/switcher.py
@@ -265,6 +265,12 @@ class Actions:
     def switcher_focus(name: str):
         """Focus a new application by  name"""
         app = actions.user.get_running_app(name)
+
+        # don't process silly things like "focus i"
+        if len(name) < 3:
+            print("switcher_focus skipped: len({}) < 3".format(name))
+            return
+
         app.focus()
 
         # Hacky solution to do this reliably on Mac.
--- a/code/switcher.py
+++ b/code/switcher.py
@@ -180,8 +180,7 @@ def update_lists():
 
         words = get_words(name)
         for word in words:
-            if word and word not in running:
-            # if word and word not in running and len(name) > 3:
+            if word and word not in running and len(name) > 3:
                 running[word.lower()] = cur_app.name
 
         running[name.lower()] = cur_app.name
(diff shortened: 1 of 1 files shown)


knausj85:
I believe this fix might be better around here
https://github.com/knausj85/knausj_talon/blob/a31c7f162dd17c51a789fb7cfea43ad7621567b8/code/switcher.py#L182
to prevent the short words from entering the dictionary in the first place?

adam-edison:
Agreed. I'll put the change there.

knausj85:
you may need to restart talent after changing this file if you're not on the current beta; that's probably why the commands are being recognized

adam-edison:
Even after restarting, the "focus" command is not working, and "running list" shows an empty list. And even after I switch back to my feature branch without this change, the same behavior occurs. Scratching my head...

knausj85:
does that include apps that are launched after talon?

adam-edison:
Just by performing that test, the problem disappeared. I think it's necessary to exit Talon Beta before switching branches.

adam-edison:
The fix in the place indicated does not seem to filter the "i" application...

image

code/switcher.py
```diff
@@ -180,7 +180,7 @@ def update_lists():
 
         words = get_words(name)
         for word in words:
-            if word and word not in running:
+            if word and word not in running and len(name) > 3:
```
knausj85, line 183: you need to check word, rather than name

knausj85: should probably maintain >= too

adam-edison: Wow, I feel embarrassed to have moved that fix so incorrectly. Sorry about that!

adam-edison: I would have tracked that down in my troubleshooting if I had separated out the "no commands are working" symptom from my previous efforts. Lessons learned.

adam-edison:
Working now.

image

adam-edison:
Should I squash this branch for you?

code/switcher.py
knausj85: no worries, this happens to everyone. thanks for taking a look at this

PR #360, opened 2021-02-18
Add a little help overlay to the mouse grid

timo:
it shows the default mouse commands like lefty, mentions you can
add modifiers, and allows to turn the help overlay off for the
future

image


splondike:
Looks good to me, might help reduce confusion for new users.

splondike:
I haven't actually seen any confusion around how this works in the Slack in recent months. Maybe it's working well enough for people without the help?

I'm just going to close this out I think. People have the same problem of learning commands for all the functionality.

PR #363, opened 2021-02-21
make firefox.talon file, move non-OS-specific stuff there

rntz:
This deduplicates a bunch of firefox stuff by having a non-operating-systems specific .talon file.

### DIFF
--- a/apps/firefox/firefox.linux.talon
+++ b/apps/firefox/firefox.linux.talon
@@ -2,14 +2,8 @@ os: linux
 app: Firefox
 app: firefox
 -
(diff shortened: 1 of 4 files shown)


apps/firefox/firefox.linux.talon
```diff
@@ -2,15 +2,9 @@ os: linux
 app: Firefox
 app: firefox
 -
-tag(): browser
-tag(): user.firefox
```
rntz, line 6: This tab appears to be used nowhere, so I removed it.

PR #374, opened 2021-03-07
Automatically recognize java file in editors

zhang-chris:


apps/jetbrains/jetbrains.py
```diff
@@ -153,8 +154,18 @@ def idea_grab(times: int):
 
 @ctx.action_class("win")
 class win_actions:
+    def filename():
+        title = actions.win.title()
+        result = re.search(r"[A-Za-z-_]+\.java", title)
```
knausj85, line 159: I didn't catch this when initially poking at Adam's branch, but we'll want to take a moment to make this generic, and not java-specific. This does leave a lot of potential filenames out as-is - anything with a number etc.

I opted to avoid regex in e.g. vscode.
https://github.com/knausj85/knausj_talon/blob/cf723e1cbe934c65f8ed88fb32160b3e8de75bdd/apps/vscode/vscode.py#L38
since it always displays things like so

```
extension.ts - command-server - Visual Studio Code
```

Does jetbrains do something similar? I haven't used it in some time.

zhang-chris: you're right this needs to be generic... my excuse is I haven't coded in a long time :)

on windows and mac for java intellij:
`untitled - Main.java [optional-module-name]` or `untitled - Main.java`
 so it can either be the last or second to last.

on windows for python pycharm (haven't tried on mac but I assume its similar):
`pythonPython - main.py`

I will change so it checks either of the last 2 elements.

How come in your implementation for vscode you didn't split by space instead and just grab the first element. Then you won't need specified mac or not right?

knausj85: It simply didn't occur to me to do it that way for VSCode at the time. Derp!

zhang-chris:
imo The default implementation should contain what I've written in jetbrains.py and neither jetbrains.py nor vscode.py  should need to implement  these functions,  but that is a bigger effort.  I'd be happy to help if I could get pointed to where to start and can create a new pull request for it

rntz:
It looks to me like we actually don't have a default implementation of win.filename or win.file_ext in knausj. To make your implementation the default, all that would need to change is that the context `ctx` would need to match anything rather than specifically jetbrains, eg:

```
emptyctx = Context()
@emptyctx.action_class("win")
class win_actions:
    def filename():
        title = actions.win.title()
        result = title.split(" ")

        for word in result:
            if "." in word:
                print(word)
                return word

        return ""

    def file_ext():
        return actions.win.filename().split(".")[-1]
```

The only question is whether we want this to be the default and if so where we would put it – obviously apps/jetbrains isn't the right place.

knausj85:
To be honest, we've played the default implementation game in the past, and I'm not eager to try again. 

Any assumption re: formatting/ordering is quite fragile and application/platform dependent, so I opted after many failures to go app-specific with no default implementation of filename or file_ext. In the early days, we had a fairly robust regex ( `re.compile("[^\\/]\.(\S*)\s*")`) for code.language... but it was prone to false positives.
 
Let's take a few examples of where the proposal will generate false positives:

```
# running a python file from the command prompt
Command Prompt - python  test.py

#viewing a python file in github in firefox
knausj_talon/abbreviate.py at master · knausj85/knausj_talon [github.com] — Mozilla Firefox
```

Both of these will activate the python commands, and one provides an incorrect filename. This isn't too big a deal for me personally, but for Talon newbies I think it's quite problematic.

Further, on Mac OS X, you run into stuff like `~/.talon/user` in the title for Finder windows.

zhang-chris:
If we can limit the scope to only apply for editor applications instead of any then we can avoid those false positive  situations. How difficult would this be?

knausj85:
Since tags are a thing now, I suppose we could consider using an `user.editor` tag on the default implementation. Might work ok with the regex.

knausj85:
I'm going to merge this one; opened https://github.com/knausj85/knausj_talon/issues/379 for more consideration.

zhang-chris:
Noticed an extra print, let me remove real quick.

Issue #367, opened 2021-02-27
re work vocabulary csv functionality


knausj85:
posting from slack:
> 
if you want to make "bind list to csv" less cursed, make a utility function that loads a csv and returns it as a dict, then write this instead:
ctx = Context()
ctx.lists['user.vocabulary'] = load_wordlist_csv('vocabulary.csv')

stop using a lock and punching lists into another file

if you resource.open or resource.read to load the csv, it will just reload your current file when the resource changes, so you don't need to implement a reloader. (edited)

knausj85:
> so I checked out what the current list loader does
> a lot of the logic is around creating the file if it doesn't exist yet
> so, that should be what the utility function does
> 1. load a dictionary from a file using resource.open()
> 2. create the file using the default if necessary
> 3. return the result
> 
> then to actually put it in a context, vocabulary.py can just do that
> 
> also unless there are multiple users of this function, it should probably just live in vocabulary.py?

rntz:
I am using bind_list_to_csv in places other than vocabulary.py in my own config. But maybe if it were simpler I could just duplicate the code and not worry about it.

rntz:
As long as we're rewriting this code we might as well fix #339 as well.

rntz:
Planning on taking a shot at this on my live stream on Sunday.

knausj85:
I've already got this locally, just need to merge.

knausj85:
Note to self: need #356 merged first.

PR #386, opened 2021-03-15
#367 Initial re-work of CSV file support.

knausj85:
@rntz Please test.


code/user_settings.py
```diff
@@ -13,37 +11,33 @@
...
-    path = SETTINGS_DIR / file_name
+def get_list_from_csv(
+    filename: str, headers: Tuple[str, str], default: Dict[str, str] = {}
+):
+    """Retrieves list from CSV"""
+    path = SETTINGS_DIR / filename
+    assert filename.endswith(".csv")
 
-    # Create the file if it doesn't exist
-    if not SETTINGS_DIR.is_dir():
```
rntz, line 28: Is there a reason this code to auto-create the settings directory got dropped? Seems like we'll need this for first-time users.

code/user_settings.py
```diff
@@ -13,37 +11,33 @@
...
+    # ability to reload this script for us when the resource changes
+    with resource.open(str(path), "r") as f:
+        rows = list(csv.reader(f))
 
+    # print(str(rows))
     mapping = {}
     if len(rows) >= 2:
         actual_headers = rows[0]
         if not actual_headers == list(headers):
             print(
```
rntz, line 45: Might want to use logging.warning here instead of just printing.

code/user_settings.py
```diff
@@ -56,95 +50,11 @@ def _load_csv_dict(
                 output, spoken_form = row[:2]
                 if len(row) > 2:
                     print(
```
rntz, line 58: Ditto re logging.warning.

rntz (commented):
Haven't actually tested yet, just a read-through.

code/user_settings.py
knausj85: It's there:
https://github.com/knausj85/knausj_talon/blob/50f55ef7abb829421cc494fae0a8bdeb3688f8df/code/user_settings.py#L11

code/user_settings.py
knausj85: if this is useful for you I would vote to write up an issue,  and we can address this more broadly.   I generally have very little output,  so haven't bothered

rntz:
did some basic testing, seemed to work, I should probably try testing on a clean install though

knausj85:
A clean install seems a bit overkill- simply deleting the settings folder while talon isn't running would probably suffice

PR #392, opened 2021-03-25
Add repeater command that makes more sense (for newbies anyway)

Chiel92:
Add a "n times" repeater command that is easier to pronounce for most people and makes more sense naturally. It is also more consistent with the `repeat that` command.


knausj85:
i think this is a good idea, though this might conflict a bit with the default operator command, `op times`. i will have to test it out a little bit when i have time.

splondike:
I've played around with this a little bit and it looks like it works great! Haven't had any conflicts. I might actually prefer it to the existing repeater system.

PR #393, opened 2021-03-25
Add git clean command

Chiel92:


misc/git.talon
```diff
@@ -15,6 +15,7 @@ git checkout master: "git checkout master\n"
 git checkout <user.text>: "git checkout {text}"
 git cherry pick: "git cherry-pick "
 git clone: "git clone "
+git clean everything: "git clean -dfx" # Leave \n out for confirmation since the operation is destructive
```
knausj85, line 18: I don't believe .talon supports in-line comments. This will cause an error on this line, and I think the command will be unavailable.

Chiel92: Oh my bad. I didn't test it after committing the comment and the syntax highlighting in VSCode suggested it was fine. I'll fix it.

PR #396, opened 2021-03-26
Add search engines

pokey:
Adds support for commands such as "google hunt hello world" to search for arbitrary text and "google this" to search for text currently selected by cursor.

Note: the specific search engines might not be generic enough?  For example I'm using Amazon Smile UK

Maybe should use a csv?

### DIFF
--- a/code/search_engines.py
+++ b/code/search_engines.py
@@ -1,0 +1,33 @@
+from .user_settings import get_list_from_csv
+from talon import Module, Context
+from urllib.parse import quote_plus
+import webbrowser
+
+mod = Module()
+mod.list(
+    "search_engine",
+    desc="A search engine.  Any instance of %s will be replaced by query text",
+)
+
+_search_engine_defaults = {
+    "amazon": "https://www.amazon.com/s/?field-keywords=%s",
+    "google": "https://www.google.com/search?q=%s",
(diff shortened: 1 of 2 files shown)


rntz:
Yeah, this seems like a good candidate for a settings/ CSV. We should probably default to Amazon.com, since we probably have more users in the US than any other country?

rntz:
Also, why "hunt" rather than "search"?

knausj85:
CSV seems like the best solution here. Best course is probably just to go ahead and merge #386 first, seems to work well on Windows and Mac.

pokey:
I thought "hunt" was standard; it's used [in `find_and_replace.talon`](https://github.com/knausj85/knausj_talon/blob/master/text/find_and_replace.talon#L3-L11) and "file hunt" is used in a few editors.  But I'm fine with "search" if you think that will be more intuitive for users.  

I'll refactor to use #386 and switch to amazon.com before merging; also lmk if you think I should switch to "search" per above or change any of the default names, eg "wiki" -> "wikipedia".  I tend to go terse but I know knausj style tends to be more verbose so happy to fit with conventions here

knausj85:
Let's stick with hunt for now. There was some reason I moved away from search. 

"wiki" is fine

PR #403, opened 2021-03-30
Anaconda virtual environment management

khornlund:
Added anaconda commands


apps/anaconda/anaconda.talon
```diff
@@ -0,0 +1,39 @@
+tag: terminal
```
fidgetingbits, line 1: ideally you you should add an `user.anaconda` tag here then it's up to the user to enable that tag in the terminal file, the users that aren't using this tool don't have all of the commands loaded every time a terminal is open. an example of this er reference would be https://github.com/knausj85/knausj_talon/blob/master/apps/kubectl/

the tag can be created in a `anaconda.py` file (see `kubectl.py`)

khornlund: Like this?

fidgetingbits: I think that looks fine, you'll just have to remember to actually set the tag in your `terminal.talon` wherever you want, but on the actual knausj repo it can stay disabled by default

PR #404, opened 2021-03-31
Do not redefine the `keys` symbol

EdJoJob:
The `keys` capture on line 102 and the `keys` dictionary on line 240
were being conflated, such that I could not use `key(escape)` in
a `.talon` file

### DIFF
--- a/code/keys.py
+++ b/code/keys.py
@@ -233,9 +233,9 @@ if app.platform in ("windows", "linux"):
     alternate_keys["menu key"] = "menu"
     alternate_keys["print screen"] = "printscr"
 
-keys = {k: k for k in simple_keys}
-keys.update(alternate_keys)
-ctx.lists["self.special_key"] = keys
+special_keys = {k: k for k in simple_keys}
+special_keys.update(alternate_keys)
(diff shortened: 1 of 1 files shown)


fidgetingbits:
I could be wrong, but I don't think they should actually conflict? the `@mod.capture` decorator should put it as member of the module class which should cause them to be in a different namespace I thought? if they were to conflict like that I think a lot of other things would break, because a lot of files end up having this type of problem iirc.

that said I guess you actually tested it and changing it fixed your problem? 

all that said I think we should merge this end move away from having these types of name conflict either way, because they also confuse python linters.

EdJoJob:
It was either the change, or just the re-interpretation of `keys.py` from the file watcher that reset it so that `escape` was a valid key again.

knausj85:
This change makes sense regardless, but we don't define the key action here. As such, I don't think this was the cause.

`actions.key` is owned by Talon proper. The only way to break this would be to override for the action, something like the below

```
@mod.action_class("main")
class MainActions:
    def key(s: str):
        s = s.replace("hyper-", "ctrl-cmd-alt-shift-"))
        actions.next(s)
```

that broke something.

Issue #397, opened 2021-03-27
csv refactor


knausj85:
- [ ] All CSVs should probably be in settings and utilize the new CSV loading stuff. e.g. switcher.py's overrides
- [ ] make a data directory for stuff like cursors, templated CSVs
- [ ] template all the default CSVs maybe. e.g. data/homophones.csv.default -> settings/homophones.csv when not present
- [ ] Implement some sort of backwards compatibility / deprecation plan for current users

knausj85:
- [ ] Evaluate the functionality from cursorless 
https://github.com/cursorless-dev/cursorless-talon/blob/main/src/csv_overrides.py

PR #414, opened 2021-04-10
#397 CSV refactor

knausj85:
#397
- Template all CSVs in /Data folder, rather than hard coding in python. These will get copied to /Settings if they don't already exist
- If CSVs exist in the legacy location, move them. Otherwise, create the CSVs from the template
- Move HiddenCursor into /Data folder
- The format and order of the CSVs unfortunately varies slightly. For back compat, parameterized the differences in get_list_from_csv() for now.
- Abbreviate.py now uses a CSV.


knausj85:
@pokey FYI, this is the CSV refactor.

pokey:
Cool. Are you just waiting on a review before you merge or is it still WIP?

knausj85:
It's ready for review from my perspective.

timo:
any reason in particular we don't put the alphabet in a csv file?

knausj85:
i am a little hesitant to change keys for backwards compatibility reasons. i really haven't thought about how to handle it though

code/homophones.py
```diff
@@ -22,17 +22,35 @@
...
 mod.list("homophones_canonicals", desc="list of words ")
 
-
 main_screen = ui.main_screen()
 
 
-def update_homophones(name, flags):
-    if name != homophones_file:
-        return
+def get_homophones_from_csv(filename: str):
```
splondike, line 28: Could this use get_list_from_csv as well? Superficially they look quite similar.

splondike (commented):
For the upgrade path, are we sure it's necessary? I cloned the master branch locally (repo A) then forked off that (repo B). I moved homophones.csv and app_names in A, then made changes to homophones.csv and app_names in my B. I then tried to rebase B on A. git forced me to commit my fork's modifications prior to doing this. After the rebase my changes were applied to the moved location (and the original one was gone).

So from that it seems like git should take care of this if that's what folks are using to manage their knausj.

code/homophones.py
knausj85: Homophone's logic is slightly different for parsing the CSV. It builds 
- a single list with only the first element of each homophone listed. No remapping.
- separately builds a dictionary of all the homophones -> alternates for selection-based shenanigans

The rest of the logic is, of course, identical. I was a bit lazy since I didn't see a quick way to consolidate these, and didn't immediately see any use cases for this outside of homophones. Thoughts?

knausj85:
> For the upgrade path, are we sure it's necessary? 

I'm happy to defer to you guys on this one. I was being cautious, particularly for those that may live on the gitless edge.

code/homophones.py
splondike: Ah right, homophones isn't a CSV with a fixed number of columns I see. Probably not worth merging them in that case.

splondike:
OK, so let's say that the git folks don't need it. How do the gitless upgrade their stuff? Here are a few options:

1. They have a copy of knausj with some untracked changes. They try to copy in new versions of files without using git in an ad-hoc fashion (likely to break things, but oh well).
2. They have a copy of knausj with some untracked changes. When they want to update they clone the latest knausj and then copy across the files they remember modifying (or re-apply their modifications to the new clone).
3. They treat knausj as a bunch of functionality they can extract from to build their own repository.

So without the upgrade path categories 1 would not have their modified lists picked up. Category 3 sounds like they're savvy and customised enough that they wouldn't be unduly impacted to me. Category 2 would likely go looking for the place they had originally put their CSV and find there wasn't a default one there anymore.

I suppose categories 1 and 2 might appreciate an upgrade path after thinking it through. Happy to keep it in. Maybe we would want to put an arbitrary date in a code comment so we could know that in 6 months or so we had decided it was OK to remove the legacy path handler or something.

fidgetingbits:
Sadly I still haven't got a chance actually try this, but I guess we may want to hold off/or reconsider until we know whether or not the new `.talon` lists will become available? 

One random thought about `abbreviate.py` becoming a csv is I feel like it is probably a more general purpose useful list for everybody, verse something like vocabularies which is a lot more personal, so people will sort of miss out by it not getting updated if we push more updates to the master repo. but maybe there are enough in there already that it's good enough. I have a quite a few more in my branch for instance, and I imagine a lot of them would be general purpose.

knausj85:
Yeah, I was thinking to wait for talon-lists too now

That’s a really good point re: abbreviate.py. I imagine people may branch still for recognition issues. In either case, .talon-lists would provide a cleaner, more consistent  mechanism for those needing to override.

knausj85:
Sorry I missed this week's session. It does probably make sense to reserurrect this PR.

knausj85:
I'm going to close this until we have a moment to take a peek at the similar CSV functionality in Cursorless
https://github.com/cursorless-dev/cursorless-talon/blob/main/src/csv_overrides.py
I do have the existing PR functionality up-to-date in my personal branch though.

PR #417, opened 2021-04-23
vscode: allow edit.find argument to be optional

rntz:


apps/vscode/vscode.py
```diff
@@ -48,13 +48,13 @@ def filename():
 
 @ctx.action_class("edit")
 class edit_actions:
-    def find(text: str):
+    def find(text = None):
```
pokey, line 51:
```suggestion
    def find(text: Optional[str] = None):
```

pokey (approved):
Looks good with minor comment

apps/vscode/vscode.py
rntz: I don't think the type signature is necessary here, as the type signature is already given when the action is declared. However, looking at actions.list() it looks like the type signature is incorrect:

```
edit.find(text: str = None)
  Open Find dialog, optionally searching for text
```

rntz: Filed an issue: https://github.com/talonvoice/talon/issues/295

pokey: Fair enough 🙂

Issue #419, opened 2021-04-24
Automatically enable 'control mouse' when 'wheel gaze' is enabled (and eye tracker is present)


codecat555:
When an eye tracker is present, invoking 'wheel gaze' should automatically enable 'control mouse' if it is not enabled already. Of course, it does raise the question of what should then happen when 'wheel stop' is subsequently invoked...prob just let things be rather than try to remember how 'control mouse' was enabled and then disable it automatically.

PR #420, opened 2021-04-25
Automatically enable 'control mouse' when 'wheel gaze' is enabled

codecat555:
(and eye tracker is present). (#419)

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -210,26 +210,22 @@ class Actions:
 
     def mouse_scroll_stop():
         """Stops scrolling"""
-        global control_mouse_forced
         stop_scroll()
-        if control_mouse_forced and config.control_mouse:
-            toggle_control(not config.control_mouse)
-            control_mouse_forced = False
 
     def mouse_gaze_scroll():
         """Starts gaze scroll"""
         global continuous_scoll_mode
-        global control_mouse_forced
         continuous_scoll_mode = "gaze scroll"
-
-        # enable 'control mouse' if eye tracker is present and not enabled already
-        if eye_mouse.tracker is not None and not config.control_mouse:
-            toggle_control(not config.control_mouse)
-            control_mouse_forced = True
 
         start_cursor_scrolling()
         if setting_mouse_hide_mouse_gui.get() == 0:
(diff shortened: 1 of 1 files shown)


knausj85:
leaving this enabled if it wasn't already enabled may not be ideal for those that use zoom mouse, so we may want to track the state after all?

codecat555:
I defer to your judgement on this. I can work on it if you think it's the right thing to do, but you don't sound completely sure. Are there any similar features already in the code, and do they behave that way?

Looks like that check would go into mouse_scroll_stop(), but what's the right mechanism to use for storing the state?

I guess we would also need to handle the case where the user toggles 'control mouse' off again before saying 'wheel stop' or clicking the stop button. Or, perhaps 'wheel stop' should be invoked automatically if 'control mouse' is toggled off...?

codecat555:
Should the state should be stored in the module context?

Something like this maybe?
```

ctx = Context()
ctx.settings['control_mouse_forced'] = True

```

codecat555:
After some study, I figured all I needed was a global variable. How does it look now?

I'm not sure why issue number 420 is showing up above now rather than 419. Has something to do with my latest commit, but I don't know what I did wrong...not familiar with github processes.

codecat555:
I found that sometimes wheel gaze is cancelled unexpectedly (at least, I didn't expect it). Seems to happen sometimes when my gaze is blocked and even once when not. After reviewing the code, I can't say what's happening. However, I have handled this case by moving part of my change into `stop_scroll()`...and that change also required a corresponding change in `mouse_gaze_scroll()`.

codecat555:
I added code to `stop_scroll()` to print the call stack , so I could catch the 'unexpected' invocations I mentioned previously.

Turns out `stop_scroll()` was being called from `on_pop()` - seems talon heard a pop that I hadn't meant to issue. I'm surprised this isn't included in the console debug output, seems like it should be.

Anyways, I have `user.mouse_enable_pop_stops_scroll` set to the default of 1, so `stop_scroll()` was called - mystery solved.

knausj85:
this looks good, thanks for taking the time

PR #422, opened 2021-04-25
formatters_reformat_selection: handle case that nothing was selected

rntz:


code/formatters.py
```diff
@@ -263,6 +263,9 @@ def formatters_reformat_last(formatters: str) -> str:
     def formatters_reformat_selection(formatters: str) -> str:
         """Reformats the current selection."""
         selected = edit.selected_text()
+        if not selected:
+            print("Asked to reformat selection, but nothing selected!")
```
pokey, line 267: We don't have a warning system to use for this kind of thing? If not then obv out of scope here but might be a useful thing to make at some point

pokey (approved):
Looks good with one question

code/formatters.py
rntz: there is Python's logging.warning() but @knausj85 seemed to prefer print()?

knausj85: I'm fine with logging.warning - my suggestion on another PR was to open an issue so we remember switch to logging.warning more broadly if its useful to folks. 

I previously hadn't really bothered since I run with zero output unless explicitly debugging something.

knausj85:
i am going to go ahead and merge this one as is, please open an issue if you guys feel we should move to using logging.warning etc broadly and we can get it done

PR #423, opened 2021-04-28
added tab restore shortcut

gpunktschmitz:
added shortcut to restore previously closed tab

### DIFF
--- a/apps/firefox/firefox.talon
+++ b/apps/firefox/firefox.talon
@@ -21,3 +21,5 @@ tab search <user.text>$:
   insert("% {text}")
   key(down)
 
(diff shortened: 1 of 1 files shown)


knausj85:
i believe this is actually already covered by this command in the generic tab commands?
https://github.com/knausj85/knausj_talon/blob/a366ab95077d7fd65a6dd4895ab70a25ee8bff68/misc/tabs.talon#L7

for reference, i generally try to stick to commands that are firefox-specific in this file.

gpunktschmitz:
my bad! thanks for looking into this and explaining!

Issue #381, opened 2021-03-12
Investigate github actions


knausj85:
Two initial priorities:
1. testing formatters
2. generating cheatsheets

PR #426, opened 2021-05-08
Implement proof of concept formatter tests

splondike:
This pull request implements a proof of concept for running unittests outside of the Talon environment (e.g. via github actions) while making sure we don't disturb things when running within Talon. You can see I've got a single test case on one of the formatters running and passing.

Some things to consider:

1. Is the built in unittest module fine or do we want to use pytest?
2. Are we ever likely to want to run tests under the actual Talon runtime? I'm not sure how to do that immediately, but I think it'd be some weird REPL thing currently. If we might want that, then we might consider how to do that a little bit here to make sure we're not going to make that difficult later.
3. Does this run fine on Windows and Mac?

If everybody thinks this looks like a good addition I'll add something to the README and merge it in. The next step down this journey (aside from writing more tests) would be setting up #381 and integrating with that.


dwiel:
pytest is much more mature and has many additional useful features. it
would also allow you to get rid of the run_test.py

otherwise excited for this to make it in to the repo!

splondike:
OK, I'll give pytest a try on the weekend.

Yeah, I don't have a compelling case for testing within the Talon runtime currently, I won't worry about it for the moment. I don't expect we'll have too many tests in this repo anyway.

splondike:
Okay so I have swapped it over to using pytest. You can run it using `pytest`.

There was a bit of a complication setting it up because pytest uses the built in Python code module (unlike unittest). This means that by the time my test code is running and wants to import from the knausj code module the other one is already registered. I worked around this using a symlink. See the `conftest.py` file for a bit of an explanation.

If this looks good to y'all now I'll add something to the README about how to run it and then merge it.

tests/stubs/README
```diff
@@ -0,0 +1,3 @@
+This directory contains stubs of Talon APIs. It is included via run_tests.py so we can run our test cases outside the Talon environment.
```
splondike, line 1: Faulty comment.

conftest.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Makes the code/ module available as knausj_code/ for use in tests. pytest uses
+    # the builtin python code module which prevents us from importing the knausj one.
+    # I've elected not to rename code as this would cause merge errors for forks that
+    # have modified it.
+    # This implementation just creates a simple symlink. An alternative might be to
+    # look at customising the python module loader:
+    # https://docs.python.org/3/reference/import.html
+    # I've had a go at relative imports without success.
+    symlink_src = os.path.join(curr_dir, "code")
```
splondike, line 28: From aegis:

```
just need a PathFinder in sys.meta_path I think
New
8:39
looks something like this:

class PluginFinder(PathFinder):
    @classmethod
    def find_spec(cls, fullname, path=None, target=None):
        return super().find_spec(fullname, path, target)
sys.meta_path.insert(0, PluginFinder)

(edited)
8:40
if you're passed a None path, you need to use the fullname to figure out the file path. otherwise, check if the path makes sense for a path you are willing to load from (edited) 
8:40
and find_spec can return None if you couldn't find anything
perhaps:

- from knausj_code import formatters
+ from user.knausj_talon.code import formatters

8:42
so then you take any fullname starting with user.knausj_talon. and rebase it to the root of your repo
```

splondike:
@lunixbochs Re-implemented the pytest import manipulations as an importlib loader. It's a little bit clunky with the root module loader, but seems to work OK. Ready for re-review.

lunixbochs:
should the draft window tests use pytest too?

splondike:
Yeah, probably should though pytest happily runs the unittest format as well. Sounds like nobody has objections to the shape of this PR so I'll do that README modification and merge it in on the weekend. Will probably swap the draft window tests to pytest also then.

After that I'm going to look into the formatters a bit more, I'm not sure they do the right thing with symbols. And after that maybe Github Actions so we can have a CI.

PR #428, opened 2021-05-13
updated unread shortcut

gpunktschmitz:
offical documentation still states 'ctrl-j' but does not work on linux (v4.15.0) ¯\_(ツ)_/¯


knausj85:
FWIW, ctrl-/ will display shortcuts in slack itself -hopefully more up-to-date than any web docs.

Issue #431, opened 2021-05-20
Enable pop-to-click during 'wheel gaze' (when setting_mouse_enable_pop_stops_scroll not active)


codecat555:
I'd like to use pop for clicking while using 'wheel gaze'. I have a change for that and am wondering if it might interfere with any use cases I haven't considered.

knausj85:
This was merged, closing

PR #432, opened 2021-05-20
Enable pop-to-click during 'wheel gaze'

codecat555:
I'd like to use pop for clicking while using 'wheel gaze'. Does this change interfere with any other use cases?


codecat555:
#431

PR #435, opened 2021-05-26
fix the order of the replacement example

xdrie:


rntz:
I don't think this is correct, although it depends on the behavior you want. Note that the order in the csv files is opposite that in Python, or a line `cache,cash` in words_to_replace.csv correspond with a dictionary entry `"cash": "cache"` in Python.

xdrie:
> I don't think this is correct, although it depends on the behavior you want. Note that the order in the csv files is opposite that in Python, meaning a line `cache,cash` in words_to_replace.csv correspond with a dictionary entry `"cash": "cache"` in Python.

Oh I had misunderstood that, thank you! I think it would be great if you noted that in the comment, for future viewers. Thanks!

PR #439, opened 2021-05-28
Update formatters.talon

astropuffin:
all the other help commands start with "help", so adding that option to make it easier for new users.


splondike:
Looks good to me.

PR #443, opened 2021-05-29
switcher menu

astropuffin:
DONOTMERGE - Only works on Windows currently.

Throwing my 2c into the ring on this cause I needed it locally. The problem is that it's challenging to switch between two windows with the same name (e.g. two instances of chrome with different tabs, or two instances of vscode with different repos open). This merely opens up the same `alt-tab` menu and let's you choose with your voice.

I would LOVE to get this working on mac, but I couldn't figure out how. I'm not sure what the equivalent for linux would be.

See also: https://github.com/knausj85/knausj_talon/pull/73 https://github.com/knausj85/knausj_talon/issues/373

### DIFF
--- a/code/switcher.py
+++ b/code/switcher.py
@@ -303,6 +303,13 @@ class Actions:
         else:
             ui.launch(path=path)
 
+    def switcher_menu():
+        """Open a menu of running apps to switch to"""
+        if app.platform == "windows":
+            actions.key("alt-ctrl-tab")
+        else:
+            print("Persistent Switcher Menu not supported on " + app.platform)
+
     def switcher_toggle_running():
         """Shows/hides all running applications"""
         if gui.showing:
--- a/misc/window_management.talon
+++ b/misc/window_management.talon
@@ -4,7 +4,7 @@ window last: app.window_previous()
 window close: app.window_close()
 focus <user.running_applications>: user.switcher_focus(running_applications)
 # following only works on windows. Can't figure out how to make it work for mac. No idea what the equivalent for linux would be.
-focus: user.switcher_menu()
+focus$: user.switcher_menu()
 running list: user.switcher_toggle_running()
 launch <user.launch_applications>: user.switcher_launch(launch_applications)
 
--- a/misc/window_management.talon
+++ b/misc/window_management.talon
@@ -3,6 +3,8 @@ window next: app.window_next()
 window last: app.window_previous()
 window close: app.window_close()
 focus <user.running_applications>: user.switcher_focus(running_applications)
+# following only works on windows. Can't figure out how to make it work for mac. No idea what the equivalent for linux would be.
+focus: user.switcher_menu()
 running list: user.switcher_toggle_running()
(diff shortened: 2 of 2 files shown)


splondike:
I'm a Linux tiling window manager person, so I don't have these issues :D.

What does the existing 'window next' command do for you in Windows? It seems possible that you could say "focus chrome" and then "window next" and it might swap to the next chrome window. This could be as fast as using the Windows built in switcher maybe?

Getting something like this working cross platform might be a bit of a pain because the window is normally hidden when you unpress the modifier I think. You could do a hack where you only press the key down in talon, but then we'd need a way to unpress that at the right time.

rntz:
i suggest end-anchoring this command, `focus$: ...`, to avoid collision with the regular `focus APPNAME` command.

astropuffin:
As far as I can tell, `window next` does nothing on windows. Adding $ was a great idea.

splondike:
Hmm OK. On Gnome 3 (probably the most common Linux desktop environment) it looks like the best analogue would be tapping the super key. This shows an exposé type view which lets you tab between windows on a virtual desktop. But you might as well just say 'super' to trigger that.

Since the exposé thing originated on OSX and I happen to have a work laptop with that on it I took a look there also. For some reason it looks like they haven't implemented any way to navigate around it using the keyboard. So we'd have to have a much more significant integration there (e.g. screen scraping or maybe the accessibility API).

I tried out my alt+tab hack on Gnome 3 as well via Talon. It sort of worked, but as expected left you with the alt key pressed which isn't really desirable. The switcher UI didn't seem to show up in the ui.apps() list either so we can't easily detect when it's gone to release the alt key. So I don't think this is viable.

As such, I'm not sure we can include a solution for OSes other than Windows. I'm happy enough to merge this as is since the voice command and action name you've chosen seem logical, folks can just override them for their particular setup if they like.

splondike:
@astropuffin I'm happy to merge this if you're ready?

knausj85:
@astropuffin I use https://neosmart.net/EasySwitch/ for the "window next" functionality on Windows. Its shortcut is wired up by default, for better or worse, in the repo.

I thought Expose on Mac used to be keyboard-friendly... really quite unfortunate. Though we could build a cross-platform version via Talon pretty easily I think. `expose <app>` -> imgui (quick and easy) or canvas with screencaps (more work) for apps with multiple windows.

splondike:
@astropuffin It sounds like the work to make this cross platform is not small. So are you interested in merging this just for windows? That sounds like a reasonable start to me; others could potentially make something for the other platforms at a future date.

astropuffin:
Sorry for the delay, life has been... a lot. It's pretty great as is for Windows without having to install anything. There has _got_ to be a way to make this work well enough for Mac. I'm ok either with merging as is or waiting until we can figure out cross platform (at least for Mac).

splondike:
No worries! I've merged this in since I don't think we're going to add the other platforms any time soon.

PR #444, opened 2021-05-30
Add new voice commands for inserting datestamps/timestamps

same-moon:
e.g. these voice commands can insert a datestamp (2021-05-30) or a timestamp (2021-05-30 16:31:39) or a high resolution timestamp (2021-05-30 16:38:50.394604) 

Originally I used words like 'date insert', but command mode ended up ignoring 'date' and then pressing the insert key on my keyboard some of the time, which put me in 'overwrite' mode in my editor, so I went to thesaurus.com and found 'fill' as a synonym for 'insert' and used that instead.


nriley:
I just use 'insert date' or 'insert time' (https://github.com/nriley/knausj_talon/blob/nriley/text/date_time.py). Unfortunately once you start supporting dates you run into many problems with localized date formats.

Not sure what the problem is with `date insert`; I just swapped the words around in mine and it worked, so perhaps we should figure out what is going wrong for you before replacing it with a less logical word.

same-moon:
@nriley I do think we should keep it as [object][verb] (e.g. date fill instead of fill date) as per the standard.

I'm open to changing 'fill' to 'insert'.  I tried it just now locally, and it didn't fail for me like it did before.  Of course, I'd want to change the filenames at the same time.

I agree there are many different formats people might like for dates.   I just went with the default python string conversions of datetime.date or datetime.datetime objects here, which felt to me like some reasonable default to get started.

lunixbochs:
At the action level you can expose strftime to let people customize it easily

In general I'm not sure I see the point in making "insert X" instead of "format X" actions, as "format X" gives you way more flexibility, and you can just call insert() on the talonscript side.

lunixbochs:
At the action level you can expose a `strftime` equivalent action to let people customize it easily from the .talon file side.

In general I'm not sure I see the point in making "insert X" instead of "format X" actions, as "format X" gives you way more flexibility, and you can just call insert() on the talonscript side. 

e.g.
```
def date_insert():
    s = str(datetime.now())
    actions.insert(s)
date fill: date_insert()
```

vs

```
def date_now() -> str:
    return str(datetime.now())
date fill: insert(date_now())
```

The latter provides much more flexibility.

same-moon:
@lunixbochs makes sense; I updated the PR, trying to follow your suggestion, let me know what you think

lunixbochs:
Here's what a strftime api might look like:
```
def time_format(fmt: str=None) -> str:
    now = datetime.now()
    if fmt is None:
        return now.isoformat()
    return now.strftime(fmt)

def time_format_utc(fmt: str=None) -> str:
    now = datetime.utcnow()
    if fmt is None:
        return now.isoformat()
    return now.strftime(fmt)
```

```
date insert:
    insert(user.time_format("%Y-%m-%d"))
date insert UTC:
    insert(user.time_format_utc("%Y-%m-%d"))
timestamp insert:
    insert(user.timestamp_now("%Y-%m-%d %H:%M:%S"))
timestamp insert high resolution:
    insert(user.time_format("%Y-%m-%d %H:%M:%S.%y"))
timestamp insert UTC:
    insert(user.timestamp_now_utc("%Y-%m-%d %H:%M:%S"))
timestamp insert UTC high resolution:
    insert(user.timestamp_now_utc_hires("%Y-%m-%d %H:%M:%S.%y"))
```

same-moon:
@lunixbochs looks good, I'll update the pull request in a moment, I'll just change your trailing '%y' in the high res commands to '%f' (the latter is "Microsecond as a decimal number, zero-padded on the left.").

lunixbochs:
Whoops I meant %f, brain error tabbing between docs and terminal :)

lunixbochs:
I think you can use a style like the following to document the fmt argument separately

```
    def time_format(fmt: str=None) -> str:
        """Return current local date or datetime formatted as a str using 'fmt', which
should be a .strftime() style format string; if 'fmt' is not
specified, iso format is used"""
```
->
```
    def time_format(fmt: str=None) -> str:
        """
        Return the current time, formatted.
        fmt: strftime()-style format string, defaults to ISO format.
        """
```

misc/datetimeinsert.py
```diff
@@ -0,0 +1,24 @@
+import datetime
+from talon import Context, actions, Module
+
+mod = Module()
+apps = mod.apps
```
fidgetingbits, line 5: this line it looks like you don't need

same-moon: @fidgetingbits good point, I updated the pull request

same-moon:
Anyone have any further comments?

lunixbochs:
wait, wrong line, didn't need either adjacent line (`apps = mod.apps` or `ctx = Context()`)

same-moon:
@lunixbochs good point, I removed another spurious line

PR #445, opened 2021-05-30
Fix command server

pokey:
This PR switches to file-based RPC for the VSCode command server to improve security, and enables fallback commands using the command palette so that no extension is strictly necessary

Works with https://github.com/pokey/command-server/pull/3 on the VSCode side

### DIFF
--- a/apps/vscode/command_client.py
+++ b/apps/vscode/command_client.py
@@ -1,126 +1,235 @@
-import requests
-import time
+import getpass
+from dataclasses import dataclass
 import json
-from typing import Any
-from talon import Module, actions, app
+import time
 from pathlib import Path
 from tempfile import gettempdir
+from typing import Any, List
+from uuid import uuid4
 
-is_mac = app.platform == "mac"
+from talon import Module, actions, Context
+
+
+STALE_TIMEOUT_MS = 10_000
 
 mod = Module()
+
+ctx = Context()
+mac_ctx = Context()
+
+ctx.matches = r"""
+app: vscode
+"""
+mac_ctx.matches = r"""
+os: mac
+app: vscode
+"""
 
 
 class NotSet:
     def __repr__(self):
         return "<argument not set>"
 
-def run_vscode_command_by_command_palette(command: str):
-    """Execute command via command palette. Preserves the clipboard."""
-    # Clip is noticeably faster than insert
-    if not is_mac:
-        actions.key("ctrl-shift-p")
-    else:
-        actions.key("cmd-shift-p")
 
-    actions.user.paste(f"{command}")
+def run_vscode_command_by_command_palette(command_id: str):
+    """Execute command via command palette. Preserves the clipboard."""
+    actions.user.command_palette()
+    actions.user.paste(command_id)
     actions.key("enter")
 
 
+def write_json_exclusive(path: Path, body: Any):
+    """Writes jsonified object to file, failing if the file already exists
+
+    Args:
+        path (Path): The path of the file to write
+        body (Any): The object to convert to json and write
+    """
+    with path.open("x") as out_file:
+        out_file.write(json.dumps(body))
+
+
+@dataclass
+class Request:
+    command_id: str
+    args: List[Any]
+    wait_for_finish: bool
+    return_command_output: bool
+    uuid: str
+
+    def to_dict(self):
+        return {
+            "commandId": self.command_id,
+            "args": self.args,
+            "waitForFinish": self.wait_for_finish,
+            "returnCommandOutput": self.return_command_output,
+            "uuid": self.uuid,
+        }
+
+
+def write_request(request: Request, path: Path):
+    """Converts the given request to json and writes it to the file, failing if
+    the file already exists unless it is stale in which case it replaces it
+
+    Args:
+        request (Request): The request to serialize
+        path (Path): The path to write to
+
+    Raises:
+        Exception: If another process has an active request file
+    """
+    try:
+        write_json_exclusive(path, request.to_dict())
+    except FileExistsError:
+        stats = path.stat()
+
+        modified_time_ms = stats.st_mtime_ns / 1e6
+        current_time_ms = time.time() * 1e3
+
+        if abs(modified_time_ms - current_time_ms) < STALE_TIMEOUT_MS:
+            raise Exception("Another process has an active request file")
+        else:
+            print("Removing stale request file")
+            path.unlink()
+            write_json_exclusive(path, request.to_dict())
+
+
 def run_vscode_command(
-    command: str,
+    command_id: str,
     *args: str,
     wait_for_finish: bool = False,
-    expect_response: bool = False,
+    return_command_output: bool = False,
     decode_json_arguments: bool = False,
 ):
-    """Execute command via vscode command server."""
+    """Runs a VSCode command, using command server if available
+
+    Args:
+        command (str): The ID of the VSCode command to run
+        wait_for_finish (bool, optional): Whether to wait for the command to finish before returning. Defaults to False.
+        return_command_output (bool, optional): Whether to return the output of the command. Defaults to False.
+        decode_json_arguments (bool, optional): Whether to decode JSON arguments. Defaults to False.
+
+    Raises:
+        Exception: If there is an issue with the file-based communication
+
+    Returns:
+        Object: The response from the command, if requested.
+    """
     # NB: This is a hack to work around the fact that talon doesn't support
     # variable argument lists
-    args = list(
-        filter(
-            lambda x: x is not NotSet,
-            args,
-        )
-    )
+    args = [x for x in args if x is not NotSet]
+
     if decode_json_arguments:
         args = [json.loads(arg) for arg in args]
 
-    port_file_path = Path(gettempdir()) / "vscode-port"
+    username = getpass.getuser()
 
-    if not port_file_path.exists():
-        if args or expect_response or decode_json_arguments:
+    communication_dir_path = Path(gettempdir()) / f"vscode-command-server-{username}"
+
+    if not communication_dir_path.exists():
+        if args or return_command_output or decode_json_arguments:
             raise Exception("Must use command-server extension for advanced commands")
         print("Port file not found; using command palette")
-        run_vscode_command_by_command_palette(command)
+        run_vscode_command_by_command_palette(command_id)
         return
 
-    original_contents = port_file_path.read_text()
+    request_path = communication_dir_path / "request.json"
+    response_path = communication_dir_path / "response.json"
+
+    uuid = str(uuid4())
+
+    unlink_if_exists(response_path)
+
+    request = Request(
+        command_id=command_id,
+        args=args,
+        wait_for_finish=wait_for_finish,
+        return_command_output=return_command_output,
+        uuid=uuid,
+    )
+
+    write_request(
+        request,
+        request_path,
+    )
 
     # Issue command to VSCode telling it to update the port file.  Because only
     # the active VSCode instance will accept keypresses, we can be sure that
     # the active VSCode instance will be the one to write the port.
-    if is_mac:
-        actions.key("cmd-shift-alt-p")
-    else:
-        actions.key("ctrl-shift-alt-p")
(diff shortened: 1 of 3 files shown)


apps/vscode/command_client.py
```diff
@@ -0,0 +1,188 @@
...
+        args = [json.loads(arg) for arg in args]
+
+    port_file_path = Path(gettempdir()) / "vscode-port"
+
+    if not port_file_path.exists():
+        if args or wait_for_finish or expect_response or decode_json_arguments:
+            raise Exception("Must use command-server extension for advanced commands")
+        print("Port file not found; using command palette")
+        run_vscode_command_by_command_palette(command)
+        return
```
pokey, line 56: Here's where we fallback to command palette

apps/vscode/command_client.py
```diff
@@ -0,0 +1,188 @@
...
+    nonce = decoded_contents["nonce"]
+
+    response = requests.post(
+        f"http://localhost:{port}/execute-command",
+        json={
+            "commandId": command,
+            "args": args,
+            "waitForFinish": wait_for_finish,
+            "expectResponse": expect_response,
+            "nonce": nonce,
```
pokey, line 100: Here's where we add the nonce

apps/vscode/command_client.py
```diff
@@ -0,0 +1,188 @@
...
+
+    if expect_response:
+        return response.json()
+
+
+@mod.action_class
+class Actions:
+    def vscode(
+        command: str,
+        arg1: Any = NotSet,
```
astropuffin, line 124: Why not use `*args`?

pokey: Not supported by talon

lunixbochs (changes requested):
Here's an overall review, though I made a suggestion on Slack that should be considered before moving forward with this approach.

apps/vscode/command_client.py
```diff
@@ -0,0 +1,196 @@
...
+    def __repr__(self):
+        return "<argument not set>"
+
+def run_vscode_command_by_command_palette(command: str):
+    """Execute command via command palette. Preserves the clipboard."""
+    # Clip is noticeably faster than insert
+    if not is_mac:
+        actions.key("ctrl-shift-p")
+    else:
+        actions.key("cmd-shift-p")
```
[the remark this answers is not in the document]
pokey: Not sure I follow; can you elaborate?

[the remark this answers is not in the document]
lunixbochs: Instead of hardcoding a keyboard shortcut in the middle of your code, prefer using an action like "command_palette()" that is overridden on Mac (instead of internally switching based on app.platform).

apps/vscode/command_client.py
```diff
@@ -0,0 +1,196 @@
...
+    # the active VSCode instance will be the one to write the port.
+    if is_mac:
+        actions.key("cmd-shift-alt-p")
+    else:
+        actions.key("ctrl-shift-alt-p")
+
+    # Wait for the VSCode instance to update the port file.  This generally
+    # happens within the first millisecond, but we give it 3 seconds just in
+    # case.
+    start_time = time.monotonic()
```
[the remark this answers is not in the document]
lunixbochs: Minor update, actions.sleep() will be much more precise than time.sleep() on all platforms now.

apps/vscode/command_client.py
```diff
@@ -0,0 +1,196 @@
...
+            command,
+            arg1,
+            arg2,
+            arg3,
+            arg4,
+            arg5,
+            wait_for_finish=True,
+        )
+
+    def vscode_json_and_wait(
```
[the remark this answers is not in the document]
pokey: removed this function as it's no longer used

apps/vscode/command_client.py
```diff
@@ -1,126 +1,235 @@
...
 from pathlib import Path
 from tempfile import gettempdir
+from typing import Any, List
+from uuid import uuid4
+
+from talon import Module, actions, Context
 
-is_mac = app.platform == "mac"
+
+STALE_TIMEOUT_MS = 10_000
```
lunixbochs, line 13: I mentioned this in the other review, I think I'd recommend a slightly longer timeout, as the timeout will never be used when everything is functioning correctly.

apps/vscode/command_client.py
```diff
@@ -1,126 +1,235 @@
...
+        request (Request): The request to serialize
+        path (Path): The path to write to
+
+    Raises:
+        Exception: If another process has an active request file
+    """
+    try:
+        write_json_exclusive(path, request.to_dict())
+    except FileExistsError:
+        stats = path.stat()
```
lunixbochs, line 84: You might want to hoist the rest of this function out of the exception handler, otherwise Python can produce a chained traceback which might not be helpful here. If the write succeeds in the `try` section, just `return`, then the `except FileExistError:` path can just `pass`, and you can put the rest of the code at the outermost indentation level of the function

apps/vscode/command_client.py
```diff
@@ -1,126 +1,235 @@
...
-    if is_mac:
-        actions.key("cmd-shift-alt-p")
-    else:
-        actions.key("ctrl-shift-alt-p")
+    actions.user.trigger_command_server_command_execution()
+
+    try:
+        decoded_contents = read_json_with_timeout(response_path)
+    finally:
+        unlink_if_exists(request_path)
```
lunixbochs, line 164: Instead of unlinking the request on the Python side, the vscode side can unlink the request once it's done reading it. That will make it more obvious which side failed if there are some stale files sitting around, and makes it less likely to accidentally double-execute a request.

apps/vscode/vscode.py
```diff
@@ -47,7 +52,7 @@ def filename():
 
 @ctx.action_class("edit")
 class edit_actions:
-    def find(text = None):
+    def find(text=None):
         if is_mac:
```
lunixbochs, line 56: Not really related to the command server changes but uses of `is_mac` can be moved to the mac_ctx.

apps/vscode/command_client.py
```diff
@@ -1,126 +1,235 @@
...
-        timeout=(0.05, 3.05),
-    )
-    response.raise_for_status()
+        except FileNotFoundError:
+            # If not found, keep waiting
+            pass
 
-    actions.sleep("25ms")
+        actions.sleep(sleep_time)
+        sleep_time *= 2
```
lunixbochs, line 210: You might want to clamp the value of `sleep_time` so you don't overshoot the timeout.
Perhaps, before the loop, set `timeout_time = now + 3.0`.
Then clamp sleep time to something like this after doubling it:
```
sleep_time *= 2
time_left = timeout_time - time.perf_counter()
sleep_time = max(0.0005, min(sleep_time, time_left))
```

Otherwise, you can only timeout after one of these total times:

```
acc = 0
sleep_time = 0.0005
for i in range(15):
    acc += sleep_time
    sleep_time *= 2
    print(f"{i} sleep={sleep_time:.3f}s total={acc:.3f}s")
```

```
0 sleep=0.001s total=0.001s
1 sleep=0.002s total=0.002s
2 sleep=0.004s total=0.004s
3 sleep=0.008s total=0.007s
4 sleep=0.016s total=0.015s
5 sleep=0.032s total=0.032s
6 sleep=0.064s total=0.064s
7 sleep=0.128s total=0.128s
8 sleep=0.256s total=0.256s
9 sleep=0.512s total=0.512s
10 sleep=1.024s total=1.024s
11 sleep=2.048s total=2.048s
12 sleep=4.096s total=4.096s
13 sleep=8.192s total=8.192s
14 sleep=16.384s total=16.384s
```

lunixbochs (changes requested):

apps/vscode/command_client.py
```diff
@@ -1,126 +1,235 @@
...
-            args,
-        )
-    )
+    args = [x for x in args if x is not NotSet]
+
     if decode_json_arguments:
         args = [json.loads(arg) for arg in args]
 
-    port_file_path = Path(gettempdir()) / "vscode-port"
+    username = getpass.getuser()
```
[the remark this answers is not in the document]
pokey: I don't think the UID approach will be portable to Windows, so would make this code more complicated, and I'm not sure I see the downside of using environment variables in this situation.  We're not using this username for access control, just a convenient way to make the communication dir user-specific

apps/vscode/command_client.py
pokey: VSCode does unlink the request once it's done reading; this code is just supposed to make talon clean up after itself in case of a broken VSCode.  But I do see your point around double-executing a request.

How about if instead of silently unlinking, we throw an exception if the file exists?  Then I guess we could either unlink the file or not before throwing the exception.  I'd be inclined to unlink, because otherwise on next request we'll throw an exception saying that another talon instance is running because there's a non-stale request file.  But as you mention, that's slightly less defensive against double-executing a request.

So I'd say either way I'd push for throwing an exception if request file still exists, but could go either way wrt unlinking.

lunixbochs: There's no guarantee the new file is related to your request, and this state generally shouldn't happen. I really don't think Python should unlink the request file. If vscode stops unlinking the request file for some reason, you should fix vscode. Python unlinking the request file removes an indicator something is wrong with the system. It also violates the exclusive lock guarantees for detecting 2+ in-flight requests, because you can feasibly write w/ 'x', unlink, write w/ 'x' without vscode having processed anything yet.

apps/vscode/command_client.py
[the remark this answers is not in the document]
lunixbochs: My concern is about quality, not access control. I've personally had users mess up a lot of stuff in their environment you wouldn't expect.

On Windows, the temp dir is user-specific anyway, so we don't need to suffix there. I also found this about `userInfo().username` on the node side:

> The only problem is, in Windows 10, it returns the first name of the owner account that has been used (just a heads up).

My vote would be to use getuid if available, otherwise don't suffix the path at all, roughly like this:

```
suffix = ""
if hasattr(os, "getuid"):
    suffix = "-{os.getuid()}"
communication_dir_path = Path(gettempdir()) / f"vscode-command-server{suffix}"
```

On the node side, you can call `userInfo()` and do something like this:
```
if (info.uid >= 0) path += `-${info.uid}`
```
or whatever, because
> On Windows, the uid and gid fields are -1, and shell is null.

apps/vscode/command_client.py
```diff
@@ -210,10 +222,14 @@ def read_json_with_timeout(path: str) -> Any:
...
         actions.sleep(sleep_time)
-        sleep_time *= 2
-        if time.perf_counter() - start_time > VSCODE_COMMAND_TIMEOUT:
+
+        time_left = timeout_time - time.perf_counter()
+
+        if time_left < 0:
             raise Exception("Timed out waiting for response")
 
+        sleep_time = min(sleep_time * 2, time_left)
```
lunixbochs, line 231: You should clamp the minimum sleep time as per my example as a precaution, because if there's just the right amount of slip between the perf_counter clock and the actual sleep() time, you could end up looping extra times with very small sleeps.
e.g.
`sleep_time = max(min(sleep_time * 2, time_left), INITIAL_SLEEP_TIME_SECONDS)`

lunixbochs (changes requested):

apps/vscode/command_client.py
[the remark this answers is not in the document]
pokey: this is great thanks @lunixbochs

apps/vscode/command_client.py
```diff
@@ -0,0 +1,315 @@
...
+        return
+
+    request_path = communication_dir_path / "request.json"
+    response_path = communication_dir_path / "response.json"
+
+    # Generate uuid that will be mirrored back to us by command server for
+    # sanity checking
+    uuid = str(uuid4())
+
+    unlink_if_exists(response_path)
```
pokey, line 156: @lunixbochs any objection to this one?  Technically it should have been removed after last request, so something is wrong if it exists here.  But there is something nice about ensuring a clean slate here

lunixbochs: Yep, I object to this one. The state machine is pretty well designed and this would hide bugs.

PR #449, opened 2021-06-02
Add grave functionality

xavier630:


text/symbols.talon
```diff
@@ -4,6 +4,10 @@ double dash: "--"
 (bracket | brack | left bracket): "{"
 (rbrack | are bracket | right bracket): "}"
 triple quote: "'''"
+(triple grave | triple grave accent):
+    insert("```")
+(grave | grave accent):
+    insert("`")
```
xavier630, line 10: I've used the insert function rather than a direct string value. Let me know if you see any issues with this.

lunixbochs:
I think this is inconsistent with the naming of back tick elsewhere in knausj

xavier630:
How do we feel about having both backtick and grave? As in I add backtick here as an extra option to each of these commands. Or would you prefer that we cut grave entirely?

text/symbols.talon
xavier630: Now that I see backtick in the punctuation, I imagine these single commands are not needed since the punctuation dictionary covers them

fidgetingbits:
I'd be fine with adding grave, but mostly because I use it in my own repo anyways. I use the term "gravy" for ``` and I find it works well too fwiw. I think the only inconsistency would be in `keys.py`. but we could just add grave to that as well I think. it would have to be agreed on by most of the maintainers if it was as actually switched completely I think because it will affect a lot of people using the repo.

knausj85:
I do like grave & gravy, and am not opposed to either adding or replacing myself.

xavier630:
Have updated the PR to take everyone's feedback into account


One Q I had was with the different styles of
```
(triple grave | triple back tick | gravy):
    insert("```")
```
vs
```
(triple grave | triple back tick | gravy): '```'
```

I have favoured using the insert function here since that provides an extra adapter layer for future modifiability, but if this isn't the standard we are working towards, let me know

knausj85:
My general rule: insert if multi-line command, or otherwise on new line. More readable.

PR #457, opened 2021-06-06
[Minor] Make settings more readable

xavier630:
I was having quite a bit of trouble reading the settings, so added whitespace between settings


splondike:
Yeah, is a little bit difficult to read currently without syntax highlighting. LGTM.

PR #458, opened 2021-06-06
Add 'go bottom', 'go top' commands

xavier630:
Likely more intuitive for beginners than `go way down`, `go way up`


awulkan:
Serenade has the commands "file start" and "file end", which are also pretty intuitive.

PR #460, opened 2021-06-09
Fix broken "panel terminal" command for VS Code

hiddefolkertsma:
See https://github.com/microsoft/vscode/issues/106535. Just started using VS Code and `workbench.panel.terminal.focus` is broken, it does not focus the terminal. The proper command is `workbench.action.terminal.focus`. The other commands with `panel` still work but may break in the future, too.


splondike:
Will leave for folks who use VS Code to confirm, but looks good to me.

awulkan:
I can confirm that "workbench.action.terminal.focus" is the correct command.

pokey (approved):

PR #465, opened 2021-06-18
add websites.csv and 'open <website>' command

rntz:
As long as we have search engines, we may as well have websites. This adds a csv of common websites and a command `open <website>` that opens the given website in the user's preferred web browser using the `webbrowser` module. Since this is so similar to the search engine functionality I've folded them together and renamed the files.


misc/websites_and_search_engines.talon
```diff
@@ -0,0 +1,5 @@
+open {user.website}: user.open_url(website)
```
pokey, line 1: "open" is probably the right verb here, tho I have a slight hesitation having it at top-level, as there are various "open" commands in a number of contexts

Seems unlikely to run into ambiguity / misrecognitions, but figured I'd mention. 

Could potentially make it a bit longer, eg "open site", "launch site", "website", etc

But I don't feel super strongly on this one; "open" is fairly intuitive and short

pokey (approved):
LGTM with minor comment

misc/websites_and_search_engines.talon
knausj85: Yeah, I'm slightly hesitant on open too.

rntz:
I've been using 'open' for this locally for a long time. I also use 'open'
instead of 'launch' for running apps (but on Linux there's no autodetection
of apps so for me that's just a list of apps I shoved in a talon file).
It's worked great for me, but I understand the hesitation.

If we need another verb, I'd suggest 'browse' since it opens in your web
browser. Not a strong opinion though.

knausj85:
This seems to be fine in practice, merging. Can revisit if needed.

PR #470, opened 2021-06-26
Add generic create_spoken_forms function

pokey:
See https://github.com/knausj85/knausj_talon/issues/413


knausj85:
@pokey - I made a few tweaks from testing around this in a few different contexts. Let me know what you think.

code/create_spoken_forms.py
```diff
@@ -70,62 +71,90 @@ def create_spoken_forms(
...
+
+        else:
+            terms = full_forms
+
         terms = [
             term
             for term in terms
-            if term not in words_to_exclude and len(term) >= minimum_term_length
+            if (term not in words_to_exclude and len(term) >= minimum_term_length)
+            # always keep the full forms, even if < min term length?
```
pokey, line 126: i think so, because otherwise there's no way to refer to this one right?

pokey (commented):
LGTM @knausj85

code/create_spoken_forms.py
knausj85: Yeah, a simple test case with a folder name of "1" for instance with the file manager stuff

knausj85:
The last piece I'd forgotten about was help.py. This revealed one small potential issue with including abbreviations... i.e. slack_win.talon becomes slack window... sigh. Not sure what to do about it, will need to revisit.

pokey:
Yeah this reminds me of the WhatsApp -> "whats application" thing

knausj85:
@pokey last set of tweaks, let me know what you think. My only remaining hesitation is the abbreviation stuff... but if you think it's good enough as-is I will ship it.

apps/file_manager.talon
```diff
@@ -14,20 +14,22 @@ go user: user.file_manager_open_directory(path.user_home())
 go back: user.file_manager_go_back()
 go forward: user.file_manager_go_forward()
 (go parent | daddy): user.file_manager_open_parent()
-^follow <number>$:
-    directory = user.file_manager_get_directory_by_index(number - 1)
+^follow numb <number_small>$:
```
rntz, line 17: Something seems wrong here - why is this PR to add create_spoken_forms changing the file_manager commands?

code/extensions.py
```diff
@@ -0,0 +1,43 @@
...
+    "dot mark down": ".md",
+    "dot shell": ".sh",
+    "dot vim": ".vim",
+    "dot see": ".c",
+    "dot see sharp": ".cs",
+    "dot com": ".com",
+    "dot net": ".net",
+    "dot org": ".org",
+    "dot us": ".us",
+    "dot U S": ".us",
```
rntz, line 19: Arguably, `.com`, `.net`, & `.us` are not file extensions (they are TLDs). Is this worth worrying about?

(`.org` is a file extension, because of org-mode in emacs)

apps/file_manager.talon
knausj85: With the new create_spoken_forms, the old commands conflict / don’t work. If there’s a strong preference, I don’t mind moving the file manager changes to another PR

code/extensions.py
knausj85: Is there a use case for splitting these  TLDs into their own list?

knausj85:
I'll be merging this one on Sunday if there are no material concerns. I've got a bunch of handy stuff waiting in the wings for this one.

PR #473, opened 2021-06-27
Added draft editor. Use VSCode as a draft window

AndreasArvidsson:
Use an editor, by default VSCode, as a draft window.

in Firefox/slack/word etc..
1) "draft this" / "draft all" / "draft line"
2) Switches to vscode and opens a new untitled file
3) Make changes
4) "Draft submit"
6) Switches back to original window and pastes over the draft content with the changes

Only supports vscode for now, but would be easy to add support for additional editors


draft_editor/draft_editor.py
```diff
@@ -0,0 +1,71 @@
...
+    actions.app.tab_close()
+
+def focus_window(window: ui.Window):
+    """Focus window and wait until finished"""
+    window.focus()
+    t1 = time.monotonic()
+    while ui.active_window() != window:
+        if time.monotonic() - t1 > 1:
+            raise RuntimeError(f"Can't focus window: {window.title}")
+        actions.sleep("50ms")
```
pokey, line 71: This should be an action, and should probably go into `switcher.py`.  In your repo you can define the action wherever you want

draft_editor/draft_editor.py
```diff
@@ -0,0 +1,71 @@
...
+        "VSCodium",
+        "Codium",
+        "code-oss"
+    }
+    for w in ui.windows():
+        if w.app.name in editor_names:
+            return w
+    raise RuntimeError("VSCode is not running")
+
+def new_file():
```
pokey, line 57: I think I'd argue that we should just use the action `app.tab_open`, and map that to `actions.user.vscode("workbench.action.files.newUntitledFile")` for VSCode.  Make sense? cc/ @knausj85

draft_editor/draft_editor.py
```diff
@@ -0,0 +1,71 @@
...
+    actions.user.vscode("workbench.action.files.newUntitledFile")
+
+def close_file():
```
pokey, line 60: **nit:** I think I'd just inline this one

draft_editor/draft_editor.py
```diff
@@ -0,0 +1,71 @@
...
+    actions.mode.disable("user.draft_editor")
+    actions.edit.select_all()
+    if save:
+        actions.edit.copy()
+    close_file()
+    focus_window(active_window)
+    if save:
+        actions.edit.paste()
+
+def get_editor():
```
pokey, line 44: I think I'd prefer to just use `actions.user.switcher_focus` here, because I believe this impl could grab a background VSCode window, instead of just switching to VSCode and using whatever is in front.  At least I have a recollection of this happening when I used this original version

draft_editor/draft_editor.py
```diff
@@ -0,0 +1,71 @@
...
+    def draft_editor_save():
+        """Save draft editor"""
+        close_editor(True)
+
+    def draft_editor_discard():
+        """Discard draft editor"""
+        close_editor(False)
+
+
+def close_editor(save: bool):
```
pokey, line 34:
```suggestion
def close_draft_window_and_maybe_submit(submit_changes: bool):
```

draft_editor/draft_editor.py
```diff
@@ -0,0 +1,71 @@
+from talon import Module, actions, ui
+import time
+
+mod = Module()
+mod.mode("draft_editor", "Mode to show if the draft editor is open")
```
pokey, line 5:
```suggestion
mod.mode("draft_editor", "Indicates whether the draft editor has been activated")
```

pokey (commented):
Cool!  Left comments, mostly about ways to remove the local functions or replace them with actions

draft_editor/draft_editor.py
AndreasArvidsson: I thought about that but aegis comment the other day was that a self sustaining folder was the prepared way to make something "semimodular" until we get real modules. If prepared one can easy integrate it more with knausj like the switcher.

draft_editor/draft_editor.py
AndreasArvidsson: Sounds good. My thinking was that maybe someone wanted to use an editor without tabs, but that might be far fetched.

draft_editor/draft_editor.py
AndreasArvidsson: Its there so we can do it context specific for other editors. in the future. For now you are completely right that its a bit verbose.

draft_editor/draft_editor.py
AndreasArvidsson: Is there such a thing as a background window? Applications yes, but windows?
If that is the case then background=False could probably be used.

pokey: Sorry I don't mean background, but on a Mac, when you do your approach, it focused a random window, whereas if you just focus the app, it leaves the window focussed that was focussed previously

AndreasArvidsson: Ah got it. No problem I can fix that.

AndreasArvidsson: Should be fixed now. I use the latest window for the app.

draft_editor/draft_editor.py
AndreasArvidsson: I resolved it by using tab_open and tab_close. That way we should be getting app specific actions that work.

draft_editor/draft_editor.py
AndreasArvidsson: Now using using tab_open and tab_close.

draft_editor/draft_editor_open.talon
```diff
@@ -2,5 +2,5 @@ mode: user.draft_editor
 app: vscode
```
knausj85, line 2: You may consider using a tag here instead, so apps can assert the tag without branching the file

draft_editor/draft_editor.py
```diff
@@ -0,0 +1,67 @@
...
+@mod.action_class
+class Actions:
+    def draft_editor_open():
+        """Open draft editor"""
+        global active_window
+        active_window = ui.active_window()
+        editor = get_editor()
+        has_selected_text = actions.edit.selected_text() != ""
+        if has_selected_text:
+            actions.edit.copy()
```
knausj85, line 18: I don't know if this is "better" per se, but maybe use `clip.set_text` since we've already got the text.

```
selected_text = actions.edit.selected_text()

if selected_text != ""
   actions.clip.set_text(selected_text)
```

AndreasArvidsson: That would definitely work but what would be the upside? Is using clip.set_text more efficient than doing just a copy?

The third option would be just to do.
```
if selected_text != "":
    actions.user.paste(selected_text)
```

knausj85: I like your third option, good call.

draft_editor/draft_editor.py
pokey: @AndreasArvidsson this still doesn't work for me.  It doesn't pick the most recent window.  Why not just use [`switcher_focus`](https://github.com/knausj85/knausj_talon/blob/b6f4939b3ce39785871c9e8ec88bbf91232c1090/code/switcher.py#L266-L276)?  That doesn't change the active window; just focuses the app, and is battle-tested across platforms.

pokey (commented):
This is pretty close imo, but I did need to add a sleep when switching back, and see also inline comment.  See [these changes](https://github.com/knausj85/knausj_talon/commit/d75f878f9068e4ab78861d6212a00e8427b3dbb7) for the extra sleep (couldn't figure out how to open PR against yours for some reason; fork tracking is confused)

draft_editor/draft_editor.py
AndreasArvidsson: 3 reasons
1. switcher_focus takes a signel app name. The draft editor needs to work for all different vscode names
2. switcher_focus doesn't wait for app to switch on other os then mac.
3. Switching back to the specific window is important.

It already works fine on windows and linux and I can update it to work better on mac but using switcher_focus is not a good fit for the above reasons.

draft_editor/draft_editor.py
AndreasArvidsson: Pushed this change.

draft_editor/draft_editor_open.talon
AndreasArvidsson: My reasoning was that since this file is only active when you already are in draft mode. ie mode `user.draft_editor` is enabled and if you want to change to another editor you have to update the python file there was not any gain in using another tag in this file.

draft_editor/draft_editor.py
pokey: So I don't feel strongly enough to stand in the way of this PR; it is good to go from my perspective.  But I will say wrt each of your points above:

1. That should be an easy fix to `switcher_focus` that we might want to consider, no?
2. That's also an easy fix
3. I'm not sure how that would be affected by using `switcher_focus`, you could use the specific-window version here just like you do today

But as above, not a showstopper imo; happy to merge if @knausj85 is happy with this PR

AndreasArvidsson: Great. If the appropriate functionality is available, or is made available, anywhere else in this repository I have no objections of using that instead.

AndreasArvidsson: I have now made the necessary changes to the switcher so the draft editor can use those.

AndreasArvidsson:
Some updates:
* use a @mod.scope to set draft_editor_running. Now global draft commands are only active if the draft editor is running.
* editor names is configurable via a csv setting. Settings doesn't support lists.
* Updated switcher to support focusing on app or window with wait for switch.
* Global draft commands are disabled in the editor itself. ie you cant draft from the draft editor.

draft_editor/draft_editor.py
```diff
@@ -0,0 +1,91 @@
...
+    "Visual Studio Code",
+    "Code",
+    "VSCodium",
+    "Codium",
+    "code-oss"
+]
+
+setting_editor_names = mod.setting(
+    "draft_editor",
+    type=str,
```
knausj85, line 17: Is there any reason not to use a list here for the setting itself, and set it to the default list? As it stands, the user would have to restart talon/touch this file for edits of the csv to take effect

AndreasArvidsson: Settings do not support lists. Tried it but got an error so I changed to a csv instead.

knausj85: Oh, that's unfortunate. I believe settings do support dictionaries (e.g. dictate.word_map). While not ideal, maybe try that to preserve the reloading sweetness?

AndreasArvidsson: I tried that as well but talon script only support string and numbers what I can tell. But that is not the only problem to get the reloading you want. I actually register an app at startup.
https://github.com/knausj85/knausj_talon/blob/d7f256fdd2725d89bb751109102c113e7a782c9b/draft_editor/draft_editor.py#L39

We would need to have a setting change event and dynamically update this app regular expression every time. For now I'm personally satisfied with having to restart when changing something as seldom as this. It is of course your call.

knausj85: I can live with it for now, thanks

AndreasArvidsson: :)

PR #475, opened 2021-06-27
Use auto_lang mode to detect if lang is forced or not

AndreasArvidsson:
Solves the problem that code.language doesn't update when forcing a language and until you change focus and the app title updates you have 2 language context active at the same time.

This is active if user.javascript OR code.language == javascript.
```
mode: user.javascript
mode: command
and code.language: javascript
```

Change it to this and the problem is solved sine auto_lang and user.javascript is mutex.
```
mode: user.javascript
mode: user.auto_lang
and code.language: javascript
```

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -61,9 +61,6 @@ extension_lang_map = {
     ".vimrc": "vimscript",
 }
 
-# flag indicates whether or not the title tracking is enabled
-forced_language = False
-
 
 @mod.capture(rule="{user.code_functions}")
 def code_functions(m) -> str:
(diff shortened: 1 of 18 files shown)


code/code.py
```diff
@@ -87,11 +84,9 @@ def code_libraries(m) -> str:
 class code_actions:
     def language():
         result = ""
-        if not forced_language:
-            file_extension = actions.win.file_ext()
-
-            if file_extension and file_extension in extension_lang_map:
-                result = extension_lang_map[file_extension]
+        file_extension = actions.win.file_ext()
```
knausj85, line 87: I feel like there's scenarios where both auto_lang and force lang could be active with this as-is. Maybe worth checking the auto_lang mode here for sanity?

AndreasArvidsson: In which scenario are you thinking about?
As the code is written now you disable auto_lang when a specific language mode is forced/enabled and you enable auto_mode when forced language is cleared/disabled. So they should be mutually exclusive.

knausj85: Sorry, you're correct. Handled by the context checks.

```
mode: user.auto_lang
and code.language: javascript
```

AndreasArvidsson: Yes exactly. As far as I can tell this should work.

knausj85 (approved):

PR #477, opened 2021-06-28
Jetbrains: Don't use proxies on request

donnysmit:
On my work station I use proxies, this gives me a 503 service unavailable when using Webstorm commands. Making sure that proxies are not used fixes the problem.


lunixbochs:
does that mean this is fixed? https://github.com/talonvoice/talon/issues/185 cc @nriley

nriley:
Nope, I am still needing to patch Talon on Windows as it is not reading the system proxy settings. @donnysmit - what platform are you using?

donnysmit:
> Nope, I am still needing to patch Talon on Windows as it is not reading the system proxy settings. @donnysmit - what platform are you using?

Windows 10, I'm pretty sure I'm not using a socks proxy which might be the reason why I am not having the same issues

Issue #322, opened 2021-01-09
folder structure in `apps` (super nitpicky, trivial thing)


ym-han:
I noticed that the OS folders (`mac`, `linux`, etc) are currently at the same level as folders for (presumably) OS-independent apps like `jetbrains` and `vscode`. Here's a quick suggestion: how about putting the latter into a folder like `OS_independent`, so that the only folders at that level are that and the folders for the specific OSes? Right now it feels like a category mistake to me to have specific apps like `vscode` be at the same level as `mac` and `linux`.

lunixbochs:
why not structure like this:

apps/slack/win.talon, apps/slack/linux.talon, apps/slack/mac.talon, apps/slack/slack.talon

knausj85:
Alright, the migration of actions from talon -> python seems like a reasonable time to do something about this.

For multi-platform apps I was leaning towards:

```
chrome.talon # self-explanatory
chrome.py  # common stuff, action declarations and such
mac.py # mac-specific action implementations  etc+ app definition maybe
win.py # win-specific action implementations  etc + app definition maybe
```

It's sometimes handy to be able to quickly compare the actions for debugging purposes and such.

@pokey @splondike @rntz @dwiel

pokey:
@knausj85 Those would all be in an `apps/chrome/` directory?

Seems reasonable to me

knausj85:
Yeah, those would all be in an apps/chrome/ directory.

I guess the default app.talon / edit.talon implementations should probably be moved into 

```
code/app/mac.py 
code/app/win.py
```

Or something too.

I’m not a huge fan of the OS_Independent idea proposed in the original issue personally-I’d vote to either flatten (no platform subdirs in apps) or keep as-is. I’m not too picky about the folder structure, so speak now. 🤗

rntz:
Are we sure that we'll never have platform-specific talon files? I think we currently do for Firefox, for example. ISTR we'd rather not name those things like `firefox/mac.talon` because of some interaction with the help system.

knausj85:
There will very likely still be platform-specific commands, and `firefox_mac.talon` is still  reasonable. `/mac.py` (or something more explicit) is  just for action definitions that have to live in python, and don't interact with help.  

Generally, I believe most platform-specific Talon files are solely used for action definitions at the moment? Reality might be different in folks branches. Slack is an exception, at the moment.

PR #482, opened 2021-07-02
Action migration for v0.2 (.talon -> .py)

knausj85:
Also some work for #322 

1. Multi-platform apps are laid out (mostly) as follows:

```
chrome.talon # commands live here
chrome.py  # any common code support e.g. action declarations, app name defintions
mac.py # mac-specific action implementations
win.py # win-specific action implementations 
linux.py # linux-specific action implementations
```

2. Platform-specific apps moved to apps/platforms/ for convenience

3. The default app.talon / edit.talon implementations were moved into /code:
```
code/platforms/win/edit.py 
code/platforms/win/app.py
```


fidgetingbits:
what's the thinking for "_Platform Specific" as the folder name? it seemed confusing to me at first, also seems hard to say if you actually had to speak it and also prone to having to escape space depending where you're working. to me since `code/` now has `code/platform/win`, it would be more clear if it was also `apps/platform/win` or maybe `apps/_platform/win` if the score is meant to indicate that it's not an application itself?

knausj85:
It's just an old habit due to how Windows orders folders in Explorer etc, which is probably bad idea. 

`apps/platform/win` is probably better

apps/eclipse/eclipse.talon
```diff
@@ -7,144 +7,116 @@ tag(): user.line_commands
...
 #note: vscode has no explicit mode for multiple cursors
-# action(user.multi_cursor_add_above): 
-# action(user.multi_cursor_add_below): 
-# action(user.multi_cursor_add_to_line_ends): 
-# action(user.multi_cursor_disable): 
-# action(user.multi_cursor_enable): 
-# action(user.multi_cursor_select_all_occurrences): 
-# action(user.multi_cursor_select_fewer_occurrences): 
-# action(user.multi_cursor_select_more_occurrences): 
+# action(user.multi_cursor_add_above):
```
fidgetingbits, line 14: i guess these can get removed

apps/platforms/mac/sarafi/safari.py
```diff
@@ -0,0 +1,75 @@
...
+apps = mod.apps
+mod.apps.safari = """
+os: mac
+and app.bundle: com.apple.Safari
+"""
+
+ctx.matches = r"""
+app: safari
+"""
+ctx.tags = ['browser', 'user.tabs']
```
fidgetingbits, line 14: i've noticed in a few places above that tags are asserted like this, rather than in the .talon file. i'm just curious if this became the preferred way to do it? i usually do it from the .talon

apps/slack/slack_win.talon
```diff
@@ -6,15 +6,8 @@ app: slack
 tag(): user.messaging
```
fidgetingbits, line 6: above chrome was still chrome.win.talon earlier and this is now slack_win.talon, and then we have win.py. Is this the idea going forward to have app.win.talon or app_win.talon, or should it now be apps/slack/win.talon? i'm pretty indifferent to any of the three, but may as well be consistent

apps/windbg/windbg.talon
```diff
@@ -6,114 +6,11 @@ mode: user.windbg
...
-
-# Convenience
-action(user.debugger_clear_line):
-    key("ctrl-a backspace")
-##
-# Windbg specific functionality
-##
-
+    
+    # Type inspection
```
fidgetingbits, line 13: no longer relevant comment

lang/c/c.talon
```diff
@@ -135,17 +48,17 @@ push brackets:
     key(enter)
     key(enter)
     edit.up()
-
-# Declare variables or structs etc.
-# Ex. * int myList
+    
+    # Declare variables or structs etc.
+    # Ex. * int myList
```
fidgetingbits, line 53: not really that important  for this review i guess, but i've noticed a few similar places where suddenly the comments are indented like this. i didn't flag them, but suddenly curious if it's a side effect of the migration script or something?

lang/vimscript/vimscript.talon
```diff
@@ -14,37 +13,13 @@ settings():
...
-  insert("elseif ")
-action(user.code_state_else):
-  insert("else")
-
-action(user.code_private_function): "function "
-action(user.code_protected_function): "function "
-action(user.code_public_function): "function "
+    
+    
+    ###
```
fidgetingbits, line 18: these comments are no longer relevant

apps/win/mintty/mintty.talon
```diff
@@ -1,17 +0,0 @@
-app: mintty
--
-tag(): terminal
-tag(): user.file_manager
-tag(): user.generic_terminal
-tag(): user.git
```
fidgetingbits, line 6: i'm not sure if this added somewhere else, but this tag assertions are missing from mintty.py. i'm not sure if this is a side effect of once the actions being gone the original file is removed?  i'm not sure if this is affecting either apps as well. the guake.py file seemed to have them added as `apps/platform/linux/guake.py
7:ctx.tags = ['user.git', 'user.kubectl', 'user.tabs', 'terminal']`, so maybe something did mess up with the script here?

apps/platforms/mac/sarafi/safari.py
fidgetingbits: after reviewing the rest of the files i noticed this seems to be part of the migration script when the associated talon file no longer has anything so get deleted

fidgetingbits (commented):
i've gone through all of the changed files and made a few comments. it seems sane in general, i'm running a merged copy in my fork at the moment in it seems to be working without problem, but obviously there is a minimum amount of stuff i'm actually hitting code coverage wise. i guess the biggest problem is potentially the missing tag migrations that i noted

lang/c/c.talon
lunixbochs: yes this is probably a bug in the migration script

apps/win/mintty/mintty.talon
lunixbochs: On my machine, talon created mintty_actions.py. Looks like @knausj85 merged that file manually to mintty.py and didn't copy the tags.

apps/slack/slack_win.talon
knausj85: The thinking here is:
1. mac.py/win.py => Define all platform-specific actions/lists/whatever
2. chrome.talon => for cross-platform apps, hopefully we take the time to create the actions necessary to have a single .talon file with defined commands. Note there is no chrome.talon at the moment
3. slack_mac.talon => since the .talon files appear in the help system, just naming these "mac.talon" isn't terribly useful. Until we can get around to item 2 for apps like slack, this is probably the least-worst option

apps/win/mintty/mintty.talon
knausj85: Good catch, looks like I only missed mintty

apps/slack/slack_win.talon
knausj85: I guess I'm also open to re-working how the help system works if there's some strong preferences here.

knausj85:
Thanks much for reviewing this one @fidgetingbits.

lang/c/c.talon
knausj85: I'm pretty indifferent here fwiw. I'd be tempted to call the migration script Good Enough(tm) as-is

knausj85: I've cleaned these up. There weren't many.

apps/slack/slack_win.talon
fidgetingbits: thanks for explaining. im fine with that for now really. longer term i guess it would be nice to not specify mac in the help output. i got confused about chrome.win.talon... it was actually a file being deleted.

nriley:
A couple bits that need fixing from this merge that I noticed when merging my fork:

platforms/mac/sarafi — typo, needs renaming

apps/linux/kde-console.talon — needs to be migrated

PR #483, opened 2021-07-03
Updated screenshot with flash and pictures directory 

AndreasArvidsson:
* Remove app notification and instead use a brief white screen flash.
* By default store images in pictures folder instead of desktop. 
* Updated file name with application title when capturing a window to file. Makes it possible to categorize images later.
* On Linux use built in selection tool instead of third party utility. (Requires beta/v2 to get printscr button working).

### DIFF
--- a/code/screenshot.py
+++ b/code/screenshot.py
@@ -60,9 +60,9 @@ def clipboard_rect(rect: ui.Rect):
 
 def get_screenshot_path(title: str = ""):
     if title:
-        title = f" | {title.replace('.', '_')}"
+        title = f" - {title.replace('.', '_')}"
     date = datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
-    filename = f"Screenshot | {date}{title}.png"
+    filename = f"Screenshot {date}{title}.png"
     folder_path = screenshot_folder.get()
     path = os.path.expanduser(os.path.join(folder_path, filename))
     return os.path.normpath(path)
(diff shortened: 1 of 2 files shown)


knausj85 (approved):

code/screenshot.py
```diff
@@ -1,83 +1,79 @@
...
     def screenshot_window_clipboard():
-        """takes a screenshot of the window and saves it to the clipboard"""
-        img = screen.capture_rect(ui.active_window().rect)
-        clip.set_image(img)
+        """Takes a screenshot of the active window and saves it to the clipboard"""
+        clipboard_rect(ui.active_window().rect)
+
+
+def screenshot_rect(rect: ui.Rect, title: str = ""):
+    flash_rect(rect)
```
knausj85, line 51: I'm slightly hesitant on the flash at the moment due to this issue https://github.com/talonvoice/talon/issues/248, which would probably prevent the entire command from working

AndreasArvidsson: Yes understand that. Probably should wait with this until that bug is fixed then. I only tested this on windows and linux and it's a great feature... if it's working of course.

AndreasArvidsson: Wouldn't above issue already make it so that the capture doesn't work since the capture already relies upon ui.Rect? The flash just uses the same rect as the capture itself. Or is is just a problem with the canvas?

knausj85: You're correct-sorry about that.

AndreasArvidsson: No problem :)

PR #485, opened 2021-07-06
Add README for command client

pokey:


settings/words_to_replace.csv
```diff
@@ -0,0 +1,30 @@
+Replacement,Original
```
xavier630, line 1: Looks like these various sittings files were accidentally committed

pokey: Nice catch thanks @xavier630

PR #490, opened 2021-07-10
Add "cancel" and "ignore"

pokey:
- "cancel" allows you to cancel a command that you realize you are flubbing, eg "air bat crap cancel"
- "ignore" allows you to say something (eg to a human) that you don't want talon to hear, eg "ignore hey Jerry"


rntz:
uh, do these commands do anything other than produce notifications?

PR #493, opened 2021-07-14
Override play pause for Windows

SCdF:
The `play` key doesn't work on Windows, instead we want `play_pause`.


SCdF:
I am also unable to find the function `user.media_set_volume` which is used for `set volume <number>`, but I'm not really sure how this hooks up. It doesn't work on windows, but I can't tell from the code if it works on the platforms.

SCdF:
I realised this puts an extra item ("media windows") in the help pane, which we almost certainly don't want. Consider this PR a draft I guess? Looking for guidance on how to do this correctly (I'm guessing it involves writing Python to do some indirection or something).

rntz:
Hm, does anyone know whether `key(play_pause)` works on Mac & Linux? If it does we should just use that. If not, yeah, overriding commands in talon files isn't a great solution, we should probably have an action, maybe user.media_play_pause, and implement it appropriately per-platform.

SCdF:
> Hm, does anyone know whether key(play_pause) works on Mac & Linux?

I tested this, and my keyboard's play/pause button works on Windows, MacOS over [barrier](https://github.com/debauchee/barrier), and Linux (Ubuntu 20.04) via a VM under Windows.

Interestingly my keyboard is an ErgoDox EZ, which is remappable, and in their configuration tool they only provide one option for play/pause, which they call "play" in the tool, but in their source code is mapped to `KC_MEDIA_PLAY_PAUSE`. 

I've updated the PR to reflect this. To fire the `play` key on MacOS and Linux to see what it does I'd have to install Talon on each (or work out another way to fire it) which is a step I have not taken.

knausj85:
I think the best approach is probably an action for this one. I get this error on my Macbook:

```
 in script at /Users/jeffknaus/.talon/user/knausj_talon/misc/media.talon:7:
 > key(play_pause)
InvalidKeyErr: play_pause
```

rntz:
It appears that the `play_pause` key exists on linux (it doesn't produce an error when pressed) but it doesn't appear to do anything, while `key(play)` has the expected behavior.

SCdF:
Apologies it took me so long to get to this. I've moved it into an action. If people who use mac and linux could test this that would be greatly appreciated.

knausj85:
looks great, thanks

PR #497, opened 2021-07-18
Tidy up generic browser commands including making them [noun -> verb]…

xavier630:
… where possible

https://github.com/knausj85/knausj_talon/issues/459

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ https://github.com/knausj85/knausj_talon/blob/master/text/symbols.talon
 
 Try using formatters by saying e.g. `snake hello world`, which will insert hello_world
 
-Mutliple formatters can be used togther, e.g. `dubstring snake hello world`. This will insert "hello_world"
+Multiple formatters can be used together, e.g. `dubstring snake hello world`. This will insert "hello_world"
 
 Formatters (snake, dubstring) are defined here
 https://github.com/knausj85/knausj_talon/blob/master/code/formatters.py#L146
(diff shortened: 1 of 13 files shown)


rntz:
I'm not sure I like this reordering. I guess it is more noun-verbish, but it converts a bunch of commands with a shared prefix ("show <blah>") into commands with shared suffixes ("<blah> show"). I find shared prefixes easier to remember -- I think this is most of the value of noun-verb order, honestly: eg. "tab (close|open|restore|next|last)" is noun-verb, but also shared-prefix, I suspect it's the latter that's important.

Independently, I also dislike turning "bookmark bar" into "bookmark bar show" and "dev tools" into "dev tools show" -- if there's only one available action, it's easier to leave it implicit. Could fix this by making the "show" optional, though -- `dev tools [show]`.

xavier630:
> I'm not sure I like this reordering. I guess it is more noun-verbish, but it converts a bunch of commands with a shared prefix ("show $FOO") into commands with shared suffixes ("$FOO show"). I find shared prefixes easier to remember -- I think this is most of the value of noun-verb order, honestly: eg. "tab (close|open|restore|next|last)" is noun-verb, but also shared-prefix, I suspect it's the latter that's important.

Happy either way here. One important note is that I haven't actually removed any of the verb noun commands so it should at least work the old was if you prefer that

> Independently, I also dislike turning "bookmark bar" into "bookmark bar show" and "dev tools" into "dev tools show" -- if there's only one available action, it's easier to leave it implicit. Could fix this by making the "show" optional, though -- dev tools [show].

Similar comment to above and that I haven't actually removed the original commands for these - but good idea to roll them into one command with an optional part - I'll do that

xavier630:
I've now rolled `dev tools show` in `bookmark bar show` each into a single command in which the words show is optional as you suggested 👍

knausj85 (approved):

xavier630:
I think one of the repository owners will need to approve this

splondike:
Looks good to me, merging.

PR #498, opened 2021-07-21
Add right-drag command to Talon

xavier630:
Adds a command to hold down right-click. This is useful for things like games where right click is used to pan the camera.

Some thoughts/design decisions here:
* the `mouse_drag` function was quite complicated since it had several responsibilities: starting drags and ending drags. Once you add a second type of drag this logic gets quite complex since the function now needs to start two types of drags, be able to end two types of drags and be able to switch between two types of drags in the case where one is enabled but not the one requested.
* In order to simplify it and keep functions as modular as possible, I moved ending drags into a separate function. Now `mouse_drag` is purely responsible for starting mouse drags. Drags can now be ended by calling `end drag` and with `touch`

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -148,20 +148,25 @@ class Actions:
         if eye_zoom_mouse.zoom_mouse.enabled:
             eye_zoom_mouse.zoom_mouse.on_pop(eye_zoom_mouse.zoom_mouse.state)
 
-    def mouse_drag():
+    def mouse_drag(button: int):
         """(TEMPORARY) Press and hold/release button 0 depending on state for dragging"""
         # todo: fixme temporary fix for drag command
         button_down = len(list(ctrl.mouse_buttons_down())) > 0
         # print(str(ctrl.mouse_buttons_down()))
         if not button_down:
             # print("start drag...")
-            ctrl.mouse_click(button=0, down=True)
+            ctrl.mouse_click(button=button, down=True)
             # app.notify("drag started")
         else:
+            # End any active drags
             # print("end drag...")
-            ctrl.mouse_click(button=0, up=True)
+            self.end_drag(0)
+            self.end_drag(1)
 
         # app.notify("drag stopped")
+    def end_drag(button: int):
+        """ Releases the specified mouse button """
+        ctrl.mouse_click(button=button, up=True)
 
     def mouse_sleep():
         """Disables control mouse, zoom mouse, and re-enables cursor"""
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -150,20 +150,8 @@ class Actions:
 
     def mouse_drag(button: int):
         """(TEMPORARY) Press and hold/release button 0 depending on state for dragging"""
-        # todo: fixme temporary fix for drag command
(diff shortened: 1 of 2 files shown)


code/mouse.py
```diff
@@ -148,20 +148,25 @@ def mouse_trigger_zoom_mouse():
...
         if not button_down:
             # print("start drag...")
-            ctrl.mouse_click(button=0, down=True)
+            ctrl.mouse_click(button=button, down=True)
             # app.notify("drag started")
         else:
+            # End any active drags
             # print("end drag...")
-            ctrl.mouse_click(button=0, up=True)
+            self.end_drag(0)
```
xavier630, line 163: We could maybe read the active buttons down and only push 'up' for the active ones rather than all every time. Quite a minor optimisation though

misc/mouse.talon
```diff
@@ -6,8 +6,10 @@ touch:
 	mouse_click(0)
 	# close the mouse grid if open
 	user.grid_close()
```
xavier630, line 8: This is added because it's quite hard to stop a right drag

Without this only `righty` and `right drag` stop it whereas users are likely used to `touch` stopping drag commands since `touch` stops the left drag command

code/mouse.py
knausj85: At this point, you can safely look at ctrl.mouse_buttons_down() to end the correct drag(s). The relevant fix to make the button indices consistent across platforms was released in v0.2.

It might make sense to make end_drag parameterless?

misc/mouse.talon
```diff
@@ -47,8 +49,12 @@ midclick:
 	mouse_click()
 	# close the mouse grid
 	user.grid_close()
-drag: 
-	user.mouse_drag()
+drag:
+	user.mouse_drag(0)
+	# close the mouse grid
+	user.grid_close()
+right drag:
```
knausj85, line 56: This is pretty nitpicky, but I'd vote to stay with "righty drag" for consistency

code/mouse.py
```diff
@@ -148,20 +148,13 @@ def mouse_trigger_zoom_mouse():
...
-            # print("start drag...")
-            ctrl.mouse_click(button=0, down=True)
-            # app.notify("drag started")
-        else:
-            # print("end drag...")
-            ctrl.mouse_click(button=0, up=True)
+        ctrl.mouse_click(button=button, down=True)
 
-        # app.notify("drag stopped")
+    def end_mouse_drag(button: int):
```
knausj85, line 155: I'd say stick with the "mouse_{name here}" convention, so mouse_drag_end would be preferable.

xavier630:
Hey @knausj85, thanks for the feedback. I think I've implemented all three of your suggestions. Let me know what you think.

fidgetingbits:
this seems reasonable to me. the only thing I might guess is an issue is now there is `left drag`,  `right drag`, and `drag`. I think because `left` is also a default keyboard command, there's actually some risk that talon will pick up this as keyboard `left` and then mouse `drag` instead of `left drag`, although I think it should favor the longer of the two commands? for now I think it should be fine so will merge, but if it becomes a problem we'll have to revert to `lefty drag` and `righty drag` only

PR #502, opened 2021-07-22
Fix windows regressions

timo95:
Some windows apps broke with the recent .talon -> .py transition.


knausj85:
Thanks for fixing these!

Issue #505, opened 2021-07-25
establish generic_terminal.talon and deprecate terminal.talon (both for linux and mac)


DonnerWolfBach:
As far as I can see, a generic talon grammar like [in the guidelines for contributions](https://github.com/knausj85/knausj_talon#guidelines-for-contributions) should be used whenever possible. And while there is generic_terminal, it is not really used. The terminals on linux use the [terminal tag](https://github.com/knausj85/knausj_talon/blob/master/apps/platforms/linux/terminal.talon) and there's a seperate and mostly redundant [apple_terminal](https://github.com/knausj85/knausj_talon/blob/master/apps/platforms/mac/terminal/terminal.talon). I [discussed briefly with knausj](https://talonvoice.slack.com/archives/CKYAFAADA/p1627161632181100?thread_ts=1627161305.177400&cid=CKYAFAADA) whether a refactor would be necessary and it seems like it is.

My proposal:

1. establish generic_terminal [in the guidelines for contributions](https://github.com/knausj85/knausj_talon#guidelines-for-contributions)
2. deprecate terminal.talon (both the linux and the mac version) and move to generic_terminal
3. For operating system specific commands (like maybe "resume" in [apple_terminal](https://github.com/knausj85/knausj_talon/blob/master/apps/platforms/mac/terminal/terminal.talon)) create files like apple_terminal or linux_terminal

knausj85:
This sounds reasonable, thanks. I would note generic_terminal is used for most of terminal support on Windows, which may or may be useful for adding support.

fidgetingbits:
I agree that we should move to `generic_terminal.talon` and we can get rid of `terminal.talon`, but i don't entirely understand why does this would involve removing the `terminal` tag (mentioned in your title, but not in the proposal)? Other things like the `generic_browser.talon` also have a `browser` tag for instance. Something needs to assert the `terminal` tag in order to trigger `generic_terminal.talon` anyway no? 

[...]

PR #514, opened 2021-07-29
Knausj generic terminal establishment

DonnerWolfBach:
solves #505


apps/platforms/win/powershell/power_shell.py
```diff
@@ -97,6 +94,10 @@ def terminal_clear_screen():
     def terminal_run_last():
         actions.key("up enter")
 
+    def terminal_rerun_search(command: str):
+        actions.key("ctrl-r")
```
fidgetingbits, line 98: TIL powershell supports this lol

fidgetingbits (approved):
looks good to me

knausj85 (approved):

knausj85:
Looks great, once you think it's ready we will merge it

DonnerWolfBach:
I think there are some things left to do, like deciding wether terminal_list_all & co should be defined in the terminal specific file or in generic_terminal.py.

One question about the existing terminals: should they all be first transitioned to the new "architecture" or should this be done in later PRs?

apps/platforms/win/powershell/power_shell.py
DonnerWolfBach: To be honest, me too.

knausj85:
Generally, things should work -- so the existing terminals should be migrated where it's reasonable. This does introduce some commands that will not work for e.g. gnome-terminal and guake as-is. I personally can live with this though.

I'm firmly on the fence re: "terminal_list_all & co should be defined in the terminal specific". I'm fine with the current approach, but we could have a `generic_unix_shell` tag for these super common implementation for terminals to opt into, as fidget suggested.

PR #523, opened 2021-08-01
Add Pound Sterling (£) Sign

xavier630:
This adds a symbol for the british currency, the pound sterling.

Relates to https://github.com/knausj85/knausj_talon/issues/512

This also sets us up as @rntz suggested to be able to use the term `{currency} sign` for various other currencies.

Since this removes one of the commands which creates a `#` symbol, I'd love to get multiple approvals and welcome any discussion.

### DIFF
--- a/code/keys.py
+++ b/code/keys.py
@@ -171,7 +171,6 @@ symbol_key_words = {
     "plus": "+",
     "tilde": "~",
     "bang": "!",
-    "dollar": "$",
     "down score": "_",
     "under score": "_",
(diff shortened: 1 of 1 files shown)


rntz:
I'm okay with this, but it does change the meaning of the "pound" command, so I'd like to get @knausj85 ’s opinion at least.

knausj85:
I'm okay with this. I'll merge in a few days if no-one objects.

knausj85 (approved):

PR #524, opened 2021-08-01
Add a right wipe (ripe) command to Talon

xavier630:
Currently `forward delete` exists but is quite tricky to say often.

### DIFF
--- a/misc/standard.talon
+++ b/misc/standard.talon
@@ -21,8 +21,9 @@ undo that: edit.undo()
 redo that: edit.redo()
 paste match: edit.paste_match_style()
 file save: edit.save()
-wipe: key(backspace)    
-(pad | padding): 
+wipe: key(backspace)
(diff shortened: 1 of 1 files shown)


xavier630:
Even though it's only one sound off wipe, from some brief testing I find that it is almost always detected correctly

rntz:
We already have "clear right" for this. It's common and expected to customize your editing commands, but I'm not sure we want to add such customizations to base knausj_talon unless they're definitely worth it. I'm also very nervous about ripe/wipe. It may work consistently with your accent, microphone, and choice of speech engine, but that doesn't guarantee it will work for everyone. There are certain accents which make "r" and "w" sounds quite similar, so I don't think we should make two commands that differ only by those sounds.

xavier630:
Yeah that's reasonable. I'll close this and keep it for just my fork.

PR #527, opened 2021-08-01
Fix select word on Windows and Linux

xavier630:
Fixes the select word edge case [here](https://github.com/knausj85/knausj_talon/issues/526).

The edge case involves having the cursor just before the very first character of the word and the `select word` command selecting the word before the current word.

It would be great if any windows or linux users could give this a quick test since I'm not able to run it.

### DIFF
--- a/code/platforms/win/edit.py
+++ b/code/platforms/win/edit.py
@@ -125,12 +125,9 @@ class EditActions:
         #action(edit.select_paragraph):
         #action(edit.select_sentence):
     def select_word():
-        actions.key('ctrl-left ctrl-shift-right')
-        #action(edit.selected_text): -> str
-        #action(edit.sentence_end):
-        #action(edit.sentence_next):
-        #action(edit.sentence_previous):
-        #action(edit.sentence_start):
+        actions.edit.right()
+        actions.edit.word_left()
+        actions.edit.extend_word_right()
     def undo():
         actions.key('ctrl-z')
     def up():
--- a/code/platforms/linux/edit.py
+++ b/code/platforms/linux/edit.py
@@ -125,12 +125,9 @@ class EditActions:
         #action(edit.select_paragraph):
         #action(edit.select_sentence):
     def select_word():
-        actions.key('ctrl-left ctrl-shift-right')
-        #action(edit.selected_text): -> str
-        #action(edit.sentence_end):
-        #action(edit.sentence_next):
-        #action(edit.sentence_previous):
-        #action(edit.sentence_start):
+        actions.edit.right()
+        actions.edit.word_left()
+        actions.edit.extend_word_right()
     def undo():
         actions.key('ctrl-z')
     def up():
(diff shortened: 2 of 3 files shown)


rntz:
This seems to work for me on linux.

rntz (approved):

code/platforms/win/edit.py
```diff
@@ -125,12 +125,9 @@ def select_none():
         #action(edit.select_paragraph):
         #action(edit.select_sentence):
     def select_word():
-        actions.key('ctrl-left ctrl-shift-right')
-        #action(edit.selected_text): -> str
-        #action(edit.sentence_end):
-        #action(edit.sentence_next):
-        #action(edit.sentence_previous):
-        #action(edit.sentence_start):
+        actions.edit.right()
```
xavier630, line 128: Quick flag that this strategy gets duplicated 3 times (once for each OS) and could probably be hoisted up one level of abstraction to the generic browser level avoid duplication

knausj85:
This doesn't seem to work on Windows, I'll have to take a peek later.

rntz:
Do we have any windows users who can take a look at this? @splondike ?

splondike:
I'm a Linux user also.

codecat555:
I tried this change on windows and it does seem to work as advertised.

Testing raised a question in my mind about what exactly constitutes a word in this context. I haven't looked to see how it's done in the code, but here are some test cases that got me thinking:

```
-merry-go-round
merry:go:round
merry/go/round
```
How words are defined usually depend on both the app and the user. That's one point, but beyond the scope of this change.

On the other hand, what should happen when the user says `select word` and the cursor is:
1. before the separator character?
2. after the separator character?
3. at the beginning of a line that starts with a word followed by some sort of separator character?
4. at the beginning of a line that ends with a word proceeded by some sort of separator character?

These various cases gave me some unexpected results on the data set above, and the results were inconsistent between apps (e.g. the text box in this pull request and the windows notepad.exe editor).

codecat555:
Oh yeah, I forgot some other cases:

5. one space to the left of a word which is separated from the previous word by more than one space?
6. one space to the right of a word which is separated from the next word by more than one space?
7. exactly between two words with an even number of spaces on either side.


note: assuming left-to-right reading order, which is yet another variable to consider here.

tararoys:
-In Windows,
  - In notepad,
  -  with the cursor positioned at the beginning of a word, 
  - this branch will select the word and not the previous word.

-In Windows,
  - In Sublime,
  -  with the cursor positioned at the beginning of a word, 
  - this branch will select the word and not the previous word.

-In Windows,
  - In firefox,
  - in a div set to contenteditable
  -  with the cursor positioned at the beginning of a word, 
  - this branch will select the word and not the previous word.

Based on these three tests I say the patch does what it says it does in Windows.

rntz:
Going to merge this since it seems like it behaves better than the status quo and I couldn't find corroboration of problems on windows.

PR #531, opened 2021-08-01
add phones word/word left/word right commands

rntz:
useful for applying homophones to the word left of/right of/at cursor

### DIFF
--- a/code/edit.py
+++ b/code/edit.py
@@ -26,3 +26,13 @@ class Actions:
             actions.edit.paste()
             # sleep here so that clip.revert doesn't revert the clipboard too soon
             actions.sleep("150ms")
+
+    def words_left(n: int):
+        """Moves left by n words."""
+        for _ in range(n):
+            actions.edit.word_left()
+
+    def words_right(n: int):
+        """Moves right by n words."""
+        for _ in range(n):
+            actions.edit.word_right()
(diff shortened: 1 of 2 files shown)


knausj85:
I would suggest adding an optional `<number_small>` to those phones word left/right commands.

rntz:
The idea being to select the nth word in that direction, so eg if you're after "foo bar baz" `phones two words left` would select "bar"? That should be doable...

xavier630:
Not in scope for this PR but we should make a generic selection generator that applies selection before applying a command since it's a pattern that keeps coming in more places and all of the selections are identical (`phones, cut, copy, snake, hammer` etc all have/will eventually have the same selections `left, right, word, line, all, etc` variants that all need to be maintained and are basically identical).

Will raise an issue tomorrow.

rntz:
I have some thoughts on that, but you should check out the text navigation commands in https://github.com/knausj85/knausj_talon/blob/master/text/text_navigation.talon, which may already do some of what you're imagining.

rntz:
ok, I've updated this to allow eg "phones second word left" or "phones third word right". it did involve adding some new actions for moving n words left or right (had to do this instead of using core.repeat() because needed to handle moving 0 words right/left, which is the default).

PR #533, opened 2021-08-02
Add Evernote module to Talon

xavier630:
There are a couple of things I'd like to add to this module eventually, but the main pain point of using Talon with Evernote is that it's a text editor but `slap` often gets jumbled. Adding the key delay fixes that. Hopefully this helps others.

I played around with different delays for a while to try to find one that made slap always work while introducing minimal delay. If anyone else ends up having issues with this though, we can always increase it later on.

Note that this is an issue that I have on Evernote for Mac. In the past I found the Windows app to be much faster and we may end up wanting to split this into a Mac-only delay.

### DIFF
--- a/apps/evernote/evernote.talon
+++ b/apps/evernote/evernote.talon
@@ -2,5 +2,5 @@ app: evernote
 -
 settings():
     # Necessary to stop commands like 'slap' getting jumbled
-    key_wait = 4.0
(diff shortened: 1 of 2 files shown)


rntz:
If you make this mac-specific then I'll merge it.

xavier630:
Sounds good, will do that

PR #534, opened 2021-08-02
Add a command to copy the url of the current page

xavier630:
I was finding that this was a really common workflow and wanted to speed it up. I figured others could benefit from it.

This current implementation is just for Chrome, but is generic enough to work with all browsers once we have a shared place to put it. I just didn't want to duplicate it four times without the shared place. For context see (https://github.com/knausj85/knausj_talon/issues/532).


apps/generic_browser.talon
```diff
@@ -1,6 +1,7 @@
 tag: browser
 -
 (address bar | go address | go url): browser.focus_address()
+(address copy | url copy | copy address | copy url ): user.address_copy()
```
xavier630, line 4: Quite a few different commands here as I try to stay consistent with the existing commands in the file but also meet the desired object -> verb standards

rntz:
This probably doesn't need to be an action; since it will have the same implementation across ask browsers so far, you can put it in generic_browser.talon.

xavier630:
> This probably doesn't need to be an action; since it will have the same implementation across all browsers so far, you can put it in generic_browser.talon.

Hmm is this the current stance? I was under the impression we were trying to move implementations away from .talon files.

rntz:
*action* implementations in .talon files are deprecated and eventually will stop working, those look like this:

```
action(edit.undo): key(ctrl-/)
```

But voice commands in .talon files aren't going away, and not every voice command needs to be implemented by a single action. This would be a fairly normal voice command:

```
copy address:
  browser.focus_address()
  sleep(50ms)
  edit.copy()
```

xavier630:
I imagine the downside is that this becomes hard to overwrite for specific cases or to call in another action later? But yep can do if that's preferred 👍

xavier630:
@rntz it's now all in .Talon files

rntz (approved):

PR #535, opened 2021-08-02
Make slap command work with Slack

xavier630:
Without this Slack currently submits messages and comments rather than creating a new line.

The implementation involves can turning slap into an action and then of overwriting that action for Slack alone.


code/line_commands.py
```diff
@@ -44,3 +44,7 @@ def camel_right():
     def line_clone(line: int):
         """Clones specified line at current position"""
 
+    def slap():
+        """Move cursor to the right by camel case/subword"""
+        actions.edit.line_end()
+        actions.key("enter")
```
xavier630, line 50: Quick flag that it was hard to decide if this belonged in `line_commands.py` or `edit.py` since they both had very similar (overlapping?) commands

apps/slack/mac.py
```diff
@@ -9,6 +9,10 @@
 
 @ctx.action_class("user")
 class UserActions:
+    def slap():
```
rntz, line 12: What is the conceptual action that user.slap() is actually encapsulating here? It ought to be named after that instead of something ambiguous like "slap". Also it needs an accurate docstring.

code/line_commands.py
rntz: It should probably be in edit.py, since it has nothing to do with line numbers.

rntz (changes requested):
Besides the changes I requested, this is a significant change to a basic command, so I'd appreciate a second opinion.

apps/slack/mac.py
xavier630: Yeah, good call. And nice catch on the docstring.

xavier630:
@rntz, I've renamed the action, updated the docstring and moved it to edit.py

Let me know if you have any other thoughts 👍

apps/slack/slack.py
```diff
@@ -13,3 +14,9 @@
 os: mac
 and app.bundle: com.tinyspeck.slackmacgap
 """
+
+@ctx.action_class("user")
+class UserActions:
```
xavier630, line 19: @rntz, I think that this is actually applying outside the Slack context. Any tips on how to make this override only apply to Slack?

pokey:
```py
ctx.matches = r"""
app: slack
"""
```

code/edit.py
```diff
@@ -36,3 +36,8 @@ def words_right(n: int):
         """Moves right by n words."""
         for _ in range(n):
             actions.edit.word_right()
+
+    def insert_blank_line_below():
```
pokey, line 40: Is this not the same as [`edit.line_insert_down`](https://github.com/knausj85/knausj_talon/blob/dded1d6e83c9cedad12ce7928e5c317b25f8306f/code/platforms/mac/edit.py#L87-L88)?

apps/slack/slack.py
xavier630: Ah, the context needs to be defined as well as the mod? Makes sense - thank you

xavier630:
@pokey, added the context and hooked into the basic edit command 👍

fidgetingbits (commented):
I might be missing something, but isn't this missing the default slap behavior now if the action isn't defined? if so this will break slap everywhere outside of slack? i don't use it, but I feel other people probably do so would have to add a default action for common editor apps (or wherever people are using it) that preserves the old behavior? or was that the case already with line_end() action not being defined in most?

xavier630:
I don't think I fully understand your worries. 

Slap works everywhere outside Slack and default behaviour for `insert_line_down` already existed without needing to be 
implemented in this PR. I've now moved it up to a more generic level rather than an operating system level but I think that either way it should work.

Let me know if I'm missing something here.

fidgetingbits:
OK, I misunderstood. thought it was a new action

PR #536, opened 2021-08-02
implement cancel command properly

rntz:


code/cancel.py
```diff
@@ -0,0 +1,15 @@
+from talon import speech_system, actions
+import logging
+
+cancel_phrase = "cancel cancel".split()
+
+def pre_phrase(d):
+    logging.info(d)
```
rntz, line 7: Urk, I should remove this line.

misc/cancel.talon
```diff
@@ -0,0 +1,6 @@
+# the actual behavior of "cancel cancel" is implemented in code/cancel.py
+# it allows you to prevent a command from executing by ending it with "cancel cancel"
```
pokey, line 2: Maybe a quick NB that if they want to use a different phrase they also need to change it in the Python file? Alternatively could unify to single source of truth with a capture

pokey (approved):
LGTM

misc/cancel.talon
rntz: Added a note

PR #538, opened 2021-08-03
`quote that` and `inside quotes` changes

xavier630:
1. Adds a command to insert single quotes around text
2. Makes the command for single quotes and double quotes consistent with [keys.py](https://github.com/knausj85/knausj_talon/blob/master/code/keys.py#L159-L201)
3. Adds a 'string' option when making single quotes in order to be consistent with the 'string' command

Because this changes the behavior of the existing `quote that` and `inside quotes` commands, I'd love to get a few reviews and some discussion.

### DIFF
--- a/text/symbols.talon
+++ b/text/symbols.talon
@@ -41,7 +41,7 @@ inside (bracket | braces):
 inside percent:
 	insert("%%")
 	key(left)
-inside quotes:
+inside (quotes | string):
 	insert("''")
 	key(left)
 inside (double quotes | dubquotes):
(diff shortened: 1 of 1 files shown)


text/symbols.talon
```diff
@@ -62,7 +65,10 @@ angle that:
 percent that:
     text = edit.selected_text()
     user.paste("%{text}%")
-quote that:
+(quote | string) that:
```
rntz, line 68: `string that` is already a command, since `string` is a formatter and `<user.formatters> that` is a command. so probably just make this "quote that".

rntz (commented):
lgtm, except for a niggle about "string that" already being a command.

text/symbols.talon
xavier630: Good call, thanks

xavier630: Okay, amended

PR #540, opened 2021-08-03
Ignore settings directory

SCdF:
As settings are now locally generated we should ignore the directory it creates

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 # Dev stuff
 __pycache__
 *.sw?
+# Locally generated
+settings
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 # Dev stuff
 __pycache__
 *.sw?
+# Locally generated
+settings
```
rntz, line 5: should be /settings; we only want to ignore the specific settings directory in the root, not all files named "settings"

rntz (changes requested):

rntz (approved):

PR #542, opened 2021-08-03
Add support for skipping occurrence

SCdF:


code/multiple_cursors.py
```diff
@@ -24,6 +24,9 @@ def multi_cursor_select_fewer_occurrences():
     def multi_cursor_select_more_occurrences():
         """Adds cursor at next occurrence of selection"""
 
+    def multi_cursor_select_skip_occurrences():
```
pokey, line 27: I find this action name a bit confusing.  How about `multi_cursor_skip_occurrence` instead?

pokey (commented):
this looks good to me other than renaming action for clarity

code/multiple_cursors.py
SCdF: okey dokey

SCdF:
Function name change

pokey (approved):

PR #543, opened 2021-08-03
Add SNAKE_CAPS_SUPPORT

SCdF:


rntz:
You can already do this with `all caps snake`, and basically anywhere that accepts a formatter also accepts a list of formatters: eg. `user.formatted_text(whatever, "ALL_CAPS,SNAKE_CASE")` should work.

it is a bit unintuitive that you have to do it in that order, though. I think that's because `snake` downcases things on its own, so the order in which the formatters are applied matters.

I'm going to close this. Reopen if you don't think combining formatters covers your use case, or open an issue to discuss whether/how formatters should work differently if you think having to give them in a specific order is annoying.

PR #544, opened 2021-08-03
Tap support

SCdF:
Allows you to tap modifier keys, for example to "tap alt(er)" to show the
window menu, or "tap super" to show the OS menu (eg Start Menu)


rntz (approved):

rntz:
I'm happy with this PR, but since it adds a new global command I'd like input from some other maintainers - @pokey @splondike thoughts?

In particular, is `tap` the right word here? One alternative would be `press`. Another would be to have no verb at all and just allow saying `super` or `control` to press those keys -- unfortunately `alt` and `odd` are a bit similar so I think this is a bad idea to do by default. So I think `tap` is fine; any other opinions?

If nobody objects, I'll merge this in a few days.

fidgetingbits:
Seems fine to me. I think I would prefer an explicit command rather than just saying the key name, but maybe just because it's something I don't do often. My only reservation about using `tap` is I can imagine some occasional overlap with `trap` (a least on my set up I'm sure that will happen), in that sense I would opt for `press`.

rntz:
@SCdF can you change tap to press? If you do then I will merge this.

Issue #528, opened 2021-08-01
Functionality to strike through text


xavier630:
I use strike-through a lot in my day-to-day, and I think a very valuable formatter and `{x} that` command would be the ability to strike through text like below:

~Here is some text~.

rntz:
If by strikethrough, you mean surrounding with tildes, `~example~`, then this could be done by adding a formatter for surrounding text with tildes. I'm not sure what the right name for it would be, though, since putting things between tildes is not a standard convention for strike-through -- AFAICT it's used by github flavored markdown and Slack. Perhaps something straightforward like "between tildes" would be a good name. Then you could say "between tildes that" to strike out something.

I don't have time to do this, but you can probably figure out how to do it by taking a look at code/formatters.py and misc/formatters.talon.

xavier630:
I was thinking it could be done in a generic application-agnostic way (since your tilde idea works for Github and Slack but not many other places that I know of like text editors or Notes/Messaging apps).

I looked into unicode a bit and it looks [semi-promising](https://stackoverflow.com/a/27235933/8091600). I naively assumed that every common character would have a strikethrough equivalent unicode character to convert to but that's not the case.

RE doing the work, don't worry, I plan to try the unicode approach myself this week, I just raised the issue to track it. If I have no luck though I'll echo back.

splondike:
Had a call with rntz, knaus, pokey, nicholas riley. We'd suggest an approach where the strikethrough command is activated via a tag and uses an action that implements the strikethrough action (which would be e.g. to wrap in ~ or to press a keyboard shortcut). The default action would be to wrap in ~.

PR #545, opened 2021-08-03
Adds strikethrough functionality to Talon..

xavier630:
..as a formatter

Relates to https://github.com/knausj85/knausj_talon/issues/528

**Example**

> This is regular text
> t̶h̶i̶s̶ ̶i̶s̶ ̶t̶e̶x̶t̶ ̶w̶i̶t̶h̶ ̶t̶h̶e̶ ̶s̶t̶r̶i̶k̶e̶t̶h̶r̶o̶u̶g̶h̶ ̶f̶o̶r̶m̶a̶t̶t̶e̶r̶ 

**Of note:**

1. Pasting has to be used for writing formatters (`actions.user.paste()`) since Talon can't insert special characters in some cases. This is generally fine and we do it all over the place, but may be slightly vexing for those with clipboard history managers. 
I have an [issue](https://github.com/knausj85/knausj_talon/issues/539) open to make formatters specify whether they need pasting or are happy with insertion.

2. This adds the `no format` formatter in order to undo strikethrough (and other future formatters)
3. The formatter doesn't yet preserve case (since all formatters call `unformat` by default which clears case). We can track this as future work.
I have an [issue](https://github.com/knausj85/knausj_talon/issues/539) open to make formatters specify whether they need `unformat` applied.

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -103,6 +103,7 @@ def every_word(word_func):
     return formatter_function
 
 
+# TODO - these tuples getting quite complicated, we should make a class with a clear, documented structure instead
 formatters_dict = {
     "NOOP": (SEP, lambda i, word, _: word),
     "DOUBLE_UNDERSCORE": (NOSEP, first_vs_rest(lambda w: "__%s__" % w)),
@@ -297,25 +298,30 @@ class Actions:
 def strikethrough_character(character):
     return character + u'\u0336'
 
+
 def strikethrough_word(word):
     return "".join([strikethrough_character(character) for character in word])
 
+
 symbols_to_ignore_when_unformatting = [
-    u'\u0336' # Strikethrough character
+    u'\u0336'  # Strikethrough character
 ]
+
 
 def unformat_text(text: str) -> str:
     """Remove format from text"""
 
     # Remove certain symbols which don't signify a new word
+    unformatted = text
     for symbol in symbols_to_ignore_when_unformatting:
         unformatted = text.replace(symbol, '')
 
-    # Replace all remaining symbols with spaces
+    # Replace symbols with spaces
     unformatted = re.sub(r"[^a-zA-Z0-9]+", " ", unformatted)
 
     # Split on camelCase, including numbers
-    unformatted = re.sub(r"(?<=[a-z])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])|(?<=[a-zA-Z])(?=[0-9])|(?<=[0-9])(?=[a-zA-Z])", " ", unformatted)
+    unformatted = re.sub(r"(?<=[a-z])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])|(?<=[a-zA-Z])(?=[0-9])|(?<=[0-9])(?=[a-zA-Z])",
+                         " ", unformatted)
     # TODO: Separate out studleycase vars
     return unformatted.lower()
 
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -103,7 +103,7 @@ def every_word(word_func):
     return formatter_function
 
 
-# TODO - these tuples getting quite complicated, we should make a class with a clear, documented structure instead
+# TODO - these tuples getting quite complicated, we should make a formatter class instead
 formatters_dict = {
     "NOOP": (SEP, lambda i, word, _: word),
     "DOUBLE_UNDERSCORE": (NOSEP, first_vs_rest(lambda w: "__%s__" % w)),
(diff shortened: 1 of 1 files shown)


code/formatters.py
```diff
@@ -148,6 +156,7 @@ def formatter_function(i, word, _):
     "allcaps": formatters_dict["ALL_CAPS"],
     "alldown": formatters_dict["ALL_LOWERCASE"],
     "camel": formatters_dict["PRIVATE_CAMEL_CASE"],
+    "no format": formatters_dict["NOOP"],
```
rntz, line 159: Is there a reason why you added this in this PR? Seems like an unrelated change. Submit it in a different PR if you think that it's useful, best not to mix unrelated changes together unless they're really minor (like whitespace or comment formatting fixes).

xavier630: Yeah I explain it in the description. It's to undo formatting like strikethrough and later bold.

rntz: Oops, shoulda read the description more carefully. Sorry!

xavier630: I'll remove it for now since maybe other formatters cover it but definitely related directly to this

code/formatters.py
```diff
@@ -112,6 +113,13 @@ def formatter_function(i, word, _):
         first_vs_rest(lambda w: w, lambda w: w.capitalize()),
     ),
     "PUBLIC_CAMEL_CASE": (NOSEP, every_word(lambda w: w.capitalize())),
+    "STRIKETHROUGH": (
+        NOSEP,
+        first_vs_rest(
+            lambda w: strikethrough_word(w),
+            lambda w: strikethrough_character(" ") + strikethrough_word(w),
```
xavier630, line 120: Should maybe call strikethrough word on the " " actually for consistency

code/formatters.py
```diff
@@ -112,6 +113,13 @@ def formatter_function(i, word, _):
         first_vs_rest(lambda w: w, lambda w: w.capitalize()),
     ),
     "PUBLIC_CAMEL_CASE": (NOSEP, every_word(lambda w: w.capitalize())),
+    "STRIKETHROUGH": (
+        NOSEP,
+        first_vs_rest(
+            lambda w: strikethrough_word(w),
+            lambda w: strikethrough_word(" ") + strikethrough_word(w),
```
xavier630, line 120: As a minor optimisation, we could probably store the result of `strikethrough_word(" ")` as a variable so that it doesn't get rerun on every lambda call.

I don't think it really matters as much though from a performance perspective and it probably reduces understandability.

nriley:
I was worried that this would affect screen readers but, at least with VoiceOver, it doesn't. It doesn't recognize the strikethrough characters as such but it also doesn't cause the problems I've seen with other "formatters" that use Unicode characters that look like letters but often aren't.

splondike:
Hmm, the unicode approach is a little inconsistent for me. Puts hyphen like things between characters in gedit, works fine in gVIM, misses spaces in my terminal (kitty with the Hack font), works well in github/stackoverflow, OK in slack. Deleting struck through text takes two backspaces rather than 1 per character.

That all said, I think this is fine because if the application in question has 1st party strikethrough support you can just override this pretty easily. E.g. the following for Abiword:

```
strikethrough that: key(ctrl-k)
strikethrough <phrase>:
    key(ctrl-k)
    insert(phrase)
    key(ctrl-k)
```

@xavier630 Happy to merge this in if you're ready.

rntz:
I'm pretty dubious of this. Strike-through via Unicode is weird and almost always produces weird looking results for me, and I'm not sure that treating it as a formatter if what you really want is a rich-text like feature (toggling strike through in Slack, eg) is the right approach.

Also, overriding commands by binding the same command in a special context is generally to be avoided in favour of overriding an action. But that's incompatible with formatters, which makes me think formatters are not the right place for this feature.

rntz:
I'd be okay with a strike-through command that invoked a corresponding action. The default implementation could even be this Unicode approach. The benefit over treating it as a formatter is that it would be easier to override. It could be worth making a related group of rich-text commands (for bold, italic, underline, etc). Since not all applications support this we might want a tag for it. I think this would be better than cramming these features into formatters, which (despite the name) I think are really about producing plain-text output.

Does that seem reasonable? Agree/disagree?

splondike:
Yeah, above about making a tag activated rich text thing with default implementation for strikethrough involving unicode sounds reasonable to me.

splondike:
Had a call with rntz, knaus, pokey, nicholas riley. We'd prefer an approach where the strikethrough command is activated via a tag and uses an action that implements the strikethrough action (which would be e.g. to wrap in ~ or to press a keyboard shortcut). The default action would be to wrap in ~.

@xavier630 Since that's pretty much a rewrite of this PR, would you be happy to close it out and just keep the existing issue?

splondike:
As mentioned above, the consensus was that this would be better implemented as an action rather than a formatter since a lot of the time it'll be natively implemented as a shortcut key.

Closing out this pull request, further deliberation will happen in #528.

PR #547, opened 2021-08-03
Feature/GitHub.com (#1)

ViacheslavKudinov:
Added more github.com shortcuts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 # Dev stuff
 __pycache__
 *.sw?
+
+# editor stuff
+.idea
+.vscode
--- a/apps/web/github.talon
+++ b/apps/web/github.talon
@@ -18,8 +18,9 @@ go to code: insert("gc")
 go to issues: insert("gi")
 go to pull requests: insert("gp")
 go to wiki: insert("gw")
-
-# network graph
+go to actions: insert("ga")
+go to projects: insert("gb")
+go to discussions: insert("gg")
 
 # source code browsing
 find file: key(t)
@@ -27,6 +28,8 @@ jump to line: key(l)
 switch (branch|tag): key(w)
 expand url: key(y)
 (show|hide) [all] in line notes: key(i)
+open blame [view]: key(b)
+(show|hide) annotations: key(a)
 
 # issues
 create [an] issue: key(c)
@@ -53,5 +56,17 @@ other parent commit: key(o)
 mark as read: key(y)
 mute thread: key(shift-m)
 
-# pull request list
+# issue or pull request list
 open issue: key(o)
+create issue: key(c)
+
+# issues and pull requests
+request reviewer: key(q)
+set milestone: key(m)
+set assignee: key(a)
+set label: key(l)
+
+# actions
+go to workflow [file]: insert("gf")
+toggle timestamps [in logs]: key(shift-t)
+toggle fullscreen [logs]: key(shift-f)


.gitignore
```diff
@@ -1,5 +1,10 @@
 # Dev stuff
 __pycache__
 *.sw?
+
+# editor stuff
+.idea
+.vscode
```
pokey, line 7: I wonder if we actually want to keep this in order to standardise things? Eg location of Python stubs?

apps/web/github.talon
```diff
@@ -7,51 +7,87 @@ win.title: /github.com/
 focus search: key(s)
 go to notifications: insert("gn")
 go to dashboard: insert("gd")
-show keyboard shortcuts: key(?)
+keyboard shortcuts show: key(?)
+selection move down: key(j)
+selection move up: key(k)
+selection toggle: key(x)
+selection open: key(o)
+# legacy [verb noun] commands to be removed at a later time
```
pokey, line 15: I have mixed feelings bout leaving these, because people may just continue using them and still be surprised when they are removed later. 

How do we want to handle these deprecations in general? Just remove them? Replace them with a notification? Leave them with a note like has been done here? Cc/ @rntz @knausj85 

Would be interesting to use the code that generates help to generate a diff of commands that have been added / removed between two commits so people can easily check what has changed when they merge

apps/web/github.talon
```diff
@@ -7,51 +7,87 @@ win.title: /github.com/
...
+blame [view] open: key(b)
+(show|hide) annotations: key(a)
+# legacy [verb noun] commands to be removed at a later time
+expand url: key(y)
+switch (branch|tag): key(w)
+find file: key(t)
 
 # issues
-create [an] issue: key(c)
+[an] issue create: key(c)
```
pokey, line 45:
```suggestion
issue create: key(c)
```

apps/web/github.talon
```diff
@@ -7,51 +7,87 @@ win.title: /github.com/
...
 # source code browsing
-find file: key(t)
+file find: key(t)
 jump to line: key(l)
-switch (branch|tag): key(w)
-expand url: key(y)
+(branch|tag) switch: key(w)
+url expand: key(y)
 (show|hide) [all] in line notes: key(i)
+blame [view] open: key(b)
```
pokey, line 37: I'd lean towards "show"? Also skeptical of the optional "view"

apps/web/github.talon
```diff
@@ -7,51 +7,87 @@ win.title: /github.com/
...
+create issue: key(c)
+
+# issues and pull requests
+reviewer request: key(q)
+milestone set: key(m)
+assignee set: key(a)
+label set: key(l)
+
+# actions
+go to workflow [file]: insert("gf")
```
pokey, line 91: Should this not be with the other "go to" commands?

apps/web/github.talon
```diff
@@ -7,51 +7,87 @@ win.title: /github.com/
...
+toggle timestamps [in logs]: key(shift-t)
+toggle fullscreen [logs]: key(shift-f)
```
pokey, line 93: How come not object-verb for these? Also not sure the optional captures are helpful

apps/web/github.talon
```diff
@@ -7,51 +7,87 @@ win.title: /github.com/
...
-# pull request list
+# issue or pull request list
+issue open: key(o)
+issue create: key(c)
+# legacy [verb noun] commands to be removed at a later time
 open issue: key(o)
+create issue: key(c)
+
+# issues and pull requests
+reviewer request: key(q)
```
pokey, line 85: Wow really going full object-verb here 😂. Feels awkward but I guess consistent. @rntz @knausj85 so I guess we're going all in on object-verb word order?

pokey (commented):
Overall looks good but left a bunch of comments / questions

apps/web/github.talon
ViacheslavKudinov: As I know this command is actual only if we are in specific actions run like this one https://github.com/ViacheslavKudinov/knausj_talon/runs/3263071796?check_suite_focus=true
It is not similar to other `go to` in the main repository view.

apps/web/github.talon
ViacheslavKudinov: View is probably needs to be not optional.
Terminology/phrase is used from official github shortcuts help page https://docs.github.com/en/get-started/using-github/keyboard-shortcuts#source-code-browsing.
Didn't want to invent something new, not how it was described in doc.

apps/web/github.talon
ViacheslavKudinov: Fixed.

apps/web/github.talon
ViacheslavKudinov: Fixed.

apps/web/github.talon
ViacheslavKudinov: Set `blame view open` according to phrase is used to describe this shortcut 
> Open blame view. For more information, see "Tracing changes in a file."

apps/web/github.talon
ViacheslavKudinov: Changed to `go to workflow`.

.gitignore
ViacheslavKudinov: Reverted back.

apps/web/github.talon
ViacheslavKudinov: @pokey Could you check now, please?

ViacheslavKudinov:
Updated according to remarks.

PR #549, opened 2021-08-04
Make sure homophones always retain case

xavier630:
There is an edge case in https://github.com/knausj85/knausj_talon/issues/501 where homophones picked from the GUI always became lower case no matter what the case of the original word was.

**Example of new case matching when picking from the GUI**
image

**In terms of integration with Talon formatters:**
I spent a bit of time working on making this work with formatters from format.py, but ultimately decided that it was unnecessary since the basic `capitalise` and `upper` formatters cover most (all?) `phones` use cases for now. It shouldn't be very hard to extend to use the Talon formatters later if we ever see a need.

### DIFF
--- a/code/homophones.py
+++ b/code/homophones.py
@@ -61,11 +61,11 @@ PHONES_FORMATTERS = [
     lambda word: word.upper(),
 ]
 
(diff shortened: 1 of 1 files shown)


splondike:
Looks good! Merging.

Issue #454, opened 2021-06-02
[Bug - Chrome] Tab close {nth} closes only one tab


xavier630:
Likely because some sort of sleep needs to be built in to the ordinal form of the tab close command (or a different chrome shortcut needs to be used to select multiple tabs before pressing cmd-w)

Occurring in Chrome 91.0.4472.77 for me.

splondike:
Hmm, "tab close third" works for me on Linux in both Chrome 90.0.4430 and Firefox 88.01. Does the "DECODED" line in Talon show it as having recognised the command in its entirety?

xavier630:
Yep, it shows as understanding the command and closes up to two tabs, but it doesn't seem to close more than two tabs for me.

Mac OS 10.15.7

![image](https://user-images.githubusercontent.com/9077250/120810745-5cd41c80-c55c-11eb-9336-a247ea0946d9.png)

lunixbochs:
Try putting a sleep in your close tab action, after the key is pressed. The problem with sleeps is they vary person to person based on stuff like machine speed and load.

xavier630:
In terms of concrete actions, is it worth putting a sleep into the default library so that this works for every one out of the box?

lunixbochs:
> putting a very small sleep into the default library

What I mean is it's hard to tune such a sleep to actually work for everyone, so you typically need to overshoot the sleep time if you do that.

splondike:
I think it would be fine for you to experiment with how long of a sleep you need to make it behave the way you expect. You can override the tab close action inside the `apps/chrome/chrome.mac.talon` file in your pull request. If the sleep time is < 250ms say I don't think it'd be an issue.

I realize we have some pending pieces of work to migrate all the action implementations from .talon to .py files using the migrator that lunixbochs has made, as well as stop using dots in filenames. I think we'll do those as a single big commit though so anything you do here will be rolled up into that later on.

xavier630:
It could make more sense to make it repeat delays (if we wanted to implement them in a generic way) a possible annotation on actions so that you can take advantage of app/os contexts

PR #553, opened 2021-08-04
Add a small delay to tab close on Chrome

xavier630:
Fixes `tab close` command when repeating across a large number of tabs.

Relates to https://github.com/knausj85/knausj_talon/issues/454.

While this fixes this command for this issue specifically, I'm still supportive of a generic repeat delay framework rather than handling each case individually like this.

For those wondering why there is a wrapper: I had to add an extra layer of abstraction in order to not need to write the specific OS commands (cmd-w vs ctrl-w, etc) while rewriting the function.

https://user-images.githubusercontent.com/9077250/128190337-a8784b91-f11f-46ec-80f7-666415b0416a.mov

### DIFF
--- a/apps/chrome/chrome.py
+++ b/apps/chrome/chrome.py
@@ -34,6 +34,10 @@ class user_actions:
         else:
             actions.key("ctrl-9")
 
+    def tab_close_wrapper():
+        actions.sleep("180ms")
+        actions.app.tab_close()
+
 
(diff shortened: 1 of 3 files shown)


fidgetingbits:
This seems reasonable even though the sleep might not be long enough for some people I suppose. I agree longer term something like a repeat/delay framework as you mentioned in #454 would be nicer.

PR #554, opened 2021-08-04
Add more git commands, including git switch commands

rntz:
Also makes `git new branch` use git switch instead of git checkout, since switch is a more user-friendly way to switch & create new branches.


misc/git.talon
```diff
@@ -59,15 +58,26 @@ git reset soft: "git reset --soft "
 git reset hard: "git reset --hard "
 git restore: "git restore "
 git restore staged: "git restore --staged "
-git remote show origin: "git remote show origin\n"
+git restore source: "git restore --source="
```
xavier630, line 61: I don't think this `=` is intentional

xavier630: Actually maybe it is, sometimes one gets used and sometimes one doesn't
image
image

misc/git.talon
```diff
@@ -59,15 +58,26 @@ git reset soft: "git reset --soft "
...
 git show: "git show "
 git stash pop: "git stash pop\n"
 git stash: "git stash\n"
 git stash apply: "git stash apply\n"
 git stash list: "git stash list\n"
 git stash show: "git stash show"
 git status: "git status\n"
+git switch [<user.text>]: "git switch {user.formatted_text(text or '', 'DASH_SEPARATED')}"
+git switch master: "git switch master "
+git switch main: "git switch main "
```
xavier630, line 76: Could maybe include `develop` if you're including `master` and `main`

fidgetingbits: what's the point of having explicit master and main if they're already covered by the first rule that uses `[<user.text>]`? `develop` will also already be covered by it.

pokey: Just to get talon to favour these I would think.  I'd consider putting an `enter` on these two tho maybe that's a bit aggressive?

fidgetingbits: Makes sense about making talon favor it, may as well keep it then. I use \n for checkouts on my repo already iirc so I'm fine with adding them. I guess sometimes people will legitimately want to add arguments after in which case it will be a bit harder to use, but I suspect that's quite rare.

rntz:
gonna just merge this as-is, if folks want to fiddle with the newline-ness later (or remove the '=' for aesthetic/consistency reasons) we can do that separately - I prefer no-newline by default for basically all commands, but that is a bikeshed for another time.

PR #555, opened 2021-08-05
Allow `window_next` and `window_last`..

xavier630:
..to switch to Mac spaces (fullscreen apps).

Relates to https://github.com/knausj85/knausj_talon/issues/456

Currently when one fullscreens an app into a space, it's quite hard to access again with Talon. This makes it so that it's part of the `window_next`, `window_last` rotation so you can easily switch back and forth between windows even if they use spaces.

If anyone disagrees with altering an existing command like this, we can always make this into a new command.

Apologies for the whitespace editions in the same PR, I needed to format the file it to fix my additions and threw the other fixes in too.

### DIFF
--- a/code/platforms/mac/app.py
+++ b/code/platforms/mac/app.py
@@ -63,3 +63,4 @@ def switch_window_by_offset_from_current(offset):
         if window is active_window:
             index_of_new_window = (index + offset) % len(windows)
             windows[index_of_new_window].focus()
+            return
--- a/code/platforms/mac/app.py
+++ b/code/platforms/mac/app.py
@@ -56,7 +56,7 @@ def switch_window_by_offset_from_current(offset):
     # Handle case where empty windows with no title show up in the results
     windows = list(filter(lambda local_window: len(local_window.title) > 0, windows))
 
-    # Sort by title since they get reordered on every switch and we want to switch windows in a fixed order
+    # Sort by title as the windows get reordered on every switch and we want to switch windows in a fixed order
     windows.sort(key=lambda local_window: local_window.title)
 
     for (index, window) in enumerate(windows):
--- a/code/platforms/mac/app.py
+++ b/code/platforms/mac/app.py
@@ -1,33 +1,45 @@
 from talon import Context, actions, ui
+
 ctx = Context()
 ctx.matches = r"""
 os: mac
 """
 
+
 @ctx.action_class('app')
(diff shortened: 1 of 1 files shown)


code/platforms/mac/app.py
```diff
@@ -1,34 +1,66 @@
...
+
+def switch_window_by_offset_from_current(offset):
+    active_window = ui.active_window()
+    windows = ui.active_app().windows()
+
+    # Handle case where empty windows with no title show up in the results
+    windows = list(filter(lambda local_window: len(local_window.title) > 0, windows))
+
+    # Sort by title as the windows get reordered on every switch and we want to switch windows in a fixed order
+    windows.sort(key=lambda local_window: local_window.title)
```
xavier630, line 60: In theory there's an edge case where two windows have the same title which makes it hard to determine which one is the active one (since the other params like pid and app are the same). 

I'm hoping this is solved in terms of exact matching by using the `is` keyword, but I don't have a good way to solve it for the sorting

Maybe @lunixbochs is able to add an extra identifier to the `ui.window` objects (or is there already an extra ID I'm missing)

SCdF:
FWIW the existing "desk <n>" functionality should allow you to switch between spaces, and in my crossplatform [workspace support](https://github.com/knausj85/knausj_talon/pull/513) PR "desk next" or "desk last" should allow you to switch between spaces, which would include full screened apps. 

(I'm not saying this PR is invalid btw, how window next|previous works is its own thing, just that there is "change spaces" functionality)

xavier630:
👍  I think this + desks will complement each nicely. Good flag on desks being a workaround even without this PR.

nriley:
Please don't replace the existing commands, which I use quite a bit — windows are windows and Spaces are spaces on Mac. "desk" seems fine for a short cross-platform way to refer to it.

pokey:
Could also add a setting that determines whether it's this behaviour or the old behaviour and have it off by default

xavier630:
Happy either way here. I really love this feature, but I don't want to make anything worse for anyone.

If it's not valuable for others, I'll close this in a week or two.

nriley:
I don't want to get in the way of you having the actions in knausj, so if the "desk" actions don't do what you want please feel free to add them.  Just would prefer they not replace existing actions.  You can always change what you call them locally (as I've done plenty…)

PR #560, opened 2021-08-08
Initial markdown support

pokey:


fidgetingbits:
Looks mostly good to me. Recently most of these language files switched to using `mode: user.auto_lang`, instead of `mode: command`, so probably worth matching unless there some reason you didn't. Also could do: 

```
level <number_small>: 
   key("#:{number_small}")
   key(space)
```

and not limit to any arbitrary number?

pokey:
> Looks mostly good to me. Recently most of these language files switched to using `mode: user.auto_lang`, instead of `mode: command`, so probably worth matching unless there some reason you didn't.

Ah right. Haven't done a merge in a while. Will patch that up

> Also could do: 
> 
> 
> 
> ```
> 
> level <number_small>: 
> 
>    key("#:{number_small}")
> 
>    key(space)
> 
> ```
> 
> 
> 
> and not limit to any arbitrary number? 

It would simplify things, tho fwiw I don't think markdown supports more than six levels, so it does add some meaningless commands. Could go either way

fidgetingbits:
I didn't realize it was limited to 6. May as well just leave it as is then.

PR #565, opened 2021-08-13
Fix editor commands being sent when popup windows are open

hbk619:
As related to this issue
https://github.com/anonfunc/intellij-voicecode/issues/7
when a non-editor window of a Jetbrains product is open we still
mark Jetbrains commands as active, meaning things like copy/paste
or 'clear line' go to the editor, rather than the popup window.

This was previously implemented using callbacks for the context in
the original Talon.
https://github.com/anonfunc/talon-user/blob/master/apps/jetbrains.py#L303

These have been replaced with scopes. The current scope updates when
the UI fires an event. This does mean if you open the 'actions' dialogue
(via the command 'please') then it doesn't trigger a UI event and the Jetbrains
context is still active. Other popups do trigger a UI event. I'm not sure what the performance hit
would be running the check every second or two via cron so opted for this
instead. Happy for this to be changed if there is no performance issue :-)

### DIFF
--- a/apps/jetbrains/jetbrains.py
+++ b/apps/jetbrains/jetbrains.py
@@ -138,6 +138,28 @@ and app.exe: rider64.exe
 """
 
 
+@mod.scope
+def is_real_jetbrains_editor():
+
+    window = ui.active_window()
+    active_app = ui.active_app()
+    is_editor = False
+    bundle = active_app.bundle or active_app.name
+    if bundle in port_mapping:
+        # XXX Expose "does editor have focus" as plugin endpoint.
+        # XXX Window title empty in full screen.
+        title = window.title
+        if len(title) == 0 or title == " ":
+            print(f"Window is Jetbrains product but not an editor: {window}")
+        else:
+            is_editor = True
+
+    return {"is_jetbrains_editor": is_editor}
+
+
+ui.register('', is_real_jetbrains_editor.update)
+
+
 @mod.action_class
 class Actions:
     def idea(commands: str):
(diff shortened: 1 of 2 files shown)


xavier630:
Hey, does this mean copy-paste will now work in the terminal/readme? If so that's amazing

hbk619:
I was running Pycharm 2019 and it seemed to work, having tested with RubyMine 2021.2 it doesn't :( There doesn't seem to be any UI events triggered when you _focus_ the terminal but a `win_title` event is triggered when you give the editor focus (although the title doesn't appear to update) so there's no way to tell where you have focus (unless you can request it via the plugin, @anonfunc might know more about that as I guess from the comment I copied from his original code it was on a todo list at some point!).

splondike:
Despite the code being fairly low cost I'd like to restrict when the scope update code is called so it doesn't run at all for us non-jetbrains users. If you did that then you could have it run every second or something. The way to do this with current Talon would be to add a setting, and use `my_setting.register("", my_event_handler)` to listen for changes to the setting. This would then toggle on/off a `cron.interval("1s", ...)` call when the setting was active. You can scope setting values so that they're only active within a particular context, so you'd just enable this setting for the existing jetbrains context (meaning your poller would only activate then). You could probably also replace the use of a scope with just a tag; tags are generally [preferred to scopes](https://talon.wiki/unofficial_talon_docs/#scopes) for this kind of binary thing.

@hbk619 If I'm reading this issue right, it sounds like this PR didn't totally solve the issues you were having. Are you still interested in getting it merged, or do you think we should just close it out?

splondike:
Closing this out.

Issue #562, opened 2021-08-09
Migrate existing terminal implementations to new architecture from #505


DonnerWolfBach:
In #505 a new "architecture" for how terminal should be supported/implemented was defined. However, not all existing implementations where migrated to this new style yet. This issue will contain a list of the missing and already migrated terminals.

Linux:
- ~~kde-konsole~~
- gnome terminal
- guake
- termite

Mac:
- apple terminal (the default terminal)
- iterm

Windows:
- mintty
- powershell
- windows_terminal
- command processor
- wsl/ubuntu

The implementation of kde-konsole serves as a reference of how it should/could be done.

DonnerWolfBach:
@fidgetingbits should we remove termite from the list and add alacritty instead?

fidgetingbits:
may as well just have both termite and alacritty I think, just because it's deprecated doesn't mean people won't use it since it's still stable, and no real harm in keeping it imo.

rntz:
Is this done yet? What's its current state if not?

DonnerWolfBach:
@rntz Still work in progress, but powershell and windows powershell is finished and waits to be merged: #711

PR #568, opened 2021-08-15
Apple terminal: cleanup & migrate to generic_unix_shell commands

### LINKED ISSUE #551: Tab last/next in Mac terminals should switch tabs
`tab next` and `tab last` don't do anything in the Mac terminal.

For a quick look at the code, it looks like the tag is implemented, so I suspect it just needs some custom keybindings.
https://github.com/knausj85/knausj_talon/blob/master/apps/platforms/mac/terminal/terminal.talon#L8

Latest knausj master, August 2nd 2021
Talon 0.2.0-8
Terminal version 2.10 (433)

DonnerWolfBach:
Should fix #551.
Also part of #562.


apps/platforms/linux/kde-konsole/kde-konsole.talon
```diff
@@ -13,6 +13,6 @@ tag(): user.git
 tag(): user.anaconda
```
knausj85, line 13: I'm going to remove this file since it doesn't really belong in this PR.

PR #569, opened 2021-08-16
Prevent talonscript language files from being active in sleep mode

AndreasArvidsson:
Adding `mode: command` to all talonscript language files to prevent them from being active in sleep mode.


fidgetingbits (approved):

lunixbochs:
Please don't overload app: for this

PR #570, opened 2021-08-17
Add list/line reversal functionality

xavier630:
This was a use case I needed that I thought might be useful for other people as well.

It uses paste since many editors (eg Github) infer things like adding extra bullet points when inserting one line at a time.

### DIFF
--- a/code/edit.py
+++ b/code/edit.py
@@ -36,3 +36,13 @@ class Actions:
         """Moves right by n words."""
         for _ in range(n):
             actions.edit.word_right()
+
+    def reverse_lines():
+        """ Reverse the order of the currently selected lines."""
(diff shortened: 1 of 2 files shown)


splondike:
When did this use case come up for you? I think the main time I reverse text is with the '-r' argument to the `ls` command in Linux.

xavier630:
For me it's lists - it is natural to write them with the oldest thing at the top, but, if they are long lived, you often want to flip them so that you don't need to scroll to the bottom of the list to add elements.

Quite specific, so happy to close this and keep it on my personal fork.

splondike:
Use case makes sense, but as you say it might be a bit too specific for knausj_talon (considering that people would have to know it exists as well). I'll close this one out.

PR #576, opened 2021-08-27
Add better support for the java programming language

michaelpog:
* Add all types for java
* Add more operators for java
* Add modifiers for java
* Add shortcuts for methods
* Add arrays

### DIFF
--- a/lang/java/java.talon
+++ b/lang/java/java.talon
@@ -1,0 +1,56 @@
+mode: user.java
+mode: user.auto_lang
+and code.language: java
+-
+tag(): user.code_operators
+tag(): user.code_comment
+tag(): user.code_block_comment
+tag(): user.code_generic
+
+action(user.code_state_import):
+    insert("import ")
+
+# Types Commands
+type {user.code_types}: 
+    insert(code_types)
+    key("space")
+
+type ({user.java_common_classes}|{user.java_generic_data_structures}): 
+    insert(java_common_classes or java_generic_data_structures)
+    key("space")
+
+boxed [type] {user.java_boxed_types}: 
+    insert(user.java_boxed_types)
+    key("space")
+
+generic [type] {user.java_generic_data_structures}:
+    insert(java_generic_data_structures)
+    insert("<>")
+    key("left")
+
+# Arrays
+type {user.code_types} array:
+    insert(user.code_types)
+    user.code_operator_subscript()
+
+type ({user.java_common_classes}|{user.java_generic_data_structures}) array:
+    insert(java_common_classes or java_generic_data_structures)
+    user.code_operator_subscript()    
+
+[state] {user.java_access_modifiers}: 
+    insert(user.java_access_modifiers)
+    key("space")
+
+[state] {user.java_other_modifiers}: 
+    insert(user.java_other_modifiers)
+    key("space") 
+
+op array:
+    user.code_operator_subscript()
+
+op new:
+    insert("new ")
+op plus plus:
+    insert("++")
+
+code_block_comment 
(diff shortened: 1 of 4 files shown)


pokey:
This looks good, tho I'd be inclined to rebase it onto #586 and figure out how it will interact with the new more generic type stuff

michaelpog:
Thank you for looking into it @pokey . I'm not sure how it's related to #586.
Would you mind clarifying please?

pokey:
#586 uses `code_type` instead of a Python-specific list, and introduces actions for inserting types that can be overridden by specific programming languages. It would be good to try to do the same thing to a certain extent for Java, eg define the `code_type` list for the Java context and override the insert type / return type actions rather than having a new talon command. Don't need everything to be generic, but at least the basic type insertion would be good to use this new mechanism. Make sense?

michaelpog:
Let me know if this is enough. I'll let the renaming from code_types to code_types and everything else, be taken care of by the other pr.

lang/java/java.talon
```diff
@@ -0,0 +1,56 @@
...
+[state] {user.java_other_modifiers}: 
+    insert(user.java_other_modifiers)
+    key("space") 
+
+op array:
+    user.code_operator_subscript()
+
+op new:
+    insert("new ")
+op plus plus:
```
splondike, line 53: I notice you've also added 'op increment'. Should we just remove 'op plus plus' here?

lang/java/java.talon
```diff
@@ -0,0 +1,56 @@
...
+    insert("++")
+
+code_block_comment
```
splondike, line 56: Leftover line.

lang/java/java.talon
```diff
@@ -0,0 +1,56 @@
...
+# Arrays
+type {user.code_types} array:
+    insert(user.code_types)
+    user.code_operator_subscript()
+
+type ({user.java_common_classes}|{user.java_generic_data_structures}) array:
+    insert(java_common_classes or java_generic_data_structures)
+    user.code_operator_subscript()    
+
+[state] {user.java_access_modifiers}:
```
splondike, line 40: C#, Python, and javascript all use 'pro static funky' type things to make functions (and assume a 'void' return type in the case of C#). Maybe the java stuff should work the same way as them? There's no direct support for class access modifiers in the C# implementation either.

I guess the difference would be between saying "public static boxed integer camel my fancy method" -> "public static Integer myFancyMethod|" and "pub static my fancy method" -> "public static void myFancyMethod(|)"and then having to go back and change the void to an integer in the second case.

lang/java/java.talon
```diff
@@ -0,0 +1,56 @@
+mode: user.java
+mode: user.auto_lang
+and code.language: java
+-
+tag(): user.code_operators
+tag(): user.code_comment
+tag(): user.code_block_comment
+tag(): user.code_generic
+
+action(user.code_state_import):
```
splondike, line 10: Do this .py style, implementing actions in .talon files is deprecated.

lang/java/java.talon
```diff
@@ -0,0 +1,56 @@
...
+    insert("import ")
+
+# Types Commands
+type {user.code_types}: 
+    insert(code_types)
+    key("space")
+
+type ({user.java_common_classes}|{user.java_generic_data_structures}):
```
splondike, line 18: I'd say maybe swap your .py stuff to put all the unboxed stuff into code_types along with the common classes and generic data structures, and then you could just fall back to the 'type *' command defined in programming.talon.

splondike (commented):
`user.code_types` has been renamed to `user.code_type` as part of the referenced PR (#586), so this PR doesn't actually work properly when merged in (e.g. "type short" doesn't insert "Short"). I've put in some other comments as well for your consideration.

michaelpog:
Comments are addressed

splondike:
Looks good, thanks for the updates!

PR #577, opened 2021-08-27
Add more words to homephones

michaelpog:
Added the words:.
* bed
* end,and
* were,whir
* ward,word


rntz:
This PR needs to be squashed before merging, as it has a lot of unrelated commits.

Besides that this PR does lean toward a pretty loose definition of homophone -- code/cold and were/where aren't homophones in my accent -- but if they're causing folks trouble I see no reason not to include them. Any objections?

michaelpog:
@rntz I'm embarrassed to admit that I didn't know that the words were and where are actually pronounced differently. You live you learn. So I'll definitely remove it. 
For the words  code and cold,  whenever I am in the dictation mode  it takes me so many attempts to get the word code recognized as  code and not as cold  that I was kind of forced to add it for convenience.  However I do agree that it is definitely on the looser  side of defining a homophone. I don't mind removing it and just keeping it for myself if nobody else has issues with this word

michaelpog:
I squashed all the commits as requested

rntz:
The point of homophones is to help people correct things when they're misinterpreted, so I think a looser definition is fine as long as we don't end up cluttering the homophones list an absurd amount.

PR #578, opened 2021-08-28
Add copy-and-immediately-paste command.

codecat555:


rntz:
What is this intended to do? Won't it generally simply copy the selection, then overwrite it?

Also, the command should probably be "copy paste" not "copypaste" although in all likelihood the latter will still work.

codecat555:
Thanks for taking a look. I use this command frequently while working in a terminal; I run a command and then immediately use part of the output to construct a new command.

So, for example, I might run 'ls', copy one of the file names from the output and then immediately paste it again, which enters it onto the command line. Or, 'git status' followed by 'git add/rm'.

I created it after getting tired of constantly saying 'copy that, paste that' and it works well for me. copypaste as one word is sufficiently unique that I haven't seen any misrecognitions with it. The action is flexible since it uses edit.copy() and edit.paste(), which can be overridden.

Do you want me to add 'copy paste' as an alias, perhaps?

rntz:
Do you have any examples of using this outside a terminal? My hesitation is
that in most contexts this command will do nothing useful, although I can
see it's very useful in a terminal. So I think either:

1. We should only bind this command in contexts where it is useful (eg. in
terminal), or
2. We should find some generalisation of it that is useful in many contexts.

Regarding the command name, just rename the command to "copy paste" instead
of "copypaste", no need for an alias. But this is a nitpick, really.

codecat555:
Yes, I agree.

I did try using it in firefox yesterday - because keystrokes populate the search box. However, it don't work when pasting - prob by design. Anyways, once you highlight something on the page it will already appear in the search box when you bring it up.

So, I'll move this into generic_terminal.

rntz (approved):
lgtm. I'll merge this next week unless someone objects.

splondike:
LGTM

PR #579, opened 2021-08-28
Implement edit.find for windows terminal.

codecat555:
synonyms:

buffy - "buffer search"
queer eye - pun on 'query'

### DIFF
--- a/apps/platforms/win/windows_terminal/windows_terminal.py
+++ b/apps/platforms/win/windows_terminal/windows_terminal.py
@@ -22,6 +22,9 @@ class AppActions:
 class EditActions:
     def paste(): actions.key('ctrl-shift-v')
     def copy():  actions.key('ctrl-shift-c')
+    def find(text: str):
+        actions.key('ctrl-shift-f')
+        actions.insert(text)
 
 @ctx.action_class("user")
 class UserActions:
--- a/apps/platforms/win/windows_terminal/windows_terminal.py
+++ b/apps/platforms/win/windows_terminal/windows_terminal.py
@@ -22,9 +22,10 @@ class AppActions:
 class EditActions:
     def paste(): actions.key('ctrl-shift-v')
     def copy():  actions.key('ctrl-shift-c')
-    def find(text: str):
+    def find(text: str = None):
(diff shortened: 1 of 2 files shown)


apps/platforms/win/windows_terminal/windows_terminal.py
```diff
@@ -22,6 +22,9 @@ def tab_open():  actions.key('ctrl-shift-t')
 class EditActions:
     def paste(): actions.key('ctrl-shift-v')
     def copy():  actions.key('ctrl-shift-c')
+    def find(text: str):
```
rntz, line 25: `text` can be optional, use `text: str = None` and handle that case.

apps/platforms/win/windows_terminal/windows_terminal.py
```diff
@@ -22,6 +22,9 @@ def tab_open():  actions.key('ctrl-shift-t')
...
+        actions.key('ctrl-shift-f')
+        actions.insert(text)
```
rntz, line 27: I suggest `if text: actions.insert(text)`

apps/platforms/win/windows_terminal/windows_terminal.talon
```diff
@@ -18,3 +18,13 @@ kill all:
     insert("y")
     key(enter)
     
+
+(find it | buffy | queer eye)$:
```
rntz, line 22: we probably shouldn't have so many synonyms for one command. One way is preferable to many.

`find it` seems fine to me. Or, elsewhere it looks like we use `hunt` for this functionality, eg. in `text/find_and_replace.talon`.

Also, there's no need to end-anchor this version (`$`), is there? Could be useful to say `find it snake hello world`, for example.

apps/platforms/win/windows_terminal/windows_terminal.talon
```diff
@@ -18,3 +18,13 @@ kill all:
...
+    key(ctrl-shift-f)
```
rntz, line 23: Could just call `edit.find()` once you handle the `None` case in its implementation.

rntz (changes requested):

PR #580, opened 2021-08-28
Add tabby alias (for tab).

codecat555:
For me, talon often turns "tab" into "ten". This alternative always seems to work.

### DIFF
--- a/code/keys.py
+++ b/code/keys.py
@@ -235,6 +235,7 @@ alternate_keys = {
     #'junk': 'backspace',
     "page up": "pageup",
     "page down": "pagedown",
+    "tabby": "tab",
 }
 # mac apparently doesn't have the menu key.
 if app.platform in ("windows", "linux"):


rntz:
Hm, how do people feel about this? @pokey @splondike 

I'm not opposed, and tab is an easily misrecognized word, but it would be adding a new key name. We do tend to advise people to just customize their own keys.

splondike:
@codecat555 Are you using conformer/beta or the public version of Talon? If it's the public version, then this issue might go away in the next couple of months when the new voice engine comes out to public.

I'd probably be inclined to think that it might make more sense for you to just have this change in your personal configuration. If other people had a similar issue with tab, they'd have to go and read the code to figure out that there was an alias anyway.

I've done a similar thing for the 'enter' key in my setup; 'return' works better for me. I've actually set this up so that it patches the 'self.special_key' list without actually needing to edit the knausj code. You can see what I've done below:

```
from talon import Context, registry, app


ctx = Context()
ctx.matches = r"""
os: linux
"""

def _add_keys():
    """
    Put this in a launch listener so it runs after knausj
    """

    # Add return as a first class alias for enter
    my_list = {
        k: v
        for k, v in registry.lists['user.special_key'][0].items()
    }
    my_list['return'] = 'enter'
    ctx.lists['self.special_key'] = my_list
app.register("launch", _add_keys)
```

You should just need to change 'return' to 'tabby' and 'enter' to tab in the above. You might also need to change the "os: " matcher to match whatever you're using. Do you think something like that would work for you?

codecat555:
Conformer/beta. I thought this might be of general use, but I have no problem keeping it in my own tree.

splondike:
Oh, that's unfortunate that you're already using the newer engine.

I guess it's hard to know if this change would be useful to others especially since discoverability wouldn't be that great. This PR should show up in searches, so if anybody else runs into the same issue and has Github set up they could put a comment in here and we can revisit the problem. In the meantime, how about we just close it out?

PR #581, opened 2021-08-28
Let windows with 'VIM' in title be recognized as vim.

codecat555:
Happy to adopt some other unique token other than 'VIM'.


splondike:
Looks good, GVIM and VIM both have this token in the title by default.

PR #582, opened 2021-08-28
Add generalized support for parsing filename extension out of window title, as a fallback measure.

codecat555:
Any objection to defining default default behavior for this action?

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -79,12 +79,28 @@ def code_libraries(m) -> str:
     """Returns a type"""
     return m.code_libraries
 
+def get_file_ext():
+    """Returns file name extension. If default method fails, try to extract the extension from the window title."""
+    """To override the fallback behavior, override the win.file_ext() action in the target context."""
+    file_extension = actions.win.file_ext()
+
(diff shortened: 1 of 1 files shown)


codecat555:
@lunixbochs made some comments (below) on this change, so I'll move this PR to draft while I rework things.

> there are definitely going to be false positives, and the extension isn't guaranteed to be the last thing on the line (you might get an ext like .py - vim - 80x24 or something, as well as weird stuff like .com (any website or file path might end up as an extension)
>
> I'd maybe also suggest you implement win.file_ext yourself, using actions.next() to call the original method, instead of inlining a helper in code.py
> 
> I think it merits discussion and maybe an approach for testing / measuring false positives
> maybe a handful of people volunteer to run a script for a few weeks that measures window titles that trigger it vs the original file_ext action
>

splondike:
Since this has been hanging around for a while without activity, I might close it out if that's OK with you @codecat555 ?

codecat555:
Sorry, I didn't realize this was still pending.

PR #583, opened 2021-08-28
Fix-up wsl path detection mechanism.

codecat555:
1. if the window title being set for a new wsl session contains more than
one colon, the path extraction code returns invalid path. this is especially
problematic because the default wsl title looks like this (at least on my
system):

`Title: WSL Ubuntu-20.04: wsl_user@some_machine: /mnt/c/Users/some_user/AppData/Roaming/talon/user/knausj_talon`

2. when a new session is started, the window title for a new wsl session
initially does not contain a valid path...due to temporary instability in
the value. this was causing a 'No such file or directory' error to
bleed out onto the console because the subprocess stderr was not caputured.
It's an error that is actually of no consequence...except that it made me
go searching for "the problem". so, the code captured stderr and issues a
warning in this case instead.

### DIFF
--- a/apps/platforms/win/wsl/wsl.py
+++ b/apps/platforms/win/wsl/wsl.py
@@ -1,8 +1,17 @@
 from talon import Context, Module, actions, imgui, settings, ui, app
+from talon.debug import log_exception
 import os
 import subprocess
+import logging
+import sys
 
 mod = Module()
+# Note: these context matches are specific to ubuntu, but there are other
+# distros one can run under wsl, e.g. docker. for that matter, there are
+# multiple ubuntu distros available. we need a more general way of detecting
+# the current distro, and a way for the user to specify which distro to use
+# for any particular operation. perhaps implement a generic_wsl module and
+# then layer various distros on top of that?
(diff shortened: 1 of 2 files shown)


codecat555 (commented):
In the first commit, I exposed the error messages being generated by the wsl session used to retrieve the current path. The errors that I then saw led to the other changes here.

In particular, during testing I started seeing some hex-encoded errors which made me curious as to why they were not being decoded properly. Turns out they were encoded with UTF-16, which indicates those errors were coming from WSL itself rather than from the host os. I added code to decode these errors cleanly.

Also during testing I found that the wsl operation sometimes fails without throwing any error and then works on retry. So, I added code to do that. Note: one such case is when the session is first coming up and yet the window title is not yet populated with the path. In this case, the window title is just the distro name. So, the code detects that and silently retries. Other errors leading to retry will be logged.

Some other details are documented in the code and commit messages.

codecat555:
I have reorganized my changes into a series of incremental commits that I will submit as separate pull requests. I'm going to close this one and start fresh with a new PR.

Issue #398, opened 2021-03-27
programming language refactor?


knausj85:
Many things in python.py should maybe be made generic across languages, declared in code.py

- [ ] Type list
- [ ] Exception list
- [ ] Doc string fields
- [ ] Maybe re-consider making operators a list

fidgetingbits:
something worth noting for python as I happen to be doing python stuff to day is that those declarations are also broken:
ex: 
```
is type {user.python_type_list}:
    insert(": {python_type_list}")
```

should be:  `<user.python_type_list>`

I'm not really sure how that got in there because I was using those commands heavily before...

I do agree that this would be useful to have generically across languages, I do have a bunch of extended type stuff added in C already as well, like being able to switch between what types you actually want to use for a given project: https://github.com/fidgetingbits/knausj_talon/blob/master/lang/c/c.py . I could see that potentially being useful for doc strings as well, in some people switch between sphinx or doxygen, etc

unfortunately I've been quite busy with work and also due to health I just haven't been able to contribute back into the main repo lately :(

knausj85:
> something worth noting for python as I happen to be doing python stuff to day is that those declarations are also broken:
> ex:
> 
> ```
> is type {user.python_type_list}:
>     insert(": {python_type_list}")
> ```
> 
> should be: `<user.python_type_list>`

this functionality works fine for me as is. there's really no need for a capture that i can see?

fidgetingbits:
ah interesting, it wasn't working on mine at all yesterday until I changed it to capture in it started working fine

wenkokke:
My Talon user directory moves a lot of the language definitions to lists loaded from CSVs: common function names, operator names, imports, type names, snippet names, etc.

This might be an easier way to generalise, without writing huge Python files?

nriley:
Closing this as we've done most of this now.

PR #586, opened 2021-09-01
Use `code_type` for python and typescript types

pokey:
Part of #398

### DIFF
--- a/lang/typescript/typescript.py
+++ b/lang/typescript/typescript.py
@@ -13,6 +13,16 @@ and code.language: typescript
 #     "string": ".ToString",
 # }
 
+ctx.lists["user.code_type"] = {
+    "boolean": "boolean",
+    "integer": "int",
+    "string": "string",
+    "null": "null",
+    "undefined": "undefined",
+    "number": "number",
+    "any": "any",
+}
+
 
 @ctx.action_class("user")
 class UserActions:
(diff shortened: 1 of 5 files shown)


lang/typescript/typescript.py
```diff
@@ -284,3 +294,8 @@ def code_public_function(text: str):
 
     #     actions.user.code_insert_function(result, None)
 
+    def code_insert_type_annotation(type: str):
+        actions.insert(f": {type}")
+
+    def code_insert_return_type(type: str):
+        actions.insert(f" => {type}")
```
pokey, line 301: One issue here is that this syntax is for defining return types in function type expressions (eg `(foo: string) => string`), but not how you define return type when  declaring a function (eg `foo(): string`).  I wonder what's the right thing to do here 🤔 cc/ @brxck @AndreasArvidsson

AndreasArvidsson: I don't have a good solution for this problem. Personally I just have `type` at its own thing and not in combination with arrow or colon.

pokey: I'm leaning towards using `: ` here, and then people can just use "op lambda type string" for the arrow version. Seems more likely to be what you'd expect

rntz (approved):

rntz (approved):

PR #587, opened 2021-09-02
Fix macOS Slack unread command

auscompgeek:
cmd-j doesn't seem to do anything, and I presume this is meant to jump to "all unreads" like I think the Windows/Linux one does?

https://github.com/knausj85/knausj_talon/blob/4ac7ba1f6b72f79d6e463b88a288f553627ae446/apps/slack/slack_win.talon#L24


rntz:
Can anyone on mac confirm that this works? @pokey ?

PR #588, opened 2021-09-04
Added all the keyworks to trigger keyboard presses from 'keys.py' to the help menus so they can be accessed via the help menus.

tararoys:
Made every key name in keys.py accessible via the help menu.  This included adding help menus for 
- number keys
- symbol keys
- modifier keys
- simple keys
- alternate keys
- arrow keys
- function keys

although formatters are not techically keys, I added a menu for formatters as well so that formatters would now work exactly the same way as all the other help menus work.  I left the old 'format help' toggle, is probably confusing and should probably be deleted. 

-formatters.  

- Commands to open each of these menus were added to help.talon, and following the convention of the help context menu, if you open a new menu, it closes all of the old ones.


misc/help.talon
```diff
@@ -1,4 +1,18 @@
 help alphabet: user.help_alphabet(user.get_alphabet())
+help symbols: user.help_symbol_key_words(user.get_symbol_key_words())
```
splondike, line 2: For me help symbols overflows the bottom of the screen. Maybe you could do some kind of pagination stuff like 'help context'. If not, you might be able to improve the situation by merging synonyms together, e.g. on my setup 'angle', 'left angle', and 'less than' all map to '<' so that could be formatted '<: angle, left angle, less than' and this would make the window wider so that more could fit on the screen at a time.

misc/help.talon
```diff
@@ -1,4 +1,18 @@
...
+help numbers: user.help_number_key_words(user.get_number_key_words())
+help punctuation: user.help_punctuation_words(user.get_punctuation_words())
+help modifier: user.help_modifier_words(user.get_modifier_words())
+help simple keys: user.help_simple_keys(user.get_simple_keys())
```
splondike, line 6: I wonder if simple keys, alternate keys, f keys, and arrows could be merged to 'help special keys'? Would people want to refer to them distinctly?

code/help.py
```diff
@@ -472,17 +603,145 @@ def help_alphabet(ab: dict):
...
         #     )
         # )
-        gui_context_help.hide()
-        gui_alphabet.hide()
+        hide_all_help_guis()
         gui_alphabet.show()
         register_events(False)
         actions.mode.enable("user.help")
 
+    def help_number_key_words(ab: dict):
```
splondike, line 611: You could probably merge all these 'help_\*' and 'gui_\*' functions/imgui instances into a single action since you're passing in the help contents anyway.

splondike (commented):
I like this PR; happy to merge it in if you like. I've also put some suggestions inline for your consideration.

splondike:
Nice job!

PR #589, opened 2021-09-06
Support named arguments for r

pokey:
Might be useful for common named arguments in python too

- [x] Get good starting list of names

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -5,7 +5,7 @@ mod = Module()
 mod.list("code_functions", desc="List of functions for active language")
 mod.list("code_types", desc="List of types for active language")
 mod.list("code_libraries", desc="List of libraries for active language")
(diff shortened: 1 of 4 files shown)


splondike:
Hmm, I'm not sure this would come up too often in the web programming I mostly do; there are too many different names. This looks like it might be good for heavy users of particular frameworks within a language (e.g. R, Pandas in Python, TensorFlow in Python etc.). I reckon you'd have to customise it pretty heavily for your use case though.

My inclination is that this would be something that you'd probably just write yourself ah-hoc if you needed it rather than having it in knausj. What do you think?

pokey:
Tbh I don't see it being particularly useful in Python, as the names of the args will vary a lot, but I'm told they're pretty standard for r. Maybe we should get a second opinion from another r programmer?

rntz (approved):

PR #596, opened 2021-09-13
Hide the selected word from the list of homophone options

xavier630:
Example:
image

The homophones window currently shows the word that is selected (and in a random position in the list) which increases the cognitive load when trying to find the option you want. This PR filters it out as an option.

### DIFF
--- a/code/homophones.py
+++ b/code/homophones.py
@@ -107,7 +107,7 @@ def raise_homophones(word_to_find_homophones_for, forced=False, selection=False)
 
     if (
             is_selection
-            and len(active_word_list) == 1
+            and len(active_word_list) == 2
             and quick_replace
             and not force_raise
     ):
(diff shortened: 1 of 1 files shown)


xavier630:
https://github.com/knausj85/knausj_talon/issues/530

xavier630:
https://github.com/knausj85/knausj_talon/issues/552

splondike:
I'd suggest sorting the phones-d word last as you suggested in the issue since it looks like this function is also used for the "phones you" command. In that case you'd want it to include 'you' as well.

xavier630:
Ah, because you don't have a way to tell which variant of the homophone has been said when someone says `phones you`? 
 
That makes sense, I've never actually used phones proactively like that, always in a reactionary way.

Let's keep it as the final option 👍 .

splondike:
Sounds like a good plan :+1: .

xavier630:
All right, always at the bottom now 👍

splondike:
Looks good! Merging.

PR #597, opened 2021-09-13
Add phones exit button and make all phones options clickable

xavier630:
image

### DIFF
--- a/code/homophones.py
+++ b/code/homophones.py
@@ -118,7 +118,6 @@ def raise_homophones(word, forced=False, selection=False):
     show_help = False
     gui.show()
 
-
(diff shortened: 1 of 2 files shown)


xavier630:
https://github.com/knausj85/knausj_talon/issues/530

splondike:
This seems good, but I'd probably just label it as 'phones hide' matching the existing voice command. It doesn't look like we have 'exit' used as a way of closing this kind of thing elsewhere in the voice commands (we have 'hide', 'off', 'close' in other places). We do have it for shutting down a program, but only one instance I think. It'd be good to keep the number of 'close' type verbs down, and preferably reduce them!.

xavier630:
Sounds reasonable, will update it to hide 👍

splondike:
Nice one :).

PR #601, opened 2021-09-18
Improve WSL error logging

codecat555:
The first commit here is a simple optimization.

The second commit exposes error messages that may be generated during wsl path detection and which were previously hidden. Part of the code was refactored in order to provide for common error handling. There is code to detect and decode a class of errors formatted in utf-16-le which may be generated by the wsl engine itself (rather than the guest distros).

This change is the first of several for this module which I would like to contribute.  Primary goals include cleanly supporting multiple distros and integration with the broader terminal/Linux framework present in knausj_talon. In the long run, it would be nice to refactor this code into a more OO implementation.

### DIFF
--- a/apps/platforms/win/wsl/wsl.py
+++ b/apps/platforms/win/wsl/wsl.py
@@ -1,8 +1,17 @@
 from talon import Context, Module, actions, imgui, settings, ui, app
+from talon.debug import log_exception
 import os
 import subprocess
+import logging
+import sys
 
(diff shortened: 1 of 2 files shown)


codecat555:
Isn't this a bug fix rather than an enhancement?

splondike:
Oh, I interpreted it as a logging enhancement. I've swapped it to a bug.

PR #604, opened 2021-09-26
Optimise context-sensitive dictation to avoid peeking left if unnecessary

rntz:
Context-sensitive dictation peeks left and right to get context for spacing and capitalizing inserted text. Depending upon the text inserted, this context may not be necessary. This PR avoids looking for context on the left if there is no circumstance where we need to insert space on the left or add capitalization.

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -67,15 +67,12 @@ no_space_before = re.compile(r"""
   | ['"] (?: $ | [\s)\]}\-'".,!?;:/] )
   )""", re.VERBOSE)
 
-# no_space_before = set("\n .,!?;:-_/%)]}")
-# no_space_after = set("\n -_/#@([{")
+def omit_space_before(text: str) -> bool:
+    return not text or no_space_before.search(text)
+def omit_space_after(text: str) -> bool:
+    return not text or no_space_after.search(text)
 def needs_space_between(before: str, after: str) -> bool:
-    return (before and after
-            and not no_space_after.search(before)
-            and not no_space_before.search(after))
-    # return (before != "" and after != ""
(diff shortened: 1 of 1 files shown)


code/dictation.py
```diff
@@ -182,19 +179,18 @@ def dictation_insert_raw(text: str):
 
     def dictation_insert(text: str) -> str:
         """Inserts dictated text, formatted appropriately."""
-        # do_the_dance = whether we should try to be context-sensitive. Since
-        # whitespace is not affected by formatter state, if text.isspace() is
-        # True we don't need context-sensitivity.
-        do_the_dance = (setting_context_sensitive_dictation.get()
-                        and not text.isspace())
```
rntz, line 189: We should no longer need the `text.isspace` check because it should be handled by `omit_space_before` and `omit_space_after`.

code/dictation.py
```diff
@@ -182,19 +179,18 @@ def dictation_insert_raw(text: str):
...
+            and not (omit_space_before(text)
+                     and auto_capitalize(text, "sentence start")[0] == text)):
             dictation_formatter.update_context(
                 actions.user.dictation_peek_left(clobber=True))
         text = dictation_formatter.format(text)
         actions.user.add_phrase_to_history(text)
         actions.insert(text)
         # Add a space after cursor if necessary.
-        if not do_the_dance or not text or no_space_after.search(text):
+        if not context_sensitive or omit_space_after(text):
```
rntz, line 193: Likewise, the `not text` check has been rolled into `omit_space_after`.

splondike (approved):
LGTM

Issue #598, opened 2021-09-14
Gaze scrolling relative to the cursor position 


daanvdh:
Current gaze scrolling uses the middle of the window as reference point, for determining to scroll up or down. If you work in an editor with multiple views inside that window, it's not always possible to scroll in a specific view. That's why I want to have an alternative implementation using the curser as reference point instead. The cursor should stay in the same location (using zoom mouse), then gazing up or down from the cursor should scroll up or down repectively. 

Let me know if this is something others see usefull to have in this repo as well.

PR #605, opened 2021-09-28
#598 Gaze scrolling relative to the cursor posiiton

daanvdh:
#598

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -14,8 +14,9 @@ from talon import (
     ui,
 )
 from talon_plugins import eye_mouse, eye_zoom_mouse
-from talon_plugins.eye_mouse import config, toggle_camera_overlay, toggle_control
+from talon_plugins.eye_mouse import mouse
 
+main_screen = ui.main_screen()
 key = actions.key
 self = actions.self
 scroll_amount = 0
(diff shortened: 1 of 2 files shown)


rntz:
@knausj85 hey, can this be merged / should it be worked on / should we close it? I don't use an eye tracker so it's a bit opaque to me. I'm trying to give old PRs some attention so they don't get forgotten.

splondike:
It's been more than a year since this was put up and we asked for help reviewing it (since I don't think we have a regular maintainer that uses an eye tracker).

Given that, I don't think we have the capacity to test this. As such I think I'd be in favour of closing it to reduce the length of the stagnant PR list. It's unfortunate since it sounds useful and has obviously had a bit of work put in. We can leave the issue open (which will reference the code in this PR), so if somebody wishes to pick it up and find some testers in the future they can dust off the existing code.

deepbluezen:
Potential user review at request of splondike from slack. Tested on beta v0.3.1-57-g0dd4 (57), Win10. I use control mouse routinely.
I can see the use case it's solving for someone who uses zoom mouse and it worked effectively for me across a number of programs that I use which are multipaned such as blender or unity. For me I wouldn't use it as it would have to much friction dropping out of the control mouse mode and I don't have an issue just clicking on drag bars or using the wheel up/down commands.
I don't have a feel for how much of the community are using one eyetracking mode over the other? Even if it wouldn't be of use to me I would be happy to support testing if it was taken forwards.

splondike:
Given deepbluezen wouldn't use this themselves and it would take a bit of effort for us to coordinate re-implementing and testing this on the current knausj main I think I'll close this out.

As stated in my last comment if somebody wants to push this forward we've still got the original issue and the code isn't lost, just closed.

PR #607, opened 2021-10-01
Vscode terminal test

pokey:
This is a POC of exposing state from VSCode, eg terminal focus.

Couldn't spam key every half second because that cancelled hovers, etc

- [ ] If you switch to application and then quickly try to switch again, you get strange behaviour because it will be mid-update
- [ ]  Some strangeness / flakiness with timeouts
- [ ] The quick-open wait code is strangely slow, when it should be faster than the arbitrary sleep
- [ ] How do we update in response to keypresses?
   - For example presssing enter or escape could change the state
   - Could override insert / key to do a debounced cron.after with update


pokey:
Yes sorry for the long-open PR. It is making its way to the top of my list. Getting close. Happy to close and then reopen when I'm ready to keep things cleaner

splondike:
Yeah, maybe let's close it out for now.

PR #623, opened 2021-10-20
deprecate "m grid" for "mouse grid" forget tag

timo:
the tag for activating "m grid" proved to be a very bad idea.


splondike:
Discussed removing the tag (making M grid global) and decided that that could result in too many false positives so we'll keep it in for the moment.

PR #624, opened 2021-10-20
simple plural support

SCdF:
If the word you are looking is not found, but ends in an s, hope that it's a plural and check to see if the singular form exists so you can use those (with s's added on).

This only affects words that end in an s and don't otherwise exist in the homophones csv.

This gives us "cursors/cursers", "duals/duels", "minors/miners" etc for free, instead of having to carefully select the singular form or bloating the homophones list. It will obviously get stuff wrong, but imo it's better to have some wrong choices in your list then no choices at all.

### DIFF
--- a/code/homophones.py
+++ b/code/homophones.py
@@ -90,6 +90,8 @@ def raise_homophones(word_to_find_homophones_for, forced=False, selection=False)
 
     word_to_find_homophones_for = word_to_find_homophones_for.lower()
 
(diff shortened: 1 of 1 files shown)


splondike:
This seems like a good idea to me. I'm happy to merge it in, but would you be able to drop in a comment somewhere around that block saying roughly what the code's doing (checking for non plural forms in homophones.csv)? As pokey said there are several ways of pluralising in English, so it may be a thread that somebody wants to improve on in the future.

PR #627, opened 2021-10-22
Improve snap to vertical screen

timo95:
I tested this with two 16:9 screens. Maybe the difference between dpi zoom levels should be taken into account in the future. One additional test combining 16:9 and 4:3 would also be nice.

### DIFF
--- a/code/window_snap.py
+++ b/code/window_snap.py
@@ -74,18 +74,17 @@ def _move_to_screen(
     # TODO: Test vertical screen with different aspect ratios
     # Does the orientation between the screens change? (vertical/horizontal)
     if (src.width / src.height > 1) != (dest.width / dest.height > 1):
-        # Rotate proportional size and position with screen
-        proportional_width = dest.width / src.height
-        proportional_height = dest.height / src.width
+        # Horizontal -> vertical or vertical -> horizontal
+        # Retain proportional window size, but flip x/y of the vertical monitor to account for the monitors rotation.
         if src.width / src.height > 1:
-            # Horizontal to vertical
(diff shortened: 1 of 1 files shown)


splondike:
This is some fairly involved code, are you able to put in a bit more of a description here in the PR about what the current master behaviour is and what you're changing it to? Putting some more in the code about what the intent of the various blocks is would be good also.

@jcaw FYI. It looks like you had a big hand in the original implementation.

timo95:
I added a few more comments to the code.

Current master:
- Scale x/y and width/height proportionally with monitor size. This breaks with vertical monitors, since it changes the window aspect ratio with the monitors (example: monitors 16:9 -> 9:16, window 2:1 -> 1:2). This makes small windows like terminals unusable.

My change:
- Test for change between monitor orientation (horizontal -> vertical or vertical -> horizontal). If it does not change keep the old behavior.
- Switch x/y of vertical monitor when computing window size
- Handle too big windows by deforming them while keeping the area the same
- Proportional positioning is harder because the width/height is no longer proportional to the monitor: Keep the free space to the left of the window proportional to the free space to the right of the window. Respectively for the top/bottom free space. This should be equivalent to the old behavior, but more general.

-> Cases where the monitor orientation does not change aren't influenced at all
-> Small windows keep their aspect ratio (proportional to the monitors)
-> If the window was fully in the visible area it will still be fully in the visible area
-> Fullscreen windows are still fullscreen

splondike:
Thanks, it seems to work quite well from my testing. The code comments should help if folks want to further refine it in the future.

PR #631, opened 2021-10-26
Parse url from browser title

timo95:
Should work with any addon for any browser that displays the whole url in the window title.


splondike:
Thanks, good addition!

PR #632, opened 2021-10-27
Add support for controlling multiple WSL distros and prune one error …

codecat555:
This change is a bit more involved than the other changes I've submitted recently.

Primarily, it adds support for controlling multiple wsl distros.

Secondarily, it establishes a convention for the format of wsl window titles - this bit may require some discussion/adjustment.

Finally, it prunes away one error case which, in fact, is the original reason I started poking around in this module in the first place.

The changes and the comments should explain the details - if not, let me know.


apps/platforms/win/wsl/wsl.py
```diff
@@ -4,25 +4,171 @@
...
-app: ubuntu
+
+# note: this context match is intentionally made more complex so that it is more specific
+# than the context defined in apps/win/windows_terminal/windows_terminal.py (and thereby
+# takes precedence).
+ctx.matches = fr"""
 app: windows_terminal
-and win.title: /Ubuntu/ 
+and tag: user.wsl
+tag: user.wsl
```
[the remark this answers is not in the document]
knausj85: I think the `windows_terminal` check here is unnecessary for the context, no? It should probably just specify the tag, and let any application assert it correctly as appropriate.

apps/platforms/win/wsl/wsl.talon
```diff
@@ -1,4 +1,8 @@
-os: windows
--
-
-(wsl|weasel) reset path detection: user.wsl_reset_path_detection()
+os: windows
+title: /^WSL:/
```
knausj85, line 2: I'm a little leery of a global regex based on the windows title creating false positives, though it's probably very unlikely in practice. 

For clarity, is your thinking to assert the wsl tag here, or in an app-specific manner as done in ubuntu.talon?

apps/platforms/win/wsl/wsl.py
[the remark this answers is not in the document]
knausj85: Never-mind, I missed the comment. Makes sense.

apps/platforms/win/wsl/ubuntu.talon
```diff
@@ -1,11 +1,12 @@
 app: ubuntu
```
knausj85, line 1: I believe this app definition got remove?

knausj85: Since this is no longer ubuntu-specific, it may make sense to rename to wsl.talon

codecat555:
@knausj85 - thanks for your comments! I'm busy working on something else at the moment, but I will circle back to this shortly and work on incorporating your suggestions.

codecat555:
Finally circling back to this, sorry for the delay.

apps/platforms/win/wsl/ubuntu.talon
codecat555: Right. I removed that line but I'm not sure I'm not sure about moving the contents of this file into `wsl.talon`. Because I'm not sure if the definitions here are appropriate to all possible wsl distros, though they certainly are applicable to more than just Ubuntu.

There are configurations one can run under wsl which would not include the `/mnt` auto-mounts or all the commands controlled by the `user.file_manager`, `user.generic_unix_shell` or `user.git` tags. In fact, I think it might be the more usual case for many docker-desktop instances. Also, I wonder if future distros will all necessarily conform in this regard.

I suppose we could redefine the `go <user.letter>` using an action, and that way it could always be redefined for less generic contexts. I guess we could also use actions to set the tags. Is that the right way to go?

We could follow the 'generic terminal' pattern of a multi-level hierarchy. That is, we could rename `ubuntu.talon` to something like `generic_linix.talon` or `generic_wsl.talon`. Then, for non-conforming wsl contexts, new .talon files could always be added to override the generic definitions.

Just thinking out loud...

apps/platforms/win/wsl/wsl.talon
codecat555: Yes, you're right. TBH, I did actually hit at least one false positive while working on this change...a web page whose title just happened to match.

My thinking was to put the context match here in the .talon file so it would be accessible to the user without having to dig into wsl.py.

I explored matching on the app name instead of window title, and hit some snags:

First, there's wsl.exe - what you get if you run wsl from the taskbar search.

Then, there are the various distro exe's under C:\Users\me\AppData\Local\Microsoft\WindowsApps. Matching on those names means you need to keep updating your .talon context match as distros as added/removed.

Then, there are the various terminal emulator exe's that may be used to host a distro session, e.g. windows terminal, conemu, etc. Another wrinkle is that these same executables may also host non-wsl sessions like powershell.

And you can run wsl from a command.exe or powershell session.

Rather than matching on the window title or exe name, I thought about rigging up a callback mechanism from inside the distro to a talon listener of some sort. As a way of setting the user.wsl tag, I mean. Perhaps just writing to a shared file, for example.

I decided using the window title was best because there are other reasons people often do that...namely, to keep track of your sessions. Docker is a sort of a meta-distro with many flavors. And, one might want to run custom distros or multiple copies of the same distro (e.g. dev, staging). The window title is a handy way of doing that and so it's familiar for many of us.

If we do use the window title, what is the best string/format to use?

In any case, I'm happy to follow your lead on this.

apps/platforms/win/wsl/ubuntu.talon
```diff
@@ -1,11 +1,11 @@
-app: ubuntu
 app: windows_terminal
-and win.title: /Ubuntu/
+win.title: /^WSL:/
```
knausj85, line 2: What if instead we require the user.wsl tag as part of the context in the relevant talon files, and leave the tag activation to user-specific for now? We could add comments about your current approach, or include the example commented out.

This keeps it simple, and prevents the issues with the regex out of the box.

codecat555: Sounds good to me. I'll make it so and push the changes some time in the coming days.

knausj85:
Looks good, thanks for the contribution & your patience.

PR #633, opened 2021-10-27
Make wheel tiny amount relative to setting

timo95:
- Make wheel tiny amount relative to setting (what fraction should this default to?)
- Add setting for wheel left/right amount


splondike:
Thanks!

PR #635, opened 2021-10-30
Remove two empty files

JelleAalbers:
This PR removes two empty files from the repo. I'm just starting out with talon; sorry if these empty files actually serve a secret purpose :-)


rntz:
I'm going to merge this, it seems sensible and those files are certainly doing nothing right now. Maybe they were there as stubs to remind us to do something in future? @knausj85 any idea?

knausj85:
This was probably just an oversight, thanks for deleting them.

Issue #389, opened 2021-03-19
Let words_to_replace replace phrases not just single words


rntz:
I should come up with a motivating example for this because I'm sure I've brushed up against this limitation at some point.

rntz:
Fixed by #641.

PR #636, opened 2021-10-31
Abbreviations inside formatters and other commands

JelleAalbers:
This allows reverse abbreviations inside formatters and other commands:
  * _funky read brief text_ → def read_txt():
  * _all caps snake brief command something_ → CMD_SOMETHING
  * _say fancy brief technology comma cool things brief etcetera_ → fancy tech, cool things etc

The implementation is to add several _brief SPOKEN_FORM_  → ABBREVIATION entries to talon's `user.vocabulary` map in a command-mode context. 

Some notes:
  * I'm too new to Talon to know how appropriate / evil messing with `user.vocabulary` is. Maybe there are better solutions? I tried using `dictate.word_map`, but that does not support phrases, as noted in #389.
  * Comments in `code/vocabulary.py` indicate that `user.vocabulary` phrases are given high recognition priority. Since the added phrases all have multiple words starting with 'brief', I imagine this is fine?
  * Surprises can happen if the literal word 'brief' is intended. For example, _snake brief phrase_ → brief_phrase is fine, but _snake brief command_ → cmd, and  _say please make a brief entry_ → please make a ent
  * I left dictation mode alone here, but we could instead update `user.vocabulary` in the global context. That would help dictate things like the third example at the top of the PR.
  * Would it be worth it to also support the long version of the reverse-abbreviate command (_abbreviate X_) inside other commands?

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -12,7 +12,7 @@ setting_context_sensitive_dictation = mod.setting(
     desc="Look at surrounding text to improve auto-capitalization/spacing in dictation mode. By default, this works by selecting that text & copying it to the clipboard, so it may be slow or fail in some applications.",
 )
 
-@mod.capture(rule="({user.vocabulary} | <word>)")
+@mod.capture(rule="({user.vocabulary} | <user.abbreviated> | <word>)")
 def word(m) -> str:
     """A single word, including user-defined vocabulary."""
     try:
@@ -20,18 +20,23 @@ def word(m) -> str:
     except AttributeError:
         return " ".join(actions.dictate.replace_words(actions.dictate.parse_words(m.word)))
 
-@mod.capture(rule="({user.vocabulary} | <phrase>)+")
+@mod.capture(rule="({user.vocabulary} | <user.abbreviated> | <phrase>)+")
 def text(m) -> str:
     """A sequence of words, including user-defined vocabulary."""
     return format_phrase(m)
 
-@mod.capture(rule="({user.vocabulary} | {user.punctuation} | <phrase>)+")
+@mod.capture(rule="({user.vocabulary} | <user.abbreviated> | {user.punctuation} | <phrase>)+")
 def prose(m) -> str:
     """Mixed words and punctuation, auto-spaced & capitalized."""
     text, _state = auto_capitalize(format_phrase(m))
     return text
 
-
+@mod.capture(rule="brief {user.abbreviation}")
+def abbreviated(m) -> str:
+    """A reverse abbreviation inside another command"""
+    return m.abbreviation
+
+
 # ---------- FORMATTING ---------- #
 def format_phrase(m):
     words = capture_to_words(m)
--- a/code/abbreviate.py
+++ b/code/abbreviate.py
@@ -1,5 +1,3 @@
-# XXX - would be nice to be able pipe these through formatters
-
 from talon import Context, Module
 
 mod = Module()
@@ -64,6 +62,7 @@ abbreviations = {
     "device": "dev",
     "dictation": "dict",
     "dictionary": "dict",
+    "difference": "diff",
     "direction": "dir",
     "directory": "dir",
     "distribution": "dist",
@@ -78,6 +77,7 @@ abbreviations = {
     "enumerate": "enum",
     "environment": "env",
     "escape": "esc",
+    "estimate": "est",
     "etcetera": "etc",
     "example": "ex",
     "exception": "exc",
@@ -216,3 +216,14 @@ abbreviations = {
 
 ctx = Context()
 ctx.lists["user.abbreviation"] = abbreviations
+
+# In command mode, recognize abbreviations inside formatters and other commands
+#   'all caps snake brief command something' -> CMD_SOMETHING
+#   'funky open brief text' -> def open_txt():
+from .vocabulary import additional_words
+ctx_cmd = Context()
+ctx_cmd.matches = "mode: command"
+ctx_cmd.lists["user.vocabulary"] = {
+    **additional_words,
+    **{"brief " + spoken_form: abbreviation
+       for spoken_form, abbreviation in abbreviations.items()}}
--- a/code/abbreviate.py
+++ b/code/abbreviate.py
@@ -216,14 +216,3 @@ abbreviations = {
 
 ctx = Context()
 ctx.lists["user.abbreviation"] = abbreviations
-
-# In command mode, recognize abbreviations inside formatters and other commands
-#   'all caps snake brief command something' -> CMD_SOMETHING
-#   'funky open brief text' -> def open_txt():
-from .vocabulary import additional_words
-ctx_cmd = Context()
-ctx_cmd.matches = "mode: command"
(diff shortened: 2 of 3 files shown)


rntz:
I think this is a useful feature, although I would like input from other maintainers on it.

However, I don't think this is the right way to implement it. It involves a cross-python-script import, which we generally try to avoid because it reduces resilience to errors in user scripts. Also, it deeply breaks the abstraction boundary of vocabulary.py - someone editing vocabulary.py would have to avoid renaming `additional_words` or otherwise changing up how vocabulary.py edits the word list.

A better way would be to roll this feature into dictation.py (see below) by editing the captures that define `<user.word>`, `<user.text>`, and `<user.prose>` to include a `brief {user.abbreviation}` branch.

https://github.com/knausj85/knausj_talon/blob/3391fc9c6af2c86f3951b9edf0687880848b9b47/code/dictation.py#L23

pokey:
Worth testing what this one does to DFA too. I added a bunch of this kind of stuff to my formatters and then removed it because my DFA exploded

JelleAalbers:
Thanks for the quick feedback!

Great idea @rntz. I just tried adding a new `{user.inline_abbreviation}` option (containing the abbreviations with 'brief' added to their spoken forms) to the captures you mentioned; this works fine and avoids the relative import. If I add `brief {user.abbreviation}` it would also capture the word 'brief', but maybe there is some anti-capturing syntax to avoid that? Just let me know which version you and other maintainers would prefer.

Thanks for the heads-up @pokey. I followed your instructions [here](https://talonvoice.slack.com/archives/CCEE45RT7/p1635427836108300?thread_ts=1635402432.099900&cid=CCEE45RT7) to measure the compile time on issuing a first command when switching to a new application. (I don't always see this triggering compilation though -- maybe Talon does some caching?)

|           | gedit | vscode | vscode + cursorless |
| --- | --- | --- | --- |
| Master branch | 72 ms | 118 ms | 361 ms |
| This PR, extending `user.vocabulary` | 147 ms | 318 ms | 644 ms |
| This PR, `inline_abbreviation` in `dictation.py` captures | 187 ms | 405 ms | 687 ms |

Interestingly it seems to impact richer contexts more, and extending `user.vocabulary` is slightly cheaper than modifying the three dictation.py captures.

I don't think there is much/any extra delay for subsequent commands, though I'm not sure how to benchmark that accurately.

rntz:
Yeah, given those DFA times I think we probably shouldn't ship this, unfortunately :(. 644ms is pretty bad. There are some changes upcoming in Talon that should reduce the impact large lists have on dfa compilation time, hopefully we can merge this after those land, but no guarantees when that will happen.

rntz:
Btw, the approach I'd take to doing this is to use an intermediate capture rather than a list. That avoids creating another large list just to add "brief" to the start of things. Something like:

```
@mod.capture(rule="({user.vocabulary} | <user.abbreviation> | <phrase>)+")
def text(m) -> str:
    """A sequence of words, including user-defined vocabulary."""
    return format_phrase(m)
    
@mod.capture(rule="brief {user.abbreviation}")
def abbreviation(m) -> str:
    """DOCSTRING HERE"""
    return m.abbreviation
```

I haven't tested this but I think it should work.

JelleAalbers:
Sure, I understand. Your version works great and is certainly cleanest. I just pushed that for future reference; feel free to close this or leave it open for later.

splondike:
We'll leave this open until we get a change in DFA times and can then revisit. Tagging and leaving open.

rntz:
DFA times have improved in beta but we're now waiting on those changes being available in public talon to merge this.

rntz:
Hopefully those improvements will go out in 0.3 soonish but we might still want to somehow make this conditional on using w2l instead of dragon, since dragon can't handle large lists.

rntz:
Ok, the DFA improvements are now in public. So the only thing holding this up is the question of not wrecking Dragon users. Ideally we can gate that on `ctx.matches = "speech.engine: wav2letter"` somehow.

code/dictation.py
```diff
@@ -12,26 +12,31 @@
...
     text, _state = auto_capitalize(format_phrase(m))
     return text
 
-

+@mod.capture(rule="brief {user.abbreviation}")
+def abbreviated(m) -> str:
+    """A reverse abbreviation inside another command"""
+    return m.abbreviation
+
```
rntz, line 38: We should test whether Dragon performs OK using this definition. If it doesn't, we should make this a contextual override for the capture instead of its global definition, in a context matching `speech.engine: wav2letter`.

rntz (changes requested):

code/dictation.py
nriley: @wolfmanstout are you still using Dragon and would you be able to take a look?

wolfmanstout: I'm not using Dragon any more regularly. Last I checked it was already really slow. If no one else has Dragon I can try to find some time to test, but I don't have much time these days so it may not be this weekend.

pokey: Hmm do we know anyone else using Dragon these days?  Maybe we should ping on slack to see if anyone is willing to help out and test this branch?

splondike:
phillco's going to push this change along some by testing it with Dragon to confirm that it can't handle it.

If indeed it can't then we can look at conditionally implementing the capture based on the `speech.engine: wav2letter` matcher rntz identified; you can have different rules for captures in different contexts (see the wiki [captures vs lists in voice commands](https://talon.wiki/unofficial_talon_docs/#captures-vs-lists-in-voice-commands) section).

code/dictation.py
```diff
@@ -12,26 +12,31 @@
     desc="Look at surrounding text to improve auto-capitalization/spacing in dictation mode. By default, this works by selecting that text & copying it to the clipboard, so it may be slow or fail in some applications.",
 )
 
-@mod.capture(rule="({user.vocabulary} | <word>)")
+@mod.capture(rule="({user.vocabulary} | <user.abbreviated> | <word>)")
```
splondike, line 15: Mind elaborating what advantage there is to having <user.abbreviated> in the <user.word> capture? My read is this change would just let you say 'word brief difference' and get 'diff' output. You could also just say 'brief difference' and get the same thing.

splondike:
@phillco This ticket hasn't really moved in almost a year. I'm thinking we just close it out since nobody seems to be pushing it along. What do you reckon?

rntz:
I would like to see this merged, actually. I think it just dropped off the radar, but it seems useful.

rntz:
If nobody can try it with Dragon soon, then I think we should just assume Dragon can't handle it on the basis of prior experience and use the speech.engine approach.

code/dictation.py
fidgetingbits: I think the reasoning here is that if there are commands that take `<word>` as an argument, like for instance `proud <word>`, this allows you to use the abbreviated version as the `<word>`. It seems in that case it's worth keeping it.

knausj85:
This is out of date enough that I think it's easier to close and reimplement, which I'll do now.

PR #637, opened 2021-11-04
Correct bundle name for the edge browser on mac

SCdF:
As of today, on Intel Macs anyway.

### DIFF
--- a/apps/edge/edge.py
+++ b/apps/edge/edge.py
@@ -10,6 +10,8 @@ os: windows
 and app.name: Microsoft Edge
 os: windows
 and app.exe: msedge.exe
+os: mac
+and app.bundle: com.microsoft.edgemac
 """
 
 ctx.matches = r"""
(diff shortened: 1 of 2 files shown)


knausj85:
Thanks, though this isn't the ideal way to address this one. I think you should instead a `microsoft_edge` app definition for the Mac in this file:

https://github.com/knausj85/knausj_talon/blob/0ef6e37c11ce748d5e0ed3cf6812434b9c624e2a/apps/edge/edge.py#L6

The `talon copy context pie` command (from talon_helpers.talon) will attempt to generate the correct application definition on a per-platform basis for the focused application to the clipboard, and you can just paste it into that file.

The advantage of defining an app in this way: we can update the app definition in one place in python, and re-use everywhere (.talon files, .py). 

[See firefox for example](https://github.com/knausj85/knausj_talon/blob/0ef6e37c11ce748d5e0ed3cf6812434b9c624e2a/apps/firefox/firefox.py#L14)

SCdF:
Oh cool I didn't know about that, will do and test tmrw

SCdF:
Done, and tested on an Intel Mac. @knausj85 please check it out :-)

knausj85:
Looks good, thanks.

PR #638, opened 2021-11-05
Fix issue with initial continuous scroll speed

daveweij:
Currently the initial speed of the continuous scroll can be increased if you do any manual "wheel up/down" actions.
For example call "wheel down 20 times" and then start "wheel downer". It will go a lot faster. 

This change should fix that. This change is consistent with the code for starting the gaze scroll.

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -181,7 +181,6 @@ class Actions:
     def mouse_scroll_down_continuous():
         """Scrolls down continuously"""
         global continuous_scoll_mode
-        stop_scroll()
         continuous_scoll_mode = "scroll down continuous"
         mouse_scroll(setting_mouse_continuous_scroll_amount.get())()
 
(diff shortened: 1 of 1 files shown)


code/mouse.py
```diff
@@ -181,6 +181,7 @@ def mouse_scroll_down():
     def mouse_scroll_down_continuous():
         """Scrolls down continuously"""
         global continuous_scoll_mode
+        stop_scroll()
```
knausj85, line 184: I believe this will break the ability to speed up the continuous scroll via repeaters as-is, no?

This should probably be fenced somehow, maybe via the `continuous_scoll_mode` variable or something, though it probably needs to be reset properly in stop_scroll

knausj85 (changes requested):

code/mouse.py
daveweij: No, speeding up is still possible using "wheel up", "wheel down" as this does not pass via the mouse_scroll_down/up_continuous.

daveweij: Or is "wheel downer" also supposed to speed up scrolling?

daveweij: I have updated the PR to allow speeding up with both "wheel down" and "wheel downer"

knausj85:
Thank you

PR #640, opened 2021-11-06
Browser address implementations for Safari and Google Chrome on Mac.

nriley:
Use accessibility where possible, otherwise AppleScript (which would generate an additional permission prompt the first time).  On Safari use the address bar (if it's visible); on Chrome use the first AXWebArea.

### DIFF
--- a/apps/chrome/chrome.py
+++ b/apps/chrome/chrome.py
@@ -1,4 +1,5 @@
 from talon import ctrl, ui, Module, Context, actions, clip, app
+from talon.mac import applescript
 
 ctx = Context()
 mod = Module()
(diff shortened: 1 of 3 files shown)


apps/chrome/chrome.py
```diff
@@ -38,9 +39,27 @@ def tab_close_wrapper():
...
         actions.app.tab_close()
 
+def chrome_app():
+    return ui.apps(bundle="com.google.Chrome")[0]
 
 @ctx.action_class("browser")
 class browser_actions:
+    def address() -> str:
+        try:
+            window = chrome_app().windows()[0]
```
knausj85, line 49: I believe you can simplify this to `ui.active_app.windows()[0]`, no?

apps/chrome/chrome.py
```diff
@@ -1,4 +1,5 @@
 from talon import ctrl, ui, Module, Context, actions, clip, app
+from talon.mac import applescript
```
auscompgeek, line 2: You probably want this change to go in `apps/chrome/mac.py`, right?

apps/chrome/chrome.py
nriley: Because this is triggered a lot (on every window/tab/title change) I was worried it might trigger after the current frontmost/active app changed so I wanted to address the browser directly.

apps/chrome/chrome.py
nriley: Makes sense, will fix. (Can you tell I don't use Chrome regularly?)

apps/chrome/chrome.py
knausj85: It’s not a big deal, but I don’t think that’s a concern here since the Chrome/Safari app has to be active for this implementation of address() to be active at all. I believe the browser scope is evaluated sometime later.

nriley: I was worried about a race condition if the user changes apps while the script is running. But maybe I am being overly cautious.

knausj85: Fair enough, works for me.

PR #641, opened 2021-11-06
Replacing phrases post-recognition

JelleAalbers:
This allows phrases of more than one word to be specified in `words_to_replace.csv` for replacement after recognition. E.g. get hub -> github. There are two open issues requesting this: https://github.com/knausj85/knausj_talon/issues/389 and https://github.com/knausj85/knausj_talon/issues/509. 

You can already achieve multi-word 'replacement' _during_ recognition by altering `user.vocabulary` via `additional_words.csv` (see https://github.com/knausj85/knausj_talon/issues/507). However, this affects recognition priority, and a large vocabulary increases DFA times.

Post-recognition phrase replacement also opens up other applications. For example, it offers an easy and DFA-free way to implement https://github.com/knausj85/knausj_talon/pull/636 (populate the phrase replacement map with 'brief abbreviation -> spoken_form' entries). I left that out of this PR.

Implementation:
  * This adds a new action `user.replace_phrases`, and uses it instead of `dictate.replace_words` in `dication.py` and `formatters.py`. User code using `dictate.replace_words` will continue to work for single-word replacement.
  * The work is done by a `PhraseReplacer` utility in a new file. It should work quickly even with large lists of replacement phrases that share common beginnings (foreshadowing things like https://github.com/knausj85/knausj_talon/pull/636). If multiple replacements are possible, phrases starting earlier have priority, then longer phrases.
  * `PhraseReplacer` stores phrases in three levels of dictionaries, indexed by first word, then number of subsequent words, then the actual subsequent words. Since dictionary lookups are O(1), checking if a replacement needs to happen is fast and essentially independent of the 'phrase vocabulary' size. (Unless you make an evil vocabulary where all phrases start with the same word and have unique lengths.)

Happy to hear ideas for alternate implementations! I'm still learning Talon and this codebase.


rntz:
Thanks for making a PR for this, this feature has been on my wishlist for a while! I haven't looked at this in detail yet, but in general we try to avoid cross-imports between python files. It looks like the phrase replacer is only used in dictation.py, so I think it should probably just be moved into dictation.py rather than getting its own file. If we find more use cases for it down the line that can't be accommodated there, then we can rethink this or maybe even ask aegis to add this feature to talon core.

JelleAalbers:
Sure! I now moved the code into `vocabulary.py` where `PhraseReplacer` was previously imported (I assume that's what you meant; the `user.replace_phrases` action is used in both `dictation.py` and `formatters.py`).

I was thinking of some use case like defining extra phrases to replace in some context only. But for that, I guess you'd currently have to import `phrases_to_replace` from `vocabulary.py` anyway, unless you want to replace rather than add to the list of phrases to replace.

JelleAalbers:
Just let me know if you'd like to see any changes to this, or if it's not quite what you're looking for. It still works on the current master, and the issue does come up on the slack from time to time (e.g. [here](https://talonvoice.slack.com/archives/C9MBPTXD4/p1639493843331400) and [here](https://talonvoice.slack.com/archives/C7ENXA7C4/p1637009960192100?thread_ts=1637009629.191400&cid=C7ENXA7C4)).

code/vocabulary.py
```diff
@@ -1,3 +1,5 @@
+from typing import Dict, Union
```
rntz, line 1: looks like `Dict` isn't used

code/vocabulary.py
```diff
@@ -90,3 +97,86 @@
...
+    Args:
+      - phrase_dict: dictionary mapping recognized/spoken forms to written forms
+    """
+
+    def __init__(self, phrase_dict: Dict[str, str]):
+        # Index phrases by first word, then number of subsequent words n_next
+        phrase_index = dict()
+        for spoken_form, written_form in phrase_dict.items():
+            words = spoken_form.split()
+            first_word, n_next = words[0], len(words) - 1
```
rntz, line 116: You might want to handle the case that `spoken_form.split()` is an empty list here, probably by logging a warning (`logging.warning("blah")`) and ignoring that item. Otherwise if a user screws up their csv file by adding an empty spoken form, all replacements will fail to work.

code/vocabulary.py
```diff
@@ -90,3 +97,86 @@
...
+            input_text = input_text.split()
+        input_words = tuple(input_text)
+
+        output_words = []
+        first_word_i = 0
+        while first_word_i <= len(input_words) - 1:
+            first_word = input_words[first_word_i]
+            next_word_i = first_word_i + 1
+            # Could this word be the first of a phrase we should replace?
+            for n_next, phrases_n_next in self.phrase_index.get(first_word, dict()).items():
```
rntz, line 141: In order for this to prefer replacing longer phrases, you seem to be relying on the iteration order being the same as the order you used when you made the dictionary on line 123-126. I don't think this is guaranteed (https://docs.python.org/2/library/stdtypes.html#dict.items), especially not across different Python implementations (this matters - Talon used to use PyPy!).

It's probably better to just use `sorted()` here to sort the relevant potential matches by length. Or you could use a list of pairs instead of a dictionary for the second-level index.

rntz (changes requested):
Looks mostly good except for a few small changes, in particular making sure you sort the replacements longest-first. Sorry for the long delay here. Also it's likely this PR will clash/merge-conflict with #696 , so whichever gets merged second will need some small revising.

code/vocabulary.py
JelleAalbers: Good point, done!

code/vocabulary.py
JelleAalbers: Changed to using an OrderedDict now -- should be as explicit as sorting during iteration, but faster for large vocabularies.

You're right that ordered dictionary iteration used to be a cPython implementation detail, but it was [blessed](https://docs.python.org/3/whatsnew/3.7.html#summary-release-highlights) into the language spec in python 3.7. OrderedDict still has advantages though (e.g. https://stackoverflow.com/questions/50872498).

code/vocabulary.py
JelleAalbers: It's used in a type annotation of `PhraseReplacer.__init__`, https://github.com/JelleAalbers/knausj_talon/blob/replace_phrases/code/vocabulary.py#L113; unless I'm missing something

JelleAalbers:
Thanks for your comments, and no worries about the delay!  Happy to adjust after #696 if you like, and feel free to squash-merge if you prefer the cleaner history.

rntz:
I pushed some changes, mainly (1) adding a fallback call to `actions.dictate.replace_words`, just in case, since this code is on a critical path for text input; (2) just using a list instead of an OrderedDict, since the only thing we do is iterate over it; (3) making the type of `user.replace_phrases` consistent with `dictate.replace_words`.

I tested it out by merging with my local knausj and everything seems to work okay, so I'm happy to merge if you are :).

rntz (approved):

JelleAalbers:
Great, those are nice simplifications / additions for robustness. Yes, after building the index we just iterate over the second level, so converting it to a list makes complete sense.

PR #642, opened 2021-11-07
python: Add match statement support

auscompgeek:
This changes `state switch` and `state case` to insert the keywords for [PEP 634: Structural Pattern Matching](https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching), added in Python 3.10.


rntz:
Thanks!

PR #643, opened 2021-11-07
web: Explain what to configure in window titles

auscompgeek:
The [previously suggested Firefox extension](https://addons.mozilla.org/en-US/firefox/addon/domain-in-title/) only appends `[hostname]` to the tab title. However, `code/browser.py` in #631 requires both the scheme and host to be present.

This replaces that suggestion with [a Firefox extension](https://addons.mozilla.org/en-US/firefox/addon/keepass-helper-url-in-title/) that prepends the URL to the window title (not the tab title) and adds some guidance on how to configure it to work with these scripts.

### DIFF
--- a/apps/web/window_titles.md
+++ b/apps/web/window_titles.md
@@ -1,8 +1,17 @@
-These talon files mostly match based on the url of the target site, however by
-default talon cannot find these URLs. To get the URL listed in the tab name,
-and thus the win.title and browser.url, you need to use a plug in:
-Firefox: https://addons.mozilla.org/en-US/firefox/addon/domain-in-title/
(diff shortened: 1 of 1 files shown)


nriley:
If we can figure out how to implement something like #640 for more browsers/platforms (or even if we don't), it might make sense for us to mention this as an alternative too. For example, s of Firefox 87, Firefox on Mac does support accessibility access but it's not on by default as it is in Safari/Chrome (requires you set accessibility.force_disabled to -1, which seems to be documented [almost nowhere](https://apple.stackexchange.com/questions/404841/get-url-of-opened-firefox-tabs-from-terminal)). Firefox on Mac is not usefully scriptable for this purpose unlike Safari/Chrome.

splondike:
Thanks for the improved extension options! I reworked the wording a little to better explain the current state of things and to reference nriley's work on getting Safari and Chrome working by default on Mac.

PR #647, opened 2021-11-12
Basic support for SQL

RodrigoChT:
- Use the existing language system (previous implementation only used a talon file).
- Make SQL the active language in MySQLWorbench and DataGrip


lang/sql/sql.talon
```diff
@@ -0,0 +1,38 @@
...
+when: "WHEN "
+then: "THEN "
+else: "ELSE "
+end: "END"
+as: " AS "
+where: "WHERE "
+between: " BETWEEN "
+and: "AND "
+or: "OR "
+is: "IS "
```
knausj85, line 27: I'm a little hesitant on some of the short commands (as/is/or/not/in/else/end), which is generally why other languages use e.g. `state else`

knausj85:
I think there's an existing `datagrip.talon` in the Mac folder that we probably want to relocate to the new app folder as part of the this PR?

lang/sql/sql.talon
```diff
@@ -0,0 +1,38 @@
+mode: command
+and mode: user.sql
+mode: command
+and mode: user.auto_lang
+and code.language: sql
+-
+tag(): user.code_generic
+tag(): user.code_operators
+
+# Some commands have spaces around them to avoid having to continually say 'space'.
```
knausj85, line 10: I haven't done SQL in forever, so I don't remember anything. But if you need a variety of space options, it may be prudent to throw these in a list.

Then you could do things like...
```
pad <sql_list>: " {user.sql} "
```

lang/sql/sql.talon
RodrigoChT: I considered that option, but SQL (at least when dealing with relatively simple queries) tends to sound a lot like plain English. Having to say `state` before each command might be cumbersome.

lang/sql/sql.talon
RodrigoChT: The problem I see with that approach is that it would be way more cumbersome to dictate the queries.

lang/sql/sql.talon
```diff
@@ -0,0 +1,38 @@
...
+# The ones that don't have an initial space are expected to be at the start of a new line.
+# This could probably be done in a better way.
+select: "SELECT "
+star: "* "
```
auscompgeek, line 14: Note that `star` is already defined as a command as a symbol key. This would be overriding that.

auscompgeek (commented):
Looks like this now conflicts with master, since `lang/sql.talon` was moved to `lang/sql/sql.talon`.

lang/sql/sql.talon
knausj85: Makes sense, I can see both sides here. I'd like to run this by some other collaborators since I don't do SQL. @splondike @rntz @pokey

lang/sql/sql.talon
nriley: Agree shadowing commands for basic keys is not a good idea. One option might be to define a single command for `SELECT * FROM` the way I do:

https://github.com/nriley/knausj_talon/blob/nriley/lang/sql/sql.talon#L12

lang/sql/sql.talon
nriley: `IS` is not its own operator in SQL. I do this by defining separate `IS NULL` and `IS NOT NULL` commands (https://github.com/nriley/knausj_talon/blob/nriley/lang/sql/sql.talon#L20) which is what I would recommend you do as well.

For `AND` and `OR`, these are operators so should be defined as such just like any other programming language.

lang/sql/sql.py
```diff
@@ -0,0 +1,65 @@
...
+    "timestamp": "TIMESTAMP",
+    "count": "COUNT",
+    "cast": "CAST",
+}
+
+sql_join_types = [
+    "LEFT",
+    "RIGHT",
+    "INNER",
+    "OUTER"
```
nriley, line 28: Should consider adding FULL for completeness' sake.

lang/sql/sql.talon
```diff
@@ -0,0 +1,38 @@
...
+else: "ELSE "
+end: "END"
+as: " AS "
+where: "WHERE "
+between: " BETWEEN "
+and: "AND "
+or: "OR "
+is: "IS "
+not: "NOT "
+in: "IN "
```
nriley, line 29: Is `NOT` ever used without `IN` or `NULL`? Same as `IS NULL` and `IS NOT NULL` above.

Here's how I implement `IN`/`NOT IN` as operators: https://github.com/nriley/knausj_talon/blob/nriley/lang/sql/sql.py#L29

lang/sql/sql.py
```diff
@@ -0,0 +1,65 @@
...
+
+mod = Module()
+ctx = Context()
+ctx.matches = r"""
+mode: user.sql
+mode: user.auto_lang
+and code.language: sql
+"""
+
+ctx.lists["user.code_functions"] = {
```
nriley, line 11: Do you have any voice command to insert these functions? I had to do this myself (https://github.com/nriley/knausj_talon/blob/nriley/lang/sql/sql.py#L47)

lang/sql/sql.py
```diff
@@ -0,0 +1,65 @@
...
+            else:
+                text = text + "()"
+            actions.insert(text)
+            actions.edit.left()
+            
+@mod.action_class
+class module_actions:
+    def join_tables(text: str):
+        """Inserts a JOIN statement"""
+        text = " ".join([text, "JOIN  ON"])
```
auscompgeek, line 61: `ON` isn't the only keyword that follows `JOIN <table>`:

- There's no join condition to specify for a `NATURAL JOIN`.
- PostgreSQL allows for specifying an equi-join with `USING`.

Also, this line is just… weird. Why is this not just an f-string?

rntz (changes requested):
Changes were requested in prior reviews, adding this to put the "changes requested" tag into github's UI.

nriley:
Since there has been no movement in our comments in nearly 2 months I plan to clean up and propose merging my own SQL  support (https://github.com/nriley/knausj_talon/tree/nriley/lang/sql).

rntz:
Superceded by #885, reopen if you think there's something more to be discussed.

rntz:
On review it looks like #885 doesn't have an equivalent of apps/data_grip/data_grip.py and apps/mysql_workbench/mysql_workbench.py and doesn't include as many built-in functions in `code_common_function`. I'm going to reopen this as a reminder to myself to try to include those.

auscompgeek:
Assuming DataGrip was always SQL struck me as odd anyway, since it's a JetBrains IDE.

nriley:
Yeah, I didn't include those because they seemed out of scope.

As @auscompgeek wrote, DataGrip is a JetBrains IDE so it actually supports multiple languages — should not assume SQL.

Several of the "functions" in this PR are not actually functions which is why I didn't include them. There are lots of SQL dialects with incompatible syntaxes and I tried to pick some ones that were in common.

PR #648, opened 2021-11-12
Add page navigation for reader apps

timo95:
Add tags to navigate pages and chapters.

Implement pages (and chapters/tabs when applicable) for reader apps:
- Adobe Acrobat Reader DC (mac missing)
- Calibre Viewer (mac missing)
- Kindle (mac missing)
- Okular (mac missing)
- Atril
- Evince
- Nitro Reader 5
- Sumatra PDF

Since I don't have a Mac, I could only implement for Windows and Linux.

### DIFF
--- a/apps/adobe/win.py
+++ b/apps/adobe/win.py
@@ -1,0 +1,41 @@
+from talon import Context, actions
+
+# Context matching
+ctx = Context()
+ctx.matches = """
+os: windows
+app: adobe_acrobat_reader_dc
+"""
+
+
+# --- Implement actions ---
+@ctx.action_class('app')
+class AppActions:
+    # app.tabs
+    def tab_next(): actions.key('ctrl-tab')
+    def tab_previous(): actions.key('ctrl-shift-tab')
+
+
+@ctx.action_class('edit')
+class EditActions:
+    def zoom_in(): actions.key("ctrl-0")
+    def zoom_out(): actions.key("ctrl-1")
+    def zoom_reset(): actions.key("ctrl-2")
+
+
+@ctx.action_class("user")
+class UserActions:
+    # user.pages
+    def page_current():
+        actions.key("ctrl-shift-n")
+        page = actions.edit.selected_text()
+        actions.key("tab:2 enter")
+        return int(page)
+    def page_next(): actions.key("right")
+    def page_previous(): actions.key("left")
+    def page_jump(number: int):
+        if number > 0:
+            actions.key("ctrl-shift-n")
+            actions.insert(str(number))
+            actions.key("enter")
+    def page_final(): actions.key("end")
--- a/apps/adobe/win.py
+++ b/apps/adobe/win.py
@@ -18,8 +18,8 @@ class AppActions:
 
 @ctx.action_class('edit')
 class EditActions:
-    def zoom_in(): actions.key("ctrl-0")
-    def zoom_out(): actions.key("ctrl-1")
+    def zoom_in(): actions.key("ctrl-0")  # in german version
+    def zoom_out(): actions.key("ctrl-1")  # in german version TODO: differentiate languages
     def zoom_reset(): actions.key("ctrl-2")
 
 
@@ -31,11 +31,10 @@ class UserActions:
         page = actions.edit.selected_text()
         actions.key("tab:2 enter")
         return int(page)
-    def page_next(): actions.key("right")
-    def page_previous(): actions.key("left")
+    def page_next(): actions.key("ctrl-pagedown")
+    def page_previous(): actions.key("ctrl-pageup")
     def page_jump(number: int):
-        if number > 0:
-            actions.key("ctrl-shift-n")
-            actions.insert(str(number))
-            actions.key("enter")
+        actions.key("ctrl-shift-n")
+        actions.insert(str(number))
+        actions.key("enter")
     def page_final(): actions.key("end")
--- a/apps/calibre/calibre.py
+++ b/apps/calibre/calibre.py
@@ -1,0 +1,19 @@
+from talon import Module
+
+# --- App definition ---
+mod = Module()
+mod.apps.calibre = """
+os: windows
+and app.name: calibre.exe
+os: windows
+and app.exe: calibre.exe
+os: windows
+and app.name: calibre-parallel.exe
+os: windows
+and app.exe: calibre-parallel.exe
+"""
+mod.apps.calibre = """
+os: linux
+app.name: calibre
+"""
+# TODO: mac context
--- a/apps/calibre/calibre_viewer.py
+++ b/apps/calibre/calibre_viewer.py
@@ -1,0 +1,30 @@
+from talon import Module, Context, actions
+
+# --- App definition ---
+mod = Module()
+mod.apps.calibre_viewer = """
+app: calibre
+title: /E-book viewer$/
+title: /eBook-Betrachter$/
+"""
+
+# Context matching
+ctx = Context()
+ctx.matches = """
+os: windows
+os: linux
+app: calibre_viewer
+"""
+# TODO: mac implementation
+
+
+# --- Implement actions ---
+@ctx.action_class("user")
+class UserActions:
+    # user.pages
+    def page_next(): actions.key("pagedown")
+    def page_previous(): actions.key("pageup")
+    def page_final(): actions.key("ctrl-end")
+    # user.chapters
+    def chapter_next(): actions.key("ctrl-pagedown")
+    def chapter_previous(): actions.key("ctrl-pageup")
--- a/apps/kindle/win.py
+++ b/apps/kindle/win.py
@@ -15,7 +15,6 @@ class UserActions:
     def page_next(): actions.key("down")
     def page_previous(): actions.key("up")
     def page_jump(number: int):
-        if number > 0:
-            actions.key("ctrl-g")
-            actions.insert(str(number))
-            actions.key("enter")
+        actions.key("ctrl-g")
+        actions.insert(str(number))
+        actions.key("enter")
--- a/apps/kindle/win.py
+++ b/apps/kindle/win.py
@@ -1,0 +1,21 @@
+from talon import Context, actions
+
+# Context matching
+ctx = Context()
+ctx.matches = """
+os: windows
+app: kindle
+"""
+
+
+# --- Implement actions ---
+@ctx.action_class("user")
+class UserActions:
+    # user.pages
+    def page_next(): actions.key("down")
+    def page_previous(): actions.key("up")
+    def page_jump(number: int):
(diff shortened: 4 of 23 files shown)


code/pages.py
```diff
@@ -0,0 +1,22 @@
...
+@mod.action_class
+class Actions:
+    def page_current() -> int:
+        """Return current page number"""
+    def page_next():
+        """Go to next page"""
+        actions.user.page_jump(actions.user.page_current() + 1)
+    def page_previous():
+        """Go to previous page"""
+        actions.user.page_jump(actions.user.page_current() - 1)
```
pokey, line 18: Do we want to do the `min` thing here that you did for chapters?

apps/platforms/win/sumatrapdf/sumatrapdf.py
```diff
@@ -0,0 +1,52 @@
...
+    def page_previous(): actions.key("p")
+    def page_jump(number: int):
+        if number > 0:
+            actions.key("ctrl-g")
+            actions.insert(str(number))
+            actions.key("enter")
+    def page_final(): actions.key("end")
+    # user.tabs
+    def tab_jump(number: int):
+        if number < 9:
```
pokey, line 50: Shall we raise an exception otherwise so it doesn't just silently do nothing?

apps/platforms/win/nitro_reader/nitro_reader_5.py
```diff
@@ -0,0 +1,40 @@
...
+    def page_jump(number: int):
+        if number > 0:
+            actions.key("ctrl-g")
+            actions.edit.select_line()
+            actions.insert(str(number))
+            actions.key("enter alt:2")
+    def page_final(): actions.key("end")
+    # user.tabs
+    def tab_jump(number: int): pass
+    def tab_final(): pass
```
pokey, line 40: What do these do?

apps/platforms/win/nitro_reader/nitro_reader_5.py
```diff
@@ -0,0 +1,40 @@
...
+    def tab_open(): actions.key("ctrl-shift-o")
+
+
+@ctx.action_class("user")
+class UserActions:
+    # user.pages
+    def page_next(): actions.key("right")
+    def page_previous(): actions.key("left")
+    def page_jump(number: int):
+        if number > 0:
```
pokey, line 32: not sure we need this guard

apps/platforms/linux/evince/evince.py
```diff
@@ -0,0 +1,33 @@
...
+    # user.pages
+    def page_current():
+        actions.key("ctrl-l")
+        page = actions.edit.selected_text()
+        actions.key("escape")
+        return int(page)
+    def page_next(): actions.key("n")
+    def page_previous(): actions.key("p")
+    def page_jump(number: int):
+        if number > 0:
```
pokey, line 29: as above, not sure we need this guard

apps/adobe/win.py
```diff
@@ -0,0 +1,41 @@
...
+    # user.pages
+    def page_current():
+        actions.key("ctrl-shift-n")
+        page = actions.edit.selected_text()
+        actions.key("tab:2 enter")
+        return int(page)
+    def page_next(): actions.key("right")
+    def page_previous(): actions.key("left")
+    def page_jump(number: int):
+        if number > 0:
```
pokey, line 37: as above

apps/calibre/calibre.py
```diff
@@ -0,0 +1,19 @@
...
+os: windows
+and app.exe: calibre.exe
+os: windows
+and app.name: calibre-parallel.exe
+os: windows
+and app.exe: calibre-parallel.exe
+"""
+mod.apps.calibre = """
+os: linux
+app.name: calibre
```
pokey, line 17: Should this not be
```suggestion
and app.name: calibre
```

apps/calibre/calibre_viewer.py
```diff
@@ -0,0 +1,30 @@
...
+title: /E-book viewer$/
+title: /eBook-Betrachter$/
+"""
+
+# Context matching
+ctx = Context()
+ctx.matches = """
+os: windows
+os: linux
+app: calibre_viewer
```
pokey, line 16: I think this needs to be
```suggestion
os: windows
and app: calibre_viewer
os: linux
and app: calibre_viewer
```

But maybe I'm missing something?

apps/kindle/win.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+# --- Implement actions ---
+@ctx.action_class("user")
+class UserActions:
+    # user.pages
+    def page_next(): actions.key("down")
+    def page_previous(): actions.key("up")
+    def page_jump(number: int):
+        if number > 0:
```
pokey, line 18: as above

apps/okular/okular.py
```diff
@@ -0,0 +1,43 @@
...
+and app.name: okular
+"""
+# TODO: mac context and implementation
+
+# Context matching
+ctx = Context()
+ctx.matches = """
+os: windows
+os: linux
+app: okular
```
pokey, line 22: As above, don't you need to do
```suggestion
os: windows
and app: okular
os: linux
and app: okular
```

pokey (commented):
Good stuff!  See inline for some minor comments.  Would also be great to get someone with Windows to take one of these apps for a spin to double check cc/ @knausj85

apps/calibre/calibre_viewer.py
timo95: Those are equivalent.

> Each kind of requirement can be listed several times. Entries of the same kind of requirement are OR‘d together, and of different kinds are AND‘d.

See: https://talon.wiki/unofficial_talon_docs/#context-header

apps/okular/okular.py
timo95: See above

apps/platforms/win/nitro_reader/nitro_reader_5.py
timo95: Tab jump isn't possible in nitro. This overwrites the default implementation of these actions so it doesn't do random stuff.

apps/calibre/calibre.py
timo95: See below

apps/calibre/calibre_viewer.py
pokey: Oh wow TIL thanks

apps/adobe/win.py
```diff
@@ -0,0 +1,41 @@
...
+@ctx.action_class('app')
+class AppActions:
+    # app.tabs
+    def tab_next(): actions.key('ctrl-tab')
+    def tab_previous(): actions.key('ctrl-shift-tab')
+
+
+@ctx.action_class('edit')
+class EditActions:
+    def zoom_in(): actions.key("ctrl-0")
```
knausj85, line 21: Shouldn't zoom out/in be `ctrl--`/`ctrl-+`?

apps/adobe/win.py
```diff
@@ -0,0 +1,41 @@
...
+
+@ctx.action_class("user")
+class UserActions:
+    # user.pages
+    def page_current():
+        actions.key("ctrl-shift-n")
+        page = actions.edit.selected_text()
+        actions.key("tab:2 enter")
+        return int(page)
+    def page_next(): actions.key("right")
```
knausj85, line 34: This doesn't work if the cursor is in the window somewhere. Maybe do `actions.key("esc right")`?

knausj85 (changes requested):

apps/adobe/win.py
timo95: There seems to be conflicting information on the internet. At least for me it is ctrl-0/1 (german version). 
grafik

code/pages.py
timo95: I actually removed that because i didn't want the flickering of jumping from page 1 to page 1 (not really applicable in any of these apps anyways). I just forgot it in chapters.

pokey: Ok I'm not sure I follow but do what you think makes sense here; was just flagging the discrepancy

apps/platforms/win/nitro_reader/nitro_reader_5.py
pokey: I don't believe these actions have default implementations, do they?

knausj85: I added default implementations for these per-OS since they work for many apps, and the tag is required to opt-in. It seemed like a reasonable compromise.

https://github.com/knausj85/knausj_talon/blob/a05c3d2e3fb5cb8b1484a1e23c67413a681d6b17/code/platforms/win/app.py#L13

apps/adobe/win.py
knausj85: Well, that's irritating. I'm honestly not sure how to handle this sort of thing. Maybe live with this for now, and submit an issue? Open to suggestions.

apps/platforms/win/nitro_reader/nitro_reader_5.py
pokey: Ok prob just a nit then, but shouldn't we remove these definitions if they don't have default definitions?  Prob better to just let them error out, no?

apps/adobe/win.py
pokey: Yeah that's frustrating.  Not sure what the right way to handle this one is.  Is ctrl-0/1 the common way to zoom in / out on German version of Windows?  If so I guess we could add a tag for German OS or something and make that global?

apps/platforms/win/nitro_reader/nitro_reader_5.py
knausj85: Yeah, I tend to agree with that. It's probably better than no feedback.

apps/adobe/win.py
timo95: Any other app uses `ctrl-+` and `ctrl--`. Only Acrobat is weird.

timo95: But there are other shortcuts that differ per language. For example the "window next" one.

knausj85 (approved):

apps/platforms/win/sumatrapdf/sumatrapdf.py
timo95: You should open an issue for that. Multiple apps implement it like this.

apps/adobe/win.py
splondike: Maybe a tag or maybe there's a (windows?) API that could be used to determine the current language. In any case I think this can be dealt with in the next PR.

splondike:
Thanks for all the work making actions for those apps!

PR #649, opened 2021-11-12
Add shortcut support for Apple notes

RodrigoChT:


phillco (commented):
Awesome!

apps/platforms/mac/notes/notes.talon
```diff
@@ -0,0 +1,36 @@
...
+apply checklist: key(shift-cmd-l)
+increase font: key(cmd-+)
+decrease font: key(cmd--)
+indent: key(cmd-])
+unindent: key(cmd-[)
+line break: key(ctrl-enter)
+mark: key(shift-cmd-u)
+move up: key(ctrl-cmd-up)
+move down: key(ctrl-cmd-down)
+zoom in: key(shift-cmd->)
```
knausj85, line 34: I would suggest implementing a few of existing actions for some of these, so the standard global commands will work instead of defining new commands:

```
edit.zoom_in()
edit.zoom_out()
edit.indent_more()
edit.indent_less()
```

zoom in/out are defined in standard.talon, and I think some indent commands are defined in generic_editor.talon

apps/platforms/mac/notes/notes.talon
```diff
@@ -0,0 +1,36 @@
+os: mac
+and app: notes
+-
+new note: key(cmd-n)
+duplicate note: key(cmd-d)
+new folder: key(shift-cmd-n)
+toggle folders: key(alt-cmd-s)
+show main: key(cmd-0)
+show as list: key(cmd-1)
```
knausj85, line 9: I might suggest matching the existing finder commands for the view options for consistency.

from finder.talon:
```
icon view: key(cmd-1)
column view: key(cmd-3)
list view: key(cmd-2)
gallery view: key(cmd-4)
```

apps/platforms/mac/notes/notes.talon
RodrigoChT: You are right, I addressed this in the commit https://github.com/knausj85/knausj_talon/pull/649/commits/e9e4613e002841debd7f0ee66b4019071b3b18cd.

apps/platforms/mac/notes/notes.talon
RodrigoChT: Good point, I renamed a couple of commands here https://github.com/knausj85/knausj_talon/pull/649/commits/6c3df7443ae746c6df4aecae116a2a8c484086f1.

I'm not sure that `show main` and `toggle folders` have a direct match in Finder though.

RodrigoChT:
> 1. I don't think `user.edit` tag  exists in the community repository?
> 2. I'm not sure we want line_commands. Does notes support a `edit.jump_line(number)` action? If not, the vast majority of commands in line_commands.talon won't work.
> 
> The indent commands here
> 
> https://github.com/knausj85/knausj_talon/blob/182b28bb04494f55481003539536305e160295a3/text/generic_editor.talon#L97
> 
> 
> should work with the indent* actions you added, fwiw.

@knausj85, thank you for your feedback. Going point by point:

1. You are right, I removed the incorrect tag.
2. I removed the `line_commands` tag and added 2 commands specific to Notes for swapping lines.
3. I removed the `indent [more]` command, but had to keep the `indent less` command because the generic one doesn't work in Notes.

knausj85:
Looks great, thanks for the contribution and your patience

Issue #651, opened 2021-11-15
Fine grain support for programming language features


wenkokke:
I would like to refractor code.py and the generic tag into several smaller files and tags. 

Currently, the generic tag is largely geared towards imperative C++-like languages. I have been attempting to implement support for Haskell, and I have found that Haskell uses almost none of the commands defined in generic.

Furthermore, while groups of features can be modularly defined using tags, the functions used by all code tags are declared in a single monolithic class in code.py.

I suggest that we move towards a convention where the Python file which declares methods is stored next to the .talon file which uses them, which is already the convention used by the concrete implementations of programming languages.

- [x] Break up code.py into one file per tag and move them into `lang/tags/` 
      PR #652)
- [x] Break up existing tags into more fine grained sets of features
      PR  #656 (WIP)

pokey:
@wenkokke can we close this now?

wenkokke:
Fixed in #656.

PR #652, opened 2021-11-15
Break up code.py into one file per tag

wenkokke:
Implements the changes proposed in the first step of #651.

This pull request should not contain any breaking changes.


splondike (commented):
lang/snippets.talon -> code/snippets.talon should probably move to something like apps/generic_snippets.talon . It's more an editor/IDE thing than a language thing I think.

I believe the intention with the code/ folder was to just have Python in there (at least that's how it currently works).

Aside from that, this change looks good to me. It's nicer to have things broken up like this.

fidgetingbits:
agreed about snippets changing to generic_snippets.talon. also agree otherwise it looks much better this way

wenkokke:
I've migrated the code for snippets into the apps folder, under `generic_snippets`.

splondike (approved):
LGTM, will wait for a second approval to merge.

knausj85 (approved):

knausj85:
Looks good, thanks

PR #654, opened 2021-11-15
Safari shortcuts fix

atacan:
correcting some of the shortcuts for Safari


atacan:
I meant to include only 7aff6e6

PR #656, opened 2021-11-15
Break up existing tags into more fine grained sets of features

wenkokke:
Implements the changes proposed in the second step of #651.

- Moved array subscription into its own tag.
- Moved anonymous functions into their own tag.
- Moved pointer operators into their own tag.
- Moved assignment operators into their own tag, as they are not generally supported outside of imperative languages.
- Moved bitwise operators into their own tag.
- Moved graphical helpers for functions and libraries into their own tag.
- Moved `code_operator_in` into `r.py` as it is only used there.
- Moved `code_include`, `code_include_system` and `code_include_local` into `c.py` as they are only used there.
- Moved `code_type_definition` and `code_typedef_struct` into `c.py` as they are only used there.
- Moved `code_from_import` to the specific languages which implemented it. (Abstracting over the imports is a task for a future PR.)
- Moved the function and visibility commands into their own tag.
- Moved the import commands into their own tag.
- Moved commands for inserting bools into their own tag.
- Moved commands for inserting and testing null into their own tag.
- Removed `code_type_struct` and `code_state_in` they were not used anywhere.


wenkokke:
It seems like the whole visibility infrastructure (e.g., code_default_function , code_private_function, etc) is never actually used generically? The whole series of functions is defined in the generic tag, but not a single generic command actually uses them.

wenkokke:
Most languages don't actually have an analogue for Python's `from x import ys` statement, and many of the language modes, such as the C# mode, simply implement `code_from_import` as `code_import`:
```python
    def code_import():         actions.auto_insert('using  ')
    def code_from_import():    actions.auto_insert('using ')
```
It may make sense to migrate these commands to the Python mode (and any other modes for languages which do have a direct analogue).

wenkokke:
The function `code_document_string` is defined in generic, and is implemented by the Python and Ruby modes. However, it is not called anywhere, and neither `generic.talon`, `python.talon`, nor `ruby.talon` exposes a command that uses it.

splondike:
Will leave off reviewing this properly until the first commit gets merged (since this PR depends on it). I have some thoughts on some of your comments below. Bear in mind I don't actually use Talon too much for coding so I will defer to other folks' usability opinions.

* The function definition stuff (code_private_function) indeed looks like it's not being used consistently. I'd be in favour of pushing the whole function definition stuff down to the individual languages since there's heaps of variation here. E.g. 'private function foobar' -> 'def __foobar():' makes sense in Python, but what should it do in C, Java where you need return types before the name? In Ruby you can do inline visibility modifiers, but you can also do block level ones. Haskell has the optional type sig on one line, and then the name and arguments on the next.
* For the from x import y thing indeed seems very Python specific. All it does is type out 'from | import ' (positioning the cursor at the pipe). Lots of languages have a construct like this (Haskell, Scala, Ecmascript), but I don't see an action like this being especially useful for their syntax. I'd be in favour of pushing this down to the Python level.
* code_document_string is used in programming.talon (dock string command).

wenkokke:
> Will leave off reviewing this properly until the first commit gets merged (since this PR depends on it). I have some thoughts on some of your comments below. Bear in mind I don't actually use Talon too much for coding so I will defer to other folks' usability opinions.
> 
>     * The function definition stuff (code_private_function) indeed looks like it's not being used consistently. I'd be in favour of pushing the whole function definition stuff down to the individual languages since there's heaps of variation here. E.g. 'private function foobar' -> 'def __foobar():' makes sense in Python, but what should it do in C, Java where you need return types before the name? In Ruby you can do inline visibility modifiers, but you can also do block level ones. Haskell has the optional type sig on one line, and then the name and arguments on the next.
> 
>     * For the from x import y thing indeed seems very Python specific. All it does is type out 'from | import ' (positioning the cursor at the pipe). Lots of languages have a construct like this (Haskell, Scala, Ecmascript), but I don't see an action like this being especially useful for their syntax. I'd be in favour of pushing this down to the Python level.
> 
>     * code_document_string is used in programming.talon (dock string command).

Thank you for your feedback! I agree with all of your suggestions, and I did miss the `dock string` command.

rntz:
Requested changes:

1. unnest the `lang/tags/*` directories and replace with file name prefixes
2. make a tag that covers the function inserting commands like code_insert_function, code_private_function etc
3. rename code_comment_line to code_comment_line_prefix or whatever is consistent

wenkokke:
I've made the following changes following our discussion:

- Flattened the directory structure of tags into a single directory, prefixing the individual tag files.
- Renamed the function for inserting line comments to be consistent.
- Created a function tag containing the functionality around inserting functions with certain visibilities.

The only remaining commands in the imperative tag were the commands around inserting booleans and null. Rather than moving these to a "miscellaneous" tag, I have moved them into their own data tags (i.e., tags for commands relating to inserting values).

lang/tags/operators_math.talon
```diff
@@ -0,0 +1,25 @@
...
+(op | is) (less | below) [than]: user.code_operator_less_than()
+(op | is) greater [than] or equal: user.code_operator_greater_than_or_equal_to()
+(op | is) less [than] or equal: user.code_operator_less_than_or_equal_to()
+
+# logical operators
+(op | logical) and: user.code_operator_and()
+(op | logical) or: user.code_operator_or()
+
+# TODO: This operator should either be abstracted into a function or removed.
+(op | pad) colon: " : "
```
auscompgeek, line 25: I imagine this is commonly used in C++ and Java, where it's used in for-each loops, but also subclassing in C++. There's probably a better place to put this than lumping them into the maths operators.

wenkokke: It's not really an abstracted operator, who am not exactly sure where it should go. Ideally we would abstract it out and give it some sensible name. Right now it has no semantics associated with it, and simply print a spaced colon.

wenkokke: From what you have said it should be split up into several different operators, one for subclassing and one for "in" in foreach loops.

wenkokke: For now I can move it to a miscellaneous tag, since deciding what should happen to this one command is beyond the scope of this pull request.

rntz: I think it's probably better to just leave this where it is than to create a whole file for one operator.

I also don't think operators/commands need to have semantic names; just memorable ones.

auscompgeek:
> Moved code_operator_in into r.py as it is only used there.

Possibly out of the scope of this PR, but I feel like it should also be implemented for Python, since it's an operator in its own right there too.

wenkokke:
> > Moved code_operator_in into r.py as it is only used there.
> 
> Possibly out of the scope of this PR, but I feel like it should also be implemented for Python, since it's an operator in its own right there too.

It should, but not as part of this pull request.

lang/tags/operators_math.talon
wenkokke: @auscompgeek I have created a miscellaneous tag as requested.

auscompgeek (commented):
As someone who wrote some commands for CSS (which I would like to contribute), I like this overall. So many of the operator actions don't make sense in CSS, yet I still want to pull a bunch of them in.

lang/tags/operators_pointer.py
```diff
@@ -0,0 +1,19 @@
+from talon import Context, Module
+
+ctx = Context()
+mod = Module()
+
+# TODO: rename this tag to 'code_operators_pointer' for consistency?
```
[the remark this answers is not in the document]
wenkokke: I removed the comment. It was there as a suggestion to rename the operators tags to operator (singular), for consistency reasons. I'm assuming the comment got pluralized at some point with a refactoring.

pokey:
This is looking really good I think before we merge though we want to have one of the maintainers pull this and try using it locally for a little while. Any volunteers? cc/ @splondike @rntz

lang/talon/talon.talon
```diff
@@ -4,9 +4,10 @@ mode: command
 and mode: user.auto_lang
 and code.language: talon
 -
-tag(): user.code_operators
-tag(): user.code_comment
-# uncomment user.talon_populate_lists tag to activate talon-specific lists of actions, scopes, modes etcetera. 
+tag(): user.code_operators_math
```
splondike, line 7: With this change .talon files lose the 'op (equals | assign)' command. Would just need to add in the tag(): user.code_operators_assignment tag here.

lang/java/java.talon
```diff
@@ -30,24 +41,16 @@ type {user.code_type} array:
...
-    key("space") 
+    key("space")
 
 op array:
     user.code_operator_subscript()
 
 op new:
     insert("new ")
-
-# Methods
```
splondike, line 47: Noting that this addresses #662 (replaces java specific function commands with the generic ones).

splondike (commented):
This looks good, I only found one small breakage.

splondike:
I think this is good to merge after fixing the merge conflict and the small issue I noted with .talon files.

I only played around with this for a couple of minutes, but I did write a basic .talon file parser and use it to compare the command sets available to the different languages before and after this change :D.

apps/generic_snippets/generic_snippets.py
```diff
@@ -2,7 +2,7 @@
 from talon import Module, actions, app, Context, imgui, registry
 
 mod = Module()
-mod.tag("snippets", desc="Tag for enabling code snippet-related commands")
+mod.tag("generic_snippets", desc="Tag for enabling code snippet-related commands")
```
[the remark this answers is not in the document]
wenkokke: I’m happy to revert it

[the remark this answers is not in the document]
splondike: Yeah, let's do that. No particular need to change it at this time.

wenkokke:
I've made the requested changes

splondike:
Looks good, thanks for the persistence!

PR #659, opened 2021-11-17
numbers.py: disallow leading scale words

### LINKED ISSUE #634: "million", "billion", "thousand" etc should not be valid numbers
They should require a prefix such as "one", eg "one million"

rntz:
fixes #634, so bare "hundred" "thousand" "million" etc will no longer recognize as numbers.

also disallows leading "oh" in number words, in particular preventing bare "oh" from recognizing as 0. I could factor this into a separate PR if desired, but it seems to be a common pain point.

In my testing this seems to perform fine during DFA compilation and command parsing. Would appreciate further testing.

### DIFF
--- a/code/numbers.py
+++ b/code/numbers.py
@@ -169,7 +169,7 @@ def digits(m) -> int:
     """Parses a phrase representing a digit sequence, returning it as an integer."""
     return int(m.digit_string)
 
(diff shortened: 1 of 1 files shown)


splondike:
This seems to work fine for me. I haven't run into the issue it aims to address either though.

knausj85 (approved):

PR #668, opened 2021-11-21
Clip history added

bra1nDump:
description in the issue


rntz:
I'm going to wait on merging this for two reasons:

1. It looks like there is still discussion in the corresponding issue and it is not clear whether you intend for this PR to be emerged yet.
2. It would be nice to get feedback from aegis on whether we are likely to have a clipboard-update hook. If we did have that we could avoid changing every place where edit.copy and company are overridden.

apps/platforms/linux/gnome-terminal/gnome-terminal.py
```diff
@@ -42,7 +42,9 @@ class EditActions:
     def page_down(): actions.key('shift-pagedown')
     def page_up(): actions.key('shift-pageup')
     def paste(): actions.key('ctrl-shift-v')
-    def copy(): actions.key('ctrl-shift-c')
+    def copy(): 
+        actions.key('ctrl-shift-c')
```
xavier630, line 46: A `copy` action which takes the string representing keys to be pressed and calls both `keys()` and `clip_updated()` would likely make sense so that we don't need to hardcode `clip_updated()` in so many places.

xavier630: Then each app's `copy()` implementation can call `generic_copy()` with just the keys

code/clip.py
```diff
@@ -0,0 +1,34 @@
...
+def gui_clip_history(gui: imgui.GUI):
+    gui.text('Copy / cut history')
+    gui.line()
+    for idx, history_entry in enumerate(clip_history):
+        gui.text("{}. {}".format(idx, history_entry[:30]))
+
+@mod.action_class
+class ClipActions:
+    def clip_updated():
+        """Should be ran every time clip contents are updated"""
```
xavier630, line 17: Minor: "should be run.."

xavier630:
This is pretty exciting. Out of curiosity, does it survive a computer/Talon restart?

bra1nDump:
@xavier630 Definitely will not survive a restart :D

bra1nDump:
I have updated the pull request to continuously poll clip contents (default at 100ms). This way this feature only touches two files, so I think we can sacrifice a little performance. Polling interval is also configurable.

Testing this locally on Mac.

https://user-images.githubusercontent.com/12608159/146155828-ae6fa729-edb0-497b-9a2c-a5c6f97cd629.mov

Windows experience should definitely be way better with the native clip history feature. I don't have a windows computer so unable to work on that / test on Windows for now.

apps/platforms/linux/gnome-terminal/gnome-terminal.py
bra1nDump: The latest version should address this concern indirectly, resolving

rntz:
I'm hesitant to introduce polling into knausj. @lunixbochs do you know whether it's possible to watch the clipboard for changes cross-platform, either using Talon's APIs or if you'd be able to extend Talon's APIs to handle it?

Options for this PR:

1. Use polling. This would be the first global system-wide poll in knausj. There is some disagreement on how bad this would be. We could always have it disabled-by-default, of course, which would be my vote if we take this route.

2. Hook every place in knausj where we call edit.copy to adjust the clip history. Disadvantage: it doesn't handle copies outside of talon; involves a large number of changes.

3. Use Talon APIs to watch the clipboard for changes. The ideal approach if possible.

4. Drop the PR, which would be a shame, since clipboard management is very useful.

nriley:
On Mac http://nspasteboard.org has a list of "temporary" pasteboard identifiers (used by password managers etc.) that ideally would be ignored.

rntz:
I'd be happy to accept this PR if the polling was disabled-by-default. Then we could always revisit the decision about how to best implement it later.

bra1nDump:
Next iteration for this PR. Continuing with option 1, Polling, disabled by default.

- [ ] Disable feature by default. Increase polling time to 1 second.
- [ ] If disabled and 'clip history' is invoked show a message how to enable it.
- [ ] Integrate with native clip history on windows. Polling is not needed on there
- [ ] Avoid picking up sensitive data like passwords. See link in one of the comments.

AndreasArvidsson:
I have done some testing on windows using firefox to copy images
1. clip.serial() seems to always return -1 for me
2. Coping images in firefox sets text and image. ie if you only read the image if there is no text you won't get any images with firefox. I would flip that around so you check for images first.

bra1nDump:
@AndreasArvidsson Tganks for testing! Your implementation seems much more featured, I like it!

Are you also polling the clip contents? How did you implement that?

Are you okay with me taking some of your code into this pull request?

Ps. Sorry I went inactive for a long time on this. I should have some time this week for it :D

AndreasArvidsson:
@bra1nDump 
I'm not polling and instead manually call `user.clipboard_manager_update()` in a few choice places. I have two reasons for this.
1. I have quite a few places where I actually need/want to update the clip or manager. 
2. I explicitly don't want the clipboard manager to update always. `user.selected_text()` and `user.paste()` both update the clipboard and I don't want them to contaminate the clipboard manager.

Feel free to take as much code as you need; my entire repository is under the mit license for this very purpose.

No problem I know how time management can be.

lunixbochs:
btw I'm playing in the next beta with giving you the whole clipboard as a single object rather than only separate text/image

rntz:
Any progress on this in the last month?

bra1nDump:
I plan to work on it during the next weekend. Should be ready by the next grooming session. I will incorporate some of @AndreasArvidsson ideas. If anyone wants to take this over I am also open to that. Just ping me before you do :D

bra1nDump:
I am terrible with git.

Anyways this is how the current version looks like if you don't have it enabled (default):

Screen Shot 2022-04-14 at 11 14 53 PM

If you do have it enabled:

https://user-images.githubusercontent.com/12608159/163529349-6d541740-2709-4858-ad80-becf7dc0ceaa.mov

rntz:
could you reopen this? I realize we haven't gotten around to it in a long while, but it is a useful potential addition. I think, unfortunately, it just never got to the top of our priority list. This is in part because we always look at the most recent PRs first, partly because I at least only do maintenance sessions once every 2 weeks, and partly because this was marked `waiting` even though you seem to have addressed the reasons it was marked that way. I can't guarantee anything but if you leave it open we may eventually find time to get around to it.

PR #671, opened 2021-11-24
adding missing camel_left/right to the line_commands.talon

atacan:
camel_left/right functions were only in the line_commands.py python file. adding it also to the talon file


rntz (approved):

rntz:
lgtm. at some point we should find a better place for these actions/commands than line_commands, though.

PR #675, opened 2021-11-26
Fix VSCode draft window on Mac

pokey:
Switches to using tags for everything and then setting them by listening for app focus switch rather than trying to use `matches`.  Should be more robust, and also means it should reflect settings changes in real time

Also fixed timing issue on submit

- [x] Test Windows
- [x] Test Linux

### DIFF
--- a/draft_editor/draft_editor.talon
+++ b/draft_editor/draft_editor.talon
@@ -1,5 +1,5 @@
 user.draft_editor_running: True
-not app: draft_editor
+not tag: user.draft_editor_app_focused
 -
 
 draft this:
--- a/draft_editor/draft_editor.py
+++ b/draft_editor/draft_editor.py
@@ -1,8 +1,23 @@
-from talon import Module, actions, ui, app
-import time
+from talon import Module, actions, ui, Context
+
 
 mod = Module()
-mod.mode("draft_editor", "Indicates whether the draft editor has been activated")
+mod.tag("draft_editor_active", "Indicates whether the draft editor has been activated")
+mod.tag("draft_editor_app_focused", "Indicates that the draft editor app currently has focus")
+
+ctx = Context()
+tags: set[str] = set()
+
+
+def add_tag(tag: str):
+    tags.add(tag)
+    ctx.tags = list(tags)
+
+
+def remove_tag(tag: str):
+    tags.discard(tag)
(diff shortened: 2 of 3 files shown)


draft_editor/draft_editor.talon
```diff
@@ -1,5 +1,5 @@
 user.draft_editor_running: True
-not app: draft_editor
+not tag: user.draft_editor_app_focused
```
splondike, line 2: Since you're removing the mode, maybe you could remove the scope (user.draft_editor_running) as well? You could do this using:

```
tag: user.draft_editor_running
and not tag: user.draft_editor_app_focused
```

splondike (approved):

splondike:
Have tested this on VSCode on Linux, it works fine.

draft_editor/draft_editor.talon
pokey: What's wrong with the scope?

splondike: The scope works fine, it's more my impression of the semantics. My feeling about scopes is that they're for passing more complex data to matchers. Since this is just a boolean it could be a tag. Kind of like how you replaced the app name (which wasn't referring to just one app and was dynamic) with a tag; it fits better with the Talon semantics. Do you agree with that impression?

Incidentally it looks like there are only two scopes in the repo currently (one is this one).

pokey: I believe @AndreasArvidsson has argued for making both of them scopes. Do either of you feel strongly enough to push for one direction over the other?

AndreasArvidsson: I don't feel strongly one way or the other.

knausj85:
this works fine on windows  as well

draft_editor/draft_editor.talon
pokey: @splondike Ok now that this PR has been tested on all 3 platforms, I'm a bit hesitant to mess with it anymore. Are you ok with me just leaving a quick TODO comment and merging in? Maybe we can refactor some more next time we revisit, eg to combine it with the talon draft editor stuff?

splondike: Yeah, I don't mind if you'd like to merge it in as is.

PR #676, opened 2021-11-27
Command client pre-phrase signal

pokey:
Touches a file known to the VSCode command server before executing any phrase, to allow VSCode extensions to detect start of phrase and maintain consistency over the course of a phrase.

Inaugural use case is allowing cursorless to freeze a snapshot of the hats so that they don't shift over the course of a single command phrase (https://github.com/pokey/cursorless-vscode/pull/318)

### DIFF
--- a/apps/vscode/command_client/command_client.py
+++ b/apps/vscode/command_client/command_client.py
@@ -20,6 +20,10 @@ VSCODE_COMMAND_TIMEOUT_SECONDS = 3.0
 # long to sleep the first time
 MINIMUM_SLEEP_TIME_SECONDS = 0.0005
 
+# Indicates whether a pre-phrase signal was emitted during the course of the
+# current phrase 
+did_emit_pre_phrase_signal = False
+
 mod = Module()
 
 global_ctx = Context()
@@ -382,6 +386,9 @@ class LinuxUserActions:
 @global_ctx.action_class("user")
 class GlobalUserActions:
     def emit_pre_phrase_signal():
+        # NB: We explicitly define a noop version of this action in the global
+        # scope so that it doesn't do anything before phrases if you're not in
+        # vscode.
         pass
 
 
@@ -416,10 +423,7 @@ def get_signal_path(name: str) -> Path:
     return signal_dir / name
 
 
-did_emit_pre_phrase_signal = False
-
-
-def pre_phrase(_: any):
+def pre_phrase(_: Any):
     try:
         global did_emit_pre_phrase_signal
 
@@ -430,7 +434,7 @@ def pre_phrase(_: any):
         pass
 
 
-def post_phrase(_: any):
+def post_phrase(_: Any):
     global did_emit_pre_phrase_signal
     did_emit_pre_phrase_signal = False
 
--- a/apps/vscode/command_client/command_client.py
+++ b/apps/vscode/command_client/command_client.py
@@ -352,6 +352,21 @@ class Actions:
             return_command_output=True,
         )
 
+    def communication_dir_named_subdir(name: str) -> Path:
+        """
+        Get a named communication dir subdirectory.
+
+        Args:
+            name (str): The name of the subdir
+
+        Returns:
+            Path: The communication subdir
+        """
+        path = get_communication_dir_path() / "namedSubdirs" / name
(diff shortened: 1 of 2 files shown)


AndreasArvidsson:
Looks workable at first sight but I have a couple of questions
1. Why use a new file for this an not just send this information as a command?
2. What is the use case for `did_emit_pre_phrase_signal()`?

apps/vscode/command_client/command_client.py
```diff
@@ -368,3 +377,63 @@ def trigger_command_server_command_execution():
...
+    Returns:
+        Path: The signal path
+    """
+    communication_dir_path = get_communication_dir_path()
+
+    if not communication_dir_path.exists():
+        raise MissingCommunicationDir()
+
+    signal_dir = communication_dir_path / "signals"
+    signal_dir.mkdir(parents=True, exist_ok=True)
```
AndreasArvidsson, line 414: What is the point of `parents=True` when you a few lines above throw an exception if the parent/communications directory doesn't exist?

apps/vscode/command_client/command_client.py
```diff
@@ -368,3 +377,63 @@ def trigger_command_server_command_execution():
...
+
+    return signal_dir / name
+
+
+did_emit_pre_phrase_signal = False
```
AndreasArvidsson, line 419: Personally I'm not too found of global/module level variables in the middle of the file. Personally I would move this to the top.

AndreasArvidsson:
Overall I think it looks good

apps/vscode/command_client/command_client.py
pokey: Ha yeah thought about that. Just habit and then left it in case we decide to nest deeper down the road. I have no strong leaning tho; happy to remove it if you think that's better

apps/vscode/command_client/command_client.py
pokey: Was trying to keep it close to where it's used but sure fair enough. Happy to have global variables at the top be our convention

apps/vscode/command_client/command_client.py
AndreasArvidsson: I can see the case for future proofing in it wasn't really hurt. Fine either way with me.

apps/vscode/command_client/command_client.py
AndreasArvidsson: In this case I absolutely think so since it's used by the action did_emit_pre_phrase_signal() above it.

apps/vscode/command_client/command_client.py
```diff
@@ -368,3 +377,63 @@ def trigger_command_server_command_execution():
...
+
+
+def pre_phrase(_: any):
```
rntz, line 422: This should be Any (the type), not any (the function).

apps/vscode/command_client/command_client.py
```diff
@@ -368,3 +377,63 @@ def trigger_command_server_command_execution():
...
+        global did_emit_pre_phrase_signal
+
+        actions.user.emit_pre_phrase_signal()
+
+        did_emit_pre_phrase_signal = True
+    except MissingCommunicationDir:
+        pass
+
+
+def post_phrase(_: any):
```
rntz, line 433: ditto

apps/vscode/command_client/command_client.py
```diff
@@ -357,6 +358,14 @@ def trigger_command_server_command_execution():
         was written to the file.  For internal use only"""
         actions.key("ctrl-shift-f17")
 
+    def emit_pre_phrase_signal():
+        """Touches a file to indicate that a phrase is about to begin execution"""
+        pass
```
rntz, line 363: I suspect that `return None` would work here to make Talon think this method is implemented. Probably want a comment explaining that we're deliberately creating a no-op action, not an unimplemented action.

rntz (changes requested):

apps/vscode/command_client/command_client.py
pokey: I've added a comment to the global scope implementation.  I think I'd lean towards keeping that impl; otherwise it feels like we're trying to trick Talon's static analysis

PR #679, opened 2021-12-03
Add Java keywords throw, extends and instanceof. Make types consistent with everything else by adding space after it. Add annotate with phrase

michaelpog:


lang/java/java.py
```diff
@@ -92,6 +92,16 @@
 mod.list("java_modifiers", desc="Java Modifiers")
 ctx.lists["self.java_modifiers"] = java_modifiers
 
+java_keywords = {
+    "extends": "extents",
```
rntz, line 96: This should be `"extends": "extends"`, not "extents"

lang/java/java.talon
```diff
@@ -44,6 +48,8 @@ op array:
 op new:
     insert("new ")
 
+^annotate with$: insert("@")
```
rntz, line 51: This doesn't need to be anchored, it probably would work better unanchored.

lang/java/java.talon
```diff
@@ -16,6 +16,10 @@ settings():
     user.code_public_variable_formatter = "PRIVATE_CAMEL_CASE"
 
 # Types Commands
+type {user.code_type}:
```
rntz, line 19: This is already defined in `lang/tags/generic.talon`, why redefine it here? I know the pull request says you want to add a space for consistency, but what is it you're being consistent with?

lang/java/java.talon
```diff
@@ -30,8 +34,8 @@ type {user.code_type} array:
     insert(user.code_type)
     user.code_operator_subscript()
 
-[state] {user.java_access_modifiers}: 
-    insert(user.java_access_modifiers)
+[state] {user.java_access_modifiers}|{user.java_keywords}:
```
rntz, line 37: I  think you need to put parentheses here, `[state] ({user.java_access_modifiers} | {user.java_keywords})`. Otherwise "state throw" will not work.

lang/java/java.talon
```diff
@@ -30,8 +34,8 @@ type {user.code_type} array:
     insert(user.code_type)
     user.code_operator_subscript()
 
-[state] {user.java_access_modifiers}: 
-    insert(user.java_access_modifiers)
+[state] {user.java_access_modifiers}|{user.java_keywords}: 
+    insert(user.java_access_modifiers or user.java_access_modifiers)
```
rntz, line 38: You want `user.java_access_modifiers or java_keywords`

lang/java/java.py
```diff
@@ -92,6 +92,16 @@
...
+    "instanceof": "instanceof",
```
rntz, line 97: You might want to make that `"instance of": "instanceof"`?

lang/java/java.talon
rntz: Also, @knausj85 - any opinion on whether it's okay to have the java keywords bound without a preceding "state"? It looks like we don't do this for other languages, but we have done it in java.

rntz (changes requested):

lang/java/java.talon
michaelpog: What is the difference, just curious? I don't really understand how it works

lang/java/java.talon
michaelpog: Consistent with the rest of the types and Java

lang/java/java.talon
knausj85: Oops. I prefer to require state for consistency/avoiding issues with short words, but looks like something I missed previously. I’d vote to leave as-is for this PR and maybe open an issue.

lang/java/java.talon
splondike: The '^' prefix means the command has to be at the start of an utterance. I.e. '^my command: "hello "' means I can say 'my command word world' and get 'hello world' written out, but I can't say 'word world my command' (I'll get 'world' out, and then maybe a misrecognition or nothing).

The '$' suffix means the rest of the utterance after the command is discarded. So if I had 'my command$: "hello "' and said "my command word world ten dot" I'd just get 'hello ' out.

Generally you try not to use anchors since it prevents people from fluidly making utterances as long as they want. For example when editing code I'm often able to queue up a bunch of edits to one line then move to a different line and start editing there in one utterance. The main limiting factor is the potential misrecognitions from Talon requiring backtracking.

As implemented this command is the same as saying 'at sign'. Do you think something like 'annotate override' -> '@Overrides', 'annotate non null' -> '@NonNull' etc. might be a more useful pattern?

lang/java/java.talon
splondike: It looks like we're using the code_type list for Python and Typescript and hence the generic 'type' command which doesn't insert the space. I'd suggest we be consistent with that and don't override this here so that people get the same behaviour across languages.

What do you think? You could easily override type to add the space in your personal configuration and we can keep things more consistent in knausj.

michaelpog: sure

lang/java/java.talon
rntz: This still needs parentheses.

lang/java/java.talon
rntz: Should either unanchor or remove this.

rntz (changes requested):
Sorry the the delay on getting to this.

PR #680, opened 2021-12-03
Add Unix shell commands to terminal on linux and mac

michaelpog:


apps/generic_terminal/unix_shell.talon
```diff
@@ -0,0 +1,26 @@
...
+^list directory$: "ls "
+^list current directory$: user.terminal_list_directories()
+^grep$: "grep "
+^edit$: "vi " # Can be replaced with your favorite editor
+^less$: "less "
+^remove$: "rm "
+^teemux$: "tmux "
+^netstat$: "netstat "
+^which$: "which "
+^touch$: "touch "
```
auscompgeek, line 23: This will conflict with the mouse `touch` command.

apps/generic_terminal/unix_shell.talon
```diff
@@ -0,0 +1,26 @@
...
+# Generic Shell Commands. (General convension is to use the command name in "man" when possible)
+
+^change (directory|dir)$: "cd "
+^home (directory|dir)$: "~/"
+^copy$: "cp "
+^make dir$: "mkdir "
+^cancel$: key(ctrl-c)
+^(move|rename) command: "mv "
+^list directory$: "ls "
+^list current directory$: user.terminal_list_directories()
```
auscompgeek, line 15: We already have the `lisa` command in `generic_terminal.talon`.

apps/generic_terminal/unix_shell.talon
```diff
@@ -0,0 +1,26 @@
+tag: terminal
+tag: user.generic_unix_shell
+os: mac
+os: linux
+-
+# Generic Shell Commands. (General convension is to use the command name in "man" when possible)
+
+^change (directory|dir)$: "cd "
```
auscompgeek, line 8: We already have the `katie` command in `generic_terminal.talon`.

auscompgeek (commented):
I think a lot of the new commands being added here would be better served with an action which can be implemented for both Unix and Windows.

apps/generic_terminal/unix_shell.talon
michaelpog: Yes I saw it. 
But Katie and Lisa don't mean anything. 
If I added "netstat" or "which" what would be the convention to name those commands? "David" or "Michael"?
How would one remember those commands? 
I can sort of see some sound similarity between ls and lisa, but what is the similarity or relationship between katie and "change directory." 
I am in favor of replacing lisa and katie in favor of more standard and scalable command system. 
And if someone wants to remap them locally to names that mean something to them specifically they should feel free to do so.

apps/generic_terminal/unix_shell.talon
michaelpog: see my comment above

apps/generic_terminal/unix_shell.talon
michaelpog: good point. I am suggesting to disambiguate this by adding the optional "command" to it:
1) ^touch [command]$: ...
or prefix with "shell"
2) ^[shell] touch$:
What do you think?
We can establish those rules here so they can be followed going on forward

michaelpog:
I agree about the separation of unix and windows.
Though I don't have windows to test, and I honestly don't know how to map many commands like "grep", "touch" or "echo" to windows.
As a software engineer working on unix, I run those commands 100 times a day and having them the way I have them in my had is super helpful and time saving

splondike:
I think I would suggest deferring the addition of this kind of thing to people's personal Talon configuration.

My thinking is that a user of this file would likely be at least somewhat conversant with programming and so would be comfortable following [the wiki](https://talon.wiki/basic_customization/) to just make their own version of this file with whatever commands and CLI programs they used regularly. Hopefully it would take less than 10 minutes and would then fit their needs exactly. For example I have a file like this for my idiosyncratic usage which I've built up over time: https://github.com/splondike/talon_mystuff/blob/main/apps/cli.talon .

What do you think, how long did it take to write up the commands in this file for you?

FYI you might also be able to make the commands more useful by not anchoring them on the end (i.e. no trailing $). Then you could say something like 'change dir word test tab word helpers tab' in one utterance to get 'cd test/helpers/' rather than having to pause after saying 'change dir'.

apps/generic_terminal/unix_shell.talon
fidgetingbits: fwiw Katie came out of "see dee" not being reliable, so became "Kay dee". convenient attr time they were both names *shrug*

but I agree, it's not particularly sensible to most people. something generic is better

I've since changed to more explicit commands in my repo, with prefixes that roughly match what I'm working on, like file, folder, disk, net:

https://github.com/fidgetingbits/knausj_talon/blob/master/apps/platforms/linux/commandline.talon

I find it works well. things like:

file list
file touch 
file edit
folder create

might be worth considering

michaelpog:
Ok in that case am dropping this pull request

PR #686, opened 2021-12-04
Implement paste_match_style() on Windows.

wolfmanstout:


rntz:
It seems like this shortcut only performs paste-matching in a few applications. So probably we should bind this only in a context matching those applications. Will accept PR if it's updated to handle that.

rntz (changes requested):
It seems like this shortcut only performs paste-matching in a few applications. So probably we should bind this only in a context matching those applications. Will accept PR if it's updated to handle that.

pokey:
Not generic enough

PR #687, opened 2021-12-04
Added private mode that disables recording.

wolfmanstout:


rntz:
In discussion during the knausj meeting we decided not to merge this until we get some more feedback from Ryan.

Issues:
- it's not visible to the user whether they're in recording-disabled mode. It doesn't affect the menu, for example, but it does override it.
- it's not clear whether a mode or a tag or something else is the right way to implement this.

wolfmanstout:
Blocked on #454 per my conversation with Ryan.

rntz:
I converted this to a draft to make clearer that we don't want to merge it yet.

splondike:
It sounds like we'd need 1st party support to make this idea have good UX. It also seems like the associated Talon issue is likely to take a while (it's been about 1 year so far). What do you think about closing this PR out pending that 1st party support?

wolfmanstout:
Makes sense. We still have the bug to track the underlying FR.

PR #691, opened 2021-12-08
Add duplicate tab browser interface..

xavier630:
.. and Chrome implementation.

Only able to test on Mac, so not 100% certain if the same shortcut works on Windows/Linux.


code/tabs.py
```diff
@@ -11,5 +11,10 @@ def tab_final():
         """Jumps to the final tab"""
 
     def tab_close_wrapper():
-        """Closes the current tab"""
+        """
+        Closes the current tab.
```
xavier630, line 15: Unrelated comment here that I forgot to add to a previous PR. Happy to put in a fresh PR if you prefer.

code/browser.py
```diff
@@ -25,3 +26,12 @@ def address():
         # Prioritize last one if multiple are valid, return empty string if none is valid.
         tokens = (url[1: -1] if not is_url(url) else url for url in reversed(actions.win.title().split(" ")))
         return next((url for url in tokens if is_url(url)), "")
+
+
+@mod.action_class
```
xavier630, line 31: Added to the `browser` tag rather than the `tabs` tag since even though it's an action related to tabs, it's only relevant for browsers. Happy to discuss merits of each option.

xavier630: Will move to the `tabs` interface now that I've thought of some use cases beyond browsers

code/tabs.py
```diff
@@ -11,5 +11,10 @@ def tab_final():
         """Jumps to the final tab"""
 
     def tab_close_wrapper():
-        """Closes the current tab"""
+        """Closes the current tab.
+        Exists so that apps can implement their own delay before running tab_close() to handle repetitions better.
+        """
         actions.app.tab_close()
+
+    def duplicate_tab():
```
splondike, line 19: Should be 'tab_duplicate' to match naming scheme.

apps/generic_browser.talon
```diff
@@ -1,7 +1,8 @@
 tag: browser
 -
-(address bar | go address | go url): browser.focus_address()
-(address copy | url copy | copy address | copy url):
+address bar | go address | go url: browser.focus_address()
+duplicate tab | tab duplicate: user.duplicate_tab()
```
splondike, line 4: I'd move this to misc/tabs.talon instead since the action doesn't apply to just browsers. I think that duplication is probably about as common as the 'tab reopen' functionality for example which is in the misc/tabs.talon file. VSCode for example also has the ability to duplicate tabs.

splondike (changes requested):
Just a bit of renaming and moving, but otherwise looks good to me.

apps/generic_browser.talon
xavier630: Yeah good call, legacy from how I built it for the browser tag to begin with, my bad.

xavier630:
@splondike, I've updated it now, thanks for the review

splondike:
On Linux the browser.focus_address action isn't implemented, but the alt enter shortcut works in both Firefox and Chrome (assuming you don't have it bound to something else like me). focus_address can be implemented as ctrl+l for FF and Chrome on Linux, but that can be a different PR.

splondike:
FF actually has the actions implemented, I've made #700 to implement them for Linux Chrom(ium).

PR #693, opened 2021-12-10
Clean up unused formatters

xavier630:
These have been sitting around, commented out for a while, let's remove them to tidy up the code base a bit.


xavier630:
I've left the implementations intact though I'm open to removing them too.

splondike:
It seems like the implementations were active a year or more ago and then the commands were disabled.

In terms of removing the implementations, I don't really see the use of the PIRST_* ones so they could be removed in a second PR IMO. CAPITALIZE_FIRST_WORD possibly makes sense to keep though.

PR #694, opened 2021-12-10
Add a question formatter to Talon

xavier630:
Adds a question formatter `ask` to Talon.

Example:
`"Ask where do I go"` -> `"where do I go?"`

In the future, I'd like to make an `ask` formatter which auto-capitalises the first word. For now I'm keeping it quite simple and assumption-free.

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -14,6 +14,8 @@ words_to_keep_lowercase = "a,an,the,at,by,for,in,is,of,on,to,up,and,as,but,or,no
 
 DEFAULT_SEPARATOR = ' '
 SMASH_SEPARATOR = ''
+SEP = False
+NOSEP = True
 
 # The last phrase spoken, without & with formatting. Used for reformatting.
 last_phrase = ""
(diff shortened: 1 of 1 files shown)


splondike:
For me I probably wouldn't use this formatter, I'd just say "sentence where do I go <pause> question". <pause> would be half a second or so. By the same token I don't use the existing string or dubstring formatters either (I use "quote say blah blah <pause> quote"). Do you think it would add much benefit over the current approach (i.e. the one I use)?

Will leave this open thought to let others add their views.

In general I feel like there are benefits to limiting the addition of global commands to the repo if they're not likely to be used too widely. Specifically it reduces the length of automatic help windows, the sise of the more heavily viewed parts of the codebase, the maintenance burden (we have no way of knowing what commands people use once they're released, so we basically have to keep all of them), and the chance of misrecognitions. Those are the reasons I ask for justification on some of these PRs, I appreciate the effort to improve things though :).

xavier630:
Happy to let others jump in. If no one finds it useful, I can always keep this on my fork only 👍

rntz:
You should also be able to say "sentence what time is it question mark"

pokey:
After some discussion I think we decided that we should let users implement the sort of thing on their own branch

Issue #688, opened 2021-12-07
mouse pointer unexpectedly missing


lunixbochs:
Every few weeks a user manages to loses their mouse pointer and has no idea how they did it, e.g. https://talonvoice.slack.com/archives/C7ENXA7C4/p1638899043213100

This is likely due to a `curse no` recognition. (if you hit this problem, say `curse yes` to fix it)

IMO knausj_talon should put those commands behind a non-default tag.

somanythings:
thanks, that was it for me!

phillco:
Definitely agree with this, I never triggered this accidentally but it was kind of scary that I could :)

splondike:
I have run into this problem as well, definitely confused the times it has happened. This is apparently a command from the first commit in this repo (Feb 2020), so I imagine it was created by knausj before it was apparent this would be a repository used by a lot of people.

The tag option sounds good. My suggestion is:

1. Define a tag `user.mouse_cursor_commands_enable` to `code/mouse.py`. Add a commented out activation of this tag to `settings.talon`.
2. Move the existing 'curse yes|no' commands to `misc/mouse_cursor.talon`. Gate this file using the `user.mouse_cursor_commands_enable` tag.
3. Add a new 'curse no' command to `misc/mouse.talon` which just displays a notification bubble saying to enable the tag in settings.talon. This tells existing users how to fix the breakage. Put a comment on this command to say it can be removed in 6 months.

splondike:
Fix merged.

PR #697, opened 2021-12-13
Put curse yes/no commands behind tag (disabled by default)

splondike:
Addresses #688


erhoppe:
Heads up, with current .talon file parsing (this changed recently, correct?), uncommenting this tag throws an unexpected token error, which breaks the settings file parsing.

PR #700, opened 2021-12-14
Add browser action implementations for Chrome on Linux

splondike:
This lets Chrome/Chromium on Linux benefit from the generic browser commands. This PR also improves the reliability of the recently merged tab duplicate command (cc @xavier630).


knausj85:
Thanks!

PR #707, opened 2021-12-18
dictation.py: do peek_right before insertion to not break undo chain

rntz:
Currently context-sensitive dictation peeks left for context before it inserts and peeks right after it inserts to determine whether to add a trailing space. This has the unfortunate effect in some editors (eg. emacs) of separating the insertion from the trailing space in the undo-chain, so you have to undo twice: once to eliminate the trailing space, once to undo the insertion. This PR fixes this by doing all peeking before insertion. However, it has the unfortunate side-effect of making context-sensitive dictation feel slower, because we have to do more work before we can start inserting text. Suggestions welcome.

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -282,27 +282,29 @@ class Actions:
 
     def dictation_insert(text: str, auto_cap: bool=True) -> str:
         """Inserts dictated text, formatted appropriately."""
-        context_sensitive = setting_context_sensitive_dictation.get()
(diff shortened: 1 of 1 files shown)


knausj85:
FWIW, I still see this issue in e.g. vscode, so it'll probably be editor-specific. That said, I'm fine with merging this.

knausj85 (approved):

PR #711, opened 2021-12-22
Knausj windows terminal and powershell integration

DonnerWolfBach:
Part of #562 : 
migrates both the power shell and the windows terminal integration to the new "architecture"


apps/platforms/win/windows_terminal/windows_terminal.py
```diff
@@ -84,5 +84,69 @@ def file_manager_open_volume(volume: str):
...
+
+    def split_clear():
+        """Clears the current split"""
+        # also closes tab, because shortcut is the same
+        # and closing a split does mean something differnent that in a code editor like vs code
+        actions.key("ctrl-shift-w")
+
+    def split_clear_all():
+        """Clears all splits"""
+        actions.app.notify("TODO: decide whether to implement it at all since it either doesn't makes sense or closes the window/whole tab")
```
rntz, line 135: This probably doesn't need to be here, just don't implement the action and leave a TODO in the file.

rntz (changes requested):
This looks good, except that "split right" and "split down" are already defined in windows_terminal.talon, so you should remove those if you want your overrides to take effect. Also, since currently "split right" does the same as "split vertically" in your implementation, could you explain what the behavior difference is, and why you don't want to implement "split right" as key(alt-shift-=)? And the same for split down/split horizontally.

apps/platforms/win/windows_terminal/windows_terminal.py
DonnerWolfBach: ok, should I only remove the notification for the "TODO" part, or also for the other commands like "split_last" (which seem to me not possible at all)?

DonnerWolfBach: If it's only about the comment with TODO and the other ones can remain, it should be done with [this commit](https://github.com/knausj85/knausj_talon/pull/711/commits/1a4d02761a11ff55a5d17a032e42107b6c4993fd)

DonnerWolfBach:
@rntz overlooked that "split right" and co was already implemented in the .talon file.
If you want, I can change "split right" (in the .py file) to do the same as "split vertically". I think my thought at that time was, that since "split left" wasn't possible, I won't do "split right" either. But I am quite open about that.

choskeli:
I believe these changes will solve #770 with the tag "tag(): user.generic_windows_shell" being added to powershell and windows terminal talon files.

Should we also add commands to open new tabs for a given terminal type? Probably ideally this would be a csv file config or read from the windows terminal config itself, but assuming using the default profile numbers for now...e.g.

```
(open|new) powershell:
    key(ctrl-shift-1)

(open|new) (command prompt|CMD):
    key(ctrl-shift-2)

(open|new) (ubuntu|linux):
    key(ctrl-shift-3)
```

DonnerWolfBach:
@choskeli I think in my own fork I solved it with "profile <number>" since I have many more profiles. But you can open a new pull request to add your solution (or ask on of the knausj collaborators how to handle that, I am not one of the maintainers).

rntz (approved):

PR #712, opened 2021-12-22
dictation.py refactoring

rntz:
Goals of this PR:

1. Make it easier to change what "modifiers" are available in prose/dictation. Currently the meanings of "caps", "no caps", "no space" are hardcoded into `apply_formatting`, which must be kept in sync with both the `prose_modifiers` list and `dictation_mode.talon` commands.

2. Reduce duplication between prose_modifiers and dictation_mode.talon.

3. Allow curly quotes to be used without being auto-straightened, while still getting spacing right for open/close quote commands.

Progress:

- Made a global dictionary, `prose_modifiers`, that maps modifier names to DictationFormat methods; used to generate `{user.prose_modifiers}` and in `apply_formatting`. So instead of editing three places (dictation_mode.talon, the prose_modifiers list, and apply_formatters) you only need to edit two (dictation_mode.talon and the prose_modifiers dictionary).

- Replaced uses of `{user.prose_modifiers}` with a capture `<user.prose_modifier>` which produces a Callable. This avoids confusion between prose modifiers and user vocabulary that produces the same words (fixing #718).

It should be possible to re-use the prose_modifiers list to define commands in dictation_mode.talon. This will make the help for dictation_mode more confusing, unfortunately.

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -1,6 +1,6 @@
 # Descended from https://github.com/dwiel/talon_community/blob/master/misc/dictation.py
 from talon import Module, Context, ui, actions, clip, app, grammar
-from typing import Optional, Tuple, Literal
+from typing import Optional, Tuple, Literal, Callable
 import re
 
 mod = Module()
@@ -39,6 +39,10 @@ ctx.lists["user.prose_snippets"] = {
     "frowny": ":-(",
 }
 
+@mod.capture(rule="{user.prose_modifiers}")
+def prose_modifier(m) -> Callable:
+    return getattr(DictationFormat, prose_modifiers[m.prose_modifiers])
+
 @mod.capture(rule="({user.vocabulary} | <word>)")
 def word(m) -> str:
     """A single word, including user-defined vocabulary."""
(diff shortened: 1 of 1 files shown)


rntz:
@wolfmanstout I'd be interested in your feedback on this PR as it progresses.

wolfmanstout:
Nice cleanup! You might consider having the map point to functions as in '"no cap": DictationFormat.no_cap'. Then you can pass in the formatter as arg when executing it. I think you'd need to move the DictationFormat definition up, but maybe that's a good thing for readability since you're referring to its methods.

rntz:
This should fix #718 now.

rntz:
Hm, I'm not sure about pointing directly at the DictationFormatter methods. It would be simpler, but I would need to move DictationFormatter up, and I tend to prefer if user-customizable things are located near the top of the file.

rntz:
I'm marking this ready for a view because it fixes #718 and accomplishes goal (1). I'll defer goals (2-3) to a later PR if this is merged.

splondike (approved):

code/dictation.py
```diff
@@ -224,6 +225,19 @@ def format(self, text, auto_cap=True):
...
         return text
 
+    # These are used as callbacks by prose modifiers / dictation_mode commands.
+    def cap(self): self.force_capitalization = "cap"
+    def no_cap(self): self.force_capitalization = "no cap"
+    def no_space(self):
+        # This is typically used after repositioning the cursor, so it is helpful to
+        # reset capitalization as well.
+        #
+        # FIXME: this sets state to "sentence start", capitalizing the next
```
wolfmanstout, line 235: Totally open to changing/removing this. In my fork I default state to None, but I think the reasoning in my comment above is orthogonal. For example when positioning the cursor to a new text box, which is a common use of this for me, presumably folks would want to reset capitalization to whatever is their default.

wolfmanstout (approved):

code/dictation.py
rntz: hm, I guess that's true. I'm mainly concerned not with the default dictation state but with the "no space" command, for which it feels weird that "foo no space bar" should produce "fooBar". But I guess you might want "no space hello" to produce "Hello" if you've just entered a new text box.

Issue #715, opened 2021-12-23
Deduplicate JavaScript and TypeScript actions


auscompgeek:
There's a lot of code duplication between JavaScript and TypeScript. This is error prone, and inconsistencies are easy to miss, for example the bugfix in #714 after a refactor.

Given that TypeScript can, for the most part, be considered a superset of JavaScript syntax, I'd like to propose combining the code in some way.

We could add a tag, move the existing JS actions to be active under that tag, and activate that tag in both JS and TS languages.

Alternatively, we could go all in and replace the modes with tags (#603), but this might be a good stepping stone?

pokey:
Yeah I'd lean towards @AndreasArvidsson's [solution](https://github.com/knausj85/knausj_talon/issues/603#issuecomment-980106400) for this one

PR #716, opened 2021-12-26
Deduplicate JavaScript/TypeScript actions with tag

auscompgeek:
This moves the JavaScript actions to being activated by a tag, which is activated by the respective contexts for JavaScript and TypeScript.

Subsequently, this deletes the duplicated actions for TypeScript, and moves the `code_state_do` action implementation to javascript.py, where it was missing.

Open questions:

- [ ] Should we prefix the language tag?
- [x] Should we delete the settings from typescript.talon altogether? People might be working on projects with differing naming schemes, but I suspect this is rare in the JS world.

Closes: #715


splondike:
This looks good, the only thing that I would suggest is that we remove the settings from the typescript.talon file (as you suggested). People can override them in their own context if they want to (they don't need to be in the knausj file).

splondike:
Looks good to me! If the tag proves annoying then it would be easy enough to change in the future. I doubt anyone would use it external to knausj so there shouldn't be issues changing it.

PR #720, opened 2021-12-28
Add tag for C-style block comments

auscompgeek:
In #656 the block comment actions were renamed for consistency, however their implementations were not renamed. The only implementations of these were for C and Java.

This adds a common implementation for C-style block comments behind a tag, and adds the tag to C# and JavaScript.

Using a tag makes it easier for users to override `code_comment_block` for their editor (targeting the relevant languages) if needed (for example, if their editor auto-closes block comments).

This assumes that #716 will be merged, and hence does not add the tag to typescript.talon.

### DIFF
--- a/lang/tags/comment_block.py
+++ b/lang/tags/comment_block.py
@@ -1,13 +1,19 @@
-from talon import Context, Module
+from talon import Context, Module, actions
 
 ctx = Context()
 mod = Module()
 
 mod.tag("code_comment_block", desc="Tag for enabling generic block comment commands")
+mod.tag("code_comment_block_c_like", desc="Denotes usage of C-style block comments")
+
+ctx.matches = """
+tag: user.code_comment_block_c_like
+"""
+ctx.tags = ["user.code_comment_block"]
+
 
 @mod.action_class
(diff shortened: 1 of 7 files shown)


lang/tags/comment_block.py
```diff
@@ -1,13 +1,19 @@
-from talon import Context, Module
+from talon import Context, Module, actions
 
 ctx = Context()
 mod = Module()
 
 mod.tag("code_comment_block", desc="Tag for enabling generic block comment commands")
+mod.tag("code_comment_block_c_like", desc="Denotes usage of C-style block comments")
+
+ctx.matches = """
```
rntz, line 9: maybe name this `c_like_ctx`, to be more explicit about what it matches?

rntz (approved):
lgtm, although it would be nice if the `ctx` variable in comment_block.py were named more descriptively since it only matches c-like contexts

splondike (approved):

lang/tags/comment_block.py
auscompgeek: Done.

PR #722, opened 2021-12-29
Remove unused formatter implementations

xavier630:
Removes a couple of old formatter implementations which no commands map to any longer.

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -138,9 +138,6 @@ formatters_dict = {
         if i == 0 or word not in words_to_keep_lowercase
         else word,
     ),
-    "FIRST_THREE": (NOSEP, lambda i, word, _: word[0:3]),
-    "FIRST_FOUR": (NOSEP, lambda i, word, _: word[0:4]),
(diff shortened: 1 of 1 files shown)


rntz (approved):
I think these are here for backward-compatibility reasons, for people who've been using knausj for a while. I'm okay with removing them but @knausj85 might want to keep them? If we want to keep them maybe we should add a comment so that we don't get more PRs about removing them.

xavier630:
Going by @splondike's comment here, they may be okay to remove https://github.com/knausj85/knausj_talon/pull/693#issuecomment-992011936

knausj85:
There's a few people here and there using these, so I'm slightly hesitant:
https://search.talonvoice.com/search/?q=formatters_dict%5B%22FIRST_THREE%22%5D&fold_case=auto&regex=false&context=true

I don't know where to draw the line on these sorts of things

xavier630:
Happy to defer to you guys as maintainers. Given that it was already commented out, it feels a bit like dead code to me.

xavier630:
But agreed that we don't want to break functionality of forks when they pull develop, so happy for you to close

Issue #408, opened 2021-04-02
jetbrains incorrectly determines filename


dwiel:
relevant commit: https://github.com/knausj85/knausj_talon/commit/940b4fefb14a76619c17e88b88393527e82860de#diff-9ad2f77588de78f273099f7fbda1df48ffe0c689c2fd083f483f8b977c244643R160

the problem occurs when:

- the name of a project has a dot in it: `.talon - jetbrains.py`
- the name of a file does not have a dot in it

@zhang-chris was there a reason for changing the logic from "the extension is anything after the last dot" to "the filename is the first thing with a dot in it"?

to whom it may concern: yet another place where some tests would be nice. I have no idea what possible situations this code might be designed to handle.

zhang-chris:
It just didn't occur to me that the project can have a dot in it. Feel free to submit a fix for it and start iterating from the back  instead.

dwiel:
perfect thanks

Michaelt293:
I have an issue when using Intellij on a Windows machine which I resolve using -
```python
    def file_ext():
        return actions.win.filename().split(" [")[0].split(".")[-1].split(" ")[0]
```

rntz:
Can anyone confirm that this is fixed post-#724.

PR #724, opened 2022-01-01
jetbrains: Ignore project module names

auscompgeek:
IntelliJ allows for having projects with multiple modules, in a similar manner to multiple folders in a VSCode workspace.

This ignores the module name in the title, shown in square brackets after the filename.

I believe this fixes the issue described by @Michaelt293 in #408.


rntz:
I don't know much about intellij but I'm going to be optimistic and merge this.

PR #727, opened 2022-01-06
Add back close buttons to language imgui popups

splondike:
These were added in #699 but lost in the language refactorings merge.


lang/tags/functions_gui_open.talon
```diff
@@ -0,0 +1,4 @@
+tag: user.code_functions_gui_showing
+-
+# Toggle prefix to be mentally similar to the 'toggle funk' show command
+toggle funk: user.code_toggle_functions()
```
splondike, line 4: I tried to use toggle funk hide, but it almost always turned into something like "toggle funk", "caret" (Talon splitting the command incorrectly). So I just use the same name as the open command.

PR #729, opened 2022-01-06
Make formatters file more readable

xavier630:
This is a minor change to help make the formatters file more readable.


code/formatters.py
```diff
@@ -12,6 +12,11 @@
     ","
 )
 
+DEFAULT_SEPARATOR = ' '
```
rntz, line 15: I'm not convinced these globals (`DEFAULT_SEPARATOR`, `SMASH_SEPARATOR`) improve readability. They're only used in one place, so this just separates their definition from where they're used.

code/formatters.py
```diff
@@ -12,6 +12,11 @@
...
+SMASH_SEPARATOR = ''
+SEP = True
```
rntz, line 17: While we're cleaning things up, SEP and NOSEP could use a comment explaining what they mean/do.

rntz (approved):

rntz:
If anyone merges this, please squash-merge to keep the commit history uncluttered.

code/formatters.py
xavier630: https://github.com/knausj85/knausj_talon/pull/746

code/formatters.py
xavier630: Have left this as is, but if you feel strongly, we can easily adjust 👍

PR #730, opened 2022-01-07
Add emoji tag

timo95:
This is just a basic selection of common emojis.


code/emoji.py
```diff
@@ -0,0 +1,76 @@
+from talon import Module, Context
```
pokey, line 1: I'd be tempted to make these lists into csvs.  WDYT @knausj85 @rntz ?

knausj85: I'd vote to put this into a CSV as well

timo95: Like that?

splondike (approved):
Looks good to me. Seems to work fine if I enable the emoji tag.

pokey:
Sorry to be late to the party here, but how come we're not using the built-in knausj csv support?  See eg https://github.com/knausj85/knausj_talon/blob/a3e51dcf15c3158f378b0cda53dd1dd98fcdb65c/code/websites_and_search_engines.py

timo95:
I don't think the default list behavior is desirable in this case.

timo95:
I thought it better to provide a full list of useful emoji instead of letting the user fill it. And not having to provide the default lists in python is much cleaner. Though there are advantages to default lists too.

nickeubank:
Naïve question from a person who doesn't play with Talon internals very much: is this currently only enabled for slack and discord? How do I enable this globally? I tried just saying `emoji angry` and text editors and in the browser but I'm not having any luck. :/

timo95:
It should be activated globally if you create a `.talon` file with:
```
-
tag(): user.emoji
```

nickeubank:
Thanks!

PR #731, opened 2022-01-07
vscode: Fix go tab conflict with Mac desktop keys

auscompgeek:
VSCode's default keybindings for "switch to tab N" conflict with the default-ish macOS "switch to desktop N" keyboard shortcuts. This uses the command client to execute the desired commands instead.

I wasn't able to use the `openEditorAtIndex` via the command palette, so I'm being conservative and only doing this for macOS.

### DIFF
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -243,13 +243,13 @@ class UserActions:
     def tab_jump(number: int):
         if number < 10:
             if is_mac:
-                actions.key("ctrl-{}".format(number))
+                actions.user.vscode(f"workbench.action.openEditorAtIndex{number}")
             else:
                 actions.key("alt-{}".format(number))
 
     def tab_final():
         if is_mac:
-            actions.key("ctrl-0")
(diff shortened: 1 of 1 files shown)


pokey:
That's really weird it didn't work via command palette.  I thought you could always just type in the command id.  So the reasoning here then is that it's broken anyway on mac, so at least make it work when they have the command server?  If so, I'd be tempted to use `actions.user.vscode_with_plugin` instead to make it clearer that it requires the plugin

auscompgeek:
> I thought you could always just type in the command id.

Yeah, I would've thought so too. It might be because they don't have names? I was able to get the `lastEditorInGroup` command to work in the palette, and that does have a name.

> So the reasoning here then is that it's broken anyway on mac, so at least make it work when they have the command server? If so, I'd be tempted to use `actions.user.vscode_with_plugin` instead to make it clearer that it requires the plugin

Ah, yup. I missed that the `vscode_with_plugin` action exists. I'll swap that in once I'm at a computer.

pokey (approved):

PR #735, opened 2022-01-12
Scala support

SCdF:
Now that scala is also in cursorless I figured it would be helpful to share my knausj scala bindings


lang/scala/scala.talon
```diff
@@ -0,0 +1,37 @@
...
+state var: "var "
+
+state match: user.code_state_switch()
+
+state case class: "case class "
+state object: "object "
+
+string: key('"')
+char: key("'")
+block string:
```
splondike, line 30: Multiline strings with a special syntax are also in Perl, bash, Python, Ruby etc. If we end up making this more generic to encompass them at some point it'd be good to be able to preserve the command name.

I think this name is fine, but want to draw this to other maintainer's attention.

lang/scala/scala.py
```diff
@@ -0,0 +1,288 @@
...
+                text, settings.get("user.code_public_function_formatter")
+            )
+        )
+
+        actions.user.code_insert_function(result, None)
+
+    def code_private_static_function(text: str):
+        actions.skip()
+    def code_protected_static_function(text: str):
+      actions.skip()
```
splondike, line 286: Some formatting/indentation issues here.

lang/scala/scala.py
```diff
@@ -0,0 +1,288 @@
...
+    def code_false():
+        actions.auto_insert('false')
+
+    def code_type_class():
+        actions.auto_insert("class ")
+
+    def code_import():
+        actions.auto_insert("import ")
+
+    def code_private_function(text: str):
```
splondike, line 214: These are duplicated below.

lang/scala/scala.py
```diff
@@ -0,0 +1,288 @@
...
+    def code_protected_function(text: str):
+        actions.user.code_private_function()
+
+    def code_public_function(text: str):
+        actions.insert("public ")
+
+    def code_state_return():
+        actions.insert("return ")
+
+    def code_comment():
```
splondike, line 226: This isn't defined in knausj.

lang/scala/scala.py
```diff
@@ -0,0 +1,288 @@
...
+        actions.edit.left()
+        actions.edit.left()
+
+    def code_break():
+        actions.auto_insert('break')
+
+    def code_next():
+        actions.auto_insert('continue')
+
+    def code_true():
```
splondike, line 202: Not in knausj, code_insert_true is used.

lang/scala/scala.py
```diff
@@ -0,0 +1,288 @@
...
+    def code_operator_bitwise_right_shift():
+        actions.auto_insert(" >> ")
+
+    def code_operator_bitwise_right_shift_assignment():
+        actions.auto_insert(" >>= ")
+
+    def code_self():
+        actions.auto_insert("this")
+
+    def code_null():
```
splondike, line 153: knausj calls these code_insert_null, code_insert_is_null etc.

lang/scala/scala.py
```diff
@@ -0,0 +1,288 @@
+from talon import Context, Module, actions, settings
+
+ctx = Context()
+mod = Module()
+ctx.matches = r"""
+mode: user.scala
+mode: user.auto_lang
+and code.language: scala
+"""
+ctx.tags = ["user.code_operators", "user.code_generic"]
```
splondike, line 10: Redundant with the .talon file.

lang/scala/scala.py
```diff
@@ -0,0 +1,288 @@
...
+    "Double": "Double",
+    "Short": "Short",
+    "Float": "Float",
+    "Long": "Long",
+    "Boolean": "Boolean",
+    "Char": "Char",
+    "String": "String",
+    "None": "None"
+}
+mod.list("scala_common_types", desc="Scala Common Types")
```
splondike, line 25: Do you need to expose these lists to Talon? It looks like they only get used to build the code_type list.

splondike (commented):
A few things to look at, but it'll be nice to add in another language when you're done :).

lang/scala/scala.py
splondike: You probably want to look at tags/comment_line.py and tags/comment_block.py.

splondike (changes requested):
Adding request changes tag.

SCdF:
Thanks for the feedback! Looking into this: there have been lots of changes since I first wrote this. Specifically all the refactoring and new tags.

I am looking into this, but it will take time to find the time.

SCdF:
OK, I've gone through all the new code tags and implement of them all except the function gui. Let me know how it looks!

lang/scala/scala.talon
```diff
@@ -0,0 +1,42 @@
+mode: user.scala
```
splondike, line 1: I recently found out there's a bit more of an incantation needed here in the matcher to avoid binding commands in sleep mode. The reason is once you add an explicit mode, the implicit command mode filter is removed.

```
mode: command
and mode: user.scala
mode: command
and mode: user.auto_lang
and code.language: scala
```

lang/scala/scala.talon
```diff
@@ -0,0 +1,42 @@
...
+
+state val: "val "
+state var: "var "
+
+state match: user.code_state_switch()
+
+state case class: "case class "
+state object: "object "
+
+char: key("'")
```
splondike, line 39: Probably don't need the 'char' command here since we already have the quote command.

splondike (commented):
Looks good, just a couple of things in the .talon file to be addressed I think.

lang/scala/scala.py
```diff
@@ -0,0 +1,284 @@
...
+}
+
+scala_types = scala_common_types.copy()
+scala_types.update(scala_common_generic_types)
+ctx.lists["user.code_type"] = scala_types
+
+@ctx.action_class("user")
+class UserActions:
+    def code_operator_indirection():
+        actions.skip()
```
rntz, line 42: You can just leave these undefined if they don't make sense for Scala, implementing them as actions.skip() will just make them do nothing instead of producing an error, which is probably the desired behavior.

lang/scala/scala.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def code_insert_is_null():
+        actions.auto_insert(" == null")
+
+    def code_insert_is_not_null():
+        actions.auto_insert(" != null")
+
+    def code_state_if():
+        actions.insert("if () ")
+        actions.key("left")
```
rntz, line 159: Should probably use `edit.left()` here and below instead, in case the editor doesn't use arrow keys to move.

lang/scala/scala.py
```diff
@@ -0,0 +1,284 @@
...
+    def code_public_function(text: str):
+        result = "def {}".format(
+            actions.user.formatted_text(
+                text, settings.get("user.code_public_function_formatter")
+            )
+        )
+
+        actions.user.code_insert_function(result, None)
+
+    # skipping these: in scala static functions are just functions in objects
```
rntz, line 278: Likewise, if these aren't used then just don't implement them.

rntz (changes requested):

lang/scala/scala.py
```diff
@@ -0,0 +1,284 @@
+from talon import Context, Module, actions, settings
+
+ctx = Context()
+mod = Module()
+ctx.matches = r"""
+mode: user.scala
```
knausj85, line 6: We've migrated to just using a tag for these matches as part of #776. Please update when you have a moment.

and

SCdF: I've done this now! Just checking: I don't have to define the tag anywhere, just use it in the header of the talon and py file? Magic!

SCdF:
@splondike / @rntz apologize deeply for taking so long to get to this!! 

It is done now, hopefully, so please take another look at your convenience

lang/scala/scala.py
```diff
@@ -0,0 +1,269 @@
+from talon import Context, Module, actions, settings
+
+ctx = Context()
+mod = Module()
+ctx.matches = r"""
+user.tag: scala
```
splondike, line 6: This should be `tag: user.scala`

splondike:
Aside from the incorrect context in the .py file, I think this is good to merge.

lang/scala/scala.py
```diff
@@ -0,0 +1,269 @@
...
+    def code_import():
+        actions.auto_insert("import ")
+
+    def code_state_return():
+        actions.insert("return ")
+
+    def code_comment_line_prefix():
+        actions.auto_insert('// ')
+
+    def code_block_comment():
```
splondike, line 208: code_block_comment -> code_comment_block

lang/scala/scala.py
```diff
@@ -0,0 +1,269 @@
...
+        actions.insert('/*')
+        actions.key('enter')
+        actions.key('enter')
+        actions.insert('*/')
+        actions.edit.up()
+
+    def code_block_comment_prefix():
```
splondike, line 215: code_block_comment_prefix -> code_comment_block_prefix

lang/scala/scala.py
```diff
@@ -0,0 +1,269 @@
...
+        actions.auto_insert('/*')
+
+    def code_block_comment_suffix():
```
splondike, line 218: code_block_comment_suffix -> code_comment_block_suffix

splondike:
Apart from the above comments I think this is good to merge.

lang/scala/scala.py
SCdF: Arg! That'll teach me to smash it out on a break and not test it 🚯

lang/scala/scala.py
SCdF: did these change? I definitely tested each of these when I wrote them the first time weeks ago?

SCdF:
OK I think this works now? I've tested it and the tags appear to work and attach correctly

lang/scala/scala.py
splondike: Yeah, I think they did along with the big refactoring of the language features.

splondike:
Looks good, thanks!

PR #737, opened 2022-01-15
Support numbers within prose.

wolfmanstout:
Allows user to say "numb" followed by a number within prose and have it formatted in decimal.

### DIFF
--- a/modes/dictation_mode.talon
+++ b/modes/dictation_mode.talon
@@ -65,11 +65,6 @@ formatted <user.format_text>:
 ^format selection <user.formatters>$:
     user.formatters_reformat_selection(formatters)
 
-# Numbers
-numb <user.number_string>: "{number_string}"
-numb <user.number_string> colon <user.number_string>: "{number_string_1}:{number_string_2}"
-numb <user.number_string> (dot | point) <digit_string>: "{number_string}.{digit_string}"
-
 # Corrections
 scratch that: user.clear_last_phrase()
 scratch selection: edit.delete()
--- a/modes/dictation_mode.talon
+++ b/modes/dictation_mode.talon
@@ -65,6 +65,11 @@ formatted <user.format_text>:
 ^format selection <user.formatters>$:
     user.formatters_reformat_selection(formatters)
 
+# Numbers
+numb <user.number_string>: "{number_string}"
+numb <user.number_string> colon <user.number_string>: "{number_string_1}:{number_string_2}"
+numb <user.number_string> (dot | point) <digit_string>: "{number_string}.{digit_string}"
+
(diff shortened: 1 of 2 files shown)


modes/dictation_mode.talon
```diff
@@ -65,6 +65,11 @@ formatted <user.format_text>:
 ^format selection <user.formatters>$:
     user.formatters_reformat_selection(formatters)
 
+# Numbers
+numb <user.number_string>: "{number_string}"
+numb <user.number_string> colon <user.number_string>: "{number_string_1}:{number_string_2}"
+numb <user.number_string> (dot | point) <digit_string>: "{number_string}.{digit_string}"
```
rntz, line 71: 1. Any reason not just include `prose_number` inside `raw_prose`?
2. Failing that, any reason not to do `<user.prose_number>: "{prose_number}"`?

rntz (commented):
Looks ok, although see comments. I use `numeral` for this. Any pros/cons of that vs `numb`? What does Dragon use by default?

modes/dictation_mode.talon
wolfmanstout: 1. Silly me! I see no reason not to.
2. I think it's helpful to support both numerals and numbers as words.

wolfmanstout:
> Looks ok, although see comments. I use `numeral` for this. Any pros/cons of that vs `numb`? What does Dragon use by default?

Dragon uses numeral. We could certainly use that, perhaps as an alternative. Downside is two more syllables. I copied "numb" from pokey's fork. Considering "numb" will only match this when adjacent to a number I'm not worried about accidentally activating this capture.

rntz:
"he was numb to the pain" might get misheard as "he was numb two the pain"

modes/dictation_mode.talon
rntz: great, if it gets included in `raw_prose` then these commands aren't needed (iiuc?) - they'll be handled by the `<user.raw_prose>: auto_insert(raw_prose)` command.

wolfmanstout:
> "he was numb to the pain" might get misheard as "he was numb two the pain" --> "he was 2 the pain"

Fair enough I'm happy to switch to numeral. Will need to test with Dragon which sometimes behaves weirdly when you reuse its grammar.

modes/dictation_mode.talon
wolfmanstout: Yep. I'll make that change later.

wolfmanstout:
Made the changes and tested with Dragon and Conformer. Pretty small change now!

rntz (approved):

PR #738, opened 2022-01-15
Handle capitalization of e.g. and i.e. within prose.

wolfmanstout:
This works out of the box in Dragon. To test in Conformer you'll need to add something like this to additional_words.csv:
```
i.e.,I E
e.g.,E G
```

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -161,7 +161,8 @@ no_cap_after = re.compile(r"""(
 
 def auto_capitalize(text, state = None):
     """
-    Auto-capitalizes text. `state` argument means:
+    Auto-capitalizes text. Text must contain complete words, abbreviations, and
+    formatted expressions. `state` argument means:
 
     - None: Don't capitalize initial word.
     - "sentence start": Capitalize initial word.
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -178,7 +178,7 @@ def auto_capitalize(text, state = None):
     sentence_end = False
     for c in text:
         # Sentence endings followed by space & double newlines create a charge.
-        if (sentence_end and c == " ") or (newline and c == "\n"):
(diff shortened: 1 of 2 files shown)


wolfmanstout:
I'll add a test and fix for newline after period as well.

rntz:
I haven't looked at this in detail, but it doesn't look like it defines any new states accepted by `auto_capitalize`, which means it's violating one of the intended laws about `auto_capitalize` (which I really should have documented), namely that the following should be true:

```
(result, state2) = auto_capitalize(text1 + text2, state)
```
should behave the same as:
```
(result1, state1) = auto_capitalize(text1, state)
(result2, state2) = auto_capitalize(text2, state1)
result = result1 + result2
```


The goal is to make sure that no matter how the text that gets inserted is "chunked" we produce the same result. The approach I used is to make auto_capitalize should be a resumable finite state machine. Unfortunately this is probably hard to do by hand if you want it to recognize long strings like "e.g." and "i.e." (you'll get a large number of states). Maybe we should give up on this approach and just accept that we'll produce bad results in some unusual cases.

wolfmanstout:
Yes this violates that in several ways, actually. One way to avoid a bunch of state bools but meet that spec is to simply use the formatted output as the state. In order to avoid performance problems a couple more reasonable constraints might be: state size is O(1) and multiple successive calls takes time O(n) regardless of chunking. This could probably be achievable by tracking a list of strings to match and only keeping the max length of that as the state. 

Another option is to formally document the expectations this has on chunking. I agree that it feels brittle now without that.

rntz:
I don't have time to try to figure out the Right Way to do this right now, and it seems like this PR's approach would be better than nothing for most users. If you add a comment explaining under what circumstances this will break I'd be happy to merge this; I'll file an issue about fixing it in the long-term.

wolfmanstout:
Thanks, done.

PR #740, opened 2022-01-16
dictation.py: call auto_format from auto_insert

### LINKED ISSUE #725: auto_insert is documented to call auto_format
I understand if you want to override `auto_insert` too because you're adjusting the cursor position afterwards, but the general intention is for `auto_format` to do text formatting

I think the dictation `auto_insert()` implementation should potentially first `text = auto_format(text)` (which is the identity function by default) as per the docstring:
```
auto_insert(text: str)
  Insert text at the current cursor position, automatically formatting it using actions.auto_format(text)
```

https://github.com/knausj85/knausj_talon/blob/master/code/dictation.py#L401

rntz:
Fixes #725.

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -398,4 +398,5 @@ mode: dictation
 
 @dictation_ctx.action_class("main")
 class main_action:
-    def auto_insert(text): actions.user.dictation_insert(text)
+    def auto_insert(text):
+        actions.user.dictation_insert(actions.auto_format(text))


rntz:
Note that the default implementation of auto_format is a noop (it returns its argument unchanged) - so for most users this PR will cause no behavior change.

rntz:
I'm going to go ahead and merge this, if we decide on moving more work into auto_format we can always do that later.

PR #743, opened 2022-01-21
Add Rust support

purpleP:
Added initial support for Rust.

### DIFF
--- a/code/code.py
+++ b/code/code.py
@@ -25,6 +25,7 @@ extension_lang_map = {
     ".ps1": "powershell",
     ".py": "python",
     ".r": "r",
+    ".rs": "rust",
     ".rb": "ruby",
     ".s": "assembly",
     ".sh": "bash",
--- a/lang/rust/rust.py
+++ b/lang/rust/rust.py
@@ -1,0 +1,215 @@
+from talon import Module, Context, actions, ui, imgui, settings
+
+mod = Module()
+
+ctx = Context()
+ctx.matches = 'tag: user.rust'
+
+ctx.lists['user.code_libraries'] = {
+    'eye oh': 'std::io',
+    'file system': 'std::fs',
+    'envy': 'std::env',
+    'collections': 'std::collections',
+}
+ctx.lists['user.code_functions'] = {
+    'drop': 'drop',
+    'print line': 'println!',
+    'error line': 'eprintln!',
+}
+
+ctx.lists['user.code_type'] = {
+    'you eight': 'u8',
+    'eye eight': 'i8',
+    'you sixteen': 'u16',
+    'eye sixteen': 'i16',
+    'you thirty two': 'u32',
+    'eye thirty two': 'i32',
+    'you sixty four': 'u64',
+    'eye sixty four': 'i64',
+    'you size': 'usize',
+    'eye size': 'isize',
+    'boolean': 'bool',
+    'string': 'str',
+    'owned string': 'String',
+}
+
+@ctx.action_class('user')
+class UserActions:
+    def code_operator_lambda():
+        actions.insert('|| ')
+        actions.edit.left()
+        actions.edit.left()
+
+    def code_operator_subscript():
+        actions.insert('[]')
+        actions.edit.left()
+
+    def code_operator_increment():
+        actions.insert(' += 1')
+
+    def code_operator_indirection():
+        actions.auto('*')
+
+    def code_operator_address_of():
+        actions.auto('&')
+
+    def code_operator_assignment():
+        actions.auto(' = ')
+
+    def code_operator_subtraction():
+        actions.auto(' - ')
+
+    def code_operator_subtraction_assignment():
+        actions.auto(' -= ')
+
+    def code_operator_addition():
+        actions.auto(' + ')
+
+    def code_operator_addition_assignment():
+        actions.auto(' += ')
+
+    def code_operator_multiplication():
+        actions.auto(' * ')
+
+    def code_operator_multiplication_assignment():
+        actions.auto(' *= ')
+
+    def code_operator_exponent():
+        actions.auto('.pow()');
+        actions.edit.left();
+
+    def code_operator_division():
+        actions.auto(' / ')
+
+    def code_operator_division_assignment():
+        actions.auto(' /= ')
+
+    def code_operator_modulo():
+        actions.auto(' % ')
+
+    def code_operator_modulo_assignment():
+        actions.auto(' %= ')
+
+    def code_operator_equal():
+        actions.auto(' == ')
+
+    def code_operator_not_equal():
+        actions.auto(' != ')
+
+    def code_operator_greater_than():
+        actions.auto(' > ')
+
+    def code_operator_greater_than_or_equal_to():
+        actions.auto(' >= ')
+
+    def code_operator_less_than():
+        actions.auto(' < ')
+
+    def code_operator_less_than_or_equal_to():
+        actions.auto(' <= ')
+
+    def code_operator_and():
+        actions.auto(' && ')
+
+    def code_operator_or():
+        actions.auto(' || ')
+
+    def code_operator_bitwise_and():
+        actions.auto(' & ')
+
+    def code_operator_bitwise_or():
+        actions.auto(' | ')
+
+    def code_operator_bitwise_exclusive_or():
+        actions.auto(' ^ ')
+
+    def code_operator_bitwise_left_shift():
+        actions.auto(' << ')
+
+    def code_operator_bitwise_left_shift_assignment():
+        actions.auto(' <<= ')
+
+    def code_operator_bitwise_right_shift():
+        actions.auto(' >> ')
+
+    def code_operator_bitwise_right_shift_assignment():
+        actions.auto(' >>= ')
+
+    def code_operator_object_accessor():
+        actions.auto('.')
+
+    def code_state_switch():
+        actions.insert('match ')
+
+    def code_block():
+        actions.auto('{\n\n}')
+        actions.edit.left()
+        actions.edit.up()
+        actions.key('tab')
+
+    def code_import():
+        actions.auto('use ')
+
+    def code_comment_line_prefix():
+        actions.auto('//')
+
+    def code_comment_documentation():
+        actions.auto('/// ')
+
+    def code_self():
+        actions.auto('self')
+
+    def code_insert_true():
+        actions.auto('true')
+
+    def code_insert_false():
+        actions.auto('false')
+
+    def code_state_if():
+        actions.insert('if ')
+
+    def code_state_else():
+        actions.insert('else ')
+
+    def code_state_else_if():
+        actions.insert('else if ')
+
+    def code_state_return():
+        actions.insert('return ')
+
+    def code_insert_function(text: str, selection: str):
+        actions.user.paste(f'{text}({selection or ""})')
+        actions.edit.left()
+
+    def code_private_function(text: str):
+        actions.insert('fn ')
+        formatter = settings.get('user.code_private_function_formatter')
+        function_name = actions.user.formatted_text(text, formatter)
+        actions.user.code_insert_function(function_name, None)
+
+    def code_protected_function(text: str):
+        actions.insert('pub(crate) fn ')
+        formatter = settings.get('user.code_protected_function_formatter')
+        function_name = actions.user.formatted_text(text, formatter)
+        actions.user.code_insert_function(function_name, None)
+
+    def code_public_function(text: str):
+        actions.insert('pub fn ')
+        formatter = settings.get('user.code_public_function_formatter')
+        function_name = actions.user.formatted_text(text, formatter)
+        actions.user.code_insert_function(function_name, None)
+
+    def code_default_function(text: str):
+        actions.user.code_private_function(text)
+
+    def code_insert_type_annotation(type: str):
+        actions.insert(f': {type}')
+
+    def code_insert_return_type(type: str):
+        actions.insert(f' -> {type}')
+
+    def code_state_for():
+        actions.insert('for ')
+
+    def code_state_return():
+        actions.insert('return ')
--- a/lang/rust/rust.talon
+++ b/lang/rust/rust.talon
@@ -1,0 +1,55 @@
+tag: user.rust
+-
+tag(): user.code_imperative
+tag(): user.code_object_oriented
+tag(): user.code_comment_line
+tag(): user.code_comment_block_c_like
+tag(): user.code_comment_documentation
+tag(): user.code_data_bool
+tag(): user.code_functions
+tag(): user.code_functions_gui
+tag(): user.code_libraries
+tag(): user.code_libraries_gui
+tag(): user.code_operators_array
+tag(): user.code_operators_assignment
+tag(): user.code_operators_bitwise
+tag(): user.code_operators_math
+tag(): user.code_operators_pointer
+
+settings():
+    user.code_private_function_formatter = "SNAKE_CASE"
+    user.code_protected_function_formatter = "SNAKE_CASE"
+    user.code_public_function_formatter = "SNAKE_CASE"
+    user.code_private_variable_formatter = "SNAKE_CASE"
+    user.code_public_variable_formatter = "SNAKE_CASE"
+
+state unsafe: "unsafe "
+
+state struct <user.text>:
+    insert("struct ")
+    insert(user.formatted_text(text, "PUBLIC_CAMEL_CASE"))
+
+state enum <user.text>:
+    insert("enum ")
+    insert(user.formatted_text(text, "PUBLIC_CAMEL_CASE"))
+
+toggle use: user.code_toggle_libraries()
+
+state pub: "pub "
+state pub crate: "pub(crate) "
+state dyn: "dyn "
+state imp: "impl "
+state let mute: "let mut "
+state let: "let "
+state mute: "mut "
+state ref mute: "ref mut "
+state ref: "ref "
+state trait: "trait "
+state match: "match "
+op arrow: " -> "
+op dub arrow: " => "
+state use: "use "
+
+use <user.code_libraries>:
+    user.code_insert_library(code_libraries, "")
+    key(semicolon enter)
--- a/modes/language_modes.talon
+++ b/modes/language_modes.talon
@@ -7,6 +7,7 @@
 ^force markdown$: user.code_set_language_mode("markdown")
 ^force python$: user.code_set_language_mode("python")
 ^force are language$: user.code_set_language_mode("r")
+^force rust$: user.code_set_language_mode("rust")
 ^force talon [language]$: user.code_set_language_mode("talon")
 ^clear language modes$: user.code_clear_language_mode()
 [enable] debug mode:


lang/rust/rust.py
```diff
@@ -0,0 +1,185 @@
...
+mod = Module()
+
+ctx = Context()
+ctx.matches = r'''
+mode: user.rust
+mode: user.auto_lang
+and code.language: rust
+'''
+
+ctx.tags = ['user.code_operators', 'user.code_generic']
```
auscompgeek, line 12: Why not set these tags in the `.talon` file along with the rest?

... actually, do these tags even still exist?

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+    user.code_public_variable_formatter = "SNAKE_CASE"
+
+^function <user.text>$: user.code_default_function(text)
+^public function <user.text>$: user.code_default_function(text)
+
+unsafe: user.code_state_unsafe()
+
+^struct <user.text>:
+    insert("struct ")
+    insert(user.formatted_text(text, "hammer"))
```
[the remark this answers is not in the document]
purpleP: Yes, it works. Happy to change it if needed.

lang/rust/rust.py
purpleP: I don’t even know whether they exist… It’s probably copy paste from somewhere, frankly.

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+
+toggle use: user.code_toggle_libraries()
+
+state let:
+    insert("let  = ;")
+    key(left left left left)
+
+has type: insert(": ")
+ref: insert("&")
+ref mute: insert("&mut")
```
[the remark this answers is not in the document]
purpleP: Should I change it everywhere?

lang/rust/rust.py
```diff
@@ -0,0 +1,185 @@
+from talon import Module, Context, actions, ui, imgui, settings
+
+mod = Module()
+
+ctx = Context()
+ctx.matches = r'''
+mode: user.rust
+mode: user.auto_lang
+and code.language: rust
```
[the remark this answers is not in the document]
splondike: Also add to language_modes.talon.

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+tag(): user.code_operators_math
+
+settings():
+    user.code_private_function_formatter = "SNAKE_CASE"
+    user.code_protected_function_formatter = "SNAKE_CASE"
+    user.code_public_function_formatter = "SNAKE_CASE"
+    user.code_private_variable_formatter = "SNAKE_CASE"
+    user.code_public_variable_formatter = "SNAKE_CASE"
+
+^function <user.text>$: user.code_default_function(text)
```
splondike, line 30: Should just use the commands behind the code_functions tag for consistency with other languages.

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+^public function <user.text>$: user.code_default_function(text)
+
+unsafe: user.code_state_unsafe()
```
splondike, line 33: For these types of things we like to prefix the command with 'state', single word commands significantly increase the chance of misrecognitions.

Since you haven't defined the action this also probably doesn't work (see https://talon.wiki/unofficial_talon_docs/#actions). I'd suggest just putting the commands inline in Talon if you want to keep this. See c.talon's `state type deaf struct` command for example.

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+    insert("enum ")
+    insert(user.formatted_text(text, "hammer"))
+    insert(" {\n\n}")
+    edit.left()
+    edit.up()
+    key(tab)
+
+toggle use: user.code_toggle_libraries()
+
+state let:
```
splondike, line 53: I think for knausj just use `state let: "let"`. It looks like other languages with variable keywords (javascript, go) only add that keyword. It'd be good to be consistent.

lang/rust/rust.talon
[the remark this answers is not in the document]
splondike: Formatter usage looks OK to me. I'd probably suggest prefixing struct and enum with 'state' to look a bit more like other languages. Anchoring seems OK for these constructs also.

lang/rust/rust.talon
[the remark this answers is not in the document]
splondike: It's probably not necessary to change unless you feel like it, just good to know about the short syntax. Stylistically we also only use the short syntax if its the only thing in the Talonscript body (otherwise it's a bit harder to read).

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+    insert("let  = ;")
+    key(left:4)
+
+has type: insert(": ")
+ref: insert("&")
```
splondike, line 58: You can probably use the `op address of` command in operators_pointer.talon instead of `ref` (you've already implemented the relevant action). For the other one you might call it `op mute address of` for consistency and just stick it inline here in the .talon file.

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+    edit.up()
+    key(tab)
+
+toggle use: user.code_toggle_libraries()
+
+state let:
+    insert("let  = ;")
+    key(left:4)
+
+has type: insert(": ")
```
splondike, line 57: If you implement the code_insert_type_annotation action you'll get the 'is type {user.code_type}' command which does the same as this.

lang/rust/rust.py
```diff
@@ -0,0 +1,185 @@
...
+    'eye thirty two': 'i32',
+    'you sixty four': 'u64',
+    'eye sixty four': 'i64',
+    'you size': 'usize',
+    'eye size': 'isize',
+    'boolean': 'bool',
+}
+
+mod.list('builtin_types', desc='Rust builtin types')
+mod.list('code_libraries', desc='Rust std libraries')
```
splondike, line 34: Don't need to declare code_libraries here since it's already defined in the `libraries_gui.py` file. Just the ctx version is fine.

I'd suggest using the existing code_type list as well rather than defining builtin_types. That way you can benefit from the stuff in functions.talon.

lang/rust/rust.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def code_self():
+        actions.auto_insert('self')
+
+    def code_state_unsafe():
+        actions.insert('unsafe {\n\n}')
+        actions.edit.up()
+        actions.edit.tab()
+
+    def code_true():
```
splondike, line 148: code_insert_true, also use actions.insert() for this one (auto_insert goes through a formatter function).

lang/rust/rust.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def code_state_return():
+        actions.insert("return ")
+
+    def code_default_function(text: str):
+        actions.user.insert(f'fn {text}() -> () ' + '{}')
+        for i in range(10):
+            actions.edit.left()
+
+    def code_public_function(text: str):
```
splondike, line 172: Do you want to run text through a formatter first?

lang/rust/rust.talon
```diff
@@ -0,0 +1,69 @@
...
+ref: insert("&")
+ref mute: insert("&mut")
+
+use <user.code_libraries>:
+    user.code_insert_library(code_libraries, "")
+    key(semicolon enter)
+
+<user.builtin_types>: insert("{builtin_types}")
```
splondike, line 65: Just use the `type {user.code_type}` command in code_functions once you swap to using that list rather than builtin_types.

splondike (changes requested):
Some stylistic changes and other suggestions to be more consistent with other languages.

sman908:
I've addressed at least some of the comments.

@splondike @auscompgeek Could you guys give it another look?

purpleP:
I have updated the PR and addressed at least some of the comments.

@splondike @auscompgeek could you guys give it another look?

lang/rust/rust.talon
```diff
@@ -0,0 +1,64 @@
...
+
+settings():
+    user.code_private_function_formatter = "SNAKE_CASE"
+    user.code_protected_function_formatter = "SNAKE_CASE"
+    user.code_public_function_formatter = "SNAKE_CASE"
+    user.code_private_variable_formatter = "SNAKE_CASE"
+    user.code_public_variable_formatter = "SNAKE_CASE"
+
+unsafe:
+    "unsafe {\n\n}"
```
auscompgeek, line 31: What if I want to write `unsafe fn`? Seems like it'd be reasonable to just insert "unsafe " so I could say `state unsafe` then either `block` or `funky foo`.

lang/rust/rust.py
```diff
@@ -0,0 +1,216 @@
...
+        actions.insert('pub ')
+        formatter = settings.get('user.code_private_function_formatter')
+        function_name = actions.user.formatted_text(text, formatter)
+        actions.user.code_insert_function(function_name, None)
+
+    def code_state_return():
+        actions.insert('return ')
+
+    def code_insert_is_null():
+        actions.auto_insert(' != None')
```
auscompgeek, line 213: I think you have this the wrong way around.

Also, PartialEq won't be implemented for Option<T> where T doesn't implement PartialEq, right? Isn't there a method which could be used here instead?

Going further, would someone perhaps like to use null and none to refer to the null pointer and the empty Option respectively?

splondike:
Yeah looking good, just a few generic suggestions (I haven't used Rust so can't comment on auscompgeek's recent comments):

* Should add 'state' prefixes at least to 'ref' and 'ref mut' I think. I'd probably go for it in front of the enum and struct as well, but since those will be multi word it's up to you.
* Speaking on the builtin_types or rust_generic_data_structures lists, do you not like the 'type {user.code_type}' command in functions.talon combining both lists? My inclination would be to just use that existing command and remove your two custom lists. You can keep them as separate dicts and put them in to the existing code_types list if you want (see the `symbol_key_words` variable in keys.py for example).
* Don't need to redeclare code_libraries, and should probably move the code_operators tag activation into the talon line and otherwise delete the tag setting line in the Python file.

lang/rust/rust.py
purpleP: Rust doesn’t have null builtin. I’m not sure about what you suggest here.

auscompgeek: https://doc.rust-lang.org/std/ptr/fn.null.html

lang/rust/rust.talon
```diff
@@ -0,0 +1,61 @@
...
+    insert(user.formatted_text(text, "hammer"))
+    insert(" {\n\n}")
+    edit.left()
+    edit.up()
+    key(tab)
+
+toggle use: user.code_toggle_libraries()
+
+state let: "let "
+state let mut: "let mut "
```
auscompgeek, line 54: Would you say 'mut' or 'mute' for this command?

lang/rust/rust.py
```diff
@@ -0,0 +1,216 @@
...
+    def code_state_else_if():
+        actions.insert('else if {}')
+        for i in range(3):
+            actions.edit.left()
+
+    def code_state_return():
+        actions.insert("return ")
+
+    def code_insert_function(text: str, selection: str):
+        actions.user.paste(f'fn {text}({selection or ""})')
```
auscompgeek, line 181: This action is used by `code_functions_gui` to insert function calls. I don't think you want `funk print line` to insert `fn println!()`, for example.

lang/rust/rust.talon
```diff
@@ -0,0 +1,61 @@
+mode: command
+and mode: user.rust
+mode: command
+and mode: user.auto_lang
+and code.language: rust
+-
+tag(): user.code_imperative
+tag(): user.code_object_oriented
+
+tag(): user.code_comment_line
```
auscompgeek, line 10: What about block comments?
```suggestion
tag(): user.code_comment_line
tag(): user.code_comment_block_c_like
```

lang/rust/rust.py
purpleP: > https://doc.rust-lang.org/std/ptr/fn.null.html

Yeah, I’ve seen that, but it’s not even in prelude. I don’t think it will be useful in general. It’s just a regular method like any other. `Option` is at least in prelude. And no one is really checking for `None` like that in Rust anyway, so if it would up to me I wouldn’t implemented it at all.

lang/rust/rust.py
purpleP: Yeah, I will need to fix that, thank you!

sman908:
@splondike @auscompgeek I've made at least some of the changes you were suggesting. Could you take another look? One of the questions I have is that I don't really understand how `code_generic` works yet, so maybe I've missed something there.

splondike:
@purpleP It's looking pretty good to me. The following are my thoughts:

* You don't use the rust_generic_data_structures list, so you could just remove it.
* You don't need to redeclare code_libraries on line 25.
* You don't need the ctx.tags line on line 11; code_generic doesn't exist as a tag (a warning should be printed in your log) and code_functions_gui is already activated in your .talon file.

sman908:
>

> @purpleP It's looking pretty good to me. The following are my thoughts:
> 
>     * You don't use the rust_generic_data_structures list, so you could just remove it.
> 
>     * You don't need to redeclare code_libraries on line 25.
> 
>     * You don't need the ctx.tags line on line 11; code_generic doesn't exist as a tag (a warning should be printed in your log) and code_functions_gui is already activated in your .talon file.

I think I addressed all of this points now.

By the way `code_generic` and stuff was taken from `java`. So maybe it should be fixed at some point.

purpleP:
> @purpleP It's looking pretty good to me. The following are my thoughts:
> 
> * You don't use the rust_generic_data_structures list, so you could just remove it.
> * You don't need to redeclare code_libraries on line 25.
> * You don't need the ctx.tags line on line 11; code_generic doesn't exist as a tag (a warning should be printed in your log) and code_functions_gui is already activated in your .talon file.

I think I’ve addressed all of this points by now.

By the way, some of the mistakes you’ve pointed at come from copying and pasting Java support implementation. You might want to check this out and do a clean up later.

splondike (approved):
OK, looks good to me, though I don't think the 'use <...>' functionality works at the moment. @auscompgeek all look fine to you also?

lang/rust/rust.talon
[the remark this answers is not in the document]
auscompgeek: Hm, I would've expected these to use the `PUBLIC_CAMEL_CASE` et al names, rather than their spoken form.

lang/rust/rust.talon
```diff
@@ -0,0 +1,63 @@
+mode: command
+and mode: user.rust
+mode: command
+and mode: user.auto_lang
+and code.language: rust
+-
+tag(): user.code_imperative
```
auscompgeek, line 7: Did you want to implement the `code_block` action too, for the `block` command?

auscompgeek (commented):
Mostly LGTM; will test once I get home later today.

lang/rust/rust.py
```diff
@@ -0,0 +1,203 @@
...
+}
+
+@ctx.action_class('user')
+class UserActions:
+    def code_operator_lambda():
+        actions.insert('|| ')
+
+    def code_operator_subscript():
+        actions.insert('[]')
+        actions.edit('left')
```
auscompgeek, line 46:
```pytb
 > user.code_operator_subscript()
KeyError: "Action 'edit' is not declared."
```
I think you meant
```suggestion
        actions.edit.left()
```

lang/rust/rust.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def code_state_if():
+        actions.insert('if {}')
+        for i in range(3):
+            actions.edit.left()
+
+    def code_state_else_if():
+        actions.insert('else if {}')
+        for i in range(3):
+            actions.edit.left()
```
auscompgeek, line 171: Oh, `code_state_else` isn't implemented either.

lang/rust/rust.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def code_insert_return_type(type: str):
+        actions.insert(f' -> {type}')
+
+    def code_state_for():
+        actions.insert('for ')
+
+    def code_public_function(text: str):
+        actions.insert('pub ')
+        actions.user.code_default_function(text)
```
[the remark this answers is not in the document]
purpleP: You’re not supposed to, but what solution you’re suggesting?

lang/rust/rust.talon
```diff
@@ -0,0 +1,63 @@
...
+    user.code_public_function_formatter = "SNAKE_CASE"
+    user.code_private_variable_formatter = "SNAKE_CASE"
+    user.code_public_variable_formatter = "SNAKE_CASE"
+
+state unsafe: "unsafe "
+
+state struct <user.text>:
+    insert("struct ")
+    insert(user.formatted_text(text, "hammer"))
+    insert(" {\n\n}")
```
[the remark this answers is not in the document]
purpleP: Yeah, I guess talon language support shouldn’t have something that’s more like a snippet. Will make this more dumb.

lang/rust/rust.talon
```diff
@@ -0,0 +1,63 @@
...
+    key(tab)
+
+toggle use: user.code_toggle_libraries()
+
+state dyn: "dyn "
+state imp: "impl "
+state let mute: "let mut "
+state let: "let "
+state mute: "mut "
+state ref mute: "ref mut "
```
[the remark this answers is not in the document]
purpleP: It’s not only used in match statements. It can be part of a pattern matching in `let` also.

lang/rust/rust.py
```diff
@@ -0,0 +1,203 @@
...
+    def code_insert_true():
+        actions.auto_insert('true')
+
+    def code_insert_false():
+        actions.auto_insert('false')
+
+    def code_state_if():
+        actions.insert('if {}')
+        for i in range(3):
+            actions.edit.left()
```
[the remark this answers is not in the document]
purpleP: Yeah, I agree.  Structural editing should be done by things like cursorless.

lang/rust/rust.py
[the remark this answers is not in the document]
auscompgeek: Each of these should use their own formatter. You already set the settings for them, but they're unused.

lang/rust/rust.talon
sman908: > Did you want to implement the `code_block` action too, for the `block` command?

Yeah, thanks for pointing this out.

And about formatters - can you give an example of what you mean?

lang/rust/rust.talon
```diff
@@ -0,0 +1,63 @@
...
+state ref: "ref "
+state trait: "trait "
+state match: "match "
```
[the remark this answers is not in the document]
sman908: > Should we make this the implementation for `state switch` too, like I did with Python? Otherwise it currently does nothing.

If we implement `switch` how do we implement `case`?
Like that?

`_ => `

That wouldn't be very convenient as you would need to return to write the pattern.

I think that until language support is refactored some more to add more paradigms from FP like sum types and pattern matching it's better to leave it as it is. But we can implement `switch`, this wouldn't make any harm, I guess.

lang/rust/rust.py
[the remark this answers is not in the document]
sman908: > Each of these should use their own formatter. You already set the settings for them, but they're unused.

I'm sorry, I need an example from other place or from you. I'm not sure what exactly I need to type.

lang/rust/rust.py
```diff
@@ -0,0 +1,210 @@
...
+        actions.auto_insert(' >> ')
+
+    def code_operator_bitwise_right_shift_assignment():
+        actions.auto_insert(' >>= ')
+
+    def code_operator_object_accessor():
+        actions.auto_insert('.')
+
+    def code_block():
+        actions.auto_insert('{\n\n}')
```
auscompgeek, line 143: I believe the intent of `block` is to have the cursor within the braces when it ends.

code/code.py
```diff
@@ -25,10 +25,12 @@
     ".ps1": "powershell",
     ".py": "python",
     ".r": "r",
+    ".rs": "rust",
     ".rb": "ruby",
     ".s": "assembly",
     ".sh": "bash",
     ".snippets": "snippets",
+    ".sql": "sql",
```
rntz, line 33: This shouldn't be in this PR, since it doesn't add sql support.

lang/rust/rust.py
```diff
@@ -0,0 +1,217 @@
...
+    'eye sixteen': 'i16',
+    'you thirty two': 'u32',
+    'eye thirty two': 'i32',
+    'you sixty four': 'u64',
+    'eye sixty four': 'i64',
+    'you size': 'usize',
+    'eye size': 'isize',
+    'boolean': 'bool',
+    'string': 'str',
+    'owned strign': 'String',
```
rntz, line 37: this should be "owned string" not "owned strign"

lang/rust/rust.py
```diff
@@ -0,0 +1,217 @@
...
+}
+
+@ctx.action_class('user')
+class UserActions:
+    def code_operator_lambda():
+        actions.insert('|| ')
```
rntz, line 43: Maybe this should also do `actions.edit.left(); actions.edit.left()` to put the cursor between the vertical bars?

lang/rust/rust.py
```diff
@@ -0,0 +1,217 @@
...
+        actions.auto_insert('{\n\n}')
+        actions.edit.left()
+        actions.edit.up()
+        actions.key('tab')
+
+    def code_import():
+        actions.auto_insert('use ')
+
+    def code_comment_line_prefix():
+        actions.auto_insert('//')
```
rntz, line 156: You sometimes use auto_insert and sometimes insert - any reason? If not I'd suggest just using insert().

lang/rust/rust.talon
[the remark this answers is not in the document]
rntz: Yeah, this should use "PUBLIC_CAMEL_CASE" not "hammer" because if a user renames their hammer formatter this will not work for them.

lang/rust/rust.talon
```diff
@@ -0,0 +1,59 @@
...
+state arrow: "-> "
+state dub arrow: "=> "
```
rntz, line 54: 1. Why do these not have leading space?
2. Why are they "state arrow" rather than "op arrow"?

rntz (changes requested):

lang/rust/rust.py
```diff
@@ -0,0 +1,217 @@
+from talon import Module, Context, actions, ui, imgui, settings
+
+mod = Module()
+
+ctx = Context()
+ctx.matches = r'''
+mode: user.rust
```
knausj85, line 7: We've migrated to just using a tag for these matches as part of #776. Please update when you have a moment.

PR #744, opened 2022-01-21
Remove redundant `ctrl` press from Mac screenshots

xavier630:
Remove redundant `ctrl` press from Mac screenshots

### DIFF
--- a/code/screenshot.py
+++ b/code/screenshot.py
@@ -38,7 +38,7 @@ class Actions:
         if app.platform == "windows":
             actions.key("super-shift-s")
         elif app.platform == "mac":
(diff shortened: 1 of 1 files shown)


splondike:
Looks fine, will just let somebody else on OSX confirm it works fine for them also.

nriley:
Please see the comments on #745.

splondike:
Do you know if cmd-shift-4 works with the default settings on Mac? If so, then this is no worse than what's already on master.

xavier630:
Yep, it works. This really just tidies up what's on master, I think it's quite independent of any CLI migration unless there is a PR already open for it which this will clash with.

PR #745, opened 2022-01-21
Add command to open screenshot settings

xavier630:
This is an important UI on mac for taking screenshots as it's where you decide where to save it, whether it will be a video, if a timer should be used, etc.

I'm open to ideas for the command naming, the naming I went for matches the pattern in the file but could be more intuitive.

### DIFF
--- a/code/screenshot.py
+++ b/code/screenshot.py
@@ -42,6 +42,14 @@ class Actions:
         elif app.platform == "linux":
             actions.key("shift-printscr")
 
+    def screenshot_settings():
+        """Opens the settings UI for screenshots.
+        Only applies to Mac for now
+        """
+        if app.platform == "mac":
+            actions.key("cmd-shift-5")
+
+
     def screenshot_clipboard(screen_number: Optional[int] = None):
         """Takes a screenshot of the entire screen and saves it to the clipboard.
         Optional screen number can be given to use screen other than main."""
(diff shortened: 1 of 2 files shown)


splondike (approved):
Looks good to merge after somebody else using OSX verifies.

nriley:
I don't think this is a particularly good idea as the keyboard shortcuts for taking screenshots on macOS are customizable. Potentially wrapping the screencapture command line utility would be a better idea.

nriley:
I'm also happy just to submit a PR that uses screencapture if @xavier630 doesn't want to. It wouldn't be a lot of work.

xavier630:
Hey team,

Agreed with Splondike that I'd rather merge this and have a separate piece of work to convert to CLI functionality.

I'd rather not block on it, especially since this follows the existing pattern in the file: which already assumes default shortcuts are used:
image

code/screenshot.py
```diff
@@ -42,6 +42,14 @@ def screenshot_selection():
         elif app.platform == "linux":
             actions.key("shift-printscr")
 
+    def screenshot_settings():
+        """Opens the settings UI for screenshots.
+        Only applies to Mac for now
+        """
+        if app.platform == "mac":
+            actions.key("cmd-shift-5")
```
rntz, line 50: Please add an `else` branch that raises an error if we're not on mac.

rntz (changes requested):

code/screenshot.py
xavier630: @rntz I've now added this. I can't test it due to my OS, but the code is pretty basic.

splondike:
Merging this in. If somebody wants to re-implement using the CLI, the coast is now clear :).

PR #746, opened 2022-01-21
Add comment to SEP and NOSEP variables

xavier630:
As requested here: https://github.com/knausj85/knausj_talon/pull/729/files#r782138064


xavier630:
I actually think we'd get better modifiability by passing in the actual separator string to each formatter since it would mean future formatters could use their own different separators. 

eg (passes string behind the constant):
`NO_SPACES": (SMASH_SEPARATOR, every_word(lambda w: w))`
Instead of (passes boolean behind the constant):
`"NO_SPACES": (NOSEP, every_word(lambda w: w)),`

But that's a refactor for another day.

PR #748, opened 2022-01-21
Modernize slack

SCdF:
- Reversing commands to be noun verb not verb noun, as it standard
   in knausj. I'm not removing the old way for now, though I think
   we shouild
 - Dropping the "clear" command, as it is dealt with elsewhere and
   it conflicts with those other commands, e.g. "clear word left"
   being interpreted as "clear everything and type the word 'left'"
   instead of "clear the word left of the cursor"


splondike:
This PR does remove the old way I think?

The move to noun verb sounds good to me, but we should keep in the old commands for a while (particularly since Slack is so important to Talon users, some of them are probably even using the desktop app :D). To do that you can show a notify balloon when the old commands are executed but then execute them anyway. See `mouse_grid.talon` for another place we've done that.

Removing the clear command completely seems like a good idea though since its commands conflict with some more commonly used ones.

SCdF:
> This PR does remove the old way I think?

So it does! At some point I was maintaining both

I've added the changed ones back with deprecation warnings (except clear)

What do you think of "toggle <left|right> side bar"? I've left that as is, but I wonder if that should be "<left|right> side bar toggle"?

I also fixed a couple of bugs with some of the other existing commands.

splondike (approved):

splondike:
Cool, looks good to me now :). Probably the left/right side bar toggle version would fit the pattern as well if you'd like to add that :).

Otherwise will give other maintainers a day or two to look and can then merge it in.

Issue #603, opened 2021-09-22
Revisit use of modes


pokey:
### The problem

We have had issues in the past where commands are unexpectedly active in sleep mode, eg #561.  The problem is that if you put `mode: ` in a `.talon` file header, you disable the implicit `mode: command`, which in turn allows you to enable the file during sleep accidentally.

A quick [search](https://github.com/search?q=%22mode%3A+%22+repo%3Aknausj85%2Fknausj_talon+extension%3Atalon&type=Code&ref=advsearch&l=&l=) reveals that we _still_ have some commands active in sleep mode for this reason.  See eg [go.talon](https://github.com/knausj85/knausj_talon/blob/master/lang/go.talon#L1-L3), which results in `go` commands being active if talon is sleeping while you have a `go` file open.

### The solution

We may want to prefer using tags in many of the places that we use modes today.  For example, the following modes are good candidates to become tags:

- [x] `user.homophones`
- [x] `user.windbg`
- [x] `user.help`
- [x] `user.help_scope`
- [x] all programming language modes
- [ ] [admin ](https://github.com/knausj85/knausj_talon/blob/93e57d041ee9097e5a58a957cc6d30a94c161464/core/modes/modes.py#L6)
- [ ] debug
- [ ] ida

On the other hand, things like "game" or "solo" would be examples of good, general modes, because a mode should ideally be very transformative of the command set and make sense when enabled by itself.

AndreasArvidsson:
I've just replaced all of my language modes with tags instead

```
mode: command
and mode: user.java

mode: command
and mode: user.auto_lang
and code.language: java
```

Turns into
```
tag: user.java

tag: user.auto_lang
and code.language: java
```

The optimal solution is of course to revaluate code.language but until then I think that tags is the cleaner solution.

AndreasArvidsson:
@pokey @knausj85 
I've created a solution where we are back to a single line header for each language. This time using a tag.
```
tag: user.typescript
-
```
https://github.com/AndreasArvidsson/andreas-talon/blob/master/langs/javascript/typescript.talon
https://github.com/AndreasArvidsson/andreas-talon/blob/master/misc/language_mode/language_mode.py

AndreasArvidsson:
I have now made a pr to update the programming languages to use tags instead
#776

PR #752, opened 2022-01-30
Add support for terraform

2d6:
This merge request intends to add support for [terraform](https://www.terraform.io/). Since terraform is both an application and a language, I included basic bindings for both.
Terraform uses a declarative language, so I could not implement most of the mutating definitions in tag `user.code_operators_assignment`. I included this tag only for the assignment operator. I'm not sure of this is the right way to go.

### DIFF
--- a/apps/terraform/terraform.talon
+++ b/apps/terraform/terraform.talon
@@ -1,0 +1,15 @@
+tag: terminal
+and tag: user.terraform
+-
+terraform: "terraform "
+
+terraform apply: "terraform apply "
+terraform destroy: "terraform destroy "
+terraform format recursive: "terraform fmt -recursive\n"
+terraform format: "terraform fmt\n"
+terraform help: "terraform -help"
+terraform init upgrade: "terraform init -upgrade\n"
+terraform init: "terraform init\n"
+terraform plan: "terraform plan\n"
+terraform state move: "terraform state mv "
+terraform validate: "terraform validate\n"
--- a/lang/terraform/terraform.py
+++ b/lang/terraform/terraform.py
@@ -1,0 +1,156 @@
+from talon import Context, Module, actions, settings
+
+ctx = Context()
+mod = Module()
+ctx.matches = r"""
+mode: user.terraform
+mode: user.auto_lang
+and code.language: terraform
+"""
+
+types = {
+    "string": "string",
+    "number": "number",
+    "bool": "bool",
+    "list": "list",
+    "map": "map",
+    "null": "null"
+}
+
+ctx.lists["user.code_type"] = types
+
+common_properties = {
+    "name": "name",
+    "type": "type",
+    "description": "description",
+    "default": "default",
+    "for each": "for_each",
+    "count": "count",
+    "prevent destroy": "prevent_destroy",
+    "nullable": "nullable",
+    "sensitive": "sensitive",
+    "depends on": "depends_on",
+    "provider": "provider",
+    "source": "source",
+}
+
+mod.list("terraform_common_property", desc="Terraform Modifier")
+ctx.lists["self.terraform_common_property"] = common_properties
+
+module_blocks = {
+    "variable": "variable",
+    "output": "output",
+    "provider": "provider",
+    "module": "module",
+}
+
+mod.list("terraform_module_block", desc="Simple Terraform Block")
+ctx.lists["self.terraform_module_block"] = module_blocks
+
+
+@mod.action_class
+class Actions:
+
+    def code_terraform_module_block(text: str):
+        """Inserts a new module-related block of a given type (e.g. variable, output, provider...)"""
+
+    def code_terraform_resource(text: str):
+        """Inserts a new resource block with given name"""
+
+    def code_terraform_data_source(text: str):
+        """Inserts a new data block with given name"""
+
+
+@ctx.action_class("user")
+class UserActions:
+
+    def code_terraform_module_block(text: str):
+        actions.insert(text + ' ""')
+        actions.key("left")
+
+    def code_terraform_resource(text: str):
+        result = 'resource "{}" ""'.format(
+            actions.user.formatted_text(
+                text, "SNAKE_CASE"
+            )
+        )
+
+        actions.insert(result)
+        actions.key("left")
+
+    def code_terraform_data_source(text: str):
+        result = 'data "{}" ""'.format(
+            actions.user.formatted_text(
+                text, "SNAKE_CASE"
+            )
+        )
+
+        actions.insert(result)
+        actions.key("left")
+
+    def code_operator_assignment():
+        actions.insert(" = ")
+
+    def code_operator_subtraction():
+        actions.insert(" - ")
+
+    def code_operator_addition():
+        actions.insert(" + ")
+
+    def code_operator_multiplication():
+        actions.insert(" * ")
+
+    def code_operator_division():
+        actions.insert(" / ")
+
+    def code_operator_modulo():
+        actions.insert(" % ")
+
+    def code_operator_equal():
+        actions.insert(" == ")
+
+    def code_operator_not_equal():
+        actions.insert(" != ")
+
+    def code_operator_greater_than():
+        actions.insert(" > ")
+
+    def code_operator_greater_than_or_equal_to():
+        actions.insert(" >= ")
+
+    def code_operator_less_than():
+        actions.insert(" < ")
+
+    def code_operator_less_than_or_equal_to():
+        actions.insert(" <= ")
+
+    def code_operator_and():
+        actions.insert(" && ")
+
+    def code_operator_or():
+        actions.insert(" || ")
+
+    def code_insert_true():
+        actions.insert("true")
+
+    def code_insert_false():
+        actions.insert("false")
+
+    def code_operator_lambda():
+        actions.insert(" => ")
+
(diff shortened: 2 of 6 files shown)


apps/terraform/terraform.talon
```diff
@@ -0,0 +1,11 @@
+tag: terminal
+and tag: user.terraform
+-
+terraform: "terraform "
+
+terraform init: "terraform init\n"
+terraform init upgrade: "terraform init -upgrade\n"
+terraform plan: "terraform plan\n"
+terraform apply: "terraform apply "
+terraform destroy: "terraform destroy "
```
pokey, line 10: Probably smart not to have a newline here 😅

lang/terraform/terraform.py
```diff
@@ -0,0 +1,156 @@
...
+    "count": "count",
+    "prevent destroy": "prevent_destroy",
+    "nullable": "nullable",
+    "sensitive": "sensitive",
+    "depends on": "depends_on",
+    "provider": "provider",
+    "source": "source",
+}
+
+mod.list("terraform_common_properties", desc="Terraform Modifiers")
```
pokey, line 37: List names should be singular

lang/terraform/terraform.py
```diff
@@ -0,0 +1,156 @@
...
+ctx.lists["self.terraform_common_properties"] = common_properties
+
+module_blocks = {
+    "variable": "variable",
+    "output": "output",
+    "provider": "provider",
+    "module": "module",
+}
+
+mod.list("terraform_module_blocks", desc="Simple Terraform Blocks")
```
pokey, line 47: As above

lang/terraform/terraform.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def code_terraform_data_source(text: str):
+        """Inserts a new data block with given name"""
+
+
+@ctx.action_class("user")
+class UserActions:
+
+    def code_terraform_module_block(text: str):
+        actions.auto_insert(text + " \"\"")
```
pokey, line 68: Do we want `auto_insert` here? Cc/ @rntz

lang/terraform/terraform.talon
```diff
@@ -0,0 +1,27 @@
...
+tag(): user.code_data_null
+tag(): user.code_imperative
+tag(): user.code_operators_assignment
+tag(): user.code_operators_lambda
+tag(): user.code_operators_math
+
+state {user.terraform_module_blocks}:
+    user.code_terraform_module_block(user.terraform_module_blocks)
+
+^resource <user.text>:
```
pokey, line 17: Why front-anchored?

lang/terraform/terraform.talon
```diff
@@ -0,0 +1,27 @@
...
+    user.code_terraform_resource(text)
+
+^data [source] <user.text>:
```
pokey, line 20: As above

pokey (commented):
Good stuff. Left a few minor comments

Side note: not sure if you're a cursorless user, but I've been thinking terraform support in cursorless might be cool to experiment with because of all the repetition in terraform, so structurally aware multiple cursor operations could be pretty powerful in that context

apps/terraform/terraform.talon
2d6: Yeah, that was the reasoning behind not including the newline in the modifying calls. Do you think I should remove the newline from `plan` as well?

lang/terraform/terraform.py
2d6: Ah, I guess I don't yet understand the difference between `insert` and `auto_insert`. I tried to mimic other files, but it's mostly cargo cult configuration, to be honest. I could not discern the difference from the wiki. Could you give me a pointer so that I don't make mistaks in the future?

lang/terraform/terraform.talon
2d6: My intuition was that front-anchoring should lead to less false positives during speech recognition. I copied this from the `funky` declaration. Do you think it's a bad idea here?

2d6:
Thank you for the code review! I tried to incorporate your recommendations in the amended commit. Let me know of there's anything else :)

> Side note: not sure if you're a cursorless user, but I've been thinking terraform support in cursorless might be cool to experiment with because of all the repetition in terraform, so structurally aware multiple cursor operations could be pretty powerful in that context

I'm not a user of cursorless, yet. Actually hadn't heard of it until your comment. I definitely need to check it out. I'm still doing my first steps with voice control at the moment.

lang/terraform/terraform.py
2d6: Thank you for the heads up. Fixed this with an amended commit.
FYI: I copied this from the java bindings, where plural is (still/erroneously?) used.

lang/terraform/terraform.py
```diff
@@ -0,0 +1,156 @@
...
+        actions.key("left")
+
+    def code_terraform_resource(text: str):
+        result = "resource \"{}\" \"\"".format(
```
auscompgeek, line 72: nit: use single quotes to avoid escaping and make this more readable
```suggestion
        result = 'resource "{}" ""'.format(
```

lang/terraform/terraform.py
```diff
@@ -0,0 +1,156 @@
...
+            actions.user.formatted_text(
+                text, "ALL_LOWERCASE,SNAKE_CASE"
```
auscompgeek, line 74: Doesn't the `SNAKE_CASE` formatter already lowercase?

lang/terraform/terraform.talon
```diff
@@ -0,0 +1,27 @@
+mode: user.code_terraform
+mode: user.auto_lang
+and code.language: terraform
+-
+tag(): user.code_comment_block_c_like
+tag(): user.code_comment_line
+tag(): user.code_data_bool
+tag(): user.code_data_null
+tag(): user.code_imperative
+tag(): user.code_operators_assignment
```
auscompgeek, line 10: > I could not implement most of the mutating definitions in tag `user.code_operators_assignment`. I included this tag only for the assignment operator.

Maybe this is a sign we need to split out the augmented assignment operators?

lang/terraform/terraform.talon
```diff
@@ -0,0 +1,27 @@
+mode: user.code_terraform
```
auscompgeek, line 1: Need `mode: command` here too, see #603.

lang/terraform/terraform.py
splondike: The difference between them (which you can see by the docs available when you put `actions.auto_insert` in the REPL) is that auto_insert also runs the text through `auto_format`. That in turn reads like it's intended for prose. In practice, knausj_talon hasn't implemented auto_format, so they're identical in function.

I think auto_insert shouldn't be used in any of the programming files.

splondike (approved):
Looks good to me. Just need to add in that `mode:command` mentioned by auscompgeek.

apps/terraform/terraform.talon
auscompgeek: Unless you expect to run plan with arguments I think it's fine.

2d6: Thanks. I thought about that, but in my experience the parameterless call is the most used one.

lang/terraform/terraform.py
2d6: Thank you for the explanation! I'm going to replace all my use of `auto_insert` by `insert`.

2d6:
Thanks to all reviewers for your explanations and corrections. I learned a lot and hope this pull request is now up to par :)

lang/terraform/terraform.talon
pokey: Cc/ @wenkokke

wenkokke: Probably? I just have a fully separate command for definitional equality in declarative languages.

lang/terraform/terraform.talon
```diff
@@ -0,0 +1,28 @@
...
+mode: command
```
auscompgeek, line 2: Not quite correct... these consecutive `mode` conditions are OR'd together, so this file is now always active when in `command` mode or when a .tf file is open in any mode. Take a look at some of the other language files (or in #603).

2d6: Oh damn! Again I oriented myself by looking at the java language files. I'm going to submit a pull request soon to fix the issues found here which are also present in the java language files. I hope it's correct now.
Also I had made another error where I used the wrong mode (`user.code_terraform` instead of `user.terraform`). Good thing you pointed me in the right direction :grimacing:

pokey: Man we really gotta implement https://github.com/knausj85/knausj_talon/issues/603#issuecomment-980106400.  This language mode stuff is getting pretty krufty 😅

apps/terraform/terraform.talon
```diff
@@ -0,0 +1,11 @@
...
+terraform state move: "terraform state mv "
```
ViacheslavKudinov, line 11: @2d6 could you, please, also add `fmt` and `validate` commands as well?
Thanks in advance.

splondike:
@2d6 Looks good to me, and I suspect other folks. Are you happy to have it merged in, or do you have more changes planned?

2d6:
@splondike That's good to hear! I'd be more than happy to have this merged. Thank you :)

PR #753, opened 2022-01-31
Expose CAPITALIZE_FIRST_WORD as 'sentence' formatter

2d6:
Hey there! Just a tiny pull request for a functionality I personally found lacking so far. This exposes the `CAPITALIZE_FIRST_WORD` formatter for usage via speech. I found this useful e.g. for writing comments or log statements without changing modes to do so. Feel free to deny this PR if you don't think it fits into your vision for this repository, or if it causes some kind of collision I am not aware of.

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -151,6 +151,7 @@ formatters_words = {
     "kebab": formatters_dict["DASH_SEPARATED"],
     "packed": formatters_dict["DOUBLE_COLON_SEPARATED"],
     "padded": formatters_dict["SPACE_SURROUNDED_STRING"],
+    "sentence": formatters_dict["CAPITALIZE_FIRST_WORD"],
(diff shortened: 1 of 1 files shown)


rntz:
This collides with the "sentence" prose formatter, unfortunately, so it would become ambiguous which one the command "sentence hello period goodbye" would use - which is bad, because they have different behaviors wrt punctuation. I'm going to close this PR but feel free to reopen/open an issue if you want to either (a) rename your formatter or (b) solve the problem some other way.

AndreasArvidsson:
You might draw some inspiration from my word formatters

https://github.com/AndreasArvidsson/andreas-talon/blob/26ad3f9e46236357d771b8a743e2899d9ab18b60/text/formatters.py#L71

xavier630:
Hey, this is definitely something I'd like to have. Agreed that we should name it so that it doesn't collide with any existing formatters/commands

PR #754, opened 2022-02-01
Add c keywords

Sam-Hornby:
This common types dictionary was created but unused before. I've renamed it to c_keywords as that more accurately describes what was inside the dictionary (though it is not a complete list of the c keywords at the moment) .  I've then tried to plug it into the .talon file the same way the c_types dictionary is.

### DIFF
--- a/lang/c/c.py
+++ b/lang/c/c.py
@@ -30,7 +30,7 @@ ctx.lists["self.c_signed"] = {
     "unsigned": "unsigned ",
 }
 
-common_types = {
+ctx.lists["self.c_keywords"] = {
     "static": "static",
     "volatile": "volatile",
     "register": "register",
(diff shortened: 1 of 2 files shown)


Sam-Hornby:
Forgot to mention in the comment, to test this I just manually spoke out the words added to check it picked them up when I was editing a c program, and then checked they weren't picked up on master. If there is a way to test these types of changes I'd be happy to try and add that to the diff

splondike:
I don't think any of the keywords you registered collide with anything significant. 'register <blah>' is used by windbg, but that's not going to be active when this thing is.

It's a shame the commands in c aren't prefixed with state like most of the other languages, but that's how it is.

In terms of how to check, there is the [sim command](https://talon.wiki/unofficial_talon_docs/#introspection-functions), but that only matches entire commands. I think the only way to do it currently is to grep the codebase. It would be possible to make a system for searching for collisions I think, but nobody's done it yet.

Merging this in.

PR #756, opened 2022-02-04
Align java with other languages

2d6:
While submitting a pull request for terraform (#752), I got some valuable feedback on how to write language files. It turns out I made some mistakes because I copied a lot from the java definition files. This pull request aims to fix the following deviations:

- use singular for lists
- use `insert` instead of `auto_insert`
- restrict to command mode

Also, I removed an apparently unused function (`user.java_access_modifiers` appears to be undefined).

### DIFF
--- a/lang/java/java.py
+++ b/lang/java/java.py
@@ -35,8 +35,8 @@ java_boxed_types = {
     "Void": "Void",
 }
 
-mod.list("java_boxed_types", desc="Java Boxed Types")
-ctx.lists["self.java_boxed_types"] = java_boxed_types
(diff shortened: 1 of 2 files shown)


splondike:
Looks good to me. I think the only thing might be to change back to docstrings on the list to the plural. I'm not sure using the singular there reads that well.

2d6:
Thanks for the review. I wasn't sure what you meant by docstrings, but I restored the `desc` parameter in the list declaration to the plural form. I hope this matches your intention. Just let me know if there's anything else to fix.

splondike (approved):
Yeah, that's what I meant :). Merging.

PR #759, opened 2022-02-07
Fix duplicate newlines in CSVs on Windows.

wolfmanstout:
See footnote: https://docs.python.org/3/library/csv.html#id3

I can confirm that I witnessed this bug and this fixes it.


splondike (approved):
Still works fine for me on Linux, approved.

PR #762, opened 2022-02-08
override dictate.replace_words instead of defining user.replace_phrases

### LINKED ISSUE #750: Override dictate.replace_words instead of using user.replace_phrases
Per discussion with aegis, instead of using a new action user.replace_phrases in vocabulary.py we should just override dictate.replace_words, since we're (I think) compatible with it except for also supporting multi-word replacement. This means if talon grows native support for multi-word replacement we could just remove our implementation, eventually.

This should be pretty easy to do, just remember to replace the fallback code that invokes dictate.replace_words with actions.next instead.

rntz:
Fix #750.

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -60,7 +60,7 @@ def word(m) -> str:
     try:
         return m.vocabulary
     except AttributeError:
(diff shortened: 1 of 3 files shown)


splondike (approved):
Seems to work fine for me.

PR #763, opened 2022-02-08
simplify _capitalize_defaults

### LINKED ISSUE #760: Do we need _capitalize_defaults now that Talon lexicon includes capitalization?

rntz:
Partially fixes #760; removes everything except month names, since we don't need them anymore with conformer's lexicon. I haven't tested this on public, though. Can someone double check that the words I removed are capitalized properly on public when removed from words_to_replace.csv?

### DIFF
--- a/code/vocabulary.py
+++ b/code/vocabulary.py
@@ -10,31 +10,15 @@ ctx = Context()
 mod.list("vocabulary", desc="additional vocabulary words")
 
 
-# Default words that will need to be capitalized (particularly under w2l).
-# NB. These defaults and those later in this file are ONLY used when
+# Default words that will need to be capitalized.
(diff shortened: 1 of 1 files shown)


splondike:
I tried all of them out on 0.2.3+Conformer and it worked fine. I'm not sure if we need to consider the other voice engines though: Dragon, Vosk? I checked that webspeech does its own capitalisation. I guess at least Dragon does also? People would probably only be using Vosk for non-english text, and these words wouldn't be useful anyway.

I also checked this continues to work with the same with `mimic("say this is wednesday") -> "this is Wednesday"`, I guess that also uses the lexicon.

So I think this is fine?

splondike (approved):

PR #764, opened 2022-02-09
Add hook for recording of macro commands

splondike:
I want to use the existing voice commands for a more advanced macro system in my personal scripts (basically the macro gets automatically saved to a file as it is recorded). The addition of this 'hook' action allows me to do that.

### DIFF
--- a/code/macro.py
+++ b/code/macro.py
@@ -1,3 +1,4 @@
+from typing import List
 from talon import actions, Module, speech_system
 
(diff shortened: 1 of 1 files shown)


rntz:
I spruced up the docstring and added an assert to check that we are recording; lgtm.

rntz (approved):

splondike:
Thanks!

PR #765, opened 2022-02-09
Add protobuf support

purpleP:
Initial protocol buffers support


auscompgeek:
Heh, nice. I've also written some protobuf support myself. Looks like this is missing a lot of what I need. I'll leave a review once I get back to a computer.

lang/proto/proto.talon
```diff
@@ -0,0 +1,19 @@
...
+state package: "package "
+state reserved: "reserved "
+state enum: "enum "
+op equals: " = "
+state import: "import "
+state import public: "import public "
+state option: "option "
+state repeated: "repeated "
+
+<user.code_insert_type>$:
```
rntz, line 18: This should probably lose the end-anchor. It's also probably better to use the list of types directly rather than a weird capture, so you could do:

```
type {user.code_type}: "{code_type}"
repeated type {user.code_type}: "repeated {code_type}"
```

lang/proto/proto.py
```diff
@@ -0,0 +1,37 @@
...
+ctx.matches = r'''
+mode: user.protobuf
+mode: user.auto_lang
+and code.language: protobuf
+'''
+
+ctx.lists['user.code_type'] = {
+    'string': 'string',
+    'bytes': 'bytes',
+    'you 64': 'uint64',
```
rntz, line 15: Numbers in spoken forms don't work, you'll need to spell them out eg "you sixty four".

rntz (changes requested):

lang/proto/proto.talon
```diff
@@ -0,0 +1,19 @@
+mode: command
+and mode: user.protobuf
+mode: command
+and mode: user.auto_lang
+and code.language: protobuf
+-
+
```
auscompgeek, line 7: What about comments?

lang/proto/proto.py
```diff
@@ -0,0 +1,37 @@
+from talon import Module, Context, actions, ui, imgui, settings
+
+mod = Module()
+
+ctx = Context()
+ctx.matches = r'''
+mode: user.protobuf
```
knausj85, line 7: We've migrated to just using a tag for these matches as part of #776. Please update when you have a moment.

and

PR #769, opened 2022-02-11
git: add special commits, remove new line after rebase

iSchluff:
Add special cases for commit -a and commit --amend

Remove automatic line after rebase to be similar to merge. Generally rebase without arguments doesn't really make sense to me.


misc/git.talon
```diff
@@ -48,7 +50,7 @@ git push origin: "git push origin "
 git push up stream origin: "git push -u origin"
 git push <user.text>: "git push {text} "
 git push tags: "git push --tags\n"
-git rebase: "git rebase\n"
+git rebase: "git rebase"
```
auscompgeek, line 53: May as well add a space?

iSchluff: Good point

Issue #669, opened 2021-11-23
Generic Talon files and their Python files should live side-by-side


wenkokke:
I would like for knausj_talon to move towards the following organizational principle:

> If a Talon file uses abstract actions, there should be a Python file with the same name in the same directory which declares them and describes their intended behavior in a comment.

I've applied this principle to the programming language modes in #652 and #656, and I think it makes the code a lot more readable and discoverable for people who are not already intimately familiar with the code base.

I am opening this issue as a place to have a discussion about this.

rntz:
I generally agree. I'm not sure how rigidly to apply this principle, but I definitely feel like the weird splitting of some functionality between `misc/` and `code/` (eg. `misc/formatters.talon` vs `code/formatters.py`) is a mistake, and it would be easier to understand/navigate if they were in the same directory.

auscompgeek:
I'd like to go a bit further and suggest that the text related files should all go in `text/`. `formatters` is a good example.

wenkokke:
Would there be interest in small piecemeal pull requests to reorganize the repository along the lines that I have organized my own repository? See: <https://github.com/wenkokke/talon-user>

```sh
apps/           # contains Talon scripts associated with applications, e.g., Firefox;
core/           # contains Talon scripts which are always running;
tags/           # contains Talon scripts which are activated by a certain tag;
tags/code/      # contains Talon scripts which are associated with a programming language, e.g., Haskell;
tags/code/tags/ # contains Talon scripts which abstract over programming language features, e.g., comments.
```

I would be happy to submit these piece by piece—or in batches—to help the restructuring.

rntz:
Yes (or at least, I'm interested), although a lot of bikeshedding is likely to ensue :). We might want to hash out the broad strokes of the reorganization in advance, either in this thread or in a community video meeting. At a glance your repository's organization looks good and thorough but more deeply nested than I tend to like.

If it's just moving files around, and we can get everyone on board with the organization scheme in advance, doing it all at once would be OK, although piecemeal would also be fine and probably easier to coordinate. If you want to change significantly more than file organization it would be good to have a list of the intended changes, and maybe do them PR-by-PR.

wenkokke:
I think it tops out at 3 for `code/tags/XXX` or `apps/XXX/platforms`?

fidgetingbits:
I'd love to see this change. Happy with the layout you already have personally, but not picky if others want to change things. 3 levels deep is fine by me.

[...]

PR #771, opened 2022-02-12
Add support for moving and resizing windows

codecat555:
This is a big change, but should be independent of any other code in the repo. I'd be happy to host a walk through if that makes it easier to review/approve.


auscompgeek:
fwiw, see #669 - this might be a good candidate for everything to live together in a `window_management` directory?

window_management/window_tweak.talon
```diff
@@ -0,0 +1,121 @@
+
+# show size and position information about the current window
+win show$:
```
rntz, line 3: Perhaps a different command name for this? `win show` sounds like it will show a window that's hidden, as opposed to show positioning info.

window_management/window_tweak.talon
```diff
@@ -0,0 +1,121 @@
...
+    user.win_show()
+
+# hide the window information window
+win hide$:
+    user.win_hide()
+
+# move the current window toward the center of the screen
+win move$:
```
rntz, line 11: Why are all of these commands end-anchored? Might be useful to remove that.

nriley:
As a reference point — here's what I use to do similar stuff (via a third-party utility). https://github.com/nriley/knausj_talon/blob/nriley/apps/platforms/mac/moom/moom.talon. Would certainly not be averse to using knausj functionality to do similar things.

settings.talon
```diff
@@ -50,5 +50,33 @@ settings():
...
+    user.win_hide_resize_gui = 0
+
+    # How long to wait (in seconds) for talon to signal completion of window move/resize requests
+    user.win_set_queue_timeout = 0.2
+
+    # How many times to retry a timed out talon window move/resize request
+    user.win_set_retries = 1
+
+    # Whether to generate warnings for anomalous events.
+    user.win_verbose_warnings = 0
```
rntz, line 79: This is a lot of settings, it would be good to double-check if there's some way to merge or remove some. But maybe there isn't!

rntz:
We took a look at this during the session today. It looks like it might need some further testing but definitely should be useful.

codecat555:
Here are the demo videos we talked about during the grooming meeting today. I'll follow-up on the comments above in the coming days. Thanks for your time!

https://youtu.be/q4gm839KhqY
https://youtu.be/EyMvwUyZN5k

window_management/window_tweak.talon
codecat555: Changed to 'window traits show' and 'window traits hide' - used 'traits' to keep the number of syllables to a minimum.

window_management/window_tweak.talon
codecat555: Fixed.

settings.talon
codecat555: I split the move and resize settings because, on my system at least, the speeds differed even though the underlying code/mechanisms were the same.

Most of these settings relate to what I call the 'continuous' commands - those commands which keep running until stopped. That is, for example, the user says 'window move'...the window starts moving and continues until the user says 'window stop'. Also, a pop-up appears showing the status and how to stop it. Very similar to how 'wheel upper' and 'wheel stop' work.

I decided to expose the values governing these 'continuous' operations as settings because it's not clear how the behavior might differ between systems (mac, windows, linux) and various hardware configurations (system speed, multiple screens of differing sizes, etc).

codecat555:
Sorry about the demo problems yesterday. I rebooted my machine and everything is back to normal; that is, the code is working as expected.

I recorded a video of my testing but it's too big to attach here. The original videos I list above are better anyways, because they explain some nuances of how the commands work.

There are some known quirks which are documented in the top of the .py files. These are all what I consider 'minor' bugs, where a valid API request produces an unexpected result for certain corner cases, like when part of a window overlaps the edge of the screen.

The one exception, IMO, is where Talon may sometimes fail to send an event notifying of a change in window size or position. As mentioned yesterday, I worked with Ryan to characterize this bug and it is tracked here - https://github.com/talonvoice/talon/issues/470.

knausj85:
I'll have to poke it this a bit later,  but I often see "ValueError: screen not found containing point" errors like the below when using the window move to pointer commands on my Mac.  I'm not sure if this is on the talon side or script side

```
2022-03-13 14:53:01 ERROR    12:                        talon/scripting/talon_script.py:590 | 
   11:                        talon/scripting/talon_script.py:297 | 
   10:                        talon/scripting/talon_script.py:557 | 
    9:                             talon/scripting/actions.py:78  | 
    8:                               talon/scripting/types.py:390 | 
    7:    user/knausj_talon/window_management/window_tweak.py:410 | compass_control.mover.resize_to_pointe..
    6: user/knausj_talon/window_management/compass_control.py:704 | return self.move_absolute(rect, rect_i..
    5: user/knausj_talon/window_management/compass_control.py:691 | result, rect = self.compass_control.se..
    4: user/knausj_talon/window_management/compass_control.py:1663| result = set_method(old_rect, rect_id,..
    3:    user/knausj_talon/window_management/window_tweak.py:259 | w.rect = rect_in.copy()
    2:                                        talon/mac/ui.py:292 | 
    1:                                        talon/screen.py:286 | 
ValueError: screen not found containing point -430, -375

    2: talon/mac/ui.py:281| 
    1: talon/screen.py:286| 
ValueError: screen not found containing point -430, -375

[The below error was raised while handling the above exception(s)]
2022-03-13 14:53:01 ERROR cb error topic="phrase" cb=<bound method SpeechSystem.engine_event of <talon.scripting.speech_system.SpeechSystem object at 0x10ff7e680>>
   30:       lib/python3.9/threading.py:930* # cron thread
   29:       lib/python3.9/threading.py:973* 
   28:       lib/python3.9/threading.py:910* 
   27:                    talon/cron.py:155| 
   26:                    talon/cron.py:106| 
   25:                     talon/vad.py:21 | 
   24: talon/scripting/speech_system.py:355| 
   23:             talon/engines/w2l.py:705| 
   22:      talon/scripting/dispatch.py:105| 
   21:      talon/scripting/dispatch.py:144| 
   20:      talon/scripting/dispatch.py:133| 
   19:          talon/scripting/rctx.py:233| # 'phrase' main:_redispatch()
   18: talon/scripting/speech_system.py:62 | 
   17:      talon/scripting/dispatch.py:105| 
   16:      talon/scripting/dispatch.py:144| 
   15:      talon/scripting/dispatch.py:133| 
   14: ------------------------------------# cron thread
   13:          talon/scripting/rctx.py:233| # 'phrase' main:engine_event()
   12: talon/scripting/speech_system.py:407| 
   11:       talon/scripting/actions.py:78 | 
   10:         talon/scripting/types.py:390| 
    9:     talon/scripting/core/core.py:100| 
    8:       talon/scripting/actions.py:78 | 
    7:         talon/scripting/types.py:390| 
    6:     talon/scripting/core/core.py:129| 
    5:       talon/scripting/actions.py:78 | 
    4:         talon/scripting/types.py:390| 
    3:     talon/scripting/core/core.py:134| 
    2:  talon/scripting/talon_script.py:691| 
    1:  talon/scripting/talon_script.py:594| 
 in script at /Users/knausj/.talon/user/knausj_talon/window_management/window_tweak.talon:52:
 > user.win_move_to_pointer(compass_direction)
ValueError: screen not found containing point -430, -375
```

codecat555:
Interesting. When that happens, is the pointer on the same screen as the active window?

Oh - I just tested 'move to pointer' across screens on my windows system and it works fine. I'll get set up to test on linux a bit later.

window_management/compass_control.py
```diff
@@ -0,0 +1,2052 @@
...
+        if dx > dy:
+            xx, xy, yx, yy = xsign, 0, 0, ysign
+        else:
+            dx, dy = dy, dx
+            xx, xy, yx, yy = 0, ysign, xsign, 0
+
+        D = 2*dy - dx
+        y = 0
+
+        for x in range(dx + 1):
```
knausj85, line 1881: rects will contains floats on Mac, so this will presently crash here when using the continuous commands. May be easiest just to cast dx/dy

window_management/window_tweak.py
```diff
@@ -0,0 +1,525 @@
...
+    gui.spacer()
+
+    gui.text(f"Width: {round(width)}")
+    gui.text(f"Height: {round(height)}")
+
+    gui.line()
+
+    gui.text(f"== Mouse ==")
+
+    gui.text(f"Position: {ctrl.mouse_pos()}")
```
knausj85, line 134: On Mac, ctrl.mouse_pos() will provide floats. We'll probably want to format this to 2 decimals places or something. Otherwise, the window will probably be unusable on Mac as the mouse moves

knausj85:
> Interesting. When that happens, is the pointer on the same screen as the active window?
> 
> Oh - I just tested 'move to pointer' across screens on my windows system and it works fine. I'll get set up to test on linux a bit later.

@codecat555 I only have 1 monitor. :-)

window_management/compass_control.py
codecat555: That was fast - thanks! I think the change I just pushed covers it, can you check?

codecat555:
@knausj85 Are the coordinates showing in the error (-430, -375) valid for your screen? Seems odd they would be negative when you only have one screen...perhaps there's another math/rounding error.

knausj85:
@codecat555 No, those coordinates are not valid
```
>>> ui.screens()[0].rect
Rect(0.0, 0.0, 1920.0, 1080.0)
```

codecat555:
@knausj85 Oh, that's a different case than I was considering before.

That is an explicit request to move the window up off the screen, i.e. move the center point of the 'south' border of the window up to the pointer position.

I was thinking that perhaps there was some kind of math or rounding error in translate_top_left_by_region(), but now I suspect that part is working properly and the error is due to differences in how talon handles mac screens versus windows screens.

By design, the continuous commands restrict window movements to stay within the boundaries of the screen and the other commands let the user do whatever they want.

window_management/window_tweak.py
```diff
@@ -0,0 +1,525 @@
...
+        """
+        def win_stop() -> None:
+            "Stops current window move/resize operation"
+            compass_control.continuous_stop()
+
+app.register("ready", on_ready)
+
+@mod.action_class
+class Actions:
+    def win_show() -> None:
```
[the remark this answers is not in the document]
codecat555: Yes, sounds right - and done.

fidgetingbits (approved):
I think this is probably fine as most (all?) people's concerns seem to have been fixed, and seems useful to get in. That said I don't have a way to test this because I don't use a relevant window manager, I'm hesitant to merge it myself.

codecat555:
Thanks @fidgetingbits. I think @knausj85 is still investigating a mac bug he found, and may have additional feedback as well.

codecat555:
@knausj85 Does it make sense to file that mac exception as a bug and have this code just return failure in that case? I'm going to push a change along those lines, for your consideration.

nriley:
Let me know when things are fixed on the Mac side and I can do some more testing; from what I read above it sounds like there is not much point at the moment!

codecat555:
@nriley Actually, that bug is specific to operations that move the window off the edge of the screen and I pushed a change that silently fails the operation. More testing is always good, if you have the time and inclination.

codecat555:
I'm going to revert the commit I made to ignore the mac bug that @knausj85 reported. I don't like masking errors unless the result improves the user experience in some way. I had originally thought someone might suggest an alternate message to log, but the talon exception is descriptive enough and contains a stack trace pinpointing the error path.

I would file a talon bug for this, but I don't have a mac to test with and can't respond to any questions.

knausj85:
I'd like to poke at this a bit more to see it if can be made a more reliable on Mac before merging, though time's a bit short at the moment. I really like the functionality.

auscompgeek:
Huh, looks like pre-commit's `mixed-line-ending` should go after `end-of-file-fixer` (see [run on rntz's commit](https://results.pre-commit.ci/run/github/240185541/1656769118.GoCMui3BQuSSygOmS2cUPg) then [run on the auto fix commit](https://results.pre-commit.ci/run/github/240185541/1656769167.ZdVGdHTlR0W_0xKHrkDJRw)). I'll raise a PR to fix that.

window_management/window_tweak.py
```diff
@@ -0,0 +1,617 @@
...
+
+        w = ui.active_window()
+
+        compass_control.snap(w.rect, w.id, w.screen.visible_rect, percent, direction)
+
+    def win_revert() -> None:
+        "Restore current window's last remembered size and position"
+
+        w = ui.active_window()
+        compass_control.revert(w.rect, w.id)
```
rntz, line 611: weird to be passing `w.rect, w.id` instead of just the window `w`

window_management/window_tweak.py
```diff
@@ -0,0 +1,617 @@
...
+setting_verbose_warnings = mod.setting(
+    "win_verbose_warnings",
+    type=bool,
+    default=False,
+    # window move and resize requests are not guaranteed
+    desc="Whether to generate warnings for anomalous events.",
+)
+
+
+@imgui.open(y=0)
```
nriley, line 106: Can this be displayed on the same screen as the window being moved?

window_management/compass_control.py
```diff
@@ -0,0 +1,2680 @@
...
+
+mod = Module()
+
+# # context used to enable/disable the tag for controlling whether the 'stop' command is active
+ctx = Context()
+
+# taken from https: //talon.wiki/unofficial_talon_docs/#captures
+#
+# a type for representing compass directions
+Direction = dict[str, bool]
```
rntz, line 93: this seems like a somewhat cumbersome way to represent compass directions? not sure what a better one is since I haven't read all of the code that uses it.

rntz (commented):
My main reservation is that this PR in general and compass_control.py in particular is very large (2000 lines). Is there any way it could be simplified? It doesn't seem like something this large should be necessary for what you're trying to do and I'm hesitant to take on maintenance of something this large for features I'm not sure how many people will use.

If you can't reduce the code size a lot, it would at least clarify where the complexity is coming from if you split components out into separate files. For instance, a lot seems to be coming from the continuous movement, which is a feature I'm not sold on. It would also help to remove any code for testing that isn't really needed.

## GENERAL COMMENTS

Probably want to support "left/right" as synonyms for "east/west".

In my testing on Ubuntu, `window move` appeared to move at roughly half the speed of `window move north/south/east/west`. Not sure why.

## COMMANDS

*window traits show/hide*: This is fine. Seems like it could easily be separated into its own file. The 'close' button should be named "Window traits hide"; knausj convention is to label the button with the command that pushes it. 

*window move to pointer*, *window move <user.compass_direction> to pointer*, *window size <foo> to pointer*: these seem useful! might want to have it be eg. 'window move {top,bottom,left,right,top left,...} to pointer' or similar instead of 'window move north to pointer', a bit more intuitive.

**continuous move/stretch commands**, eg *"window {move,stretch} <direction>"*: Not sold on this being useful. (Maybe pop-to-stop could improve this? But not sure how to make sure this doesn't step on the toes of pop-to-click in knausj.) It's not easy to time "window stop" to get the window where you want. It seems like what we already have in window_management.talon (snapping to corners of screen, sizing to a half/third/etc) seems more useful.

**Percentage/pixel-based commands**, eg *"window move <direction> <number> percent"*, *"... pixels"*: These seem okay but in most circumstances overly precise. Do you have motivating examples for when you'd want these?

codecat555:
I'll address the specific comments first, and then follow-up with a response to the more general discussion.

window_management/window_tweak.talon
```diff
@@ -0,0 +1,121 @@
...
+# show size and position information about the current window
+window traits show:
+    user.win_show_info()
+
+# hide the window information window
+window traits hide:
+    user.win_hide_info()
+
+# move the current window toward the center of the screen
+window move:
```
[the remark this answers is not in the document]
codecat555: `window move center` is already explicitly supported (in the following command).

This unqualified variation is just there to provide a 'default direction' and 'center' seemed the natural choice.

window_management/window_tweak.py
```diff
@@ -0,0 +1,526 @@
...
+)
+
+@imgui.open(y=0)
+def win_stop_gui(gui: imgui.GUI) -> None:
+    gui.text(f"Say 'window stop' or click below.")
+    gui.line()
+    if gui.button("Stop moving/resizing"):
+        actions.user.win_stop()
+
+@imgui.open(x=2100, y=40)
```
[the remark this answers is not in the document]
codecat555: Oops, that's a personal setting that should not have made it into this copy.

window_management/compass_control.py
```diff
@@ -0,0 +1,2680 @@
...
+        self.last_rect: dict = dict()
+
+        # tag used to enable/disable commands used to manage continuous move/resize operations, e.g. 'stop'
+        self.continuous_tag_name: str = continuous_tag_name
+        self.continuous_tag_name_qualified: str = "user." + self.continuous_tag_name
+
+        # the method to call when stopping a continuous move/resize operation, so the caller can do any additional tasks, as required
+        self.continuous_stop_method = stop_method
+
+        self.continuous_mutex: threading.RLock = threading.RLock()
```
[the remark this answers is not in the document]
codecat555: Yes, in the case of the continuous commands, multiple threads may execute this code simultaneously. It really depends on Talon internals, which are variable and opaque.

window_management/compass_control.py
```diff
@@ -0,0 +1,2680 @@
...
+
+        self.settings_map = settings_map
+        self.refresh_map = {
+            talon_setting.path: local_name
+            for local_name, talon_setting in settings_map.items()
+            if hasattr(self, local_name)
+        }
+        self.refresh_settings()
+        # catch updates
+        settings.register("", self.refresh_settings)
```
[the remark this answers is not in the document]
codecat555: I understand your point, but I also note that this implementation matches the only existing example in knausj_talon: https://github.com/knausj85/knausj_talon/blob/5b1a7372b20c8ff90560a00e38fb69e3dc6f6e71/talon_draft_window/draft_talon_helpers.py#L67.

window_management/compass_control.py
codecat555: Perhaps, not sure I agree. In any case, I just lifted that from the Talon wiki, as indicated by the comment at line 90.

codecat555:
@rntz - Thanks for your feedback. I'm hearing that you just don't like this approach - it's too big and complex for knausj_talon.

That's understandable, although I note that @knausj85 wrote 'I really like the functionality' in the comments above (and also in some personal communications).

I want to make a distinction between simplifying the code and simplifying the design. Sure, there are some ways the code could be simplified, but I don't think it would produce the result you want. I could break things out into more files and remove the debug code, but I think what you really want is a simpler design rather than a different implementation. So, let me respond to that point.

I wrote this code to solve a problem for myself and to develop my understanding of how to work with the Talon API. I approached the implementation from a top-down perspective - before writing any code, I decided on my design goals and how to achieve them. I defined the command set based on actual use cases I expected to encounter and similarly decided to split the code into two modules (specific window management functionality layered on top of generic rectangle handling).

I thought this submission would be appropriate for knausj_talon because it is of general utility and because there is nothing there that already provides the functionality.

The window snap commands are not sufficient. Sure, some people may be happy with just those commands and nothing else, but then some people also like tiled window managers (I don't). Personally, I want to be able to use my voice to do anything that I could otherwise do with mouse and keyboard alone - and that includes moving and sizing windows with precision.

There is a middle ground between basic window snap commands and the full window management functionality provided by this module. That seems to be what you want and I suppose it may be the best choice given the guidelines within which you are working.

I don't know what specific goals have been set for knausj_talon development. I do know that I personally would like it to be my repository of _professional-grade_ voice control tools.

Among other things, that means supporting professional use cases, such as precise positioning at various levels of granularity - something important to anyone doing graphic design. This makes even more sense when you consider applying compass_control.py to applications other than window management.

Ideally, from my perspective, knausj_talon code would allow for casual (i.e. simple) usage while, at the same time, also providing for more rigorous applications. Hence, we have 'window move' (simple) AND 'window move south 327 pixels' (precise).

codecat555:
At this point (after 6 months in review), I think it's safe to say this code is not a good fit for knausj_talon at this time. I don't have any time or appetite for splitting it up and submitting smaller chunks. I'm pretty happy with the way it is now.

Anyways, this code has no dependencies that require it to live in knausj_talon and anyone who wants it can get it from my personal repo: https://github.com/codecat555/talon-window-tweak. So, unless I hear otherwise, I'll just close out this pull request after the next grooming meeting.

pokey:
Yeah I'd be tempted to have this be a separate standalone user file set that we add to some kind of "awesome Talon" list, like eg the souped up mouse grid.  I believe we have a couple of these lists lying around cc/ @tararoys @2shea

PR #776, opened 2022-02-14
Updated language context to use tags instead of modes 

AndreasArvidsson:
Instead of the following header
```
mode: command
and mode: user.c
mode: command
and mode: user.auto_lang
and code.language: c
```

We can now just use
```
tag: user.c
```

Also took the liberty to add a few new file endings when I was in there.

@rntz @knausj85 @pokey 
Would probably be nice to get this one merged before the other pr regarding new languages

PR #777, opened 2022-02-16
C++ support

dgrunwald:
This initial version started out as a copy of the C mode, but has some major new features:
  * commands for many C++ keywords
  * list of standard type names (`type standard vector`)
  * list of standard function names (`call standard move`)
  * a type name parser with [cdecl](https://cdecl.org/)-like features:
     * e.g. `pointer to function returning pointer to array of int` turns into `int(*(*)())[]`
  * `include string` will insert `#include <string>` below the existing includes

### DIFF
--- a/lang/cplusplus/cplusplus.py
+++ b/lang/cplusplus/cplusplus.py
@@ -63,6 +63,12 @@ ctx.lists["self.cpp_simple_types"] = {
     "typename": "typename",
 }
 
+# Words that trigger the standard namespace logic
+ctx.lists["self.cpp_standard"] = {
+    "standard": "std",
+    "stood": "std",
+}
+
 # Types in std namespace
 ctx.lists["self.cpp_standard_types"] = {
     # Language Support / Utility Library
@@ -606,6 +612,7 @@ mod.list(
     "cpp_type_bit_width", desc="Typical bit-width of C++ fixed width integer types"
 )
 mod.list("cpp_simple_types", desc="Common C/C++ types")
+mod.list("cpp_standard", desc="Words that trigger the standard namespace logic")
 mod.list("cpp_standard_types", desc="C++ types in namespace std")
 mod.list(
     "cpp_user_types",
@@ -665,8 +672,8 @@ def cpp_fixed_width_integer_type(m) -> str:
     rule=rm_newlines(
         """
             [{self.cpp_type_sign_modifiers}] {self.cpp_simple_types}
-          | standard {self.cpp_standard_types}
-          | standard <self.cpp_fixed_width_integer_type>
+          | {self.cpp_standard} {self.cpp_standard_types}
+          | {self.cpp_standard} <self.cpp_fixed_width_integer_type>
           | {self.cpp_user_types}
           | <self.cpp_fixed_width_integer_type>
         """
@@ -674,7 +681,7 @@ def cpp_fixed_width_integer_type(m) -> str:
 )
 def cpp_simple_type(m) -> str:
     "Returns a string"
-    if m[0] == "standard":
+    if m[0] == "std":
         return "std::" + m[1]
     else:
         return " ".join(list(m))
@@ -831,16 +838,16 @@ def cpp_unqualified_standard_generic_type(m) -> str:
     rule=rm_newlines(
         """
         {user.code_functions}
-        | standard {user.cpp_standard_functions}
-        | standard {user.cpp_standard_range_algorithms}
-        | standard ranges {user.cpp_standard_range_algorithms}
+        | {user.cpp_standard} {user.cpp_standard_functions}
+        | {user.cpp_standard} {user.cpp_standard_range_algorithms}
+        | {user.cpp_standard} ranges {user.cpp_standard_range_algorithms}
         | {user.cpp_user_functions}
         """
     ),
 )
 def code_functions(m) -> str:
     """Returns a function name"""
-    if m[0] == "standard":
+    if m[0] == "std":
         if m[1] == "ranges":
             return "std::ranges" + m[2]
         else:
--- a/lang/cplusplus/cplusplus.py
+++ b/lang/cplusplus/cplusplus.py
@@ -1,0 +1,1132 @@
+from talon import Context, Module, actions, settings
+import re
+
+mod = Module()
+
+ctx = Context()
+ctx.matches = r"""
+tag: user.cplusplus
+"""
+
+ctx.lists["self.cpp_pointers"] = {
+    "pointer": "*",
+    "reference": "&",
+    "R value reference": "&&",
+    "universal reference": "&&",
+    "forwarding reference": "&&",
+    # C++/CLI extensions:
+    "managed pointer": "^",
+    "tracking reference": "%",
+}
+
+# Type modifiers and qualifiers
+ctx.lists["self.cpp_type_sign_modifiers"] = {
+    "signed": "signed",
+    "unsigned": "unsigned",
+}
+ctx.lists["self.cpp_type_qualifiers"] = {
+    "const": "const",
+    "volatile": "volatile",
+}
+ctx.lists["self.cpp_type_bit_width"] = {
+    "eight": "8",
+    "sixteen": "16",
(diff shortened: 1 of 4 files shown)


dgrunwald:
Note: I'm not sure if a C++ mode separate from C mode makes much sense -- only one of the modes can be active for `.h` files, so it might be better to merge both back together into a single mode.

lang/cplusplus/cplusplus.talon
```diff
@@ -0,0 +1,146 @@
+mode: command
+and mode: user.cplusplus
+mode: command
+and mode: user.auto_lang
+and code.language: cplusplus
```
knausj85, line 5: We've migrated to just using a tag for these matches as part of #776. Please update when you have a moment.

and

ym-han:
> Note: I'm not sure if a C++ mode separate from C mode makes much sense -- only one of the modes can be active for `.h` files, so it might be better to merge both back together into a single mode.

Maybe forcing (or whatever the analogue of that currently is in knausj) can help with that? I.e., have the automatic default for .h be either C or C++, and then have the user manually set the language if they want to use the other one.

lang/cplusplus/cplusplus.talon
dgrunwald: Rebased + updated for the new tags.

lang/cplusplus/cplusplus.talon
```diff
@@ -0,0 +1,142 @@
...
+type <user.cpp_type>: "{cpp_type} "
+# Alternatively, words like "const", "pointer" or "standard" also start type-entry mode.
+<user.cpp_type_prefix> <user.cpp_raw_type>:
+    ty = user.cpp_build_declarator_with_prefix(cpp_type_prefix, cpp_raw_type, '')
+    insert("{ty} ")
+{user.cpp_standard} <user.cpp_unqualified_standard_generic_type>: "std::{cpp_unqualified_standard_generic_type} "
+# 'auto' is common enough that we allow skipping the "type" prefix
+auto: "auto "
+call <user.code_functions>:
+    user.cpp_insert_call("{code_functions}")
```
rntz, line 102: I think this is normally handled by the `funk` command in lang/tags/functions_gui.talon. No need to make a second version here.

lang/cplusplus/cplusplus.py
```diff
@@ -0,0 +1,1139 @@
...
+    "float": "float",
+    "size T": "size_t",
+    "size type": "size_t",
+    "bool": "bool",
+    "boolean": "bool",
+    "typename": "typename",
+}
+
+# Words that trigger the standard namespace logic
+ctx.lists["self.cpp_standard"] = {
```
rntz, line 67: I'm not sure whether using a list here is the right thing. You could instead use `(standard|stood)` everywhere this list is used. I don't feel strongly about this, but it is a bit unusual.

rntz (commented):
I haven't read this whole PR in detail, as it's quite long, but it looks like it's got a lot of useful functionality in it! I think the main blocking issue is whether we want to maintain separate C and C++ modes or merge them.

I think it would be better to just have one mode for C/C++, as they're so similar. I guess the argument for two modes is that there's loads of stuff in C++ that doesn't make sense in C. But there's enough overlap that maintaining two entirely separate modes would be too bad, I think. Concretely, there's a lot of duplication between c.talon and cplusplus.talon (eg. the "state include" and "state define" commands, the `size_t` and `ptrdiff_t` types, functions like `memcpy`).

I think there are two reasonable approaches here:

1. Unify C and C++ mode, and just accept that it'll be possible to say things in C mode that produce C++ code rather than C code. I'm in favor of this: it's simpler, it avoids the ".h" problem, and unless misrecognitions are common it won't produce problems for the user, because they just won't say the C++ commands if they're editing C.

2. Get C++ mode to "extend" the C mode somehow. A lot of the difference is just that C++ has bigger lists (eg. many more types, libraries, etc). We could, for instance, combine c.py and cplusplus.py and define each C list first, then add the things for C++ and put them in the corresponding list, something like:

   ```python
   ctx = Context()
   c_types_list = { "character": "char", ... }
   ctx.lists["c_types"] = c_types_list
   cpp_types_list = c_types_list.copy().update({ 
       "unique pointer": "unique_ptr<>",... })
   ctx.lists["cpp_types"] = cpp_types_list
   ```

   Then we could arrange that being in C++ mode also activates the user.c tag or something like that. I dislike this option; it's more complicated and it doesn't handle `.h` files. But if we really want to treat C and C++ as separate this would at least avoid duplicating all the common functionality.

What do you think?

lang/cplusplus/cplusplus.py
dgrunwald: It's meant as a customization point since everyone pronounces `std` differently (https://twitter.com/jfbastien/status/1183894210971193345)

auscompgeek:
> Get C++ mode to "extend" the C mode somehow.

My idea for this would be to have a common context for all the shared implementations between C and C++.

```python
common_ctx = Context()
common_ctx.matches = """
tag: user.c
tag: user.cplusplus
"""
```

From there C can set its lists (and maybe any actions that don't make sense for C++), and C++ can have its own context that sets lists and adds a bunch of its own actions if need be.

lang/cplusplus/cplusplus.py
fidgetingbits: As someone that uses C all the time but will never use C++, I'd much rather them be separate or I guess preferably C++ be an extension in different files. It makes it easier to quickly find/extend C stuff, and as I am plagued by misrecognitions in general, the less potential mismatching the better.

rntz:
@fidgetingbits That is an unfortunate set of constraints, since the easiest way to have one mode extend the other would be to put them in the same python file. Otherwise it becomes hard to have the C++ lists extend those from C. Any ideas?

lang/cplusplus/cplusplus.talon
dgrunwald: Interesting, I didn't realize this.
I've been using cursorless's `call <user.letter>`. Like cursorless `call`, my `call` will wrap an existing expression with the call, e.g. in `take vest call standard move` to turn `auto x = var;` into `auto x = std::move(var);`.
`funk` can't do that (yet?).

dgrunwald:
Note that personally, I use the C++ mode for all `.cpp .c .h` files -- we actually have a part of our code base that uses `.c` file extensions (to simplify merging patches from a third-party vendor) but is being compiled as `.cpp`.
So I won't be working on the C vs. C++ modes.

fidgetingbits:
> @fidgetingbits That is an unfortunate set of constraints, since the easiest way to have one mode extend the other would be to put them in the same python file. Otherwise it becomes hard to have the C++ lists extend those from C. Any ideas?

As per discussed in one of the grooming meetings, I think it's fine to share the same python file to avoid the duplication

lang/cplusplus/cplusplus.talon
rntz: `funk wrap` does this.

AndreasArvidsson:
Personally I would probably make two separate files `c.py` and `cpp.py` in the same folder and then use relative import. That is what I do for typescript/javascript.

https://github.com/AndreasArvidsson/andreas-talon/blob/d1b08445e23383ee5b1659dd37a88d54be7bfb57/languages/javascript/typescripts.py#L2
https://github.com/AndreasArvidsson/andreas-talon/blob/d1b08445e23383ee5b1659dd37a88d54be7bfb57/languages/javascript/typescripts.py#L40
https://github.com/AndreasArvidsson/andreas-talon/blob/d1b08445e23383ee5b1659dd37a88d54be7bfb57/languages/javascript/javascript.py#L61-L95

nriley:
From the community meetup — we decided that this should likely be turned into (community) snippets for all of the complex control structures and so forth and just use individual actions for operators.

knausj85:
Converting to draft until someone has time to pick this up

nriley:
From the community backlog session we decided that there has been enough change to the "right way" to implement a language with snippets and operators that it may make sense to start from scratch in the future.

PR #781, opened 2022-02-20
replace insert_cursor with insert_between

rntz:
Fixes a few things:

1. insert_cursor was in `lang/python/python.py`, which is a weird place. This moves it to `code/insert_between.py`.

2. insert_cursor looks for a magic symbol, `[|]`, to determine where to put the cursor. This is fine for inserting constant strings but means we can't use it more widely. I've replaced it with `insert_between(before, after)`, which takes two strings and doesn't look for magic symbols.

3. Updated lots of code to use insert_between instead of manually calling edit.left() after inserting stuff.

One of the nice things about `insert_between` is that it might eventually be possible to implement it more efficiently (eg via accessibility APIs), avoiding the repeated edit.left()s.

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -329,10 +329,7 @@ class Actions:
         # so that future dictation is properly formatted.
         text = text.replace("“", "\"").replace("”", "\"")
         actions.user.add_phrase_to_history(text)
-        # we insert the text all at once in case we have an implementation of
-        # insert that is more efficient for long strings, eg. paste-to-insert
-        actions.insert(text + (" " if add_space_after else ""))
-        if add_space_after: actions.edit.left()
+        actions.user.insert_between(text, " " if add_space_after else "")
 
     def dictation_peek_left(clobber: bool = False) -> Optional[str]:
         """
--- a/code/insert_between.py
+++ b/code/insert_between.py
@@ -5,7 +5,7 @@ mod = Module()
 @mod.action_class
 class module_actions:
     def insert_between(before: str, after: str):
-        """Insert `before + after`, leaving cursor between `before` and `after`."""
+        """Insert `before + after`, leaving cursor between `before` and `after`. Not entirely reliable if `after` contains newlines."""
         actions.insert(before + after)
         for _ in after: actions.edit.left()
 
--- a/code/insert_between.py
+++ b/code/insert_between.py
@@ -1,4 +1,4 @@
-from talon import Module, actions
+from talon import Module, actions, app
 
 mod = Module()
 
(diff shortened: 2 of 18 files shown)


code/dictation.py
```diff
@@ -329,10 +329,7 @@ def dictation_insert(text: str, auto_cap: bool=True) -> str:
         # so that future dictation is properly formatted.
         text = text.replace("“", "\"").replace("”", "\"")
         actions.user.add_phrase_to_history(text)
-        # we insert the text all at once in case we have an implementation of
-        # insert that is more efficient for long strings, eg. paste-to-insert
-        actions.insert(text + (" " if add_space_after else ""))
-        if add_space_after: actions.edit.left()
+        actions.user.insert_between(text, " " if add_space_after else "")
```
phillco, line 332: Thanks, this will allow us to override this for accessibility dictation a bit more cleanly.

code/insert_between.py
```diff
@@ -0,0 +1,18 @@
...
+        actions.insert(before + after)
+        for _ in after: actions.edit.left()
+
+    # # This is deprecated, please use insert_between instead.
+    # def insert_cursor(text: str):
+    #     """Insert a string. Leave the cursor wherever [|] is in the text"""
+    #     if "[|]" in text:
+    #         actions.user.insert_between(*text.split("[|]", 1))
+    #     else:
+    #         actions.insert(text)
```
phillco, line 18: Maybe leave it in but call `app.notify` to tell the user it is deprecated? That's what was done with e.g. the older Slack actions that used verb-now

lang/talon/talon.talon
```diff
@@ -17,9 +17,7 @@ tag(): user.code_functions_gui
 dot talon: insert(".talon")
 #defintion blocks for the context
 action block:
```
auscompgeek, line 19: Heh, we can drop this command now right?

lang/python/python.talon
```diff
@@ -38,7 +38,7 @@ self taught: "self."
 pie test: "pytest"
 state past: "pass"
 
-raise {user.python_exception}: user.insert_cursor("raise {python_exception}([|])")
+raise {user.python_exception}: user.insert_between("raise {python_exception}(", ")")
```
auscompgeek, line 41: Would it be worth just dropping the parens for this? Sometimes I do want to just `raise NotImplementedError` for example without the visual noise.

lang/talon/talon.talon
rntz: Yeah, that's an issue for a separate commit/PR though.

lang/python/python.talon
rntz: Maybe? I don't want to change any actual behavior with this PR though. Submit a PR if you want.

code/insert_between.py
splondike: Yeah, maybe so. It looks like at least fidget is using it a fair bit in a non-Python context: https://search.talonvoice.com/search/?q=insert_cursor&fold_case=auto&regex=false&context=true

splondike:
This is a nice improvement. I think it's probably worth keeping in an app.notify version of the [|] command (with a deprecation timeout comment) as suggested by phillco.

code/insert_between.py
fidgetingbits: I'll make a point of switching all of mine as soon as this goes in.

PR #784, opened 2022-02-23
voice commands for reporting bugs

tararoys:
Added two commands that print out urls to the Talon and Knausj
issue trackers to encourage bug reporting.  Right now the link to
the Talon issue tracker is hidden as a pinned message on the help
channel of the slack.  Attaching it to the command 'report Talon bug'
may make it easier to find.

### DIFF
--- a/misc/talon_helpers.talon
+++ b/misc/talon_helpers.talon
@@ -10,7 +10,7 @@ talon copy name:
 talon copy executable:
     executable = app.executable()
     clip.set_text(executable)
-talon copy bundle:
+talon copy bundle:[Talon](talonvoice.com)
     bundle = app.bundle()
     clip.set_text(bundle)
 talon copy title: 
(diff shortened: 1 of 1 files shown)


auscompgeek:
Would this be better in websites.csv?

misc/talon_helpers.talon
```diff
@@ -57,3 +57,7 @@ talon dump context:
 ^talon copy active app$:
     result = user.talon_get_active_application_info()
     clip.set_text(result)
+
+^report talon bug$: 'https://github.com/talonvoice/talon/issues'
+^report knausj bug$:'https://github.com/knausj85/knausj_talon/issues'
```
rntz, line 62: We can use `user.open_url` here to go directly to the website.

Also, the naming convention suggests this should be `talon report bug` (or `talon bug report`, or both), although I'm fine with having both synonyms. I would suggest that we just have one command, which points to `https://github.com/knausj85/knausj_talon/issues`, since most users may not be able to tell what's a knausj bug and what's a talon bug.

rntz (changes requested):

PR #786, opened 2022-02-27
Added rust support

RyanRuane:
Following the existing python and c language implementations as references, rust support was set-up. Additions have also been made to extend some common tags which may find use elsewhere. For example,

* CLion was added to jetbrains
* A generic look `loop {}` was added to imperative
* Additions to document comments for nested and bulk doc comments.


lang/rust/rust.py
```diff
@@ -1,215 +1,522 @@
...
+        actions.auto_insert('true')
 
     def code_insert_false():
-        actions.auto('false')
+        actions.auto_insert('false')
 
-    def code_state_if():
-        actions.insert('if ')
+    # tag: data_null
+    # Convenience function, however, Option technically isn't null
```
auscompgeek, line 314: I had concerns about this in the original PR: https://github.com/knausj85/knausj_talon/pull/743#discussion_r790621760

auscompgeek:
The rust.py diff is large enough that GitHub collapses it by default. It's too large to actually see what the real changes are, since it rearranges all the action implementations.

lang/tags/data_null.talon
```diff
@@ -1,5 +1,5 @@
 tag: user.code_data_null
 -
-state (no | nil | null): user.code_insert_null()
+state (no | non | none | nil | null): user.code_insert_null()
```
rntz, line 3: I don't think we need both `non` and `none` here, just `none` is fine.

lang/tags/comment_documentation.talon
```diff
@@ -1,3 +1,6 @@
 tag: user.code_comment_documentation
 -
 dock comment: user.code_comment_documentation()
+block dock comment: user.code_comment_documentation_block()
+inner dock comment: user.code_comment_documentation_inner()
+inner block dock comment: user.code_comment_documentation_block_inner()
```
rntz, line 6: Are these kinds of commands universal enough to warrant putting them here? I guess block doc-comments are fairly common. I have no idea what an "inner dock comment" is.

lang/tags/imperative.py
```diff
@@ -40,6 +40,9 @@ def code_state_go_to():
     def code_state_while():
         """Inserts while statement"""
 
+    def code_state_loop():
+        """Inserts loop statement"""
+
```
rntz, line 45: This seems like the wrong place for this, I'm not sure I know many languages that have a bare "loop" construct.

lang/rust/rust.py
```diff
@@ -1,215 +1,522 @@
...
+
+@mod.capture(rule='{user.code_macros}')
+def code_macros(m) -> str:
+    """Returns a macro name"""
+    return m.code_macros
+
+
+@mod.action_class
+class Actions:
+    def code_operator_structure_dereference():
```
rntz, line 19: If these actions are only going to be used in Rust, you don't really need to abstract them like this; might be simpler just to put their implementations into the rust.talon file. Not a big deal either way though.

rntz (commented):
It's hard to review this PR b/c it's hard to distinguish genuine changes from code moving around. In future PRs please try to minimize the diff with the existing code.

1. Please check for methods you may have deleted (eg. `code_operator_address_of`). Likewise for deleted list items.

2. Please check for changes in behavior between the old code and this code, and summarize any relevant ones. (eg. `code_state_{if,else}` changed)

3. I'm a bit concerned about over-reliance on the details of editor auto-formatting, especially in the function-defining commands/actions like `code_private_function`, and others like `code_state_implements`. Can you double-check these? Generally we're ok assuming that `tab` auto-indents the current line if it needs more indentation but I'd like to avoid assuming too much more than that if possible.

4. I'm not sure whether separating out macros from functions is good. I'm not a Rust programmer, though. It seems like many macros are called like functions (eg. `println!`) and it might be easier not to have to remember a different command for invoking them, but there are also some special ways to invoke macros. Perhaps all macros should be included in the `code_functions` list as well? Your opinion?

apps/jetbrains/jetbrains.py
```diff
@@ -113,6 +113,7 @@ def idea_commands(commands):
 mod = Module()
 
 mod.apps.jetbrains = "app.name: /jetbrains/"
+mod.apps.jetbrains = "app.name: CLion"
```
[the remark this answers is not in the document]
RyanRuane: This program is the jetbrains product that best integrates support for rust. Was you prefer this be submitted as its own per for this line?

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Dev stuff
 __pycache__
 *.sw?
+.idea/
```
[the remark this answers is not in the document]
RyanRuane: I included this so long to support people using clion or any other kind of jet beans project, but I can happily move it out into a different pr. Can I just confirm you would still like this?

lang/tags/comment_documentation.talon
auscompgeek: I'd probably say they're probably not really universal. Such a doc comment in Rust is most commonly used to add docs for a module within the module. (The syntax works in any block that can be documented, so `inner` would definitely be a correct name here.)

I'm not familiar with a similar construct in JavaScript, and docstrings in Python always go inside the block it documents, so :shrug:

RyanRuane: I suspect the concepts of a block comment is sufficiently generic, however I'm not too sure about inner document comments. In rust you can document and entity atop it with a doc comment, or you can document information about it on top of its relevant member with an inner doc comment. Is there a strong preference to have the inner cases removed here and integrated into the module inside the rust file? I thought it would be okay because I had seen functions inside tags only implemented by one language before, in the case of the R language.

lang/tags/imperative.py
RyanRuane: My thinking here was that this could be overloaded to represent infinite loops in general. For example, in C you might have something like `for(; ;)  `, or `while(true) ` in java, or `while True:` in python, or simply `for {}`.  Sometimes there are multiple ways to do an infinite loop in a language, and the community may prefer a particular method as the standard, which could be placed here. this could perhaps be changed to code state infinite loop?

lang/rust/rust.py
RyanRuane: that's completely fair, I was just trying to keep these functions inside the python file, because their implementation is very similar to implementations of tags below. I'll bear this in mind for the future.

RyanRuane:
> It's hard to review this PR b/c it's hard to distinguish genuine changes from code moving around. In future PRs please try to minimize the diff with the existing code.

 Unfortunately, I develop this branch prior to the merging of the current rest code and without knowledge of it. I did move around lines in order to minimize some of the conflict, especially in the talon file, however I didn't want to compromise the flow of the implementations which are in the same order as the tag definitions in the talon file. 

> Please check for methods you may have deleted (eg. code_operator_address_of). Likewise for deleted list items.

Two of the three methods of this tag around pointers were implemented, with the third function which has the word dereference in its name, omitted. Because the concept of the referencing in rust is itself achieved with *, I neglected to implement this tag in preference of rust specific grammar. However, if it is preferred that that tag be used, I can add this back in.

> I'm not sure whether separating out macros from functions is good. I'm not a Rust programmer, though. It seems like many macros are called like functions (eg. println!) and it might be easier not to have to remember a different command for invoking them, but there are also some special ways to invoke macros. Perhaps all macros should be included in the code_functions list as well? Your opinion?

I could perhaps see merit of including them in the general functions list, however macros are quite distinct from functions in reality. As they have their own internal grammar, it's not that they can be called differently, but they get to a fine how they are called. It is for this reason that I implemented it separately, and allows three different ways for calling a macro with the most common parentheses, those being parentheses themselves for function like macros, square brackets for those that emulate listing, like `vec![]`, and braces, for those macros, like `error_chain`.

lang/rust/rust.py
```diff
@@ -1,215 +1,522 @@
...
+    'see string': 'CString',
+    'see string slice': '&CStr',
+    'option': 'Option',
+    'result': 'Result',
+    'hashmap': 'HashMap',
+    'hash set': 'HashSet',
+    'reference count': 'Rc',
+}
 
-    def code_operator_indirection():
```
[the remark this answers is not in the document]
RyanRuane: I chose not to implement this tag. Discussion can be found in the major comet below.

lang/rust/rust.py
```diff
@@ -1,215 +1,522 @@
...
+    'condition variable': 'Condvar',
+    'mutex': 'Mutex',
+    'once': 'Once',
+    'read write lock': 'RwLock',
+    'receiver': 'Receiver',
+    'sender': 'Sender',
+    'sink sender': 'SyncSender',
+}
 
-    def code_operator_address_of():
```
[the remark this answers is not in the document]
RyanRuane: I chose not to implement this tag. Discussion can be found in the major comet below.

lang/rust/rust.talon
```diff
@@ -1,31 +1,42 @@
...
 tag(): user.code_functions_gui
+
 tag(): user.code_libraries
 tag(): user.code_libraries_gui
+
 tag(): user.code_operators_array
 tag(): user.code_operators_assignment
 tag(): user.code_operators_bitwise
 tag(): user.code_operators_math
-tag(): user.code_operators_pointer
```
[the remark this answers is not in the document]
RyanRuane: I chose not to implement this tag. Discussion can be found in the major comet below.

lang/rust/rust.talon
```diff
@@ -35,21 +46,65 @@ state enum <user.text>:
...
 state let mute: "let mut "
 state let: "let "
-state mute: "mut "
-state ref mute: "ref mut "
+state (mute | mutable): "mut "
+state ref (mute | mutable): "ref mut "
 state ref: "ref "
 state trait: "trait "
 state match: "match "
-op arrow: " -> "
```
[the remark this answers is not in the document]
RyanRuane: I found it for myself at least when using talon that it is simply possible to say arrow or dub arrow and so they seemed to be needless and longer. Are these here for another reason? Perhaps to protect against the  the smaller  commands being overwritten in some context?

RyanRuane:
> I'm a bit concerned about over-reliance on the details of editor auto-formatting, especially in the function-defining commands/actions like code_private_function, and others like code_state_implements. Can you double-check these? Generally we're ok assuming that tab auto-indents the current line if it needs more indentation but I'd like to avoid assuming too much more than that if possible.

 I agree, I'm afraid this is my first talon pr, and am a bit naive to the best usage. I simply relied on things I had seen elsewhere.  where superiors solutions have been pointed out, I'll make the appropriate changes. I wasn't sure which of my implementation and the conflicting rust implementation that I had previously been unaware of,  was better, and so I tried to make as few possible changes to my own branch.

lang/rust/rust.py
```diff
@@ -1,215 +1,522 @@
...
+    # tag: libraries_gui
+
+    def code_insert_library(text: str, selection: str):
+        actions.user.paste(f'use {selection}')
+
+    # tag: operators_array
+
+    def code_operator_subscript():
+        actions.insert('[]')
+        actions.key('left')
```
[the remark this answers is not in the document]
RyanRuane: What's the idiomatic way to deal with going left multiple times?

lang/rust/rust.py
```diff
@@ -1,215 +1,522 @@
...
-        formatter = settings.get('user.code_protected_function_formatter')
-        function_name = actions.user.formatted_text(text, formatter)
-        actions.user.code_insert_function(function_name, None)
+    def code_private_function(text: str):
+        name = actions.user.formatted_text(
+            text, settings.get('user.code_private_function_formatter')
+        )
+        result = f'fn {name}() {{\n}}\n'
+        actions.user.paste(result)
+        actions.key('up:2 right:3')
```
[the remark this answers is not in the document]
RyanRuane: I was looking to place the cursor by the name, however I have elected to go with the previous implementation

lang/rust/rust.talon
```diff
@@ -35,21 +46,65 @@ state enum <user.text>:
...
+
+## specialist flow control
+if let some: user.code_insert_if_let_some()
+if let error: user.code_insert_if_let_error()
+
+## rust centric synonyms
+is (some|sum): user.code_insert_is_not_null()
+
+## for implementing
+implement: user.code_state_implements()
```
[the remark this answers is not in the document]
RyanRuane: This is to implement an implement block, where functions are added to a structure. Still I agree, and so I have changed this to be implement struct or implement structure.

RyanRuane:
I've addressed everything I gave a thumb up, and left outstanding questions for things I wasn't sure about or where I didn't know community preference. Of the things that I said I would fix that I have not, that is the auto formatting of movements keys, I would like to know what is the best way to make multiple left movements while using edit. left.

 Regarding all variants on if statements, I have left them as one liner statements. This is because it is possible to have a one line if statement whose value is assigned to a variable, and I didn't feel it was a major  problem for a person to have to hit enter when they need already to define the if clause on the same line.

lang/rust/rust.talon
```diff
@@ -1,31 +1,42 @@
...
     user.code_public_function_formatter = "SNAKE_CASE"
     user.code_private_variable_formatter = "SNAKE_CASE"
+    user.code_protected_variable_formatter = "SNAKE_CASE"
     user.code_public_variable_formatter = "SNAKE_CASE"
 
+# rust-specific grammars
+
+## for unsafe rust
 state unsafe: "unsafe "
+unsafe block: user.code_state_unsafe()
```
auscompgeek, line 36: Any reason for making this its own command instead of just saying `state unsafe block` (i.e. chaining the `state unsafe` and `block` commands)?

lang/tags/imperative.py
auscompgeek: `code_state_infinite_loop` sounds like a reasonable idea to me.

lang/rust/rust.talon
```diff
@@ -35,21 +46,63 @@ state enum <user.text>:
...
 use <user.code_libraries>:
     user.code_insert_library(code_libraries, "")
     key(semicolon enter)
+
+## specialist flow control
+if let some: user.code_insert_if_let_some()
+if let error: user.code_insert_if_let_error()
+
+## rust centric synonyms
+is (some|sum): user.code_insert_is_not_null()
```
auscompgeek, line 78: Is `sum` here for Dragon? If not and Dragon works without it, it'd be clearer to just have `is some`.

lang/rust/rust.talon
RyanRuane: My thinking was because you can have an unsafe block, or an unsafe function. I thought relying on state unsafe would work for functions, but this would be quicker for blocks

lang/rust/rust.talon
RyanRuane: done

.gitignore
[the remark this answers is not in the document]
rntz: On reflection this is fine, no need to split it out.

apps/jetbrains/jetbrains.py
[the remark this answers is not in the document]
rntz: it's fine

lang/rust/rust.py
```diff
@@ -1,215 +1,522 @@
...
+    'float sixty four': 'f64',
     'boolean': 'bool',
-    'string': 'str',
-    'owned string': 'String',
+    'character': 'char',
 }
 
-@ctx.action_class('user')
-class UserActions:
-    def code_operator_lambda():
```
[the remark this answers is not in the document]
rntz: Ah, I see. This is fine. You can include the lambda operator and tag if you think they're useful.

lang/tags/comment_documentation.talon
rntz: I think we'd prefer if these all were made Rust-specific, we can't find examples of other languages that have all four of these distinctions.

rntz (changes requested):
Apart from a few small changes, I think the only remaining question here is functions vs macros. From experience voice-coding, the fewer things you have to think about when issuing a command, the better. So I think we'd prefer if eg. the "funk" command from functions_gui.talon worked with macros like `println!`. The simplest way to accomplish this would be to include all macros into `<user.code_functions>`. The special macro commands for different brace/bracket 
styles are fine, of course.

pokey:
@RyanRuane the below is not a blocking request, but fwiw I'd be inclined to go a step further wrt the macro / function thing, and just have a single command "funk" that you use for functions, square macros, curly macros etc, and then have separate sets that you can use to indicate things that want square or curly brackets.  Something like the following:

```python
square_bracket_macros = {
    "vec!",
}

curly_bracket_macros = {
    "foo!",
    "bar!",
}
```

Then in `code_insert_function`, do something like

```python
    def code_insert_function(text: str, selection: str):
        if text in square_bracket_macros:
            code_insert_function_or_macro(text, selection, '[', ']')
        elif text in curly_bracket_macros:
            code_insert_function_or_macro(text, selection, '{', '}')
        else:
            code_insert_function_or_macro(text, selection, '(', ')')
```

This approach presumes you'll always use the same style brackets with the same macro, but if so, it's one less thing to think about while coding.  You just always say "funk foo", and don't need to worry about function vs macro / bracket style / etc.  As @rntz mentions, the less you have to think about while voice coding the better

lang/rust/rust.talon
auscompgeek, line 36: The action implementation isn't very complicated, and isn't shared anywhere. Any reason to have it in the .py file?

RyanRuane:
* Have moved block and inner documents comments into rust specific space.
* Have included all macros into the general functions and functions gui tags. 
* Now automatically deciding on whether to use brackets braces or parentheses with respect to given macros based on hardcoded designations. 
* Have corrected control flow statements formatting to avoid issues in ides such as pycharm which attempt to insert lines and chars automatically.

lang/rust/rust.py
```diff
@@ -1,215 +1,568 @@
...
 
     def code_block():
-        actions.auto('{\n\n}')
+        actions.auto_insert('{}')
         actions.edit.left()
+        actions.key('enter')
+
+    def code_state_if():
+        actions.auto_insert('if  {  }')
+        repeat_call(5, actions.edit.left)
```
auscompgeek, line 263: There's an `actions.user.insert_between` that was added in #781, which can be used here instead.

I'm not convinced these should insert the braces in the first place though. Having to say ~~`go right twice`~~ `go right third` (and counting to get it right in the first place) is a significant cognitive load for me, compared to saying `brace` or `block` if these didn't insert the braces.

RyanRuane: I didn't know about the insert between function so that is great. I've replaced use with it where appropriate. and I find that I agree with your comments on the if statement, and similar such constructs. I've removed the braces wherever a
is not inserted, such is in a for loop.

lang/rust/rust.py
```diff
@@ -1,215 +1,558 @@
...
+        actions.auto_insert(' else if ')
+
+    def code_state_else():
+        actions.user.insert_between(' else { ', ' }')
+
+    def code_state_switch():
+        actions.auto_insert('match ')
+
+    def code_state_for():
+        actions.auto_insert('for  in  {}')
```
auscompgeek, line 274: Should the implementations of `code_state_for` and `code_state_for_each` be the same? I don't think the Python implementation does this.

RyanRuane: I implemented both so that invoking either one would result in the only for statement, however in practice I have found it pointless. I will remove the fall each implementation.

PR #787, opened 2022-02-27
Simple Discord for Linux

RyanRuane:
Clone of mac implementation with command button substituted

I verified the shortcut worked, and where there are conflicts, that they match the  defaults in the discord client. It should be noted these can be can be changed in Discord, so these are only good for default settings. Additionally, at least on Ubuntu,  a couple of the shortcuts conflict with gnome 3 shortcuts, such as:

```
    def messaging_workspace_previous():  actions.key('ctrl-alt-up')
    def messaging_workspace_next():      actions.key('ctrl-alt-down')
    def discord_mentions_last():         actions.key('ctrl-alt-shift-up')
    def discord_mentions_next():         actions.key('ctrl-alt-shift-down')
```

### DIFF
--- a/apps/discord/linux.py
+++ b/apps/discord/linux.py
@@ -1,0 +1,39 @@
+from talon import Context, actions
+ctx = Context()
+ctx.matches = r"""
+os: linux
+app: discord
+"""
+
+@ctx.action_class('user')
+class UserActions:
+    # Navigation: Servers
+    def messaging_workspace_previous():  actions.key('ctrl-alt-up')
+    def messaging_workspace_next():      actions.key('ctrl-alt-down')
+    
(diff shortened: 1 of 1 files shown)


rntz (approved):

auscompgeek:
Isn't this already in the `win.py`?

RyanRuane:
Are you saying it might be better to have a combined implementation that triggers for both operating systems?

auscompgeek:
I'm saying that combined implementation already exists.

RyanRuane:
Oh yes I see it now. Because the file is labeled win, I had just assumed there was no Linux support. And oversight on my part. However, is that perhaps a better naming strategy we could have here?

PR #788, opened 2022-02-27
Terminator for Linux

RyanRuane:
Added terminator for Linux, based on the gnome terminal.

Shortcuts appropriately updated, and split window functionality implemented based on the splits tag.

### DIFF
--- a/apps/platforms/linux/terminator/terminator-linux.py
+++ b/apps/platforms/linux/terminator/terminator-linux.py
@@ -28,7 +28,7 @@ ctx.tags = [
 # --- Implement actions ---
 @ctx.action_class("user")
 class user_actions:
-    # tag: user.splits
+    # user.splits
     def split_window_right(): actions.key("alt-right")
     def split_window_left(): actions.key("alt-left")
     def split_window_down(): actions.key("alt-down")
--- a/apps/platforms/linux/terminator/terminator-linux.py
+++ b/apps/platforms/linux/terminator/terminator-linux.py
@@ -16,7 +16,6 @@ app: terminator
 """
 ctx.tags = [
     'terminal',
-    'user.terminal',
     'user.tabs',
     'user.splits',
     'user.generic_unix_shell',
--- a/apps/platforms/linux/terminator/terminator-linux.py
+++ b/apps/platforms/linux/terminator/terminator-linux.py
@@ -70,4 +70,4 @@ class EditActions:
 
     def delete_line():
         actions.edit.line_start()
-        actions.key('ctrl-a ctrl-k')
+        actions.key('ctrl-u')
--- a/apps/platforms/linux/terminator/terminator-linux.py
+++ b/apps/platforms/linux/terminator/terminator-linux.py
@@ -72,11 +72,3 @@ class EditActions:
     def delete_line():
         actions.edit.line_start()
         actions.key('ctrl-a ctrl-k')
-
-    # afaik not possible in gnome-terminal
-    def extend_left(): pass
-    def extend_right(): pass
-    def extend_up(): pass
-    def extend_down(): pass
-    def extend_word_left(): pass
-    def extend_word_right(): pass
--- a/apps/platforms/linux/terminator/terminator-linux.py
+++ b/apps/platforms/linux/terminator/terminator-linux.py
@@ -15,42 +15,42 @@ ctx.matches = r"""
 app: terminator
 """
 ctx.tags = [
-    "terminal",
-    "user.terminal",
-    "user.tabs",
-    "user.splits",
-    "user.generic_unix_shell",
-    "user.git",
-    "user.kubectl",
+    'terminal',
+    'user.terminal',
+    'user.tabs',
+    'user.splits',
+    'user.generic_unix_shell',
+    'user.git',
+    'user.kubectl',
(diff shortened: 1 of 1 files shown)


rntz (approved):

apps/platforms/linux/terminator/terminator-linux.py
```diff
@@ -0,0 +1,82 @@
+from talon import Context, actions, Module
```
auscompgeek, line 1: Naming the file just `terminator.py` would probably be fine.

apps/platforms/linux/terminator/terminator-linux.py
```diff
@@ -0,0 +1,82 @@
...
+    def copy(): actions.key('ctrl-shift-c')
+
+    def find(text: str = None):
+        actions.key('ctrl-shift-f')
+        if text:
+            actions.insert(text)
+
+    def delete_line():
+        actions.edit.line_start()
+        actions.key('ctrl-a ctrl-k')
```
auscompgeek, line 74: This seems like something that should be in `generic_unix_shell` instead, especially since this isn't the terminal emulator's behaviour but whatever program is running within.

apps/platforms/linux/terminator/terminator-linux.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # afaik not possible in gnome-terminal
```
auscompgeek, line 76: Huh, these are implemented in `gnome-terminal.py` like this? They probably shouldn't be implemented at all.

auscompgeek (commented):
Keybindings look correct to me.

apps/platforms/linux/terminator/terminator-linux.py
RyanRuane: Is that so? This deviates from the key binding used in the gnome terminal, at least. So I was concerned it was terminator specific?

apps/platforms/linux/terminator/terminator-linux.py
RyanRuane: It currently only triggers for Linux, because I have no experience with terminator on any other operating system. I was concerned that the shortcuts could be different, and so I wanted  to make it clear but this only is known to operate on Linux. I'm not one hundred percent sure, but I believe then might be a port of this program for mac.

apps/platforms/linux/terminator/terminator-linux.py
splondike: Is the ctrl-a necessary in terminator? If so, then this does look to be app specific.

auscompgeek: `ctrl-a` would just go to the beginning of the line in readline/emacs, right?

I'll note too that this is definitely wrong in a shell with vim keybindings.

RyanRuane: Here is the documentation that specifies the shortcuts, which I used as reference. If I try to use control k without first having pressed control a I get no response. https://terminator-gtk3.readthedocs.io/en/latest/gettingstarted.html

auscompgeek: That page doesn't list `ctrl-a` nor `ctrl-k`. Those are both standard readline keybindings. `ctrl-k` cuts from the cursor position to the end of the line buffer.

As an aside, `ctrl-u` cuts from the cursor to the beginning of the line buffer. Again, this is standard readline behaviour. Terminal emulators do not have a concept of a line input buffer.

RyanRuane: Yeah, on review I see that it's not one of the shortcuts I pulled by looking into the terminated documentation. It's a method I implemented because it's inside the file for the Linux gnome terminal. Considering the keybinding is different in the gnome implementation, I was sure that this should be moved to a generic editor? If so, I'll move it there.

auscompgeek: Moving this to a tag context probably warrants further discussion in a separate PR. Not everyone uses emacs keybindings in their shell.

RyanRuane: Is it the case then, that I should not implement this method all together?

apps/platforms/linux/terminator/terminator-linux.py
```diff
@@ -0,0 +1,73 @@
...
+    def copy(): actions.key('ctrl-shift-c')
+
+    def find(text: str = None):
+        actions.key('ctrl-shift-f')
+        if text:
+            actions.insert(text)
+
+    def delete_line():
+        actions.edit.line_start()
+        actions.key('ctrl-u')
```
splondike, line 73: I think the fix for readline emacs vs vim mode can happen in a different PR. We can just assume emacs mode like the other emulators in this one I think.

I installed terminator on my machine and replacing actions.key with `ctrl-k` works fine for me (i.e. the same implementation as in gnome terminal). Maybe the issue you were having is you have line_start() re-implemented or something like that?

splondike:
Tested this out with terminator on my machine, it looks fine aside from the new delete_line implementation. If that gets fixed I'm happy to merge this in.

splondike:
Thanks for that!

PR #789, opened 2022-02-28
selected_text(): use capture().text() not .get()

rntz:
Squashes some warnings from talon. My only concern here is that this API may not yet be in public talon. Can someone check?

### DIFF
--- a/code/edit.py
+++ b/code/edit.py
@@ -11,7 +11,7 @@ class EditActions:
         with clip.capture() as s:
             actions.edit.copy()
         try:
-            return s.get()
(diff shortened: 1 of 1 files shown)


auscompgeek:
It's not in public Talon.

rntz:
this should be in public talon as of 0.3, although we ought to test that.

phillco:
Did 0.3 release to windows yet?

rntz:
Apparently not :(

rntz:
it did now

pokey (approved):

PR #795, opened 2022-03-06
Fix terraform client tagging

2d6:
#776 introduced a bug if the user explicitly set the tag `user.terraform`, which already existed due to #752. Terraform is both an application and a language, which is why this collision occurred. The bug manifested in lang/terraform/terraform.py
always being evaluated if a user explicitly set the client tag. This PR renames the client tag to fix this.


splondike:
Thanks!

PR #796, opened 2022-03-06
Extend Scala language

2d6:
This pull request intends to extend the scala language definitions introduced in #735. If applied, it will

- add further types
- add modifiers
- add keywords
- remove added `enter` after `else` keyword to enable `if(foo) bar else baz` syntax

### DIFF
--- a/lang/scala/scala.py
+++ b/lang/scala/scala.py
@@ -8,31 +8,80 @@ tag: user.scala
 
 # Scala Common Types
 scala_common_types = {
-    "Byte": "Byte",
-    "Int": "Int",
-    "Double": "Double",
-    "Short": "Short",
-    "Float": "Float",
-    "Long": "Long",
-    "Boolean": "Boolean",
-    "Char": "Char",
-    "String": "String",
-    "None": "None"
+    "boolean": "Boolean",
+    "int": "Int",
+    "float": "Float",
+    "byte": "Byte",
+    "double": "Double",
+    "short": "Short",
+    "long": "Long",
+    "char": "Char",
+    "unit": "Unit",
+    "any": "Any",
+    "any val": "AnyVal",
+    "string": "String",
+    "thread": "Thread",
+    "exception": "Exception",
+    "throwable": "Throwable",
+    "none": "None",
+    "success": "Success",
+    "failure": "Failure",
 }
 
 # Scala Common Generic Types
 scala_common_generic_types = {
-    "seq": "Seq",
-    "set": "Set",
-    "list": "List",
     "array": "Array",
+    "deck": "Deque",
+    "future": "Future",
+    "list": "List",
     "map": "Map",
-    "stack": "Stack"
+    "nil": "Nil",
+    "option": "Option",
+    "queue": "Queue",
+    "seek": "Seq",
+    "set": "Set",
+    "some": "Some",
+    "stack": "Stack",
+    "try": "Try",
 }
 
 scala_types = scala_common_types.copy()
 scala_types.update(scala_common_generic_types)
 ctx.lists["user.code_type"] = scala_types
+
+# Scala Modifies
+scala_modifiers = {
+    "public": "public",
+    "private": "private",
+    "protected": "protected",
+}
+
+mod.list("scala_modifier", desc="Scala Modifiers")
+ctx.lists["user.scala_modifier"] = scala_modifiers
+
+scala_keywords = {
+    "abstract": "abstract",
+    "case class": "case class",
+    "def": "def",
+    "extends": "extends",
+    "implicit": "implicit",
+    "lazy val": "lazy val",
+    "new": "new",
+    "object": "object",
+    "override": "override",
+    "package": "package",
+    "sealed": "sealed",
+    "throw": "throw",
+    "trait": "trait",
+    "type": "type",
+    "val": "val",
+    "var": "var",
+    "with": "with",
+    "yield": "yield",
+}
+
+mod.list("scala_keyword", desc="Scala Keywords")
+ctx.lists["user.scala_keyword"] = scala_keywords
 
 @ctx.action_class("user")
 class UserActions:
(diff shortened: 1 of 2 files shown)


lang/scala/scala.talon
```diff
@@ -24,15 +24,19 @@ settings():
     user.code_protected_variable_formatter = "PRIVATE_CAMEL_CASE"
     user.code_public_variable_formatter = "PRIVATE_CAMEL_CASE"
 
-state package: "package "
+[state] {user.scala_modifier}:
```
splondike, line 27: I think I'd keep the state prefix required for these ones, we tend to stay away from single word commands in knauj public to reduce the chance of collisions with future commands and also because they tend to recognise less well.

splondike:
@SCdF Happy to lose that enter press after the else statement?

lang/scala/scala.py
```diff
@@ -8,32 +8,81 @@
...
+}
+
+mod.list("scala_modifier", desc="Scala Modifiers")
+ctx.lists["user.scala_modifier"] = scala_modifiers
+
+scala_keywords = {
+    "abstract": "abstract ",
+    "case class": "case class ",
+    "class": "class ",
+    "def": "def ",
```
SCdF, line 65: so `def` is how you define a function and so is already covered by "funk <blah>", as is I think `package` and possibly others below. @splondike not sure about default repo's philosophy regarding duplicate ways of doing things, but thought I'd flag if you weren't aware.

SCdF:
@splondike yep happy with single line else, it's definitely more correct! Made one comment about duplicate functionality as well

lang/scala/scala.py
2d6: This is mostly true, but at least in our codebase, we also sometimes use `def` for expressions without argument brackets:
```scala
def foo = "hello"
// or
override def bar = "baz"
```
afaik, `funky` inserts `()` automatically. So I'm not sure this counts as duplicate functionality. What do you think?

lang/scala/scala.talon
2d6: Good point, thank you! I amended this in the commit.

2d6:
Now that I think of it @splondike : in one of my pull requests, you mentioned that for code we should use `insert` instead of `auto_insert` (https://github.com/knausj85/knausj_talon/pull/752#r795277190). If this also applies here, I would be happy to amend that here, as well. There are several operator methods which use `auto_insert` in `scala.py`. Just let me know :)

2d6:
I think it would also make sense to extend the language mappings with the `.sbt` extension, because the scala build tool also uses scala syntax for the `build.sbt`, `plugins.sbt` etc. files.

lang/scala/scala.py
SCdF: Yep that's fair. Scala is nothing if not flexible in how you use it

SCdF:
> I think it would also make sense to extend the language mappings with the `.sbt` extension, because the scala build tool also uses scala syntax for the `build.sbt`, `plugins.sbt` etc. files.

This is a good example of scala's flexibility making building this kind of thing really hard. The sbt file is a dsl based on scala, but not really idiomatic scala, right? So a build file can look like:

```sbt
ThisBuild / organization := "com.example"
ThisBuild / scalaVersion := "2.12.15"
ThisBuild / version      := "0.1.0-SNAPSHOT"

lazy val root = (project in file("."))
  .settings(
    name := "hello"
  )
```
Which is sort of scala, but also you're not "dividing" ThisBuild by organisation for example, so if you wanted an idiomatic sbt mapping you would want to call that something else, and that new mapping would only apply to sbt files. Similarly with  `:=`, in  sbt you could map this to "op assign" wherein scala that would just be  `=`.

If you think the value of having sbt use scala definitions is additive to your experience though, even if it's a little wrong, then that's probably OK?

There are tons of DSLs in scalaland unfortunately, but at least this one is mostly scala and most people are guaranteed to use it.

_(FWIW I also did the cursorless mapping for scala, and I did not even look into sbt files or how they get parsed out, so it's likely there is work required for support there, if you use that tool)_

splondike:
Yeah, you can replace the auto_inserts with insert to be a little more correct IMO, good memory!

For the SBT stuff it sounds like maybe that should be a different language (and probably a different PR)? You could do something like has been done with typescript/javascript perhaps. Activate the user.scala tag from the sbt language and then override some select things (e.g. op assign -> ':=').

2d6:
> Yeah, you can replace the auto_inserts with insert to be a little more correct IMO, good memory!
Thank you, I amended this to the commit.

> For the SBT stuff it sounds like maybe that should be a different language (and probably a different PR)? You could do something like has been done with typescript/javascript perhaps. Activate the user.scala tag from the sbt language and then override some select things (e.g. op assign -> ':=').
This actually sounds like a great idea. Thanks for this suggestion :)

Also thank both of you for your valuable suggestions and feedback. Unless you find anything else to improve in this pull request, I currently have nothing else to add to it.

lang/scala/scala.py
```diff
@@ -8,32 +8,80 @@
 
 # Scala Common Types
 scala_common_types = {
-    "Byte": "Byte",
-    "Int": "Int",
-    "Double": "Double",
-    "Short": "Short",
```
rntz, line 14: "short" seems to have been lost, is this intentional?

lang/scala/scala.py
```diff
@@ -8,32 +8,80 @@
...
     "array": "Array",
+    "deque": "Deque",
+    "future": "Future",
+    "list": "List",
     "map": "Map",
-    "stack": "Stack"
+    "nil": "Nil",
+    "option": "Option",
+    "queue": "Queue",
+    "seq": "Seq",
```
rntz, line 40:
```suggestion
    "seek": "Seq",
```

lang/scala/scala.talon
```diff
@@ -24,15 +24,19 @@ settings():
...
 
-state package: "package "
+state {user.scala_modifier}:
+    insert(user.scala_modifier)
+    key("space")
 
-state val: "val "
-state var: "var "
+state {user.scala_keyword}:
+    insert("{scala_keyword}")
```
rntz, line 32: It's weird that modifiers get space added to them by the command while keywords include the space in them. It would be clearer if it was consistent.

rntz (approved):
Looks mostly good to me, however, looks like "short" has been lost from the scala_common_types list. Is that intentional?

lang/scala/scala.py
2d6: D'oh, good catch! I amended this.

rntz:
I'm just going to merge this, if you want to address any further concerns such as the short type or the comments I left then just make another PR.

lang/scala/scala.py
2d6: Good suggestion. I also modified the command for `Deque` from `deque` to `deck`.

lang/scala/scala.talon
2d6: You are absolutely right. I have now included this in the commit.

rntz:
oh wow, looks like you got to my concerns before I merged : )

2d6:
Phew, just in time :D Sorry for involuntarily sneaking in the changes, but I'm happy to see the pull request merged!

PR #798, opened 2022-03-10
Add Brave as a new browser applicatoin

richardmcmillen:
EDIT: This adds Brave as a new browser application. See below for the original issue description.
----

Brave is a web browser built on Chromium and so I thought it was fitting to add it as a variant of Chrome.

This is my first contribution and I'm still getting used to what the configuration parameters are used for if I haven't got it right please correct me. I see this in the contributing section in the README

>For Mac OS X, the bundle id should be used for defining app contexts, rather than the name

And I'm not sure if I've done more than I should have there.

Also I don't have access to a Windows machine (only mac) to test but some googling suggested what the .exe name would be.

### DIFF
--- a/apps/brave/brave.py
+++ b/apps/brave/brave.py
@@ -1,0 +1,50 @@
+from talon import ctrl, ui, Module, Context, actions, clip, app
+
+ctx = Context()
+mod = Module()
+
+mod.apps.brave = "app.name: Brave Browser"
+# TODO: Add other OS application identifiers
+mod.apps.brave = """
+os: mac
+and app.bundle: com.brave.Browser
+"""
+ctx.matches = r"""
+app: brave
+"""
+
+@ctx.action_class("user")
+class user_actions:
+    def tab_jump(number: int):
+        if number < 9:
+            if app.platform == "mac":
+                actions.key("cmd-{}".format(number))
+            else:
+                actions.key("ctrl-{}".format(number))
+
+    def tab_final():
+        if app.platform == "mac":
+            actions.key("cmd-9")
+        else:
+            actions.key("ctrl-9")
+
+    def tab_close_wrapper():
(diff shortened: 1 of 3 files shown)


richardmcmillen:
I've realised I should be updating https://github.com/knausj85/knausj_talon/blob/master/apps/chrome/mac.py as well and maybe some other places.

Also I've been thinking perhaps Brave should just be a new app, but there is so much shared with Chrome. I'll think about this some more.

auscompgeek:
Are there any Brave-specific features which you might want a command for? That'd be a good sign that we should have a separate app matcher.

I definitely agree with you on that a lot of the action implementations will be the same across Chromium flavours. We already have both Chrome and Edge here. It might be worth refactoring them to use a `user.chromium` tag. Doesn't have to be in this PR though.

richardmcmillen:
Thank you very much for your comment and I apologize for taking so long to provide more information here. I think you may be right in that having a chromium tag would be an an interesting avenue to investigate.  

I have started a google sheet to help me collect and compare the shortcuts a cross chromium applications:
https://docs.google.com/spreadsheets/d/1hai5fgcDth8LWU42lFf-LzYIBxLEBZm80WH3yF179_o/edit?usp=sharingdub 

I only have the mac commands recorded for edge, chrome, and brave, recorded currently. I hope to get to the other ones at some point to find what the common shortcuts there are across applications and operating systems.

richardmcmillen:
If I do decide to contribute a change to pull out chromium it would be in a new pull request so I am marking this one ready for review as is.

 - Add brave as a new application pretty much copying what was already in place for Chrome.
 - The tag matchers do not include other OS's as I don't have access to them and didn't want to guess.

splondike (approved):
Looks good, thanks!

PR #799, opened 2022-03-12
Add further git commands for diffing and merging

2d6:
I would like to contribute some further `git` commands from my personal configuration which are almost in daily use for me. The list of commands is already quite extensive, so feel free to close this pull request if you think it's too much.


misc/git.talon
```diff
@@ -28,6 +28,9 @@ git commit existing: "git commit -a\n"
 git diff (colour|color) words: "git diff --color-words "
 git diff: "git diff "
 git diff cached: "git diff --cached\n"
+git diff staged: "git diff --staged\n"
```
auscompgeek, line 31: Hmm, `git diff --cached` and `--staged` are synonyms, right? Should we just have one command?

misc/git.talon
```diff
@@ -39,6 +42,7 @@ git log: "git log "
 git log changes: "git log -c "
 git merge: "git merge "
 git merge <user.text>:"git merge {text}"
+git mergetool: "git mergetool\n"
```
auscompgeek, line 45: Should the spoken form be written as `merge tool`, for Dragon?

auscompgeek (approved):
I'd probably find these useful myself.

misc/git.talon
2d6: Oh, TIL. I did not cross-reference `--cached`. The existing one is definitely good enough, then. I am just going to retain my local alias, since it's already ingrained for me. Going to remove this.

misc/git.talon
2d6: Done, thanks :) I gather Dragon is less able to parse words it does not know yet, compared to Conformer?

splondike:
Thanks!

PR #800, opened 2022-03-13
Refactor functions_gui to functions_common

### LINKED ISSUE #732: find better name for function_gui
While commands defined in the [functions_gui.talon](https://github.com/knausj85/knausj_talon/blob/master/lang/tags/functions_gui.talon) file provide easier access to common functions (e.g. print,len or int in python) the file name seems very unintuitive to me. 

I propose renaming this and any connected file to "common_functions.talon/py" or something similar.

2d6:
Fixes #732

This commit refactors the misleadingly named functions_gui tag and the associated list, tags, and files, as @pokey suggested in #732. 
It also removes dead configuration for some languages.

As suggested in the issue, I changed the name of the function list to singular to increase readability and align with the code standard. I deviated from pokey's suggestions in that I used the tag `code_functions_common` instead of `code_common_functions` (same with the file names). I did so to conform with the pattern established by other tags e.g. `code_operators_array` or `code_data_bool`.

I was as careful as possible, but considering the large number of files I had to touch, this pull request has the potential of breaking functionality. Therefore, I would welcome a thorough review. I manually verified that the `funk` command still works for the languages I touched.

I realize that there is a similar functionality for common libraries. If this pull request is welcomed, I would volunteer to refactor that functionality as well.

### DIFF
--- a/lang/tags/functions_common.py
+++ b/lang/tags/functions_common.py
@@ -3,18 +3,18 @@ from talon import Context, Module, actions, imgui, registry, settings
 ctx = Context()
 mod = Module()
 
-mod.list("code_functions", desc="List of functions for active language")
+mod.list("code_common_function", desc="List of common functions for active language")
 
 # global
 function_list = []
 
-@mod.capture(rule="{user.code_functions}")
+@mod.capture(rule="{user.code_common_function}")
 def code_functions(m) -> str:
     """Returns a function name"""
-    return m.code_functions
+    return m.code_common_function
 
-mod.tag("code_functions_gui", desc="Tag for enabling GUI support for common functions")
-mod.tag("code_functions_gui_showing", desc="Active when the function picker GUI is showing")
+mod.tag("code_functions_common", desc="Tag for enabling support for common functions")
+mod.tag("code_functions_common_gui_active", desc="Active when the function picker GUI is showing")
 
 @mod.action_class
 class Actions:
@@ -25,7 +25,7 @@ class Actions:
         if gui_functions.showing:
             function_list = []
             gui_functions.hide()
-            ctx.tags.discard("user.code_functions_gui_showing")
+            ctx.tags.discard("user.code_functions_common_gui_active")
         else:
             update_function_list_and_freeze()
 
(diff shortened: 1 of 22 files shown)


splondike (approved):
Looks good to me, I will leave it open for a few more days (in case others want to have a look) and will then merge it in.

lang/tags/functions_common.py
```diff
@@ -25,15 +25,15 @@ def code_toggle_functions():
         if gui_functions.showing:
             function_list = []
             gui_functions.hide()
-            ctx.tags.discard("user.code_functions_gui_showing")
+            ctx.tags.discard("user.code_functions_common_gui_active")
```
rntz, line 28: This method doesn't exist, because `ctx.tags` is a frozenset (immutable) - the right approach is to set `ctx.tags` to an updated set, `ctx.tags = ...`. Since the code is already broken, no objection to merging, but would be nice to fix this at some point.

lang/tags/functions_common.py
```diff
@@ -46,13 +46,13 @@ def code_insert_function(text: str, selection: str):
...
-    if "user.code_functions" in registry.lists:
-        function_list = sorted(registry.lists["user.code_functions"][0].keys())
+    if "user.code_common_function" in registry.lists:
+        function_list = sorted(registry.lists["user.code_common_function"][0].keys())
     else:
         function_list = []
 
     gui_functions.show()
-    ctx.tags.add("user.code_functions_gui_showing")
+    ctx.tags.add("user.code_functions_common_gui_active")
```
rntz, line 55: ditto, tags.add does not exist.

lang/tags/functions_common.py
```diff
@@ -3,18 +3,18 @@
...
 
-mod.list("code_functions", desc="List of functions for active language")
+mod.list("code_common_function", desc="List of common functions for active language")
 
 # global
 function_list = []
 
-@mod.capture(rule="{user.code_functions}")
+@mod.capture(rule="{user.code_common_function}")
 def code_functions(m) -> str:
```
rntz, line 12: If we're going to rename the list, shouldn't we also rename the capture?

rntz (changes requested):
Looks mostly good, but if we're going to rename the list, we should also rename the corresponding capture.

rntz:
Also we recently merged rust support so you'll need to update that to use the new tag etc.

Issue #801, opened 2022-03-14
[Feature Request] Smooth Scrolling


douglasg14b:
It would be nice to have smooth scrolling instead of scrolling jumping.  Jumping tends to cause a loss of position.

 I have already implemented this and will have a PR that links to this request.

PR #803, opened 2022-03-16
Adds Smooth Scroll Behavior to mouse.py

douglasg14b:
This adds smooth scrolling as the default, however, it may be desirable for this to be controlled via settings to toggle between smooth/jump scrolling?

Related to #801 

<hr>

This provides consistent smooth scrolling behavior,  that when multiplied with `third`, `second`...etc has an effective scroll speed increase based on the multiplier. Ie. scrolling 4th scrolls 4x as far, but in the same amount of time as a single scroll.

## Testing:

In `mouse.talon` replace the `scroll up` & `scroll down` commands:

```
scroll down: user.mouse_scroll_down_smooth(5)
scroll up: user.mouse_scroll_up_smooth(5)
```

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -97,7 +97,12 @@ setting_mouse_wheel_horizontal_amount = mod.setting(
     default=40,
     desc="The amount to scroll left/right",
 )
-
+setting_mouse_wheel_smooth_scroll_speed = mod.setting(
+    "mouse_wheel_smooth_scroll_speed",
+    type=int,
+    default=35,
+    desc="The speed in which smooth scrolling occurs, ms per line",
+)
 continuous_scoll_mode = ""
 
 
@@ -185,6 +190,13 @@ class Actions:
         """Scrolls down"""
         mouse_scroll(amount * setting_mouse_wheel_down_amount.get())()
 
+    def mouse_scroll_down_smooth(amount: float = 1):
+        """Scrolls down smoothly"""
+        global continuous_scoll_mode
+        continuous_scoll_mode = "scroll down smooth"
+        mouse_scroll(setting_mouse_wheel_down_amount.get())()
+        start_smooth_scroll(amount)
+
     def mouse_scroll_down_continuous():
         """Scrolls down continuously"""
         global continuous_scoll_mode
@@ -200,6 +212,13 @@ class Actions:
     def mouse_scroll_up(amount: float = 1):
         """Scrolls up"""
         mouse_scroll(-amount * setting_mouse_wheel_down_amount.get())()
+
+    def mouse_scroll_up_smooth(amount: float = 1):
+        """Scrolls down smoothly"""
+        global continuous_scoll_mode
+        continuous_scoll_mode = "scroll up smooth"
+        mouse_scroll(-1 * setting_mouse_wheel_down_amount.get())()
+        start_smooth_scroll(-amount)
 
     def mouse_scroll_up_continuous():
         """Scrolls up continuously"""
@@ -311,7 +330,6 @@ def mouse_scroll(amount):
 
     return scroll
 
-
 def scroll_continuous_helper():
     global scroll_amount
     # print("scroll_continuous_helper")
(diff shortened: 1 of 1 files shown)


code/mouse.py
```diff
@@ -201,6 +213,13 @@ def mouse_scroll_up(amount: float = 1):
         """Scrolls up"""
         mouse_scroll(-amount * setting_mouse_wheel_down_amount.get())()
 
+    def mouse_scroll_up_smooth(amount: float = 1):
+        """Scrolls down smoothly"""
```
douglasg14b, line 217:
```suggestion
        """Scrolls up smoothly"""
```

code/mouse.py
```diff
@@ -320,6 +338,27 @@ def scroll_continuous_helper():
...
+        actions.mouse_scroll(y=int(scroll_amount))
+
+    def callback():
+        smooth_scroll_helper(amount, count + 1)
+    
+    cron.after(f'{setting_mouse_wheel_smooth_scroll_speed.get()}ms', callback)
+
+def start_smooth_scroll(amount):
+    print(f'start_smooth_scroll {amount}')
+    global scroll_job
```
rntz, line 359: This `global` declaration seems to be dead code.

rntz (commented):
Unfortunately I don't use the scroll commands and I find it a bit difficult to tell what's going on with this code. Would like to test it out but I don't have the time right now. Gonna label this "help wanted" and hopefully someone who is familiar with this code/uses this codepath can help.

douglasg14b:
@rntz 

It should be **very** easy to test out!

In `mouse.talon` replace the `scroll up` & `scroll down` commands:

```
scroll down: user.mouse_scroll_down_smooth(5)
scroll up: user.mouse_scroll_up_smooth(5)
```

And now you can test & see how it feels.

I didn't include commands for this since I assumed those could be hashed out in this PR discussion, and then added.

douglasg14b:
As for how this works, and the understandability. I would have prefered to refactor the class to make this easier to both read & modify, as there are some not-insignificant structural issues with how the entire mouse behavior is put together atm, but that seemed well out of scope.

**How this works:**

1. call `mouse_scroll_down_smooth(amount)`
   * We set the `continuous_scoll_mode` (Yes, the name is typod, but didn't want to modify that here)
   * We do a `mouse_scroll` (As this ALSO sets up some state we need)
   * We call `start_smooth_scroll()` which then calls `smooth_scroll_helper()` (The intermediate was there for some setup we had to do before, but probably don't need anymore)


**smooth_scroll_helper**:

smooth_scroll_helper simply recursively gets called by cron callbacks to delay each scroll action by the configured delay. Keeping track of the desired scroll count, and the number of times scrolled. When the count is reached it stops scrolling & exits.

nriley:
We tried this out. At the default settings it seems to be quite "jittery". Is there a platform / app / settings where this feels smoother? Could you consider recording a short screen video of how it's supposed to behave?

splondike:
I think there's a parameter on the actions.mouse_scroll action that the existing commands end up using which might let you do smooth scrolling out of the box. See the by_lines parameter in this REPL output:
```
>>> help(actions.mouse_scroll)
Help on ActionPath in module talon.scripting.actions:

mouse_scroll = mouse_scroll(y: float = 0, x: float = 0, by_lines: bool = False)
    Scroll the mouse wheel
    Scroll the mouse wheel
```

To progress this feature I think we should first evaluate whether that by_lines thing does the job and use it if so. If it doesn't then we should get some confirmation of whether the manual implementation in this PR is actually smooth across a few OS-es.

If both the by_lines and cron based implementations aren't smooth then it's probably not worth making any changes. If one of them is smooth I think a nice interface would be to use a setting that swaps the existing commands between 'default scroll' and 'smooth scroll' modes.

douglasg14b:
Do you have a gif of the jittery motion?

"smooth" scroll won't be perfectly smooth since it still has to scroll by increments, but 10 steps vs 1 step is still 10x smoother. The point is to make it feel like scrolling with your mouse, as opposed to jumping, as jumping is quite disorientating.

splondike:
I took a look at `by_lines`, it's not what we want. I think it just alters the units of the scroll action from something like pixels to something like lines of text. 

As for your change here, its default 35ms settings work pretty well for me on M1 OSX Ventura; any jitter is barely perceptible to me.

Linux doesn't have a great scrolling experience in general with Talon. For example a `by_lines=False` scroll amount of '10' on OSX is roughly the same as 400 on Linux and some apps seem to suppress some of the scroll events or batch them. Building on that unsteady base your feature is also a little inconsistent depending on the app. On Linux I've had much more consistency with `by_lines=True`, and I've altered my commands to use that.

I don't have access to Windows to test it there.

So in my view we could implement this feature by overriding mouse_scroll_up/down/left/right to check a setting like 'mouse_scroll_smooth' that defaults to false. If the setting is true then your code would be used.

splondike (changes requested):
I'm happy with this feature if we change it to work with the existing commands and a new setting.

douglasg14b:
Sure thing, I also have a separate implementation option from a chat on Slack a while back that may be more preferable. I plan on swinging back around to this in the near future (hopefully), just very busy right now both at and out of work.

splondike:
@douglasg14b It's been a little while, so should we just close out this PR and you can raise another one with your new implementation if/when you get a chance to come back to it?

douglasg14b:
Yeah, might as well. Thought I'd swing back around to this, but just haven't been able to.

PR #808, opened 2022-03-25
Enable window next/last on windows platform.

codecat555:
Use win32 calls to enable 'window next'/'window last' commands on the windows platform, without requiring third party software.

### DIFF
--- a/code/platforms/win/app.py
+++ b/code/platforms/win/app.py
@@ -1,10 +1,46 @@
 # defines the default app actions for windows
 
-from talon import Context, actions
+import win32gui
+from talon import Context, actions, ui
 ctx = Context()
 ctx.matches = r"""
 os: windows
 """
+
+# adapted this from code posted by @PeterLinder
+# return list of top level windows
+def get_top_level_windows():
+    def enumHandler(hwnd, resultList):
+        resultList.append(hwnd)
+
+    top_level_windows = []
+
+    win32gui.EnumWindows(enumHandler, top_level_windows)
+    return top_level_windows
+
+# focuses the n-th next or n-th previous instance of the current
+# app, depending on the magnitude and sign of the 'direction' arg
+def _focus_neighbor_window(direction: int) -> ui.Window:
+    active_window = ui.active_window()
+    active_app = active_window.app
+
+    top_level_windows = get_top_level_windows()
+    
+    # don't use .hidden for now, as the cached value may be stale (https://github.com/talonvoice/talon/issues/494#issuecomment-1059517184)
+    # app_windows = [w for w in ui.windows() if w.app.name == active_app.name and not w.hidden and w.id in top_level_windows]
+    app_windows = [w for w in ui.windows() if w.app.name == active_app.name and win32gui.IsWindowVisible(w.id) and w.id in top_level_windows]
+    app_windows.sort(key=lambda w: w.id)
+    window_count = len(app_windows)
+    # print(f'_focus_neighbor_window: app_windows: {len(app_windows)=}, {[w.id for w in app_windows]}')
+    
+    for i in range(window_count):
+        if app_windows[i].id == active_window.id:
+            target_idx = (i + direction) % window_count
+            target_window = app_windows[target_idx]
+            
+            actions.user.switcher_focus_window(target_window)
+
+            break
 
 @ctx.action_class('app')
 class AppActions:
@@ -28,11 +64,10 @@ class AppActions:
         actions.key('alt-space n')
     def window_hide_others():
         actions.key('win-d alt-tab')
-        #requires easy window switcher or equivalent (built into most Linux)
     def window_next():
-        actions.key('alt-`')
+        _focus_neighbor_window(1)
     def window_open():
         actions.key('ctrl-n')
         #requires easy window switcher or equivalent (built into most Linux)
     def window_previous():
-        actions.key('alt-shift-`')
+        _focus_neighbor_window(-1)
--- a/code/platforms/win/app.py
+++ b/code/platforms/win/app.py
@@ -1,14 +1,22 @@
 # defines the default app actions for windows
 
-import win32gui
-from talon import Context, actions, ui
+
+from talon import Context, actions, ui, app
+
+# we expect this import to succeed on windows (only)
+if app.platform == 'windows':
+    import win32gui
+
 ctx = Context()
 ctx.matches = r"""
 os: windows
 """
 
 # adapted this from code posted by @PeterLinder
-# return list of top level windows
+# this is a wrapper around the win32 EnumWindows() function, which is documented
+# here - https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumwindows.
+# you might think that 'top level windows' are non-hidden, non-background windows that are
+# not child windows...but you'd be wrong.
 def get_top_level_windows():
     def enumHandler(hwnd, resultList):
         resultList.append(hwnd)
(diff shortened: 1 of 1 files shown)


code/platforms/win/app.py
```diff
@@ -1,11 +1,47 @@
 # defines the default app actions for windows
 
-from talon import Context, actions
+import win32gui
```
rntz, line 3: `import win32gui` will cause an ImportError if we're not on windows, so you need to try/except fence this.

code/platforms/win/app.py
```diff
@@ -1,11 +1,47 @@
...
+from talon import Context, actions, ui
 ctx = Context()
 ctx.matches = r"""
 os: windows
 """
 
+# adapted this from code posted by @PeterLinder
+# return list of top level windows
```
rntz, line 11: what are top-level windows? could you add a comment here explaining the need for this?

code/platforms/win/app.py
```diff
@@ -1,11 +1,47 @@
...
+    def enumHandler(hwnd, resultList):
+        resultList.append(hwnd)
+
+    top_level_windows = []
+
+    win32gui.EnumWindows(enumHandler, top_level_windows)
+    return top_level_windows
+
+# focuses the n-th next or n-th previous instance of the current
+# app, depending on the magnitude and sign of the 'direction' arg
```
rntz, line 22: this should be a docstring instead of a comment but w/e

rntz (changes requested):
Looks good, except this will cause errors for everyone not on windows b/c the `import win32gui` will fail. You should try/except wrap it, and make sure the file runs correctly if the import fails (eg. force it to produce an error and check nothing errors downstream).

code/platforms/win/app.py
codecat555: Okay, used 'if' instead of 'try' because we don't want to ignore import errors when we are on windows (and so we would have needed an 'if' anyways).

code/platforms/win/app.py
codecat555: Don't rightly know, TBH...except that we want to exclude non-top-level windows while switching with this command. The m$ docs could be clearer on this point.

Added a (snarky) comment.

code/platforms/win/app.py
codecat555: Done.

knausj85:
This'll be awesome to get merged, though this does have some issues that easy window switcher does not. I think we should work to make it a bit more reliable first.

- [ ] Windows File Explorer in particular is a challenge since about a thousand things run under the explorer.exe, have windows, and are "visible." I ran into issues with two file explorer windows open--it would try to switch to other things. This may be more Talon side, not sure.

- [ ] I also think it's worth using easy window switcher **if** it's running via user.running to avoid exposing those users any new bugs once this does get merged.

- [ ] I've also seen a fair number of "can't focus window" errors with apps like beyond compare & vscode. I'm not sure what's going on there yet.

codecat555:
Sigh, seems windows is always a headache.

* I see what you mean about explorer.exe. I want to assume that the user will want to switch between windows showing the file system and ignore everything else - does that sound reasonable? If so, looks like we can filter out all explorer.exe windows that do not show a valid path in the title. What does EWS do?

* Makes sense to retain backwards compatibility with EWS. Can you share an example of what the code should look for in user.running to make this determination?

* Does EWS do the right thing for those "cant' focus window" apps?

knausj85:
I haven't seen these issues with EWS.

For reference, I've reimplemented EWS using user.running:
https://github.com/knausj85/knausj_talon/blob/91c5d13ca96804e3cd719eaa1f2495283bbc6629/apps/platforms/win/easy_window_switcher/easy_window_switcher.py#L17

I'll submit a PR eventually

codecat555:
I pushed a change to handle windows explorer. It seems to work for me, let me know what you see.

I guess we still need to figure out the "can't focus window" errors you've seen.

codecat555:
Oops, I forgot to test things like 'Desktop', 'Downloads', etc. I'll have to figure out how to handle those explorer.exe cases too, I guess.

codecat555:
Wow, now I see all the work that you've done in explorer.py - I'm so glad you figured this stuff out before I got to it. I'll take another stab at fixing things after I've had a chance to grok the work you've already done.

knausj85:
I almost wonder if we'd better better off exploring using this guy:
https://docs.microsoft.com/en-us/windows/win32/api/appmodel/nf-appmodel-getapplicationusermodelid?redirectedfrom=MSDN
and having Talon group windows via that. 

The app user model id appears to be what windows itself uses in e.g. the taskbar, if this documentation is to be believed.
https://docs.microsoft.com/en-us/windows/win32/shell/appids

codecat555:
Yep, app user model id does seem to be the right thing to use.

I can't seem to get SHGetPropertyStoreForWindow() to work for me, however.

Well, it returns a PyIPropertyStore reference, but GetCount() always returns 0 and I can't retrieve the AppUserModel_ID.

I tried with explorer, firefox and vscode windows.

Must be doing something wrong. I've been perusing docs and examples, and tried variations, but I don't see the problem.

When I have time later, maybe I'll try with c# or something, just in case there's a pywin32com issue.

```
ps = propsys.SHGetPropertyStoreForWindow(w.id, propsys.IID_IPropertyStore)
id = ps.GetValue(pscon.PKEY_AppUserModel_ID).GetValue()
```

```
>>> time.sleep(3); w=ui.active_window()
>>> w
Window(app=App(pid=79332, "Firefox"), title="Slack | bayarea | Talon — Mozilla Firefox")
>>> p = propsys.SHGetPropertyStoreForWindow(w.id, propsys.IID_IPropertyStore)
>>> p.GetCount()
0
>>> i = p.GetValue(pscon.PKEY_AppUserModel_ID)
>>> i
<PyPROPVARIANT object at 0x0000000045B224E0>
>>> v = i.GetValue()
>>> v
>>> not v
True
>>> not i
False
>>> not p
False
>>>
```

codecat555:
Oh, I just looped through ui.windows() and tried that code on each. All of the 290 windows on my system returned None, except for these -

```
2022-03-27 16:57:11    IO Window(app=App(pid=3064, "Windows Explorer"), title=""): MicrosoftWindows.Client.CBS_cw5n1h2txyewy!InputApp
2022-03-27 16:57:11    IO Window(app=App(pid=3064, "Windows Explorer"), title=""): MicrosoftWindows.Client.CBS_cw5n1h2txyewy!InputApp
2022-03-27 16:57:11    IO Window(app=App(pid=11128, "Application Frame Host"), title="Microsoft Store"): Microsoft.WindowsStore_8wekyb3d8bbwe!App
2022-03-27 16:57:11    IO Window(app=App(pid=11128, "Application Frame Host"), title="Movies & TV"): Microsoft.ZuneVideo_8wekyb3d8bbwe!Microsoft.ZuneVideo
2022-03-27 16:57:11    IO Window(app=App(pid=11128, "Application Frame Host"), title="Settings"): windows.immersivecontrolpanel_cw5n1h2txyewy!microsoft.windows.immersivecontrolpanel
2022-03-27 16:57:11    IO Window(app=App(pid=11128, "Application Frame Host"), title="Chris's Notebook ‎- OneNote"): Microsoft.Office.OneNote_8wekyb3d8bbwe!microsoft.onenoteim

```

codecat555:
After some more research, I suspect we need to get an interface to the taskbar so we can get at the "jump lists" defined there.

codecat555:
Well, I haven't found any suitable taskbar interfaces. Perhaps it's available through the desktop somehow, although I didn't find it before I got tired of looking.

I've got a solution for the explorer.exe case...I think. It's not going to win a prize at any beauty contest, but I'm not sure it's much different from code I see elsewhere in the repo. Let me know if you see any problems, or if there's a better way to do it - I'm receptive and ready for enlightenment. :-)

codecat555:
I _would_ like to discover a more elegant solution. I thought I might try pointing an API monitor at various explorer.exe processes while switching windows, but I expect it may be hard to find the useful bits in among all the output. Doing the same with EWS might be quicker. Not sure when I'll have time to get back to this, however.

codecat555:
The solution I have works well for me and I don't know when I might pick this back up again. I'll close this for now.

@knausj85 , thanks for spending some time on this with me, it was educational.

PR #810, opened 2022-03-26
Enable voice management of consolidated csv settings

codecat555:
Goal: provide voice commands for supplementing user vocabulary files, e.g. 'additional word' to add a new word.

Inspired by code shared by @pokey some time back.

A secondary goal is to consolidate user vocabulary settings in one place (under the settings folder).

This change makes use of the user's default application mapping to open the .csv files. I tried to avoid security pitfalls common for this sort of use case. It seems to work on windows and (Ubuntu 21.04) linux, have not tested on mac though.

I included code to move to the end of the newly-opened file and start a new line. This code is sensitive to timing variations, however. I'm looking for suggestions on how to do this better...I tried using a 'win_focus' callback, but that fails when the default application already has focus.

### DIFF
--- a/code/abbreviate.py
+++ b/code/abbreviate.py
@@ -1,11 +1,13 @@
 # XXX - would be nice to be able pipe these through formatters
 
+from .user_settings import get_list_from_csv
+
 from talon import Context, Module
 
 mod = Module()
 mod.list("abbreviation", desc="Common abbreviation")
 
-abbreviations = {
+_abbreviation_defaults = {
     "address": "addr",
     "administrator": "admin",
     "administrators": "admins",
@@ -213,6 +215,12 @@ abbreviations = {
     "what the fuck": "wtf",
     "window": "win",
 }
+
+abbreviations = get_list_from_csv(
+    "abbreviate.csv",
+    headers=("Abbreviation", "Word(s)"),
+    default=_abbreviation_defaults
+)
 
 ctx = Context()
 ctx.lists["user.abbreviation"] = abbreviations
--- a/code/exec.py
+++ b/code/exec.py
@@ -1,10 +1,16 @@
 import os
 import subprocess
 
-from talon import Module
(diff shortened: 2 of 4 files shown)


codecat555:
I just pushed a change to make edit_talon_user_csv() more flexible by allowing both absolute and relative paths, as long as they are within the talon user directory. Also added a check to make sure the target path actually exists before trying to 'execute' it.

pokey:
Travelling so haven't had a chance to look in detail, but worth mentioning https://github.com/knausj85/knausj_talon/issues/792 in case you've not seen it

codecat555:
@pokey I hadn't seen that, thanks for sharing it.

I found the code from @wolfmanstout instructive, on multiple levels. It seems to be very well done and has some sophisticated features.

I suppose the two approaches are complementary, in that both the interactive and fully-automated use cases are valid. I don't think they should conflict with each other, as long as the voice commands are kept distinct.

wolfmanstout:
@codecat555 Thanks! I do plan to upstream that code. I continue to make improvements, so make sure you are looking at the latest version if you want to use any of it yourself.

rntz:
I was hoping to get to this today but I didn't have time. Hopefully I'll be able to take a look at it at the next knausj maintenance session in two weeks.

codecat555:
@rntz Item #1 in your list is the core of what I wanted to contribute with this PR.

The other two things were sort of incidental. My longer-term goal with these was to consolidate all the user-customizable csv files in one place, as a first step on the path toward providing some mechanism for keeping those things in a location outside of knausj_talon...so users could keep them in their own personal repos.

Since I submitted this PR, I have read through the other issues open on knausj_talon and see there has been some ongoing discussion and work related to csv files in general. Also, there seems to be a Talon bug related to the function I used in this change (get_list_from_csv).

Given all that, perhaps it best to just let this one go. The changes I suggested here might fit into the other csv work that is already in process.

PR #811, opened 2022-03-28
Add basic CSS support

auscompgeek:
This adds some basic CSS support. It intentionally does not add support for the entirety of CSS, rather opting to include in lists things I have used in the past or anticipate on using.

Opening this to gauge how people feel about this approach. I see several people have been writing their own CSS support for a while. Figured I'd make a PR with something that should be maintainable in the long term.

### DIFF
--- a/lang/css/css.talon
+++ b/lang/css/css.talon
@@ -42,5 +42,3 @@ unit {user.css_unit}: insert(css_unit)
 
 [value] current color: "currentColor"
 op important: " !important"
-
-border box model: "box-sizing: border-box;\n"
--- a/lang/css/css.talon
+++ b/lang/css/css.talon
@@ -41,6 +41,6 @@ at {user.css_at_rule}: "@{css_at_rule} "
 unit {user.css_unit}: insert(css_unit)
 
 [value] current color: "currentColor"
-important: " !important"
+op important: " !important"
 
 border box model: "box-sizing: border-box;\n"
--- a/lang/css/css.talon
+++ b/lang/css/css.talon
@@ -1,0 +1,46 @@
+tag: user.css
+-
+tag(): user.code_comment_block_c_like
+tag(): user.code_functions_common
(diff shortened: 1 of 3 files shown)


lang/css/css.talon
```diff
@@ -0,0 +1,46 @@
...
+    name = user.formatted_text(text, settings.get("user.code_public_variable_formatter"))
+    insert("var(--{name})")
+
+op var: user.insert_between("var(--", ")")
+
+at {user.css_at_rule}: "@{css_at_rule} "
+unit {user.css_unit}: insert(css_unit)
+
+[value] current color: "currentColor"
+important: " !important"
```
splondike, line 44: Maybe 'op important'

lang/css/css.talon
```diff
@@ -0,0 +1,46 @@
...
+
+border box model: "box-sizing: border-box;\n"
```
splondike, line 46: border box model and value current color might be a bit idiosyncratic to your case, I wouldn't have used them that often the way I've worked with CSS.

splondike (commented):
This looks good to me and works well. I might suggest removing the two commands I indicated from the knausj version, but interested to hear your thoughts on that.

lang/css/css.talon
auscompgeek: border box model might be slightly idiosyncratic - it probably most commonly occur in CSS resets these days - notably Bootstrap makes that the default. I'm fine removing it for knausj.

I suppose for `currentColor` people might want to spell it `currentcolor` all lowercase? I'd argue it's pretty useful to set *-color properties without hardcoding them. Happy to remove that on the preference note though.

splondike: Happy to leave in the currentColor thing if you reckon it's useful. I'll drop out the border box model command and then merge this in, thanks!

splondike (approved):

PR #824, opened 2022-04-16
Add default TypeScript function declaration

brxck:
Currently it's not possible to create a plain old function declaration in TypeScript: "funky" defaults to a protected function, which only makes sense in the context of a class. This also makes the behavior of "funky" match the JS implementation.

---

"funky hello world"

```ts
// before
protected function helloWorld()

// after
function helloWorld()
```

### DIFF
--- a/lang/typescript/typescript.py
+++ b/lang/typescript/typescript.py
@@ -25,9 +25,9 @@ class UserActions:
                 text, settings.get("user.code_private_function_formatter")
             )
         )
-        
+
         actions.user.code_insert_function(result, None)
-        
+
     def code_private_function(text: str):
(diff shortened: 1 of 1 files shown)


splondike:
I get 'funky hello world' -> 'private function helloWorld()' on current master, which seems a bit more useful than protected. So I think this is probably an improvement, but it would stop people from accessing the private function thing.

I think perhaps the suggestion in #825 would be a good way of dealing with this problem. I'll let that issue sit around for a few more days (in case others want to comment) and then might look at implementing it if nobody else takes it on by then.

splondike:
This has been superceeded by #879 . Closing PR.

PR #827, opened 2022-04-21
Fixed broken Windows implementation of find().

wolfmanstout:


auscompgeek:
iirc the Linux implementation is broken in the same way.

wolfmanstout:
> iirc the Linux implementation is broken in the same way.

Fixed.

splondike (approved):
Thanks!

PR #830, opened 2022-04-24
Add some javascript functions

pokey:


lang/javascript/javascript.py
```diff
@@ -55,6 +87,7 @@ def code_state_switch():
 
     def code_state_case():
         actions.auto_insert("case :")
+        actions.key("left")
```
pokey, line 90: Bug fix

lang/javascript/javascript.talon
```diff
@@ -25,6 +26,7 @@ settings():
 
 (op | is) strict equal: " === "
 (op | is) strict not equal: " !== "
+op null coal: " ?? "
```
pokey, line 29: Open to other terms here

lang/javascript/javascript.py
```diff
@@ -6,6 +6,38 @@
...
+    "max": "Math.max",
+    "min": "Math.min",
+    "abs": "Math.abs",
+    "round": "Math.round",
+    "floor": "Math.floor",
+    "print": "console.log",
+    "values": "Object.values",
+}
+
+mod.list("code_chain_function", "Function to use in a dotted chain, eg .foo()")
```
auscompgeek, line 23: I noticed C# puts `.ToString` in the `code_common_function` list. Maybe doing the same here would be fine?

pokey: Yeah I guess we could do it that way?  These chained functions are currently accessed using "state", so I was thinking to stick with that approach

pokey: Fwiw I personally use "chain" for these

pokey: I think of them a bit differently from "funk".  For example, I'm not sure it makes sense to wrap with them, etc

lang/javascript/javascript.py
auscompgeek, line 90:
```suggestion
        actions.user.insert_between("case ", ":")
```

pokey: nice

lang/javascript/javascript.talon
david-tejada: How about `null or`?

lang/javascript/javascript.py
```diff
@@ -6,6 +6,38 @@
...
+    "entries": "Object.entries",
+    "flatten": "flatten",
+    "from entries": "Object.fromEntries",
+    "keys": "Object.keys",
+    "max": "Math.max",
+    "min": "Math.min",
+    "abs": "Math.abs",
+    "round": "Math.round",
+    "floor": "Math.floor",
+    "print": "console.log",
```
nriley, line 19: Consider log as a synonym.

lang/javascript/javascript.talon
```diff
@@ -38,14 +40,9 @@ state async: "async "
...
-state map:
-    user.insert_between(".map(", ")")
+state {user.code_chain_function}:
+    user.insert_between(".{code_chain_function}(", ")")
 
-state filter:
-    user.insert_between(".filter(", ")")
-
-state reduce:
-    user.insert_between(".reduce(", ")")
```
nriley, line 48: Consider outputting a deprecation warning on the use of `state` here then add `chain` for user.code_chain_function.

lang/javascript/javascript.talon
nriley: Agree `null or` makes more sense here.

nriley (changes requested):

lang/javascript/javascript.py
pokey: @nriley and I decided to use "chain" for this one, and then add deprecation notifications on the old "state" ones

auscompgeek: I wonder if chain is the right word. I'd be inclined to call it meth, short for method.

pokey: "meth" is...interesting 😅. Not sure that's gonna be the best word to be spamming in a workplace setting, but that aside, it's not a bad term 🤔

I'd also be inclined in the future to also use the same word for things like `.length`, which would argue for something other than "meth", eg "chain", "prop", or "field"

I still prefer "chain", as I find it intuitive, but I don't feel super strongly; I'll prob keep "chain" on my fork either way.

Cc/ @knausj85 @AndreasArvidsson @nriley any leaning here?

rntz: No real opinion about what voice keyword to use (chain seems fine), but the list should perhaps be called common_methods, since that's what they are.

AndreasArvidsson: Personally I don't separate common functions with or without `.`

lang/javascript/javascript.py
AndreasArvidsson: This kind of thing should preferably use snippets

pokey: snippets are nice, but generally knausj repo doesn't depend on them, prob in part so it can work anywhere without snippet support.  tho i have a vague recollection that you have some impl of snippets that can work in more generic contexts using a fallback?

lang/javascript/javascript.py
pokey: ah interesting.  so you just have them in the "funk" list with a `.` at the start?

lang/javascript/javascript.py
AndreasArvidsson: Yes I have a snipped system that fall backs on text insertion. It even calculates the number of arrow key steps it has to take to get to the default tab stop. 
https://github.com/AndreasArvidsson/andreas-talon/blob/ec039ba885e41c59943a722a05883e8d9460ea68/text/snippets.py#L15

lang/javascript/javascript.py
AndreasArvidsson: No. I used to have a system like this: `dot {user.code_function}: ".{code_function}"`. That way I could use all functions with or without a dot prefix. You could of course separate them into distinct lists if you want.

lang/javascript/javascript.py
knausj85: This snippet thing sounds great, just never made the time. :)

lang/javascript/javascript.py
nriley: We also have dot… for filename extensions in knausj. Please not "meth", thanks :-)

auscompgeek: Haha, fair enough re: "meth". Probably worth calling the list `code_common_method` though.

pokey: Hmm yeh "dot" is nice. Shall we go with that?

lang/javascript/javascript.py
pokey: Neat!  Filed https://github.com/knausj85/knausj_talon/issues/848 to track

lang/javascript/javascript.py
```diff
@@ -1,11 +1,44 @@
...
 mod = Module()
 ctx = Context()
 ctx.matches = """
 tag: user.javascript
 """
 
+ctx.lists["user.code_common_function"] = {
+    "abs": "Math.abs",
+    "entries": "Object.entries",
+    "flatten": "flatten",
```
auscompgeek, line 12: Just wondering where this comes from?

lang/javascript/javascript.py
```diff
@@ -1,11 +1,44 @@
...
+ctx.lists["user.code_common_member_function"] = {
+    "concat": "concat",
+    "filter": "filter",
+    "find": "find",
+    "flat map": "flatMap",
+    "for each": "forEach",
+    "includes": "includes",
+    "map": "map",
+    "push": "push",
+    "reduce": "reduce",
```
auscompgeek, line 35:
```suggestion
    "reduce": "reduce",
    "slice": "slice",
```

lang/javascript/javascript.py
```diff
@@ -1,11 +1,44 @@
...
+    "some": "some",
+    "split": "split",
+    "substring": "substring",
+    "then": "then",
```
auscompgeek, line 39: If we're including Promise#then, how about catch and finally?

lang/javascript/javascript.py
```diff
@@ -1,11 +1,44 @@
...
+mod.list("code_common_member_function", "Function to use in a dotted chain, eg .foo()")
+
+ctx.lists["user.code_common_member_function"] = {
+    "concat": "concat",
+    "filter": "filter",
+    "find": "find",
+    "flat map": "flatMap",
+    "for each": "forEach",
+    "includes": "includes",
+    "map": "map",
```
auscompgeek, line 33:
```suggestion
    "map": "map",
    "pop": "pop",
```

auscompgeek (approved):
LGTM. I have a few suggestions for more entries.

lang/javascript/javascript.py
pokey: ah right; looks like this one should be `flat`?

auscompgeek: Would that be referring to Array#flat? I guess then it should be in the other list.

pokey: ah wait nvm this is in common functions.  It's from lodash.  Shall we leave it out since it's not standard?

pokey: Deleted, since it's not standard

pokey: Is `flat` common enough to include?  I could go either way

pokey: Btw tempted to make these into csvs so users can add their own common functions

pokey:
Ok I think this one is good to go

lang/javascript/javascript.py
auscompgeek: I've honestly never seen flat used.

lang/javascript/javascript.py
```diff
@@ -1,11 +1,47 @@
...
 
 mod = Module()
 ctx = Context()
 ctx.matches = """
 tag: user.javascript
 """
 
+ctx.lists["user.code_common_function"] = {
+    "abs": "Math.abs",
+    "entries": "Object.entries",
```
auscompgeek, line 11:
```suggestion
    "entries": "Object.entries",
    "fetch": "fetch",
```

lang/javascript/javascript.py
```diff
@@ -1,11 +1,47 @@
...
+mod.list("code_common_member_function", "Function to use in a dotted chain, eg .foo()")
+
+ctx.lists["user.code_common_member_function"] = {
+    "catch": "catch",
+    "concat": "concat",
+    "filter": "filter",
+    "finally": "finally",
+    "find": "find",
+    "flat map": "flatMap",
+    "for each": "forEach",
```
auscompgeek, line 32:
```suggestion
    "for each": "forEach",
    "join": "join",
```

auscompgeek (commented):
A couple more I just found in my local changes (last ones, I promise!).

lang/javascript/javascript.talon
```diff
@@ -25,6 +26,7 @@ settings():
 
 (op | is) strict equal: " === "
 (op | is) strict not equal: " !== "
+op null or: " ?? "
```
pokey, line 29: @david-tejada mentioned on slack that accuracy for "null or" hasn't been great. I've been using "null coal", which works _ok_. @david-tejada suggested "nullish" seems to work ok, tho the fact it's not a real word makes me a little nervous. Maybe "dub quest" (short for "double question mark")?

auscompgeek: How about `null else`?

knausj85 (approved):

PR #838, opened 2022-04-29
VSCode: improve actions `selected_text` and `line_insert_down`

dgrunwald:
* Use heuristic to improve compatibility with "copy entire line if no selection" setting
 *  Fix slap command if cursor is already at end-of-line and a completion popup is open.

### DIFF
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -112,6 +112,12 @@ class EditActions:
     def line_clone():
         actions.key("shift-alt-down")
 
+    def line_insert_down():
+        actions.user.vscode("editor.action.insertLineAfter")
+
+    def line_insert_up():
+        actions.user.vscode("editor.action.insertLineBefore")
+
     def jump_line(n: int):
         actions.user.vscode("workbench.action.gotoLine")
         actions.insert(str(n))
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -118,6 +118,20 @@ class EditActions:
         actions.key("enter")
         actions.edit.line_start()
 
+    def selected_text() -> str:
+        with clip.capture() as s:
+            actions.edit.copy()
+        try:
+            txt = s.text()
+        except clip.NoChange:
+            return ""
+        if txt and txt[-1] == '\n' and sum(1 for c in txt if c == '\n') == 1 and txt not in ('\n', '\r\n'):
+            # This looks like the special case where VSCode copies the full line
+            # when given an empty selection.
+            return ""
+        else:
+            return txt
+
 
 @ctx.action_class("win")
 class WinActions:
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -124,20 +124,6 @@ class EditActions:
         actions.key("enter")
         actions.edit.line_start()
 
-    def selected_text() -> str:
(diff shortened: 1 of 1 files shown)


apps/vscode/vscode.py
```diff
@@ -112,12 +112,31 @@ def line_swap_down():
     def line_clone():
         actions.key("shift-alt-down")
 
+    def line_insert_down(): # slap
+        actions.edit.line_end()
+        actions.key("escape")  # close completion popup
+        actions.key("enter")
```
pokey, line 118: VSCode has a built-in command for this. Why not just use that instead?

apps/vscode/vscode.py
```diff
@@ -112,12 +112,31 @@ def line_swap_down():
...
         actions.edit.line_start()
 
+    def selected_text() -> str:
+        with clip.capture() as s:
+            actions.edit.copy()
+        try:
+            txt = s.text()
+        except clip.NoChange:
+            return ""
+        if txt and txt[-1] == '\n' and sum(1 for c in txt if c == '\n') == 1 and txt not in ('\n', '\r\n'):
```
pokey, line 133: Won't this break if someone has actually selected a line to copy it?

pokey (changes requested):
Left a couple comments / questions

apps/vscode/vscode.py
dgrunwald: Yes, it's a heuristic.
I think it's rare to manually copy a full line including the line terminator -- e.g. `take line` doesn't include the terminator. And a manual selection with `shift-down` will tend to include the indentation of the following line. So I think it's more likely that the selection was empty and the user didn't turn off the VSCode setting, than that the user really wanted this type of selection.

Also note that this is just about the parts of talon that care about `selected_text` -- the actual copy action will still work even if this heuristic was wrong.

pokey: Mixed feelings. I agree that it will be vastly more common for this heuristic to help than hurt, but when it does hurt, it will be quite surprising

We could do this using a VSCode action if we added one to the talon extension, but probably good to have a fallback even so

Any thoughts @knausj85 ?

knausj85: I'm a bit hesitant since eg selecting lines via the line numbers does include the terminator, so this could break kinda often depending on folks workflow (eg eye tracking, some folks can use a mouse)

I'm on the fence atm

pokey: I would be inclined to leave this part out, and instead add something to a VSCode extension, eg the Talon VSCode extension.  

What might make sense is to add something to the [knausj VSCode README](https://github.com/knausj85/knausj_talon/blob/master/apps/vscode/README.md) suggesting to turn off the setting to copy entire line, and if they don't, put the code snippet in there, or maybe link to something in the wiki

knausj85: That sounds reasonable to me!

rntz: I'm going to remove this controversial bit from the PR and then merge the rest of it.

rntz (approved):

rntz:
hey @pokey I'd appreciate you testing whether these changes to line_insert_down and line_insert_up still work. You looked at them before and didn't object so I just went ahead and merged on the assumption they do.

PR #842, opened 2022-05-04
Quickly add new vocabulary to additional_words.csv via voice command.

tararoys:
This is a voice command to quickly add highlighted words to additional_words.csv from dictation mode.  

I don't think this is necessarily the best way to implement this, because 

- I don't check to see if the word is already present in additional_words.csv
- I'm using f.write and string-concatinating a newline in order to add the file.  I saw in user_settings.py there was a csv writer, but I don't know how to use it.


nriley:
From Talon grooming session — this seems like a subset of #846. If you have any comments on that PR please share them, thanks!

Issue #792, opened 2022-03-01
Make adding to additional_words.csv quicker


bra1nDump:
My most common reason for opening user directory is adding to additional_words.csv. 

Should we:
* Option 1: add a command to add the selection additional_words
* Option 2: add a shortcut to open additional_words or other settings file

phillco:
@wolfmanstout did (1) in his fork :)

wolfmanstout:
Yeah, I made a fancy version that optionally lets you supply a spoken form, which it will test to ensure that the new vocabulary works (and if not, it adds the spoken form). It depends on some private APIs so it's not something I can push upstream yet, but I plan to factor it out into a separate repository at some point:
https://github.com/wolfmanstout/wolfmanstout_talon/blob/4cb0578bf3f2059f3f647d2ce7b41ac3da143243/code/vocabulary.py#L185

rntz:
If we wanted to do (2), is there an action that opens a particular file in the user's preferred editor? Is there even a good way to do this cross-platform? We could definitely have the action even if not, so that users can override it to their preference, but it would be neat if we could just Do The Right Thing.

nriley:
Addressed in #846 and #847.

PR #846, opened 2022-05-05
Added commands for adding vocabulary words/phrases

wolfmanstout:
Implements Option 1 in #792.

See also https://github.com/wolfmanstout/talon-vocabulary-editor/blob/main/vocabulary.py for a fancier version of this that depends on unstable APIs (for parse_phrase.py). That is intended to override this simpler implementation. There is some duplication of code, unfortunately, but I think that would be hard to avoid without exposing a bunch of functions as actions that don't need to be otherwise (i.e. the CSV functions). Open to feedback, though!

### DIFF
--- a/code/user_settings.py
+++ b/code/user_settings.py
@@ -1,7 +1,7 @@
 import csv
 import os
 from pathlib import Path
-from typing import Dict, List, Tuple
+from typing import Dict, Tuple
 from talon import resource
 
 # NOTE: This method requires this module to be one folder below the top-level
@@ -13,13 +13,13 @@ if not SETTINGS_DIR.is_dir():
 
 
 def get_list_from_csv(
-    filename: str, headers: Tuple[str, str], default: Dict[str, str] = {}
+    filename: str, headers: Tuple[str, str], default: Dict[str, str] = {}, write_default=True
 ):
     """Retrieves list from CSV"""
     path = SETTINGS_DIR / filename
     assert filename.endswith(".csv")
 
-    if not path.is_file():
+    if write_default and not path.is_file():
         with open(path, "w", encoding="utf-8", newline="") as file:
             writer = csv.writer(file)
             writer.writerow(headers)
@@ -58,3 +58,19 @@ def get_list_from_csv(
             mapping[spoken_form] = output
 
     return mapping
+
+def append_to_csv(filename: str, rows: Dict[str, str]):
+    path = SETTINGS_DIR / filename
+    assert filename.endswith(".csv")
+
+    with open(str(path), "r") as file:
+        line = None
+        for line in file:
+            pass
+        needs_newline = line is not None and not line.endswith("\n")
+    with open(path, "a", encoding="utf-8", newline="") as file:
+        if needs_newline:
+            file.write("\n")
+        writer = csv.writer(file)
+        for key, value in rows.items():
+            writer.writerow([key] if key == value else [value, key])
--- a/code/user_settings.py
+++ b/code/user_settings.py
@@ -69,8 +69,8 @@ def append_to_csv(filename: str, rows: Dict[str, str]):
             pass
         needs_newline = line is not None and not line.endswith("\n")
     with open(path, "a", encoding="utf-8", newline="") as file:
-        if needs_newline:
-            file.write("\n")
         writer = csv.writer(file)
+        if needs_newline:
+            writer.writerow([])
         for key, value in rows.items():
             writer.writerow([key] if key == value else [value, key])
--- a/code/user_settings.py
+++ b/code/user_settings.py
@@ -13,13 +13,13 @@ if not SETTINGS_DIR.is_dir():
 
 
 def get_list_from_csv(
-    filename: str, headers: Tuple[str, str], default: Dict[str, str] = {}, write_default=True
+    filename: str, headers: Tuple[str, str], default: Dict[str, str] = {}
 ):
     """Retrieves list from CSV"""
     path = SETTINGS_DIR / filename
     assert filename.endswith(".csv")
 
-    if write_default and not path.is_file():
+    if not path.is_file():
         with open(path, "w", encoding="utf-8", newline="") as file:
             writer = csv.writer(file)
             writer.writerow(headers)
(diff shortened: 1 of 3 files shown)


code/vocabulary.py
```diff
@@ -159,3 +161,54 @@ def replace_words(words: Sequence[str]) -> Sequence[str]:
...
+def _add_selection_to_csv(
+    phrase: Union[Phrase, str], type: str, csv: str, headers: Tuple[str, str]
+):
+    written_form = actions.edit.selected_text().strip()
+    if phrase:
+        spoken_form = " ".join(actions.dictate.parse_words(phrase))
+    else:
+        is_acronym = re.fullmatch(r"[A-Z]+", written_form)
+        spoken_form = " ".join(written_form) if is_acronym else written_form
+    phrases = get_list_from_csv(csv, headers=headers, write_default=False)
```
rntz, line 188: I think this is wrong, because get_list_from_csv causes the file to be watched and auto-reloads when it changes, whereas here you want a one-time read of the csv file.

code/vocabulary.py
```diff
@@ -159,3 +161,54 @@ def replace_words(words: Sequence[str]) -> Sequence[str]:
...
+    """Expands the provided spoken form and written form into multiple variants based on
+    the provided type, which can be either "name" to add a possessive variant or "noun"
+    to add plural.
+    """
+    entries = {spoken_form: written_form}
+    if type == "name":
+        entries[f"{spoken_form}s"] = f"{written_form}'s"
+    elif type == "noun":
+        entries[f"{spoken_form}s"] = f"{written_form}s"
+    return entries
```
nriley, line 175: Consider vendoring/incorporating https://github.com/jpvanhal/inflection/blob/master/inflection/__init__.py ?

code/user_settings.py
```diff
@@ -58,3 +58,19 @@ def get_list_from_csv(
...
+def append_to_csv(filename: str, rows: Dict[str, str]):
+    path = SETTINGS_DIR / filename
+    assert filename.endswith(".csv")
+
+    with open(str(path), "r") as file:
+        line = None
+        for line in file:
+            pass
+        needs_newline = line is not None and not line.endswith("\n")
+    with open(path, "a", encoding="utf-8", newline="") as file:
```
nriley, line 71: Need to ensure this does the right thing on Windows, since newline="" means the \n will be written untranslated.

code/user_settings.py
```diff
@@ -58,3 +58,19 @@ def get_list_from_csv(
...
+        if needs_newline:
+            file.write("\n")
```
nriley, line 73: csv.writer uses the Excel dialect by default which always uses `\r\n` line endings regardless of platform. Consider using `csv.writerow([])`.

nriley (changes requested):
Consolidated comments from the Talon grooming session.

code/user_settings.py
wolfmanstout: Done and tested on Windows.

code/vocabulary.py
wolfmanstout: I'm open to it, but I want to point out that this feature is primarily used for custom vocabulary words, in which case I'm not sure that normal English rules are as likely to apply, so there is some benefit to the predictability of my approach here. For example, "equipment" is listed as uncountable, but if this were a class name in a programming language I could imagine that "equipments" would be appropriate. Of course there are plenty of contrary examples such as nouns ending in "y" where you'd probably want "ies" -- my point is just that there may be some benefit to the predictability. But like I said, I definitely don't feel strongly and would be happy to add this file with attribution and do a relative import.

code/vocabulary.py
```diff
@@ -159,3 +161,54 @@ def replace_words(words: Sequence[str]) -> Sequence[str]:
...
+        spoken_form = " ".join(written_form) if is_acronym else written_form
+    phrases = get_list_from_csv(csv, headers=headers, write_default=False)
+    entries = _create_vocabulary_entries(spoken_form, written_form, type)
+    new_entries = {}
+    for spoken_form, written_form in entries.items():
+        if spoken_form in phrases:
+            logging.info(f'Spoken form "{spoken_form}" is already in {csv}')
+        else:
+            new_entries[spoken_form] = written_form
+    append_to_csv(csv, new_entries)
```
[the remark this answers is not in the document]
wolfmanstout: Good idea, updated (and adjusted the feedback that is provided).

code/vocabulary.py
wolfmanstout: I just reached out to Ryan about this, because it seems undesirable that Talon would do this when resource.open is invoked from a command resolution (vs. file initialization). In a nutshell, he didn't disagree, but he wants to get away from resource.open entirely so his advice here was to work around this by simply reusing the earlier result of get_list_from_csv, but without going through ctx to avoid issues with changing the file. That is now implemented here. I also reverted my changes to get_list_from_csv which are no longer needed.

code/vocabulary.py
```diff
@@ -159,3 +162,71 @@ def replace_words(words: Sequence[str]) -> Sequence[str]:
...
+    return entries
+
+# See https://github.com/wolfmanstout/talon-vocabulary-editor for an experimental version
+# of this which tests if the default spoken form can be used instead of the provided phrase.
+def _add_selection_to_csv(
+    phrase: Union[Phrase, str],
+    type: str,
+    csv: str,
+    csv_contents: dict[str, str],
+    skip_identical_replacement: bool,
```
wolfmanstout, line 185: FYI I added this to handle the erroneous case where the user provides a no-op replacement.

wolfmanstout (commented):
Thank you for the close review!

code/vocabulary.py
nriley: I don't feel strongly about this; perhaps we can see how people use this in practice. May help to add some comments in vocabulary.talon, perhaps discussing your intended use case and/or examples?

code/vocabulary.py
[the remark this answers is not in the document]
nriley: Would also include **what** was added to the CSV in the success path.

nriley (changes requested):
Thanks!

nriley (approved):

PR #847, opened 2022-05-09
add user.open_file action and list/commands for editing settings csvs

rntz:
Changes:

1. Adds a new `user.open_file` action which opens a file in whatever the appropriate application is.

2. Adds a list `user.talon_settings_csv` for paths to talon user settings csv files, e.g. additional_words.csv, search_engines.csv, etc.

3. Adds a command `customize {user.talon_settings_csv)` that opens the corresponding csv and jumps to end of file.

This is heavily based on #810.

Feedback requested:

1. Is `customize` the command prefix we want here? At present this will bind the ccommands:

    ```
    customize search engines
    customize websites
    customize file extensions
    customize system paths
    customize words to replace
    customize additional words
    ```

    Alternatives might be `edit` or `open` (a bit short), `add new`, `edit talon`, `edit {talon_settings_csv} csv`.

2. Is there a better place to put this command than in a new file? Maybe standard.talon?

### DIFF
--- a/code/open_file.py
+++ b/code/open_file.py
@@ -1,5 +1,6 @@
 # Actions for opening files in appropriate applications.
 import os
+import subprocess
 
 from talon import Context, Module, actions, app, ui
 
@@ -28,8 +29,8 @@ ctx.lists["self.talon_settings_csv"] = _csvs
 
 @mod.action_class
 class ModuleActions:
-    def open_file(path: str, directory: str = None):
-        """Opens a file in the appropriate application for its type. If the path is relative, opens in the given directory (defaults to user's home directory)."""
+    def edit_text_file(path: str, directory: str = None):
+        """Tries to open a file in the user's preferred text editor. If the path is relative, opens in the given directory (defaults to user's home directory)."""
         # if the path is relative, directory is relevant.
         if not os.path.isabs(path):
             directory = directory or actions.path.user_home()
@@ -37,15 +38,54 @@ class ModuleActions:
         if not os.path.exists(path):
             app.notify(f"No file found at {path}")
             raise FileNotFoundError(path)
-        # TODO:
-        # 1. error handling if ui.launch/whatever fails?
-        # 2. should we use ui.launch or subprocess? since open/xdg-open is short-lived.
+
+        # tries to open file using a subprocess
+        def subproc(args):
+            try: return subprocess.run(args, timeout=0.5, check=True)
+            except subprocess.TimeoutExpired:
+                app.notify(f"Timeout trying to open file for editing: {path}")
+                raise
+            except subprocess.CalledProcessError:
+                app.notify(f"Could not open file for editing: {path}")
+                raise
+
         if app.platform == "windows":
-            os.startfile(path, "open")
+            # If there's no applications registered that can open the given type
+            # of file, 'edit' will fail, but 'open' always gives the user a
+            # choice between applications.
+            try: os.startfile(path, "edit")
+            except OSError:
+                os.startfile(path, "open")
         elif app.platform == "mac":
-            ui.launch(path="/usr/bin/open", args=[path])
+            # -t means try to open in a text editor.
+            subproc(["/usr/bin/open", "-t", path])
         elif app.platform == "linux":
-            ui.launch(path="/usr/bin/xdg-open", args=[path])
+            # we use xdg-open for this even though it might not open a text
+            # editor. we could use $EDITOR, but that might be something that
+            # requires a terminal (eg nano, vi).
+            subproc(["/usr/bin/xdg-open", path])
         else:
             raise Exception(f"unknown platform: {app.platform}")
         actions.sleep("500ms")
+
+    # def open_file(path: str, directory: str = None):
+    #     """Opens a file in the appropriate application for its type. If the path is relative, opens in the given directory (defaults to user's home directory)."""
+    #     # if the path is relative, directory is relevant.
+    #     if not os.path.isabs(path):
+    #         directory = directory or actions.path.user_home()
+    #         path = os.path.join(directory, path)
+    #     if not os.path.exists(path):
+    #         app.notify(f"No file found at {path}")
+    #         raise FileNotFoundError(path)
+    #     # TODO:
+    #     # 1. error handling if ui.launch/whatever fails?
+    #     # 2. should we use ui.launch or subprocess? since open/xdg-open is short-lived.
+    #     if app.platform == "windows":
+    #         os.startfile(path, "open")
+    #     elif app.platform == "mac":
+    #         ui.launch(path="/usr/bin/open", args=[path])
+    #     elif app.platform == "linux":
+    #         ui.launch(path="/usr/bin/xdg-open", args=[path])
+    #     else:
+    #         raise Exception(f"unknown platform: {app.platform}")
+    #     actions.sleep("500ms")
--- a/code/open_file.py
+++ b/code/open_file.py
@@ -1,29 +1,35 @@
 # Actions for opening files in appropriate applications.
-from talon import Module, Context, actions, ui, app
 import os
+
+from talon import Context, Module, actions, app, ui
+
 
 # path to knausj root directory
 REPO_DIR = os.path.dirname(os.path.dirname(__file__))
-SETTINGS_DIR = os.path.join(REPO_DIR, 'settings')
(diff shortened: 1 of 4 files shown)


code/open_file.py
```diff
@@ -0,0 +1,42 @@
...
+
+mod.list('talon_settings_csv', desc="Absolute paths to talon user settings csv files.")
+_csvs = {
+    x: os.path.join(SETTINGS_DIR, '_'.join(x.split()) + ".csv")
+    for x in ["file extensions", "search engines", "system paths",
+              "websites", "words to replace", "additional words"]
+}
+_csvs.update({
+    "homophones": os.path.join(REPO_DIR, 'code', 'homophones.csv'),
+})
```
pokey, line 20: I think maybe this stuff would go into a separate file, along with the proposed `edit_user_file`

code/open_file.py
```diff
@@ -0,0 +1,42 @@
...
+ctx.lists['self.talon_settings_csv'] = _csvs
+
+@mod.action_class
+class ModuleActions:
+    def open_file(path: str, directory: str = None):
```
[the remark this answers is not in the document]
rntz: What's the distinction you want to draw between `open_file` and `edit_user_file`, exactly? Like, is it opening vs editing, or all files vs user files (do you mean talon user directory files?), arbitrary files vs text files, or something else?

Can you give an example of when you'd want to override `edit_user_file` but would still need the unaltered `open_file`?

code/open_file.py
rntz: Yeah, it might belong in user_settings.py

codecat555:
This looks good, I agree it's an improvement on what I submitted in #810.

code/open_file.py
```diff
@@ -0,0 +1,42 @@
...
+REPO_DIR = os.path.dirname(os.path.dirname(__file__))
+SETTINGS_DIR = os.path.join(REPO_DIR, 'settings')
+
+mod = Module()
+ctx = Context()
+
+mod.list('talon_settings_csv', desc="Absolute paths to talon user settings csv files.")
+_csvs = {
+    x: os.path.join(SETTINGS_DIR, '_'.join(x.split()) + ".csv")
+    for x in ["file extensions", "search engines", "system paths",
```
codecat555, line 15: Nitpick - it seems unnecessarily complicated to process the names in this way, the reader has to actually inspect the code to understand what's happening.

Alternatively, the list could just contain the actual values wanted, e.g. 'file_extensions.csv'.

It would make sense to do it this way if the strings were being used as user interface elements somewhere, but that's not the case here.

code/open_file.py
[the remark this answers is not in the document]
auscompgeek: Maybe we can have an `open_editor` which will always open $EDITOR, then if we add `open_file` that can open in the default application for the file type.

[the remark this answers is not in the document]
nriley: I like `edit_user_file` or maybe `edit_text_file`; for example, I also have my CSVs set to open in Excel (which I use a lot for work) but I don't ever want Talon's CSVs to do that.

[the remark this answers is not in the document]
auscompgeek: I suppose `edit_text_file` would be a clearer name than `open_editor`, SGTM.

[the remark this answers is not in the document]
pokey: I would lean slightly towards `edit_user_file` over `edit_text_file`, but I don't feel super strongly.  @rntz does the separation seem reasonable?

code/open_file.py
```diff
@@ -0,0 +1,42 @@
...
+            path = os.path.join(directory, path)
+        if not os.path.exists(path):
+            app.notify(f"No file found at {path}")
+            raise FileNotFoundError(path)
+        if app.platform == "windows":
+            os.startfile(path, 'open')
+        elif app.platform == "mac":
+            ui.launch(path="open", args=[path])
+        elif app.platform == "linux":
+            ui.launch(path='/usr/bin/xdg-open', args=[path])
```
auscompgeek, line 39: A small concern if `open_file` will stick around: `/usr/bin/xdg-open` won't exist on NixOS, but the user will likely have `xdg-open` on `$PATH`.
```suggestion
            ui.launch(path='xdg-open', args=[path])
```

codecat555: Generally, for security reasons, it considered bad form to embed non-absolute paths in code...unless that code sets $PATH. For example, if an attacker gets write access to some component of your PATH, they can put their own xdg-open executable into place and use it to gain (further) access to your account.

pokey: Would the same concern not apply to `open`, as used for the Mac branch?

codecat555: Probably, I don't know how that gets resolved. I think I copied it from elsewhere in knausj_talon.

nriley: Yeah, makes sense to use `/usr/bin/open`. Since that’s all on a read-only snapshot I don’t think we have to worry about it being elsewhere!

rntz: I think for now I'm going to go with absolute paths and if that doesn't work for someone they can either override this action or file an issue and we'll try to figure out what the Right Thing to do is.

rntz:
Ok, we rewrote this during the maintenance session to make the following changes:

1. We now use contexts instead of examining `app.platform`.
2. The action is named `edit_text_file`, since that's our current use-case for it. We can add an `open_file` command later if we need it.
3. The action doesn't do any path handling and doesn't wait for 500ms. (Instead the "customize" voice commmand calls `sleep`.) This should make it easier for users to override the action if they want it to do something else.
4. The action does error handling on each platform.
5. The action tries to open the file in a text editor if there's a natural way to do that (eg. invoking the 'edit' action on Windows, passing the `-t` flag on Mac).

Would appreciate another review.

code/open_file.py
rntz: We've rewritten this to be clearer.

code/open_file.py
```diff
@@ -0,0 +1,42 @@
...
+ctx.lists['self.talon_settings_csv'] = _csvs
+
+@mod.action_class
+class ModuleActions:
+    def open_file(path: str, directory: str = None):
+        """Opens a file in the appropriate application for its type. If the path is relative, opens in the given directory (defaults to user's home directory)."""
+        # if the path is relative, directory is relevant.
+        if not os.path.isabs(path):
+            directory = directory or actions.path.user_home()
+            path = os.path.join(directory, path)
```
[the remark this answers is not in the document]
rntz: we just ditched it entirely

code/open_file.py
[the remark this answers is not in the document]
rntz: We ended up renaming the action to `edit_text_file`, since that's our only current use-case.

pokey (approved):
Beautiful 👏

Lmk if you want me to test it out locally; otherwise feel free to merge it in

nriley:
WFM on macOS. Can't test Windows right now.

code/open_file.py
```diff
@@ -0,0 +1,80 @@
+# Actions for opening files in appropriate applications.
+import os
+import subprocess
+
+from talon import Context, Module, app
+
+# path to knausj root directory
+REPO_DIR = os.path.dirname(os.path.dirname(__file__))
+SETTINGS_DIR = os.path.join(REPO_DIR, "settings")
```
codecat555, line 9: Could wrap this in os.path.realpath(), so user can use a symlink for their settings directory.

code/open_file.py
```diff
@@ -0,0 +1,80 @@
...
+
+mod = Module()
+ctx = Context()
+
+mod.list("talon_settings_csv", desc="Absolute paths to talon user settings csv files.")
+_csvs = {
+    name: os.path.join(SETTINGS_DIR, file_name)
```
codecat555, line 16: Again, could support symlinks with os.path.realpath() here.

code/open_file.py
nriley: Not sure why this is necessary; it should support symlinks already? Also note that SETTINGS_DIR is derived in a different way in user_settings.py.

codecat555: Seems you're right, @nriley. I thought I remembered that Talon didn't follow symlinks by default, or maybe that it didn't support them if they pointed outside the user directory...something like that.

pokey: Talon's watch function doesn't work with symlinks. Maybe that's what you're thinking off?

codecat555:
Must be, because in this case no Talon api is even involved.

PR #850, opened 2022-05-10
Add action to enable/disable control mouse explicitly

mostlyjason:
Updating the behavior of the action to match the original `toggle_control` function, which allows passing a bool to enable or disable the control mouse mode. I want to enable and disable the control mouse mode via hotkeys specified in a talon file. This allows a hot key to have a consistent behavior regardless of whether the mode was initially enabled or disabled. If no bool is passed in, it maintains the current behavior of flipping the current mode.

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -130,9 +130,12 @@ class Actions:
         """Start calibration"""
         eye_mouse.calib_start()
 
-    def mouse_toggle_control_mouse():
-        """Toggles control mouse"""
-        toggle_control(not config.control_mouse)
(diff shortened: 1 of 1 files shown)


pokey (commented):
I would be tempted to have a separate action for this one, but if we want to standardise on this style of function where no-args toggles and passing a bool sets it to that value, then I'd be ok with that.  cc/ @knausj85 @rntz

knausj85:
I'm personally okay with this style.

knausj85 (approved):

Issue #851, opened 2022-05-10
formatter accuracy


lunixbochs:
This has come up a couple of times. I suspect you'll improve formatter accuracy if you make these changes:

"allcaps" -> "all cap"
"alldown" -> "all down"

The reason for this is e.g. "alldown" is not an english word, D sounds like T, so "alld" sounds like alt, and you get "alt down".
Same for allcaps. Allc sounds a bit like alt, and you get "alt cap".

allcaps has a second problem. If you say "allcaps something", the S S (cap(s s)omething) chain will blend together in your speech and you'll get something that might sound more like "alt cap something". Hence removing the S.

pokey:
I would be tempted to leave this one where @rntz left it, with "all caps", because I'd worry "all cap" would lead to some accuracy issues wrt "alt cap".  Shall we close this one out?

lunixbochs:
it's the other way around! binding it as `all caps` can cause misrecognitions of `alt cap` when you say a word starting with an s, because it's human nature to collapse a double-s, e.g. try saying "all caps something" out loud quickly.

pokey:
ahh ok yeah. For some reason it made more sense the second time you said it 🤷‍♂️😄

do we want to keep "all caps" as a synonym?  @lunixbochs in your testing did you try saying "all caps" when the term was mapped as "all cap", or just tried "all cap"?

lunixbochs:
I did test saying "all caps" when it was bound as "all cap" and it worked, still better than when only "allcaps" was bound.

A possible issue with binding both, is now you may have a problem of homophones that only differentiate by whether the word starts with an `s`.

I would suggest just binding "all cap" and re-evaluate if new complaints surface. We're still likely going to be much better off than the old "allcaps" command.

pokey:
All makes sense to me.  Opened #861 to finish addressing this issue

PR #854, opened 2022-05-11
formatters.py: put spaces in "all down", "all caps"

rntz:
addresses part of #851. doesn't change "all caps" to "all cap" because I'm less convinced that won't screw up recognition for people who are used to "all caps", needs testing. I'm gonna merge this the next time I get around to it if nobody objects, we can always go  "all caps" -> "all cap" later.

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -124,8 +124,8 @@ formatters_dict = {
 
 # This is the mapping from spoken phrases to formatters
 formatters_words = {
-    "allcaps": formatters_dict["ALL_CAPS"],
-    "alldown": formatters_dict["ALL_LOWERCASE"],
+    "all caps": formatters_dict["ALL_CAPS"],
+    "all down": formatters_dict["ALL_LOWERCASE"],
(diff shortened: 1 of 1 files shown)


code/formatters.py
```diff
@@ -124,8 +124,8 @@ def every_word(word_func):
 
 # This is the mapping from spoken phrases to formatters
 formatters_words = {
-    "allcaps": formatters_dict["ALL_CAPS"],
-    "alldown": formatters_dict["ALL_LOWERCASE"],
+    "all caps": formatters_dict["ALL_CAPS"],
+    "all down": formatters_dict["ALL_LOWERCASE"],
     "camel": formatters_dict["PRIVATE_CAMEL_CASE"],
     "dotted": formatters_dict["DOT_SEPARATED"],
     "dubstring": formatters_dict["DOUBLE_QUOTED_STRING"],
```
auscompgeek, line 131: Should we add a space here as well? I have had slight troubles with dubstring in the past.
```suggestion
    "dub string": formatters_dict["DOUBLE_QUOTED_STRING"],
```

splondike (approved):

Issue #755, opened 2022-02-04
Setting to allow pop to click without eye tracker


splondike:
Some Talon users don't have an eye tracker but would still like to use pop to click. For example they might like to use the mouse or trackpad to reposition the cursor, but avoid the strain of pressing to click or repeatedly saying 'touch'.

This issue is to allow this use case while maintaining the existing functionality.

The proposal is to change the type of the 'mouse_enable_pop_click' setting to a string. The following values would be accepted: 'off', 'with_tracker', 'on'. 'on' allows clicking at all times. 'with_tracker' is the existing control mouse functionality. The strings '0' and '1' would map to 'off' and 'with_tracker' to maintain backwards compatibility. The 'mouse_enable_pop_stops_scroll' setting would stay as is.

pokey:
So is this one just waiting for someone to implement it?

timo:
i would like to suggest making it easier to get any one of the sixty bajillion combinations of options:

* does sleeping talon turn pop-to-click off?
* does not having an eye tracker turn pop-to-click off?
* does having an eye tracker but not being in "control mouse" mode turn pop-to-click off?
* does being in wheel-gaze mode turn pop-to-click off?

in the last few weeks we've had a whole lot of questions about this, and each had a different desired behaviour.

we ought to utilise the settings system to the biggest extent possible as well, rather than asking users to modify code (seems obvious?)

There is already a pull request to turn on unconditional pop-to-click. It doesn't have a way to influence whether sleep mode turns pop-to-click off. That is currently implemented by sleep_mode.talon setting that to 0, which is not as user-configurable as I would like.

Thoughts?

pokey:
So maybe we want to have one boolean setting for each of your bullet points, saying whether the given condition is required for pop to click to be on? Eg `pop_to_click_requires_eyetracker`, `pop_to_click_disable_on_sleep`, `pop_to_click_requires_control_mouse`? Then the impl just returns false if any of these settings are true but their condition isn't met

splondike:
I haven't been closely following the pop+click discussions in Slack; but it sounds like Timo has! I'd be up for a bunch of boolean operations.

We could keep the existing 'mouse_enable_pop_click' setting where now '0' would be off and '1' would be fully enabled. We'd also have the new setting 'mouse_pop_to_click_requires_eyetracker' which would default to '1'. In this way the behaviour of the existing setting would have the same effect while allowing for more detailed control via the new settings (which could all default to '0').

[...]

PR #858, opened 2022-05-14
Provide setting to unconditionally enable pop-to-click.

codecat555:
Per aegis - https://talonvoice.slack.com/archives/C9LFDNBAQ/p1652291642060249

### DIFF
--- a/code/mouse.py
+++ b/code/mouse.py
@@ -61,6 +61,12 @@ setting_mouse_enable_pop_click = mod.setting(
     default=0,
     desc="Enable pop to click when control mouse is enabled.",
 )
+setting_mouse_enable_unconditional_pop_click = mod.setting(
+    "mouse_enable_unconditional_pop_click",
+    type=int,
+    default=0,
+    desc="Enable pop click even if 'control mouse' mode is not enabled and even if there is no eye tracker present.",
+)
 setting_mouse_enable_pop_stops_scroll = mod.setting(
     "mouse_enable_pop_stops_scroll",
     type=int,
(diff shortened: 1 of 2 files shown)


pokey:
Is it fair to say that this one partially solves #755?  Is it worth folding the rest of #755 into this PR?

codecat555:
Sorry, I'm not up to addressing all of #755 at this time. I didn't even realize that issue existed before now.

I just jumped on this opportunistically to help reduce the support burden for aegis and to streamline the slack discussions around this point. Oh, and to make things easier for people too.

pokey:
Thanks for being upfront about that; obviously work / health has to take priority

My concern with merging this PR is that we may be introducing a setting that we'll need to support going forward, when we plan to change it soon

I don't use pop to click or an eye tracker, so I don't have a particularly strong opinion, but I do think it's worth aligning on #755 before merging this one

WDYT @knausj85 @rntz

codecat555:
Well, you could accept this incremental improvement if you were able to decide which settings names you want before completing the full implementation. Whatever you guys want to do with this is fine with me.

pokey:
That seems reasonable. Let's see where discussion goes on #755

splondike:
It looks like we have a consensus in #755 around adding a new setting that just enables pop to click more or less unconditionally. As such I think the next step is to convert this PR into an implementation of the design described there.

splondike (changes requested):
Marking as 'changes requested'; we should implement the decision in #755. Either in this PR, or by closing this out and making a new one.

splondike:
Replaced this with #1139 , closing.

PR #859, opened 2022-05-17
add javascriptreact tag to typescriptreact

uriklar:
looks like a good case for adding this inheritance

### DIFF
--- a/lang/typescript/typescriptreact.talon
+++ b/lang/typescript/typescriptreact.talon
@@ -1,3 +1,4 @@
 tag: user.typescriptreact
 -
 tag(): user.typescript
+tag(): user.javascriptreact


lang/typescript/typescriptreact.talon
```diff
@@ -1,3 +1,4 @@
 tag: user.typescriptreact
 -
 tag(): user.typescript
+tag(): user.javascriptreact
```
pokey, line 4: @AndreasArvidsson just wanted to check that there wasn't some reason your original PR didn't include this tag.  Is it a problem that we get a diamond inheritance structure where we activate `javascript` twice?

pokey (commented):
This seems reasonable to me but let's wait to see if @AndreasArvidsson had some reason to leave it out originally

lang/typescript/typescriptreact.talon
AndreasArvidsson: Nah no problem like that with tags

pokey (approved):
tested locally; works fine.  Thanks!

PR #861, opened 2022-05-21
"all caps" => "all cap"

### LINKED ISSUE #851: formatter accuracy

pokey:
Fixes #851

### DIFF
--- a/code/formatters.py
+++ b/code/formatters.py
@@ -124,7 +124,7 @@ formatters_dict = {
 
 # This is the mapping from spoken phrases to formatters
 formatters_words = {
-    "all caps": formatters_dict["ALL_CAPS"],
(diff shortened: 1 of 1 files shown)


splondike (approved):
Sounds like it's probably worth the breaking change here.

PR #863, opened 2022-05-27
Add common unix utilities to the generic unix terminal

2d6:
In my daily work, I use a lot of common utilities in my terminals. I extended the `generic_unix_terminal` tag in order to make it easy to run common commands and assemble pipes.

Some of the utilities are quite short and could be replaced by spelling their names out (e.g. `df`) or dictating the word (e.g. `touch`, `true`). I opted to include those cases since I found the flow to be better this way. I assume the accuracy also benefits for words with close pronounciation, like `false`/`falls`.

For several commands, I also added flags which I commonly use. It might be argued that adding those flags introduces unnecessary clutter. Including them was down to personal preference, again. It might make sense to remove them from the `default_unix_utilities.csv` file. The inclusion of flags in the default list might also lead to clashes if the local user has a version of these utilities with differing flags. For example, I found that sometimes the utilities in OSX differ from the ones in my Linux distribution. What do you think?

Some further points about which I'm not quite sure are:
- the lack of integration with the existing terminal commands from `generic_terminal.py`.
- the pronunciation of the some of the spellings. I'm not a native speaker, so some of them might be idiosyncratic. Since I use conformer myself, I also don't know how well they would work with Dragon. I tried to use existing words, as I heard that dragon needs those.

I'm looking for advice on these topics :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,20 @@ tag(): user.tabs
 
 For instance, kubectl commands (kubernetes) aren't relevant to everyone.
 
+### Unix utilities
+
+If you have a Unix (e.g. OSX) or Linux computer, you can enable support for a number of
+common terminal utilities like `cat`, `tail`, or `grep` by uncommenting the following
+line in [unix_shell.py](tags/terminal/unix_shell.py):
+
+```
+# ctx.tags = ["user.unix_utilities"]
+```
+
+Once you have uncommented the line, you can customize your utility commands by editing
+`settings/unix_utilities.csv`. Note: this directory is created when first running Talon
+with knausj_talon enabled.
+
 ## Jetbrains commands
 
 For Jetbrains commands to work you must install https://plugins.jetbrains.com/plugin/10504-voice-code-idea
--- a/tags/terminal/unix_shell.py
+++ b/tags/terminal/unix_shell.py
@@ -6,6 +6,9 @@ ctx.matches = r"""
 tag: user.generic_unix_shell
 """
 
+# Uncomment the following line to enable common unix utilities from unix_utilities.py
+# ctx.tags = ["user.unix_utilities"]
+
 
 @ctx.action_class("user")
 class Actions:
--- a/tags/terminal/unix_utilities.py
+++ b/tags/terminal/unix_utilities.py
@@ -1,0 +1,85 @@
+from talon import Context, Module
+
+from ...core.user_settings import get_list_from_csv
+
+ctx = Context()
+mod = Module()
+
+mod.tag(
+    "unix_utilities", desc="tag for enabling unix utility commands in your terminal"
+)
+
+# Do not edit this dictionary. It is just used to initially populate 'settings/unix_utilities.csv'.
+# Edit that file instead if you want to customize your commands.
+default_unix_utilities = {
+    "ark": "awk",
+    "base sixty four decode": "base64 -d",
+    "base sixty four": "base64",
+    "concat": "cat",
+    "change mode recurse": "chmod -R",
+    "change mode": "chmod",
+    "change owner recurse": "chown -R",
+    "change owner": "chown",
+    "curl": "curl",
+    "cut": "cut",
+    "disk free human": "df -h",
+    "disk free": "df",
+    "dig": "dig",
+    "disk usage human": "du -h",
+    "disk usage": "du",
+    "echo": "echo",
+    "false": "false",
+    "find": "find",
+    "grab here": "grep -Hirn",
+    "grab": "grep",
+    "head": "head",
+    "harp top": "htop",
+    "I D user": "id -u",
+    "I D": "id",
+    "less": "less",
+    "link soft": "ln -s",
+    "link": "ln",
+    "M D five sum": "md5sum",
+    "make dear": "mkdir",
+    "paste": "paste",
+    "print format": "printf",
+    "work dear": "pwd",
+    "reverse": "rev",
+    "are sync": "rsync",
+    "said": "sed",
+    "sequence": "seq",
+    "shah two fifty six sum": "sha256sum",
+    "sleep": "sleep",
+    "sort human": "sort -h",
+    "sort numeric": "sort -n",
+    "sort reverse": "sort -r",
+    "sort unique": "sort -u",
+    "sort": "sort",
+    "S S H": "ssh",
+    "stat": "stat",
+    "pseudo": "sudo",
+    "tail follow": "tail -f",
+    "tail": "tail",
+    "tea append": "tee -a",
+    "tea": "tee",
+    "touch": "touch",
+    "translate delete": "tr -d",
+    "translate": "tr",
+    "true": "true",
+    "unique count": "uniq -c",
+    "unique": "uniq",
+    "word count characters": "wc -c",
+    "word count lines": "wc -l",
+    "word count": "wc",
+    "who": "who",
+    "who am I": "whoami",
+}
+
+unix_utilities = get_list_from_csv(
+    "unix_utilities.csv",
+    headers=("command", "spoken"),
+    default=default_unix_utilities,
+)
+
+mod.list("unix_utility", desc="A common utility command")
+ctx.lists["self.unix_utility"] = unix_utilities
--- a/tags/terminal/unix_utilities.talon
+++ b/tags/terminal/unix_utilities.talon
@@ -1,0 +1,3 @@
+tag: user.unix_utilities
+-
+core {user.unix_utility}: "{unix_utility} "


apps/generic_terminal/default_unix_utilities.csv
```diff
@@ -0,0 +1,75 @@
...
+cat,concat
+chmod -R,change mode recurse
+chmod,change mode
+chown -R,change owner recurse
+chown,change owner
+curl,curl
+cut,cut
+df -h,disk free human
+df,disk free
+dig,dick
```
knausj85, line 13: This might not go over well in an office heh

2d6: True :D The benefits of working from home. I added this alternate voice command because I am a non-native speaker. I had trouble with accuracy with `dig`, but the reason is probably because I'm german and tend to pronounce consonants a little more hard than the usual english native speaker. I reverted this.

richardmcmillen:
I also need to use a lot of these utilities in my daily work so would definitely be interested in many of these. I am on MacOS and agree figuring out compatibility between *nix's could be annoying. Perhaps focusing on simple, lowest common denominator ones first would be good? (Note: I am by no means a maintainer of this repo, just a curious outsider who could also make use of this effort :))

I am english native and can have a go at how your proposed words recognize for me!

richardmcmillen:
Should maybe some of these be in `.talon` files as well/instead? Like a lot of the `git` commands? What do you think?
```
journal [cuddle | control] follow [<user.text>]:
  insert("journalctl -fu {user.text}")
journal [cuddle | control] reverse [<user.text>]:
  insert("journalctl -ru {user.text}")
```

apps/generic_terminal/generic_unix_shell.py
```diff
@@ -1,11 +1,39 @@
...
+mod.list("unix_utility", desc="A common utility command")
+ctx = Context()
+ctx.lists["self.unix_utility"] = unix_utilities
+
+
+@mod.action_class
+class UnixUtilityActions:
+    def terminal_run_unix_utility(command: str):
+        """Runs a utility command"""
+
```
rntz, line 36: This declares the action `terminal_run_unix_utility` but doesn't implement it. Did you forget to add that to the PR?

apps/generic_terminal/generic_unix_shell.talon
```diff
@@ -0,0 +1,3 @@
+tag: user.generic_unix_shell
+-
+core {user.unix_utility}: user.terminal_run_unix_utility("{unix_utility}")
```
rntz, line 3: Do you want this command to *run* the shell command, or just type it out? I'd prefer typing it out, because you might want to add options to the command afterward; you can always say "enter" afterwards to hit the enter key and run the command.

Either way, it seems like you don't need an action for this, you can just insert the command text.

apps/generic_terminal/generic_unix_shell.py
```diff
@@ -1,11 +1,39 @@
...
 tag: user.generic_unix_shell
 """
 
+path = Path(__file__).parents[0]
+with open(path / "default_unix_utilities.csv") as f:
+    # get_list_from_csv() expects the columns in the reverse order
+    default_unix_utilities = {
+        spoken.strip(): command.strip()
+        for command, spoken in [line.split(",", 1) for line in f]
+    }
```
rntz, line 19: This might be simpler if we use Python's `csv` module:

```
    with open(...) as f:
        default_unix_utilities = list(csv.reader(f))
```

rntz (changes requested):

apps/generic_terminal/generic_unix_shell.talon
2d6: I meant text insertion of course, for pipeline composition and adding arguments, just as you said. The wording of the function was suboptimal.

> Either way, it seems like you don't need an action for this, you can just insert the command text.

D'oh, you are obviously right! Will simplify this.

2d6:
> Should maybe some of these be in `.talon` files as well/instead? Like a lot of the [`git` commands](https://github.com/knausj85/knausj_talon/blob/master/misc/git.talon)? What do you think?
I think you are right. Especially the systemd utilities feel like that, on second thought. I will go over the default ones with this in mind.

apps/generic_terminal/generic_unix_shell.py
2d6: Good point. Used this in the current revision.

apps/generic_terminal/generic_unix_shell.py
2d6: Seems that way. Anyways, as I removed the action as per your suggestion, this is not relevant anymore. I'm going to mark this specific issue as resolved.

2d6:
@richardmcmillen Thank you for your support :) I think it would be wise to limit the default CSV to a subset that works on both OSX and Linux. If you could run the commands from the default list and tell me which do not work on OSX, that would be invaluable to me. I don't have access to OSX, unfortunately.

fidgetingbits:
fwiw in my repo I have a bunch of generic service and timer commands with default actions and then have systemd-specific implementations, which might be useful depending on how you plan to break out some of it. I also have a bunch of extra commands in my commandline.talon file, which might be useful too.

apps/generic_terminal/default_unix_utilities.csv
```diff
@@ -0,0 +1,67 @@
...
+ip -br -c a,I P all brief
+ip a,I P all
+less,less
+ln -s,link soft
+ln,link
+md5sum,M D five sum
+mkdir,make dear
+paste,paste
+printf,print format
+ps -aux,processes
```
richardmcmillen, line 34: This is not compatible with MacOS `ps`

```ps -aux
ps: No user named 'x'`
```

apps/generic_terminal/default_unix_utilities.csv
```diff
@@ -0,0 +1,67 @@
...
+echo,echo
+false,false
+find,find
+grep -Hirn,grab here
+grep,grab
+head,head
+htop,harp top
+id -u,I D user
+id,I D
+ip -br -c a,I P all brief
```
richardmcmillen, line 25: `ip` is not a standard tool for macos as far as I know.

richardmcmillen:
What are peoples thoughts on incorporating changes like this MR and files from @fidgetingbits into this repository as a general question?

It seems like a lot of fairly niche commands, which could have potentially dangerous implications if unsure how to use or accidentally triggered. `rm` files etc.

I'm thinking it might be an good idea to make a lot of these commands as commandline tags (with logical groupings), which are not enabled in generic terminals by default.

Meaning as a user you would need to create a `.talon` file configuration which is active in generic terminals and enable the various tags you wish to use.

```
tag(): user.cli_grep
tag(): user.cli_rm
tag(): user.systemd
tag(): user.virsh
tag(): user.iptables
```

fidgetingbits:
I agree using tags like that is the ideal way to do it for now and how we have mostly done it in the past. I have some stuff in my repo tagged like that, and knausj already has a fair bit too, and it's easy enough to manage. Just might be a bit confusing to know about what commands exist if you are a new user but I suppose the that the tag could just be commented out by default to give a hint (or something like that).

For rm I would maybe guess people would at least expect a basic removal command to be available by default without having to fiddle wirh tags, but I agree the more complicated/dangerous stuff probably should be left to a tag.

apps/generic_terminal/generic_unix_shell.py
```diff
@@ -1,11 +1,31 @@
...
 ctx = Context()
 mod = Module()
 ctx.matches = r"""
 tag: user.generic_unix_shell
 """
 
+path = Path(__file__).parents[0]
+with open(path / "default_unix_utilities.csv") as f:
+    # get_list_from_csv() expects the columns in the reverse order
+    default_unix_utilities = {spoken: command for command, spoken in csv.reader(f)}
```
pokey, line 17: hmm i wonder if there's a way to avoid reading this every time, because after the first time it will just ignore it in favour of user csv, right? maybe `get_list_from_csv` could take a callback that it only calls if it can't find user one?  Idk I guess a callback maybe isn't very pythonic 🤔.  any thoughts @auscompgeek

2d6: I further simplified this by using a static dict to initialize the CSV. This keeps it from loading the default CSV on every launch.

2d6:
As per the discussion above, I simplified the code by providing a static dictionary for initializing the CSV. I also hid the functionality behind a tag and removed the `rm` command, alongside those not valid on OSX.

I'm not sure how to best document the hidden functionality, so that new users may find the utlitities tag. Is there a place to document all the available feature flags?

apps/generic_terminal/unix_utilities.py
```diff
@@ -0,0 +1,90 @@
+from talon import Context, Module
+
+from ...code.user_settings import get_list_from_csv
+
+ctx = Context()
+mod = Module()
+ctx.matches = r"""
+tag: terminal
+and tag: user.unix_utilities
```
2d6, line 9: I'm not sure if it makes sense to use the tag in the file where it is declared. I would like advice here.

gsingh93:
I'd love to get this merged 🙂 Any chance you could rebase so we can ping the reviewers again?

Also, I think we should probably move `ls` into here as it's a unix command and not a shell command.

apps/generic_terminal/unix_utilities.py
splondike: I think you'd just make a context without a matches block. This is the least specific match and so would let anybody override it if they wanted.

apps/generic_terminal/unix_utilities.py
```diff
@@ -0,0 +1,90 @@
...
+}
+
+unix_utilities = get_list_from_csv(
+    "unix_utilities.csv",
+    headers=("command", "spoken"),
+    default=default_unix_utilities,
+)
+
+mod.list("unix_utility", desc="A common utility command")
+ctx = Context()
```
splondike, line 89: Extra ctx definition here should be removed.

apps/generic_terminal/unix_utilities.talon
```diff
@@ -0,0 +1,4 @@
+tag: terminal
```
splondike, line 1: I wouldn't include the 'terminal' tag here, just your custom unix_utilities one. That gives users the flexibility to add it in to e.g. their IDE which might have an embedded terminal but not otherwise be tagged as one.

apps/generic_terminal/unix_utilities.py
```diff
@@ -0,0 +1,90 @@
+from talon import Context, Module
```
splondike, line 1: This file should probably go in `tags/terminal/` now since that refactoring a little while ago.

splondike (commented):
This concept looks good to me. I've left a few comments, if those get addressed I'm happy to merge this in.

In terms of how to make its existence known, I'd actually be in favour of enabling it by defaut in the unix_shell.py file. However to keep this PR uncontroversial in that respect so we can get something merged I suggest we just put it in there commented out. See the bottom of [this wiki section](https://talon.wiki/unofficial_talon_docs/#tags) if you're not sure how to do that. You could also add something to [the README section](https://github.com/knausj85/knausj_talon#terminal-commands).

Other unix utilities behind tags (e.g. kubectl) have just been enabled by default in random places like gnome_terminal.talon, likely because the nice unix_shell.py file didn't exist prior to Wen Kokke's big refactoring. IMO it would probably be better in the future if these unix utility tags were centralised into unix_shell.py.

apps/generic_terminal/unix_utilities.py
```diff
@@ -0,0 +1,90 @@
...
+and tag: user.unix_utilities
+"""
+
+mod.tag(
+    "unix_utilities", desc="tag for enabling unix utility commands in your terminal"
+)
+
+# Do not edit this dictionary. It is just used to initially populate 'settings/unix_utilities.csv'.
+# Edit this file instead.
+default_unix_utilities = {
```
splondike, line 18: The rsync one is a bit difficult for my Australian accent, but I think on the whole this is a good translation to voice. It also has a lot of the utilities I use regularly (and that aren't in the existing voice commands). Nice one!

2d6:
Hey, thanks for the detailed feedback. I kind of lost track of this PR, and am pretty busy these days. However, this weekend I plan to adress and implement the comments :)

apps/generic_terminal/unix_utilities.py
```diff
@@ -0,0 +1,90 @@
...
+ctx.matches = r"""
+tag: terminal
+and tag: user.unix_utilities
+"""
+
+mod.tag(
+    "unix_utilities", desc="tag for enabling unix utility commands in your terminal"
+)
+
+# Do not edit this dictionary. It is just used to initially populate 'settings/unix_utilities.csv'.
```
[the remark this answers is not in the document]
2d6: I thought about that and looked around how this sort of stuff is implemented in other places. Since these commands are just the initial content of `settings/unix_utilities.csv`, I figured I should take inspiration from `file_extensions.py` and `vocabulary.py`. These also use a dictionary for the initial content. My rationale is that having only one CSV file makes at less confusing which file you need to edit if you want to customize your utility commands.

Therefore I would be in favour of keeping it the wait currently is. I would be happy to follow the consensus, though.

apps/generic_terminal/unix_utilities.py
2d6: I adjusted this in the current iteration. Thanks for the suggestion :)

apps/generic_terminal/unix_utilities.py
2d6: Huh, `git rebase` was smart enough to do that automatically. Adjusted in the latest iteration.

apps/generic_terminal/unix_utilities.talon
2d6: Done

apps/generic_terminal/unix_utilities.py
2d6: Done :see_no_evil:

2d6:
Alright, I hope I was able to correctly address your input @splondike . Let me know if you want any further changes. I checked the previous change requests by rntz and am of the opinion that I integrated their input as well. Feel free to double-check.

splondike (approved):
'are sync' works better for me than 'our sync'. If it works fine for you also, perhaps we could swap to that? Otherwise I'm happy to merge this in; looks good :+1: .

2d6:
> 'are sync' works better for me than 'our sync'. If it works fine for you also, perhaps we could swap to that? Otherwise I'm happy to merge this in; looks good +1 .

@splondike Awesome! As per your suggestion, I replaced `our sync` with `are sync`. I also tried to clarify the README paragraph a little. Looking forward to the merge :)

splondike:
Thanks! Sorry it took a while, but we got there in the end :).

PR #868, opened 2022-05-30
Add support for programming language keywords

### LINKED ISSUE #807: Support chaining keywords in programming languages
## The goal

I have a [commit](https://github.com/pokey/pokey_talon/commit/aab118daa6f027e67a1075f025b7869ffde64168) on my repo that enables me to chain programming language keywords, for example `"state return null"` or `"state export default const"` , instead of `"state return state null"` or `"state export state default state const"`.  I’m trying to figure out how to reconcile this setup with @wenkokke’s fine-grained tags [commit](https://github.com/knausj85/knausj_talon/commit/73d94794bcb330374f18432b0f9f2af35082849c).

## How it works on my branch

Here are the important parts of my commit:

We first define a list and an action for keywords:
https://github.com/knausj85/knausj_talon/blob/aab118daa6f027e67a1075f025b7869ffde64168/code/keywords.py#L4-L12

Then define a spoken form that allows "state" followed by one or more keywords:
https://github.com/knausj85/knausj_talon/blob/aab118daa6f027e67a1075f025b7869ffde64168/lang/programming.talon#L33

Then each language defines its keywords, including spacing around them.  See eg Typescript:
https://github.com/knausj85/knausj_talon/blob/aab118daa6f027e67a1075f025b7869ffde64168/lang/typescript/typescript.py#L50-L77

## The problems

There’s kind of two problems to solve:

### Many tags

The first is how we handle the fact that after @wenkokke’s commit, these keywords are now spread across many tags. 
 We would somehow need to define the lists in all the separate tags, then merge all of them together in a capture or something?

### Actions vs lists
The other problem is that I just use a list of strings to insert, whereas in knausj they are actions.  The latter is a bit more flexible, but I’m not sure how to reconcile that with my approach

pokey:
Fixes #807

Adds support for keywords, such as "return", "export", etc.  Note that these differ from the usual "state" commands (eg "state if"), because they insert only the keyword, rather than building the statement, which would include adding parentheses, etc

They can be chained, so you can say eg "keyword return false"

Note that there is definitely some overlap with many of the "state" commands; might be worth thinking whether that is a problem

We may want to consider removing a few of the JS "state" commands that are just keywords https://github.com/knausj85/knausj_talon/blob/796bb9ec84e7e897f1fde141ee664b944e303a1e/lang/javascript/javascript.talon#L31-L41

### DIFF
--- a/lang/javascript/javascript.py
+++ b/lang/javascript/javascript.py
@@ -44,6 +44,32 @@ ctx.lists["user.code_common_member_function"] = {
     "then": "then",
 }
 
+ctx.lists["user.code_keyword"] = {
+    "a sink": "async ",
+    "await": "await ",
+    "break": "break",
+    "class": "class ",
+    "const": "const ",
+    "continue": "continue",
+    "default": "default ",
+    "export": "export ",
+    "false": "false",
+    "function": "function ",
+    "import": "import ",
+    "let": "let ",
+    "new": "new ",
+    "null": "null",
+    "private": "private ",
+    "protected": "protected ",
+    "public": "public ",
+    "return": "return ",
+    "throw": "throw ",
+    "true": "true",
+    "try": "try ",
+    "undefined": "undefined",
+    "yield": "yield ",
+}
+
 
 @ctx.action_class("user")
 class UserActions:
--- a/lang/tags/keywords.py
+++ b/lang/tags/keywords.py
@@ -1,0 +1,20 @@
+from talon import Module, actions
+
+mod = Module()
+
+mod.tag("code_keywords", desc="Tag for enabling commands for keywords")
+
+mod.list("code_keyword", desc="List of keywords for active language")
+
+
+@mod.capture(rule=("{user.code_keyword}"))
+def code_keyword(m) -> str:
+    return str(m)
+
+
+@mod.action_class
+class Actions:
+    def code_keyword(keywords: list[str]):
+        """Adds keywords"""
+        for keyword in keywords:
+            actions.insert(keyword)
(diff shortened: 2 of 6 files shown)


lang/javascript/javascript.py
```diff
@@ -44,6 +44,32 @@
     "then": "then",
 }
 
+ctx.lists["user.code_keyword"] = {
```
pokey, line 47: Note that some of the keywords have trailing space, depending how they're usually used

lang/tags/keywords.talon
```diff
@@ -0,0 +1,3 @@
+tag: user.code_keywords
+-
+keyword (<user.code_keyword>+): user.code_keyword(code_keyword_list)
```
pokey, line 3: Open to ideas on the prefix here.  

Fwiw I personally use "make" for keywords, statements, and operators, and then "snip" for the bigger structures where knausj tends to use "state", and they are actual vscode snippets

lang/tags/keywords.py
```diff
@@ -0,0 +1,20 @@
...
+def code_keyword(m) -> str:
+    return str(m)
+
+
+@mod.action_class
+class Actions:
+    def code_keyword(keywords: list[str]):
+        """Adds keywords"""
+        for keyword in keywords:
+            actions.insert(keyword)
```
rntz, line 20: It seems like we at least want to ensure there is whitespace between the keywords, since you basically never want to just smash keywords next to one another? Probably we should check whether they're going to have whitespace between them and if not insert a single space.

lang/tags/keywords.talon
rntz: I think we probably need to think about redesigning our programming command prefixes at some point. For now I think `keyword` is probably fine? It's very clear and specific.

rntz (approved):

PR #876, opened 2022-06-04
Add equivalent cursorless commands as options for the draft_window

richardmcmillen:
For anyone familiar with cursorless these may be baked into their memory. I learned cursorless before I used the draft window and saw the commands could be fairly easily expanded to allow for the cursorless style phrasing. Obviously this is only snapshot in time and cursorless could evolve in future but perhaps this small selection will always be compatible.

Interested to know if folks think this is an appropriate addition or not.

I can also make a small addition to the draft_window README mentioning cursorless compatibility if this is desired.

### DIFF
--- a/talon_draft_window/draft_window.talon
+++ b/talon_draft_window/draft_window.talon
@@ -6,37 +6,34 @@ settings():
   user.context_sensitive_dictation = 1
 
 # Replace a single word with a phrase
-replace <user.draft_anchor> with <user.text>:
+[change | replace] <user.draft_anchor> [with] <user.text>:
   user.draft_select("{draft_anchor}")
   result = user.formatted_text(text, "NOOP")
   insert(result)
 
 # Position cursor before word
-cursor <user.draft_anchor>:
-  user.draft_position_caret("{draft_anchor}")
-
-cursor before <user.draft_anchor>:
+[pre | cursor | cursor before] <user.draft_anchor>:
   user.draft_position_caret("{draft_anchor}")
 
 # Position cursor after word
-cursor after <user.draft_anchor>:
+[post | cursor after] <user.draft_anchor>:
   user.draft_position_caret("{draft_anchor}", 1)
 
 # Select a whole word
-select <user.draft_anchor>:
+[take | select] <user.draft_anchor>:
   user.draft_select("{draft_anchor}")
 
 # Select a range of words
-select <user.draft_anchor> through <user.draft_anchor>:
+[take | select] <user.draft_anchor> [through | past] <user.draft_anchor>:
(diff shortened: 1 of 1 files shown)


pokey:
Interesting idea. Are you familiar with the Cursorless draft window? If you say "draft all" it will copy everything into VSCode, and then "draft submit" will paste it back

richardmcmillen:
I was not aware no. I had seen around people talking about using VSCode for
drafts and found some non-knausj user files and didn't think it was part of
the community repo. I've found the relevant files in this repo now. Thanks
for pointing it out.

That will probably be what I'll use, I'll try it out soon. Feel free to
close this MR if you don't think it's appropriate.

rntz (approved):
Seems fine to me, if this makes it more convenient for cursorless users maybe we should just do it.

rntz:
@pokey why don't you decide whether to merge this or not, you're the cursorless expert.

pokey:
My guess is most Cursorless users will opt for VSCode draft editor, so I'll toss the potato to @splondike 😊

EdJoJob:
I was hoping to use this as a pathway to getting used to the cursorless commands before migrating to VSCode (Vim user). It would let me get the "voice memory" in place while doing the _other_ setup for VSCode, which I keep procrastinating.

richardmcmillen:
The reason I came to make this change was driven by trying to do the
example commands in the draft window readme and having difficulty with the
`select` and `through` recognisability, `through` especially, it always is
recognised as `three` due to my accent.

So I figured using the cursorless phrasing, as I'm sure a good bit of trial
and error has gone in to it, would be better for accuracy. With the added
benefit of being compatible, currently at least, with the cursorless
commands to reduce the mental load of those who use both.

splondike:
Sounds fine to me, thanks!

rntz:
I'm not sure how none of us noticed this before, but this PR uses square brackets where it should be using parentheses, thus making all these keywords optional. oops.

richardmcmillen:
Oh whoops, sorry about that! I can make a new PR to fix this up.

rntz:
I already did: https://github.com/knausj85/knausj_talon/pull/894

richardmcmillen:
Just another thing, the reason I came to make this change was driven by
trying to do the example commands in the draft window readme and having
difficulty with the `select` and `through` recognisability, `through`
especially, it always is recognised as `three` due to my accent.

So I figured using the cursorless phrasing, as I'm sure a good bit of trial
and error has gone in to it, would be better for accuracy. With the added
benefit of being compatible, currently at least, with the cursorless
commands to reduce the mental load of those who use both.

PR #877, opened 2022-06-04
Support Discord and Slack in browser

auscompgeek:
This adds support for the Discord and Slack webapps.

Some of the keyboard shortcuts conflict with browser shortcuts, but this may vary from browser to browser, so I figured leaving them as is for now is probably fine, e.g. <kbd>cmd-k</kbd> to open the Slack channel picker on macOS conflicts with Firefox's focus search bar shortcut.


splondike (approved):

splondike:
Thanks, it seems like a net improvement even if it's not perfect.

PR #878, opened 2022-06-04
Remove direct noting of port 10

EdJoJob:
We use `number_small` as the capture for all port commands, which includes 10.

Comment at https://github.com/knausj85/knausj_talon/pull/871#discussion_r889588800
suggests it was initially included for debugging and is no longer required.


EdJoJob:
Missed the change of `master` to `main` in my local pull

Issue #825, opened 2022-04-18
Missing explicit voice command to write private function


auscompgeek:
Currently the only way to call `code_private_function` is indirectly via `code_default_function` (except for R).

This poses a few problems:

- Python and Scala override `code_default_function` to call `code_public_function`. This currently makes it impossible to use their `code_private_function` implementations without adding a custom voice command.
- The Java implementation of `code_protected_function` creates a method without any modifiers, which is confusing.
  - It _should_ insert the `protected` access modifier, but then there would be no way to insert a method with no access modifiers (i.e. a package-private method, distinct from `protected`).
  - As a workaround, one can say `state protected`, pause, then `pro funky ...`, but this is clearly inconsistent and disrupts the flow.

I would like to propose the following:

- a new `private funky <text>` command, which calls `code_private_function`
- a new action `code_default_static_function`, to be called by `static funky <text>`, can default to calling `code_private_static_function`
- a new `private static funky <text>` command, to replace current usage of `static funky <text>`

This would allow for:

- Java: creating methods with all possible access modifiers (including package-private and `protected`) using the commands in `functions.talon`
- TypeScript: creating functions outside methods, whilst still being able to create private methods (see also #824)
- Python/Scala: actually creating private functions

splondike:
This sounds good to me.

pokey:
What might be nice is the following:

```talon
{user.function_modifier}+ funky <user.text>: user.code_function(function_modifier_list, text)
```

Then there could be a default implementation

```py
    def code_function(modifier_list: list[str], text: str):
        """Inserts function declaration"""
        formatted_text = actions.user.formatted_text(
            text, settings.get("user.code_function_formatter")
        )
        modifier_string = " ".join(modifier_list)
    
        result = f"{modifier_string} function {formatted_text}"

        actions.user.code_insert_function(result, None)
```

Individual languages could then override the implementation to look at `modifier_list` and choose a different formatter / use different template for the function

Definitely a different approach, but figured I'd throw it out there.  

It is related to #807

splondike:
I actually really like pokey's suggestion. It opens this up to more keywords like async, and even putting in return types depending on the language. To facilitate all that we'd probably make function_modifier a capture instead of a list.

The complexity of the extra abstraction could be offset by defining a 'standard' way of ordering modifiers as well as a standard list of command words. Something like this in the .talon file:

```
# This command is used for inserting a function definition with modifiers (e.g. private, protected, async) or return types.
[...]

PR #879, opened 2022-06-05
Flexible funky modifiers

splondike:
Adds the following:
* Allow flexible combination of function modifiers (static, public etc.) in funky command. Allows for easier addition of 'non standard' modifiers (e.g. async in javascript), and suppression of irrelevant ones (e.g. static in javascript).
* Adds the 'private' modifier and makes the unadorned 'funky' in Typescript emit an unmodified function.
* Updates the test stubs so that the @mod.action_class decorator registers actions in a similar way to Talon.

Addresses #825 and replaces #824 .

### DIFF
--- a/lang/javascript/javascript.py
+++ b/lang/javascript/javascript.py
@@ -248,6 +248,16 @@ class UserActions:
         actions.user.paste(text)
         actions.edit.left()
 
+    def code_default_function(text: str):
+        """Inserts function declaration without modifiers"""
+        result = "function {}".format(
+            actions.user.formatted_text(
+                text, settings.get("user.code_private_function_formatter")
+            )
+        )
+
+        actions.user.code_insert_function(result, None)
+
     def code_private_function(text: str):
         """Inserts private function declaration"""
         result = "function {}".format(
--- a/lang/tags/functions.py
+++ b/lang/tags/functions.py
@@ -1,3 +1,5 @@
+from typing import List, Union
+
 from talon import Context, Module, actions, settings
 
 ctx = Context()
@@ -12,23 +14,17 @@ mod.list(
     "code_parameter_name", desc="List of common parameter names for active language"
 )
 mod.list(
-    "code_function_modifiers",
+    "code_function_modifier",
     desc="List of function modifiers (e.g. private, async, static)",
 )
 
-ctx.lists["user.code_function_modifiers"] = ["pub", "pro", "private", "static"]
+ctx.lists["user.code_function_modifier"] = ["pub", "pro", "private", "static"]
 
 
 @mod.capture(rule="{user.code_type}")
 def code_type(m) -> str:
     """Returns a macro name"""
     return m.code_type
-
-
-@mod.capture(rule="{user.code_function_modifiers}+")
-def code_function_modifiers(m) -> str:
-    """Returns a space separated string of function modifiers"""
-    return " ".join(m)
 
 
 setting_private_function_formatter = mod.setting("code_private_function_formatter", str)
@@ -45,11 +41,15 @@ setting_public_variable_formatter = mod.setting("code_public_variable_formatter"
 
 @mod.action_class
 class Actions:
-    def code_modified_function(modifiers: str, text: str):
-        """Inserts function declaration with the given modifiers"""
-        mods = set(modifiers.split(" "))
+    def code_modified_function(modifiers: Union[List[str], int], text: str):
+        """
+        Inserts function declaration with the given modifiers. modifiers == 0
+        implies no modifiers (.talon files don't have empty list literal
+        syntax)
+        """
+        mods = {} if modifiers == 0 else set(modifiers)
 
-        if mods == {""}:
+        if mods == {}:
             return actions.user.code_default_function(text)
         elif mods == {"static"}:
             return actions.user.code_private_static_function(text)
@@ -66,7 +66,7 @@ class Actions:
         elif mods == {"pub", "static"}:
             return actions.user.code_public_static_function(text)
         else:
-            raise RuntimeError(f"Unhandled modifier set: {modifiers}")
+            raise RuntimeError(f"Unhandled modifier set: {mods}")
 
     def code_default_function(text: str):
         """Inserts function declaration"""
--- a/lang/tags/functions.py
+++ b/lang/tags/functions.py
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import Union
 
 from talon import Context, Module, actions, settings
 
@@ -41,7 +41,7 @@ setting_public_variable_formatter = mod.setting("code_public_variable_formatter"
 
 @mod.action_class
 class Actions:
-    def code_modified_function(modifiers: Union[List[str], int], text: str):
+    def code_modified_function(modifiers: Union[list[str], int], text: str):
         """
         Inserts function declaration with the given modifiers. modifiers == 0
         implies no modifiers (.talon files don't have empty list literal
--- a/lang/tags/functions.py
+++ b/lang/tags/functions.py
@@ -1,5 +1,6 @@
 from talon import Context, Module, actions, settings
 
+
 ctx = Context()
 mod = Module()
 
@@ -11,12 +12,24 @@ mod.list("code_type", desc="List of types for active language")
 mod.list(
     "code_parameter_name", desc="List of common parameter names for active language"
 )
+mod.list(
+    "code_function_modifiers",
+    desc="List of function modifiers (e.g. private, async, static)"
+)
+
+ctx.lists["user.code_function_modifiers"] = ["pub", "pro", "private", "static"]
 
 
 @mod.capture(rule="{user.code_type}")
 def code_type(m) -> str:
     """Returns a macro name"""
     return m.code_type
+
+
+@mod.capture(rule="{user.code_function_modifiers}+")
+def code_function_modifiers(m) -> str:
+    """Returns a space separated string of function modifiers"""
+    return " ".join(m)
 
 
 setting_private_function_formatter = mod.setting("code_private_function_formatter", str)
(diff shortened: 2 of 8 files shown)


lang/tags/functions.py
```diff
@@ -11,6 +11,12 @@
 mod.list(
     "code_parameter_name", desc="List of common parameter names for active language"
 )
+mod.list(
+    "code_function_modifiers",
```
pokey, line 15: Prefer singular for talon list names

```suggestion
    "code_function_modifier",
```

lang/typescript/typescript.py
```diff
@@ -18,6 +18,16 @@
 
 @ctx.action_class("user")
 class UserActions:
+    def code_default_function(text: str):
```
pokey, line 21: Why not define this one in `javascript.py`? Then typescript gets it for free

lang/typescript/typescript.py
```diff
@@ -18,6 +18,16 @@
```
pokey, line 18: Don't we want `async` in typescript?

tests/stubs/talon/__init__.py
```diff
@@ -1,12 +1,21 @@
+import inspect
```
pokey, line 1: I'd love to get a walkthrough of the knausj testing setup at some point, as we're planning to introduce cursorless talon tests soon. If you're interested, let's find a time

Cc/ @AndreasArvidsson @will-sommers @andrewdant

lang/tags/functions.py
```diff
@@ -19,6 +25,12 @@ def code_type(m) -> str:
     return m.code_type
 
 
+@mod.capture(rule="{user.code_function_modifiers}+")
+def code_function_modifiers(m) -> str:
+    """Returns a space separated string of function modifiers"""
+    return " ".join(m)
```
pokey, line 31: Why join this with strings if you're just going to split it? Could just inline the above rule into the talon file and then use `_list` suffix to get an array

pokey (changes requested):
Looks good with a few comments. I haven't reviewed the test code changes in detail because I'm not really familiar with the setup. Maybe someone else more familiar can take a look? Would also love a walkthrough if you're interested

pokey:
Also have a look at the end of [pre-commit change](https://github.com/knausj85/knausj_talon/commit/16c12cb8cf5f372df760d3af0b679832268ac2d6). If I'm reading that right, did it clobber one of your dynamic imports?

lang/typescript/typescript.py
splondike: Also javascript, but I think I'll leave that for another PR. If somebody wants to add it they'll probably need to implement a version of `code_modified_function` that does the construction of the function from its modifiers rather than use my default delegation implementation.

lang/typescript/typescript.py
splondike: Javascript gets the default implementation that passes through to code_private_function anyway.

tests/stubs/talon/__init__.py
splondike: We can try and work out the timezones some time :). The main technical trick that enables all the pass shenanigans is in /conftest.py . I don't actually remember too much about that, it's just some arcane thing to hook into the class loader.

The stubs directory just has enough approximate implementation of the Talon API to allow loading of the file under test (and to facilitate testing). I don't really have any long term plans for how to extend it; I've just hacked on it when I needed something.

lang/tags/functions.py
splondike: I wanted to have a capture rather than a raw list as they allow for more flexible overriding. Then that meant I had to split out the no modifiers command because you can't make the capture rule optional. And in .talon you can't construct an empty list. So then I could have the type be e.g. `Union[List[str], int]`, but I figured just using str like this was fine.

That all said, I can't really think of a reason somebody would want the flexibility of a capture, so I might just go to a raw list like you said :).

splondike:
Have addressed the above comments I think. The `#isort:skip` incantation warded off the robot successfully as well.

lang/tags/functions.py
```diff
@@ -11,6 +13,12 @@
 mod.list(
     "code_parameter_name", desc="List of common parameter names for active language"
 )
+mod.list(
+    "code_function_modifier",
+    desc="List of function modifiers (e.g. private, async, static)",
+)
+
+ctx.lists["user.code_function_modifier"] = ["pub", "pro", "private", "static"]
```
pokey, line 21: ok sorry i'm nitpicking a bit here, but I think I'd prefer to use more verbose internal identifiers
```suggestion
ctx.lists["user.code_function_modifier"] = {"pub": "public", "pro": "protected", "private": "private", "static": "static",}
```

lang/tags/functions.py
```diff
@@ -33,6 +41,33 @@ def code_type(m) -> str:
...
+            return actions.user.code_private_function(text)
+        elif mods == {"private", "static"}:
+            return actions.user.code_private_static_function(text)
+        elif mods == {"pro"}:
+            return actions.user.code_protected_function(text)
+        elif mods == {"pro", "static"}:
+            return actions.user.code_protected_static_function(text)
+        elif mods == {"pub"}:
+            return actions.user.code_public_function(text)
+        elif mods == {"pub", "static"}:
```
pokey, line 66: then here we'd use the more verbose identifiers from above

lang/typescript/typescript.py
pokey: Sorry–not following.  Why do we need this for typescript but not javascript?

tests/stubs/talon/__init__.py
```diff
@@ -1,12 +1,21 @@
...
 
     def __getattr__(self, name):
-        return self.dict[name]
+        for category in ("test", "module"):
+            cat_actions = self.registered_actions[category]["namespaced"]
+            if self.namespace in cat_actions:
+                if name in cat_actions[self.namespace]:
+                    return cat_actions[self.namespace][name]
+
+        print(self.registered_actions)
```
pokey, line 17: do we want to keep the `print`?

pokey (commented):
Ok basically there; a couple minor comments left.  I'm not comfortable enough with the test code to review it, so maybe @rntz or @wolfmanstout could have a look at that part?

tests/stubs/talon/__init__.py
```diff
@@ -16,30 +25,78 @@ class Actions:
...
+        if namespace == "":
+            self.registered_actions[category]["non_namespaced"][name] = func
+        else:
+            if namespace not in self.registered_actions[category]["namespaced"]:
+                self.registered_actions[category]["namespaced"][namespace] = {}
+
+            self.registered_actions[category]["namespaced"][namespace][name] = func
+
+    def __getattr__(self, name):
+        callable = None
```
wolfmanstout, line 75: Could remove this and simply `return object.__getattribute__(self, name)` in the try clause.

tests/stubs/talon/__init__.py
```diff
@@ -16,30 +25,78 @@ class Actions:
...
+        except AttributeError:
+            pass
+
+        if callable:
+            return callable
 
-    def _build_namespace_accessor(self, key):
-        return DictAccessor(self.registered_actions[key])
+        found_namespace = False
+        for category in ("test", "module"):
```
wolfmanstout, line 85: This duplicates some of the logic in RegisteredActionsAccessor. Would be cleaner to simply have a RegisteredActionsAccessor for "main" and try using it first, then falling back to RegisteredActionsAccessor(self.registered_actions, name) and letting that constructor throw an error if the namespace doesn't exist instead of doing the RuntimeError here.

tests/stubs/talon/__init__.py
```diff
@@ -16,30 +25,78 @@ class Actions:
     """
 
     def __init__(self):
-        self.reset_actions()
+        self.registered_actions = {
+            "module": {"non_namespaced": {}, "namespaced": {}},
```
wolfmanstout, line 29: All actions in talon have a namespace -- it's just that there is syntactic sugar for your convenience to refer to actions in the "main" namespace. But you can also refer to them prefixed, e.g. "main.key()". Would be simpler to remove the distinction here and simply use "main" if the namespace is empty. That will let you drop one level of the dictionary.

wolfmanstout (changes requested):
Thanks, this is some nice new functionality for the tests!

lang/typescript/typescript.py
splondike: user.code_default_function's default implementation delegates to user.code_private_function. In javascript.py's code_private_function outputs `function myFunc` which is the desired behaviour. In typescript.py code_private_function outputs `private function myFunc`. This means in typescript there was previously no way of outputting a function without modifiers (which is what you'd want to do outside a class context).

pokey: I guess I'm not seeing why do we want `code_private_function` to do one thing in javascript and another thing in typescript

tests/stubs/talon/__init__.py
splondike: This seems to work for me in .talon land, but not in python (i.e. actions.main.key("t") doesn't work in the repl).

splondike:
Have addressed the above code review comments.

tests/stubs/talon/__init__.py
```diff
@@ -1,12 +1,23 @@
...
+        for category in ("test", "module"):
+            cat_actions = self.registered_actions[category]
+            if self.namespace in cat_actions:
+                if name in cat_actions[self.namespace]:
+                    return cat_actions[self.namespace][name]
+
+        raise AttributeError(f"Couldn't find action {self.namespace}.{name}")
+
+    def __call__(self, *args, **kwargs):
+        raise RuntimeError(f"actions.{self.namespace}() is not an available action")
```
wolfmanstout, line 20: Consider adding a comment for why this is needed.

tests/stubs/talon/__init__.py
```diff
@@ -16,30 +27,66 @@ class Actions:
     """
 
     def __init__(self):
-        self.reset_actions()
+        self.registered_actions = {
+            "module": {},
+            "test": {},
+            # "contexts": []
```
wolfmanstout, line 33: Remove?

wolfmanstout (changes requested):
Thanks, the test code looks good to me now.

lang/javascript/javascript.py
```diff
@@ -248,6 +248,16 @@ def code_insert_function(text: str, selection: str):
         actions.user.paste(text)
         actions.edit.left()
 
+    def code_default_function(text: str):
```
pokey, line 251: Ok I see that you have moved this one into Javascript, so now I'm confused.  Won't that mean that there's no way of outputting a function without modifiers in Typescript, like you suggested [earlier](https://github.com/knausj85/knausj_talon/pull/879/files/16c12cb8cf5f372df760d3af0b679832268ac2d6#r895266125)?

splondike: It works as you'd expect; `function helloWorld` output by `funky hello world` in both cases. The reason is typescript.talon enables the user.javascript tag which layers in all the stuff in javascript.py.

The implementation of code_default_function prior to this PR called code_private_function. In typescript this output `private function helloWorld`, and in javascript `function helloWorld`. After this PR code_default_function (for both .js and .ts) explicitly outputs `function helloWorld`, it doesn't delegate to another action.

I moved the definition up to javascript because I thought you were objecting to where I'd put it semantically (even if the behaviour was correct). The argument there would be something like: 'typescript is a specialised version of javascript, and so common behaviour should be in the javascript level'.

pokey (approved):
Ok I think this one is good enough to merge, other than a couple comments remaining from @wolfmanstout.

I'd also be tempted to do the following, but I think it totally fine to just file a follow-up issue instead:

- [ ] Add support for [javascript privates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields), having the `#` be part of the private function formatter
- [ ] Switch default to use default function formatter or public function formatter so the above doesn't break things
- [ ] Add support for static methods and static private methods to JS
- [ ] Remove support for protected methods in javascript, including commented code
- [ ] Add support for static private methods (using typescript private) and static methods of other visibilities in typescript

splondike:
I'll let somebody else take on those javascript changes in another PR. This one was just about putting in the capability to have flexible function signature modifiers as discussed in #825.

PR #881, opened 2022-06-06
pick item from a dropdown

NikolaiKircher:
I stole this code from someone elses repository and thought this might be useful for other people as well.

This dropdown picker can come in handy in lots of places like the browser or an IDE.

I am not sure whether this might conflict with other commands?
Maybe a prefix like "dropdown" would be helpful, so that you could say "dropdown pick three" to select the third item in the dropdown.
But that makes the command less useful because it needs more time to say it.

What do you guys think about this? Also, this is my first PR.


nriley:
I use something quite similar.

https://github.com/nriley/knausj_talon/blob/b23a2b87/misc/pick.talon

I like the idea of ‘pick up’ and will probably steal it!

The main differences with mine are (a) waiting a little while between selecting and pressing enter (probably not worth doing if this works well for most people) and (b) having “pick” alone just select the item under the cursor.  I use “pick” as my command for “enter” pretty much everywhere and find it very rarely conflicts.

splondike:
I'm personally slightly inclined against adding this feature to knausj for the following combination of reasons:

* It is a 'shortcut' rather than adding new functionality. The current competition for 'pick three' is 'go down three times enter'. I have an inclination toward letting people make their own idiosyncratic shortcuts out of the existing functions, particularly if they're just a short snippet of .talon. IMO we have a problem of command discoverability in knausj already (partially addressed by cheat sheets) so should have a slight bias against new global commands, particularly for this kind of small improvement.
* It globally takes over the nice command prefix 'pick' which people might want to use for other things.
* 'pick up' is dangerous the same way enter is if misrecognised. Imagine getting misrecognised as 'pick up three' in the terminal (execute a random history command), or 'pick three' in a web form (form submitted).

None of these things is definitive in itself, but they add up for me.

It's a cool idea though. I'd be more inclined to put this kind of thing in to a list of 'ideas'/inspiration somewhere so people could get their idiosyncratic creative process started. This would probably be a new talon.wiki page.

Anyway, just my two cents. Happy to be talked out of the above :).

nriley:
I don't feel strongly given how simple the command definition is.

However, having had this command in place for 11 months I can't think of any case where it's accidentally triggered. But I realize my accent/dictation environment varies!  (Before "pick" I used "choose" which got misinterpreted as "two" quite a lot.)

Perhaps we can talk about it more tomorrow…

rntz:
This does seem useful. I'm not opposed to having commands for very generally useful short sequences of actions like this. I do worry about possible misrecognition, though; `pick` is super short. Maybe `dropdown pick` or `dropdown`?

nriley:
Another idea we had in the knausj grooming session was to implement this as an action. That way, if we have a smarter way of selecting from a list (e.g., using accessibility, adding delays or accounting for an off-by-one error) we can easily override in a particular context.

splondike:
Sounds like other folks are in favour of this, so let's put it in. As for the prefix, here are some options:

1. :grinning: `pick three` and `pick up three`.
2. :tada: `dropdown pick three` and `dropdown pick up three`
3. :heart: `dropdown three` and `dropdown up three`
4. :rocket: `pick list three` and `pick list up three`
5. :confused: Something else.

My favourite is option #4. Let's try voting by reacting to this comment with the emoji next to the option. You can of course write a comment if you have another suggestion :). We can then get this added in with the preferred option in a few weeks.

As for the action idea, I'd be happy for that to be included. If not it could be easily added later I think.

splondike:
Even split, though I guess Nikolai probably likes pick <num> as well.

What do you think about using dropdown as the prefix instead Nikolai (I could change my preference to that)? Also the idea of implementing this as an action so people can re-implement it using accessibility APIs if they want to?

rntz (approved):

rntz:
I've altered this to use "dropdown" instead of "pick". If someone wants to make an action for this we can do it in another PR. Merging.

PR #887, opened 2022-06-19
Extend basic css functionality to scss files

SandeepJoel:


code/code.py
```diff
@@ -15,6 +15,7 @@
     # 'cplusplus': 'cpp hpp',
     "csharp": "cs",
     "css": "css",
+    "scss": "scss",
```
Will-Sommers, line 18: Can you delete `sass` on line 39 below? It also looks like this is a sorted list so we may want to move this down.

Will-Sommers (commented):
Looks pretty good!

auscompgeek (approved):
Nice.

splondike (approved):

splondike:
Thanks!

PR #888, opened 2022-06-19
Add some useful micro commands and fix bug on the css `variable` command

SandeepJoel:


misc/menu.talon
```diff
@@ -0,0 +1,3 @@
+-
+pick <number_small>: key("down:{number_small-1} enter")
```
splondike, line 2: This is being discussed in #881 .

misc/standard.talon
```diff
@@ -17,7 +17,10 @@ scroll up: edit.page_up()
 scroll down: edit.page_down()
 copy that: edit.copy()
 cut that: edit.cut()
-paste that: edit.paste()
+(pace | paste) that: edit.paste()
```
splondike, line 20: I haven't had any issues with 'paste that', does pace help with your accent?

text/symbols.talon
```diff
@@ -29,6 +29,10 @@ empty escaped string:
     "\\'\\'"
     key(left)
     key(left)
+empty square:
+  insert("[]")
```
splondike, line 33: What would you use these two commands for?

splondike (commented):
I like the vscode and find_and_replace commands. Have left a couple of questions and comments on the others.

misc/standard.talon
SandeepJoel: yes

text/symbols.talon
SandeepJoel: During variable assignment of empty arrays and empty objects

misc/menu.talon
SandeepJoel: Sure I will remove them

text/symbols.talon
splondike: The other empty* commands seem to place the cursor between the inserted wrappers. From what you say it sounds like this wouldn't work for you (and the 'empty square' at least would be redundant with 'op subscript'). I'm thinking to keep things consistent we leave these two out of knausj?

lang/css/css.talon
```diff
@@ -32,8 +32,8 @@ value <user.number_string> point <digit_string> [{user.css_unit}]:
 value <user.text>: user.insert_formatted(text, "DASH_SEPARATED")
 
 variable <user.text>:
-    name = user.formatted_text(text, settings.get("user.code_public_variable_formatter"))
-    insert("var(--{name})")
+    name = user.formatted_text(text, "DASH_SEPARATED")
+    insert("var(--{name});")
```
splondike: I dunno if the trailing semicolon is what you'd always want. I know people most often use CSS variables for single-value colours, but you could use them in multi-value contexts as well.

splondike (approved):

splondike:
Thanks!

rntz:
Question: does the "pace" alternative here really increase recognition accuracy for you, @SandeepJoel ? In what contexts? It seems like there's no other similar commands that these could be misheard as...

SandeepJoel:
@rntz During many times, `paste that` would get recognized as `b10` for me. And it seems `pace` works well when we are issuing multiple commands at a faster pace, such that you don't need to emphasis on `te` ending of the `paste`

PR #891, opened 2022-06-21
Add vscode settings comment for text navigation

ShuPink:
I spent my morning trying to understand why the `text_navigation` module worked in notepad but not VSCode.

Turns out that VSCode has a setting `editor.emptySelectionClipboard` where if your cursor has nothing selected you'll select the whole line instead, which leads `navigate_right` and `navigate_left` actions to fail.

I've been wanting to use these commands for a long time but it kept going into the too hard basket.
Hope this comment might help someone else.


splondike (approved):
Thanks, hopefully it helps somebody :).

PR #893, opened 2022-06-21
formatters.py: no more "over"-removal

rntz:
I think this code is dead. Plan is to comment it out and remove it later if it doesn't cause problems.


splondike (approved):

splondike:
Yeah this is some mysterious ancestral code. Happy to try commenting it out and see what happens.

PR #894, opened 2022-06-21
oops

rntz:
Fix mistaken use of square brackets in `talon_draft_windows/draft_window.talon`.

### DIFF
--- a/talon_draft_window/draft_window.talon
+++ b/talon_draft_window/draft_window.talon
@@ -6,34 +6,34 @@ settings():
   user.context_sensitive_dictation = 1
 
 # Replace a single word with a phrase
-[change | replace] <user.draft_anchor> [with] <user.text>:
+(change | replace) <user.draft_anchor> (with) <user.text>:
   user.draft_select("{draft_anchor}")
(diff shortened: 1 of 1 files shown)


richardmcmillen:
The intent was to have cursorless style phrasing in the draft editor.

I agree that this isn't, totally correct, would the cursorless way of doing
this be?

> change air say some words

To replace the word at hint 'air' with the literal phrase 'some words'.

If you think this is different enough or confusing you can revert it.

EdJoJob (approved):

pokey:
No worries! To be clear, I am not advocating for one thing or another here, just saying how it works for Cursorless. I have no objection to doing things differently in the draft editor

I do agree with @rntz that "change ... with ..." is a bit awkward, but that doesn't necessarily mean you need to mimic Cursorless exactly. Eg switching to "change ... to ..." could be reasonable.

The main reason we don't support that grammar in Cursorless, fwiw, is that we prefer to allow people to use their own / knausj insertion commands, as they are fairly sophisticated / might be custom

But again, I'm happy to defer to @splondike / other maintainers on this one as I'm not an active user of the draft editor

PR #902, opened 2022-07-01
Fix dictation dance in vscode

pokey:
Fixes bug with dictation dance in VSCode where empty selection caused copy to get whole line and mess things up

Note: implemented in pairing session with @rntz

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -422,16 +422,15 @@ class Actions:
         # take two words left. I also tried taking a line up + a word left, but
         # edit.extend_up() = key(shift-up) doesn't work consistently in the
         # Slack webapp (sometimes escapes the text box).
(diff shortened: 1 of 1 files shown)


nriley:
Might be helpful to note that this also effects several other editors including at least Sublime Text/Merge.

I just ran into the problem with https://github.com/knausj85/knausj_talon/blob/main/text/text_navigation.py today as I had forgotten to turn this off in Sublime Merge — probably worth a bug report if not fixed in this PR.

rntz (approved):

rntz:
@nriley this PR didn't fix anything about text_navigation.py so feel free to submit that as a separate issue.

nriley:
Thanks, filed #917.

PR #903, opened 2022-07-02
symbols: Add spaces in under score and down score

cgoates:
This resolved some recognition issues for me, and seems to follow the general rules of thumb that I've heard to improve recondition.

### DIFF
--- a/text/symbols.talon
+++ b/text/symbols.talon
@@ -1,5 +1,4 @@
 question [mark]: "?"
-(downscore | underscore): "_"
 double dash: "--"
 (bracket | brack | left bracket): "{"
 (rbrack | are bracket | right bracket): "}"
--- a/code/keys.py
+++ b/code/keys.py
@@ -170,7 +170,7 @@ symbol_key_words = {
     "tilde": "~",
     "bang": "!",
     "down score": "_",
-    "under score": "_",
+    "underscore": "_",
     "paren": "(",
     "L paren": "(",
     "left paren": "(",


auscompgeek (approved):
Yeah, `down score` should absolutely have a space. `underscore`, whilst it is a real word, if adding a space helps with recognition for people it probably makes sense.

text/symbols.talon
```diff
@@ -1,5 +1,5 @@
 question [mark]: "?"
-(downscore | underscore): "_"
+(down score | under score): "_"
```
pokey, line 2: Ok so I think the problem here is that these appear here and also implicitly via keys.py / keys.talon, with different spacing. This throws off talon's precedence rules. I would just remove this line

Also, given that this one was winning, I'd be tempted to use "underscore" in keys.py

@lunixbochs does that sound reasonable?

pokey (dismissed):

text/symbols.talon
cgoates: Okay, that makes sense. I made this suggested change and that gets rid of the issue I was having as well as my original solution does. I see a few other duplications between this file and keys.py/keys.talon; is that something that should be cleaned up as well, or is there a reason to keep both around? Mostly just asking for curiosity's sake, though I'm happy to do more here if that is desired.

auscompgeek: Ah, of course. Yeah, we realistically shouldn't have duplicated command phrases.

rntz (approved):

rntz:
ok, so now we just have keys named `underscore` and `down score`. if recognition issues continue we can also add in the variant `under score`.

PR #906, opened 2022-07-02
improved golang support

asoee:
This should improve the golang support a bit.
Probably not ready for merge yet, but suggestions are welcome.


lang/go/go.talon
```diff
@@ -194,3 +269,19 @@ swipe [<user.text>] [over]:
...
+
+receiver function <user.text> for <user.text>:
+    insert("func (")
+    insert(user.formatted_text(text_2, "ABBREVIATION"))
+    insert(" *")
+    insert(user.formatted_text(text_2, "PUBLIC_CAMEL_CASE"))
+    insert(") ")
+    insert(user.formatted_text(text_1, "PUBLIC_CAMEL_CASE"))
+    insert("() {")
+    key("Enter")
```
pokey, line 287: This feels like it should be done using snippets.  Not sure it makes sense to have something this complex built out of `insert`s.  I'd be inclined to just leave this one out and then at some point maybe we have a separate set of common snippets.  I'm also a bit hesitant about the `ABBREVIATION` formatter here

lang/go/go.talon
```diff
@@ -171,7 +243,10 @@ append <user.text> [over]:
 state (air | err): "err"
 error: " err "
 loop over [<user.text>] [over]:
-    insert("forr ")
+    insert("forr")
+    sleep(100ms)
+    key(tab)
+    sleep(100ms)
```
pokey, line 249: What is this doing?

lang/go/go.talon
```diff
@@ -194,3 +269,19 @@ swipe [<user.text>] [over]:
     key("right")
     insert(", ")
     insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
+
+type <user.text> struct:
+    insert("type ")
+    insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
+    insert(" struct {")
+    key("Enter")
```
pokey, line 277: Is this relying on editor inserting a `}`?  Worth looking to see how we do this for eg function definitions / `for` loops / `if` statements in other C-style languages

lang/go/go.talon
```diff
@@ -128,11 +201,11 @@ interface <user.text> [over]:
...
 
 state string: " string "
-[state] (int | integer | ant): "int"
+[state] (int | integer): "int"
 state slice: " []"
 slice of: "[]"
-[state] (no | nil): "nil"
-state (int | integer | ant) 64: " int64 "
+[state] (nil): "nil"
+state (int | integer) 64: " int64 "
```
pokey, line 208: I find the "no" helps accuracy for me.  Presumably the "ant" is there for the same reason

lang/go/go.talon
```diff
@@ -117,6 +185,11 @@ state (start | struct | struck):
     insert(" struct {")
     key("enter")
     insert(user.formatted_text(text, "PUBLIC_CAMEL_CASE"))
+type struct <user.text>:
+    insert("type ")
+    insert(user.formatted_text(text, "PUBLIC_CAMEL_CASE"))
+    insert(" struct {")
+    key("enter")
```
pokey, line 192: This appears to be a duplicate of the type struct below with slightly different syntax.  I'd have one, possibly with a `|` to support both syntaxes

lang/go/go.talon
```diff
@@ -66,25 +113,46 @@ else <user.text> [over]:
...
     insert(user.formatted_text(text, "PUBLIC_CAMEL_CASE"))
+    insert("(\"")
+
+format print line [over]:
+    insert("fmt.Println")
+    insert("(\"")
+
+format print format [over]:
+    insert("fmt.Printf")
+    insert("(\"")
```
pokey, line 154: These should just be in "funk" list (common functions).  In particular the more common one should just be "funk print", as that is standard in many languages

lang/go/go.talon
```diff
@@ -66,25 +113,46 @@ else <user.text> [over]:
...
-    insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
+    user.insert_between("for "," {}")
 
 state for: "for "
-for <user.text> [over]:
+state for <user.text> [over]:
     insert("for ")
     insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
 
+snip for range: "forr"
```
pokey, line 132: why is this a snippet?

lang/go/go.talon
```diff
@@ -66,25 +113,46 @@ else <user.text> [over]:
...
+
 state for range: "forr "
 range <user.text> [over]:
     insert("forr ")
     insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
 
+[state] for i:
```
pokey, line 139: I'd make the "state" mandatory here to not misfire.  I'd also be inclined to have this one be in a more standard location (eg in the imperative tag) and implemented via an action, as this is something that appears across many languages.

lang/go/go.talon
```diff
@@ -66,25 +113,46 @@ else <user.text> [over]:
     key("enter")
     insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
 
-state while: "while "
```
pokey, line 69: why did you remove this one?  Fwiw this should probably become a keyword using the keyword mechanism we have for other languages

lang/go/go.talon
```diff
@@ -66,25 +113,46 @@ else <user.text> [over]:
...
+[state] if (err|error):
+    insert("if err != nil {")
+    key("enter")
+
+[state] if not (err|error):
+    insert("if err == nil {")
+    key("enter")
```
pokey, line 122: We should prob standardise on how all htese things with `{` followed by enter work, and do the same thing in all languages

Also please make "state" mandatory

lang/go/go.talon
```diff
@@ -49,6 +92,10 @@ state break: "break"
 state (chan | channel): " chan "
 state go: "go "
 state if: "if "
+state defer: "defer "
+string convert: "strconv."
+string convert to int: "strconv.AtoI("
```
pokey, line 97: probably add to common functions?

lang/go/go.talon
```diff
@@ -16,6 +52,13 @@ state comment: "// "
 #     text_with_leading(" // ")
 # ]
 # "[state] context: insert("ctx")
+
+
+state (package): "package "
```
pokey, line 57: why the parens?

lang/go/go.talon
```diff
@@ -49,6 +92,10 @@ state break: "break"
 state (chan | channel): " chan "
 state go: "go "
 state if: "if "
+state defer: "defer "
```
pokey, line 95: these could prob go into keywords tbh

lang/go/go.talon
```diff
@@ -1,8 +1,44 @@
...
+    user.code_public_function_formatter = "PUBLIC_CAMEL_CASE"
+    user.code_private_variable_formatter    = "PRIVATE_CAMEL_CASE"
+    user.code_protected_variable_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_public_variable_formatter = "PRIVATE_CAMEL_CASE"
+
+
+(variadic | spread): "..."
 logical and: " && "
 logical or: " || "
+boolean: "bool"
```
pokey, line 31: this should be in the type list, no?

lang/go/go.talon
```diff
@@ -1,8 +1,44 @@
...
+assign: " := "
+channel (receive|send): " <- "
+
+# Time conveniences
+time dot millisecond: "time.MilliSecond"
+buf i o: "bufio"
```
pokey, line 37: maybe should be a keyword?  Prob should be prefixed at least

lang/go/go.talon
```diff
@@ -1,8 +1,44 @@
...
+settings():
+    user.code_private_function_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_protected_function_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_public_function_formatter = "PUBLIC_CAMEL_CASE"
+    user.code_private_variable_formatter    = "PRIVATE_CAMEL_CASE"
+    user.code_protected_variable_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_public_variable_formatter = "PRIVATE_CAMEL_CASE"
+
+
+(variadic | spread): "..."
```
pokey, line 28: should probably be "op spread".  Do we not have a standard operator for this one?

code/formatters.py
```diff
@@ -111,6 +111,7 @@ def every_word(word_func):
         first_vs_rest(lambda w: w.lower(), lambda w: "_" + w.lower()),
     ),
     "NO_SPACES": (NOSEP, every_word(lambda w: w)),
+    "ABBREVIATION": (NOSEP, every_word(lambda w: w[0])),
```
pokey, line 114: As mentioned below, mixed feelings about this formatter.  If we do remove the use case, I'd also remove the formatter

code/formatters.py
```diff
@@ -151,6 +152,7 @@ def every_word(word_func):
     "snake": formatters_dict["SNAKE_CASE"],
     "string": formatters_dict["SINGLE_QUOTED_STRING"],
     "title": formatters_dict["CAPITALIZE_ALL_WORDS"],
+    "abbreviate": formatters_dict["ABBREVIATION"],
```
pokey, line 155: Either way, I wouldn't make it exposed as a formatter people can use, as they might as well just spell using the Talon alphabet

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
+from talon import Context, Module, actions, settings
+
+ctx = Context()
+mod = Module()
+ctx.matches = r"""
+tag: user.go
+"""
+
+# Primitive Types
+ctx.lists["self.go_types"] = {
```
pokey, line 10: Please use the standard type list that we use in eg Python and Javascript

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+    "long": "long",
+    "char": "char",
+    "string": "string",
+    "rune": "rune",
+    "void": "void",
+}
+
+ctx.lists["self.go_pointers"] = {
+    "pointer": "*",
+}
```
pokey, line 26: why do we need a list for this?

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+@mod.capture(rule="{self.go_pointers}")
+def go_pointers(m) -> str:
+    "Returns a string"
+    return m.go_pointers
+
+
+@ctx.action_class("user")
+class UserActions:
+    def code_operator_lambda():
+        actions.insert(" -> ")
```
pokey, line 47: I don't believe this is how lambdas are expressed in Go

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+        actions.insert("nil")
+
+    def code_insert_is_null():
+        actions.insert(" == nil")
+
+    def code_insert_is_not_null():
+        actions.insert(" != nil")
+
+    def code_state_if():
+        actions.user.insert_between("if ", " ")
```
pokey, line 155: What is the `" "` doing here?

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def code_state_else_if():
+        actions.user.insert_between("else if ", " ")
+
+    def code_state_else():
+        actions.insert("else ")
+        actions.key("enter")
+
+    def code_state_switch():
+        actions.user.insert_between("switch (", ") ")
```
pokey, line 165: do we want to add the `{` and possibly `}` here?  we should prob standardise across languages here cc/ @rntz 

Also looks like we don't want the `()`, no?

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def code_state_case():
+        actions.user.insert_between("case ", " :")
```
pokey, line 168:
```suggestion
        actions.user.insert_between("case ", ":")
```

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def code_state_for():
+        actions.user.insert_between("for ", " ")
+
+    def code_state_while():
+        actions.user.insert_between("for ", " ")
```
pokey, line 174: maybe worth a comment

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+        actions.key("enter")
+
+    def code_state_switch():
+        actions.user.insert_between("switch (", ") ")
+
+    def code_state_case():
+        actions.user.insert_between("case ", " :")
+
+    def code_state_for():
+        actions.user.insert_between("for ", " ")
```
pokey, line 171: As above re `" "`?

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+    def code_state_return():
+        actions.insert("return ")
+
+    def code_comment_line_prefix():
+        actions.insert("// ")
+
+    def code_insert_function(text: str, selection: str):
+        text += f"({selection or ''})"
+        actions.user.paste(text)
+        actions.edit.left()
```
pokey, line 203: a bit confused by this one; did you copy it from another lang implementation?

lang/go/go.py
```diff
@@ -0,0 +1,213 @@
...
+        actions.insert("true")
+
+    def code_insert_false():
+        actions.insert("false")
+
+    def code_import():
+        actions.insert("import ")
+
+    def code_private_function(text: str):
+        actions.insert("private")
```
pokey, line 192: This duplicate of `code_private_function`.  I believe the other one is what we want
```suggestion
```

pokey (changes requested):
Great start!  Left a bunch of small comments / questions

asoee:
> Great start! Left a bunch of small comments / questions

Thanks for the feedback.
Will go through it, when i get back from holiday.

lang/go/go.talon
asoee: This one uses a go-specific vscode snippet, tabs through the positions, and inserts the name at the right slot.
I wondered if this should go in a separate .talon file, triggered on the combination of golang AND vscode,
to allow for snippets from different editors.

lang/go/go.py
asoee: fixed

lang/go/go.talon
asoee: There is no while keyword in go: https://golangdocs.com/while-loop-in-golang

lang/go/go.talon
```diff
@@ -1,8 +1,44 @@
...
+tag(): user.code_data_bool
+tag(): user.code_data_null
+tag(): user.code_functions
+tag(): user.code_libraries
+tag(): user.code_operators_array
+tag(): user.code_operators_assignment
+tag(): user.code_operators_bitwise
+tag(): user.code_operators_lambda
+tag(): user.code_operators_math
+tag(): user.code_operators_pointer
```
auscompgeek, line 17: Are the actions for the pointer tag implemented? They seem to be missing.

```python
    def code_operator_indirection():
        actions.insert("*")

    def code_operator_address_of():
        actions.insert("&")
```

lang/go/go.py
auscompgeek: The same action inserts the parens in JS and other C-like languages, so no.

theryangeary: Go does not require them.

theryangeary: I'm suggesting adding the opening bracket of the if block, not the opening paren of the if considering. I'm not sure if that was clear.

splondike:
@asoee Are you still keen to get this merged sometime soon, or should we just close it out for the moment?

splondike:
I think I'll leave this open for another week and will close it out around the 11th Jan unless I hear otherwise.

splondike:
Closing this out since it sounds like it might not be worked on for some time.

PR #914, opened 2022-07-08
Insert between more

rntz:
use `user.insert_between` more often to compact code.

### DIFF
--- a/text/symbols.talon
+++ b/text/symbols.talon
@@ -14,41 +14,17 @@ dub arrow: "=>"
 new line: "\\n"
 carriage return: "\\r"
 line feed: "\\r\\n"
-empty dubstring:
-    '""'
-    key(left)
-empty escaped (dubstring|dub quotes):
-    '\\"\\"'
-    key(left)
-    key(left)
-empty string:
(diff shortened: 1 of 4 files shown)


text/symbols.talon
```diff
@@ -14,41 +14,17 @@ dub arrow: "=>"
...
-inside (double quotes | dubquotes):
-    insert('""')
-	key(left)
-inside (graves | back ticks):
-	insert("``")
-	key(left)
+empty dubstring: user.insert_between('"', '"')
+empty escaped (dubstring|dub quotes): user.insert_between('\\"', '\\"')
+empty string: user.insert_between("'", "'")
+empty escaped string: user.insert_between("\\'", "\\'")
```
pokey, line 20: Huh I personally use "empty" to mean don't leave cursor inside.  But I guess you're not changing anything here 🤷

pokey (approved):

text/symbols.talon
rntz: Yeah, I do the same thing, but this appears to be the default.

PR #915, opened 2022-07-08
faster dictation dance

rntz:
I optimized the fixes made in #902 to involve fewer keypresses, which will help when these are expensive (eg. on Linux :P). Mainly:

1. I removed the clobber argument to dictation_peek_left, we only ever call it with clobber=True and inserting space will already clobber the selection, so this gets rid of a space-backspace pair.

2. I used edit.extend_line_end in dictation_peek_right instead of edit.extend_right multiple times.

# Todo

- [ ] Test linux firefox

### DIFF
--- a/code/dictation.py
+++ b/code/dictation.py
@@ -437,12 +437,18 @@ class Actions:
         # Insert space to ensure something to select (see dictation_peek_left).
         actions.insert(" ")
         actions.edit.left()
-        # We select to end of line because it's usually a single keypress and we
-        # only depend on what's on the current line. Without this trick we might
-        # need to select at least three characters to the right: one for the the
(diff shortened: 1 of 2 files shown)


pokey (changes requested):
I tried this locally, and in the following case:

```
hello |world
```

with `|` being the cursor, if I say "testing", I get the following:


```
hello  testing world
```

Note the two spaces after `hello`.  This does not occur on `main`, where I get only one space, as expected

rntz:
Yeah, I think I accidentally duplicated key(" ") in dictation_peek_left when I cherry-picked this from my local changes. Good catch.

pokey (approved):

PR #918, opened 2022-07-13
A couple missing rust commands

RyanRuane:


code/create_spoken_forms.py
```diff
@@ -1,8 +1,9 @@
 import itertools
 import re
 from collections import defaultdict
+from collections.abc import Mapping
```
rntz, line 4: What's up with these changes from typing to collections.abc?

rntz (approved):
Looks okay. What's the purpose of changing from typing to collections.abc though?

code/create_spoken_forms.py
RyanRuane: good question. I think my branch may have been out of sync and not up to date. I can put it back.

pokey: Our autoformatter started doing that last week.  Apparently as of 3.9 they're starting [to deprecate](https://peps.python.org/pep-0585/#:~:text=Importing%20those%20from%20typing%20is%20deprecated) the versions from `typing` in favour of those from `collections.abc`

lang/rust/rust.talon
```diff
@@ -48,6 +48,7 @@ toggle use: user.code_toggle_libraries()
 ## Simple aliases
 borrow: "&"
 borrow mutable: "&mut "
+state (async | asynchronous): "async "
```
pokey, line 51: I use "a sink" for this, as I found recognition better.  Not sure our stance regarding these kind of games to improve recognition tho.  cc/ @knausj85 @lunixbochs

code/create_spoken_forms.py
auscompgeek: Yeah, I think this is coming from a new [pyupgrade](https://github.com/asottile/pyupgrade) release.

auscompgeek: Looks like this change was made in #921, so if you rebase this change should disappear.

lang/rust/rust.talon
auscompgeek: At some point it might be worth adding this under a common tag to make this consistent between languages.

pokey: Fwiw we support this [as a keyword](https://github.com/knausj85/knausj_talon/blob/3eab2c9525c3896718152c1f88e0f2d0ceee1237/lang/javascript/javascript.py#L48) in Javascript

RyanRuane:
rebased

auscompgeek (approved):

Issue #920, opened 2022-07-14
Inconsistent function names/descriptions in operators_assignment.py and operators_assignment.talon


chrisnicollo:
For some of the bitwise operator assignment functions, their names end in “equals” in operators_assignment.talon and in “assignment” in operators_assignment.py. It seems like the convention here is “assignment” and that the function names ending in “equals“ should be changed to end with “assignment” instead. Additionally, some of the doc comments in operators_assignment.py do not match the name of their function, as with function ‘code_operator_subtraction_assignment’ and its comment “””code_operator_subtraction_equals”””

pokey:
I believe this issue was addressed by #920

PR #921, opened 2022-07-14
#920 Inconsistent function names/descriptions in operators_assignment.py and operators_assignment.talon

chrisnicollo:
#920 
This should make it so that the bitwise assignment operators are now able to be used.

### DIFF
--- a/code/create_spoken_forms.py
+++ b/code/create_spoken_forms.py
@@ -1,8 +1,9 @@
 import itertools
 import re
 from collections import defaultdict
+from collections.abc import Mapping
 from dataclasses import dataclass
-from typing import Any, Mapping, Optional
+from typing import Any, Optional
 
 from talon import Module, actions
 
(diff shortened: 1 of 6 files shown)


chrisnicollo:
It looks like the pre-commit program wanted to push some changes to the code, but it couldn't because I didn't allow edits by maintainers. However, I have since fixed that so that it can fix the issues if the checks are run again.

pokey (approved):
Looks good to me!  Re pre-commit, maybe you didn't check the "Allow edits by maintainers" box?

chrisnicollo:
That's what it seems like to me, and I checked that box to fix it. I'm not sure if you have to run that pre-commit again to fix any errors.
Also, some of the bitwise assignment operators were never given spoken commands as far as I can tell. I could probably just follow talon convention on those to add them in, but I'm don't know whether to put them in the same pull request.

pokey (approved):
I pushed an empty commit to kick CI; seems to have done the trick

Re adding other spoken forms; go ahead if you think you'll use them!  Prob for follow-up PR tho; let's get this one merged in as it fixes a bug

Thank you!

Issue #925, opened 2022-07-17
what do we want to keep in CI?


rntz:
Applying pyupgrade in CI caused serious breakage ('say' and dictation and formatters not working): #922.

We've temporarily fixed this by removing shed from CI: #923.

Long term we probably want to add back some of what we removed. Which things and with what options?

phillco:
FYI @pokey - you might want to make a similar change to cursorless-talon

rntz:
Closing this as I think we've settled on a reasonably stable configuration after disaggregating what used to be bundled into shed. Reopen for further discussion.

PR #926, opened 2022-07-17
pre-commit-config: re-add Black and isort

phillco:
#925

This re-adds:

- the Black code formatter
- the isort import sorter

These provide most of the value we were getting from Shed and are probably the safest of what Shed was running. There are no changes to the codebase after running them.

If you want a more incremental change, just take the first commit, which adds only Black.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,6 +30,11 @@ repos:
     rev: "0.76"
     hooks:
       - id: flynt
+  - repo: https://github.com/pycqa/isort
+    rev: 5.6.4
+    hooks:
+      - id: isort
+        args: ["--profile", "black", "--filter-files"]
   - repo: https://github.com/psf/black
     rev: 22.6.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,8 +39,4 @@ repos:
     rev: 22.6.0
     hooks:
       - id: black
-        # It is recommended to specify the latest version of Python
-        # supported by your project here, or alternatively use
-        # pre-commit's default_language_version, see
-        # https://pre-commit.com/#top_level-default_language_version
         language_version: python3.9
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -30,3 +30,13 @@ repos:
     rev: "0.76"
     hooks:
       - id: flynt
+  - repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        args: ["--profile", "black", "--filter-files"]
```
auscompgeek, line 37: Thoughts on adding an isort config file instead of specifying this here?

phillco: I could go either way. Do you have a strong preference? I suppose the top level configuration file makes it easier to run `isort` yourself, although I tend to dislike the proliferation of tons of top level dotfiles.

auscompgeek: I know VSCode has an isort integration, I presume it'd respect an isort.cfg.

pokey: Isort supports pyproject.toml for config, so maybe we should use that to avoid giving isort its own config?

.pre-commit-config.yaml
```diff
@@ -30,3 +30,13 @@ repos:
...
+  - repo: https://github.com/psf/black
+    rev: 22.6.0
+    hooks:
+      - id: black
+        language_version: python3.9
```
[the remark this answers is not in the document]
rntz: it looks like it uses it for determining whether Python supports trailing commas and underscores in number literals. This seems fine.

rntz (approved):

.pre-commit-config.yaml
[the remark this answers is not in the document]
auscompgeek: Yeah, black deliberately avoids making changes to the AST, with the only exception being docstring indentation AFAIK.

phillco:
Going to merge now to ensure that we maintain formatting but will look at using `pyproject.toml` / `isort.cfg` as a followup.

PR #927, opened 2022-07-20
Save all files mapping

EdJoJob:
The platform edit utilities have `save_all` defined, but there is no verbal
mapping to that function.


rntz:
Do we have any implementations of this action anywhere? It might be confusing for people to say it if it does nothing.

pokey:
> Do we have any implementations of this action anywhere? It might be confusing for people to say it if it does nothing.

[Yes](https://github.com/knausj85/knausj_talon/search?q=save_all)!

rntz (approved):
Cool, let's do it then.

PR #930, opened 2022-07-20
window_snap.py: introduce alternative cross-screen snapping style that preserves window size

rntz:
Currently commands like `snap next` that snap windows across screens try to preserve both the relative position and size of windows. I have two screens of very different sizes and find changing the window size is usually not what I want. So I have added an alternate code path that tries to preserve the window's absolute size but keep its rough position relative to the screen. I've gated this behind a setting.

Concerns:

1. I'm not sure what the default behavior should be. I much prefer my version, but maybe that's just me? For now I've left the defaults unchanged.
2. I don't understand the existing code here very well, so I just added a big `if`; maybe there's some code sharing/simplification that could be done.
3. Maybe the setting should be named something else?


splondike (approved):
This would be a decent case for some unit tests, but I'm not too fussed either way.

Tested this on Linux and OSX and it worked fine in both cases (as you'd expect given it's just maths).

PR #932, opened 2022-07-21
add emacs support

rntz:
basic emacs support


nriley:
Thanks for starting to contribute some Emacs support!

One issue we'll run into at least on the Mac, and maybe on Linux, is that Meta is not always Alt. For example in my [preferred Emacs Mac port](https://bitbucket.org/mituharu/emacs-mac/src/master/) Command ("cmd-") is Meta. Probably we'll need to define an action that takes a placeholder like "meta-" or "M-" and maps it, perhaps based on a setting? I could even see some people would want to be able to use Esc as Meta. (Esc as Meta may actually be a reasonable default given it'll work everywhere.)

rntz:
ok, I've added a setting for how meta should be done: esc, alt, or cmd.

and spruced up a few other things.

rntz:
I need to test this locally, I've made a bunch of edits without running the result now.

apps/emacs/emacs.py
```diff
@@ -0,0 +1,165 @@
+from talon import Context, Module, actions
+import logging
+
+mod = Module()
+setting_meta = mod.setting(
+    'emacs_meta',
+    type=str,
+    default='esc',
+    desc="""What to use for the meta key in emacs. Defaults to 'esc', since that should work everywhere. Other options are 'alt' and 'cmd'.""",
```
ParetoOptimalDev, line 9: Does this mean you would have to change your emacs configuration to use escape or your talon configuration to use another choice?

apps/emacs/emacs.py
```diff
@@ -0,0 +1,165 @@
...
+        # NB. f1 works in ansi-term mode; C-h doesn't.
+        actions.key("f1")
+        if key is not None: actions.key(key)
+
+@ctx.action_class('user')
+class UserActions:
+    def split_window():               actions.key('ctrl-x 2')
+    def split_window_vertically():    actions.key('ctrl-x 2')
+    def split_window_up():            actions.key('ctrl-x 2')
+    def split_window_down():          actions.key('ctrl-x 2 ctrl-x o')
```
ParetoOptimalDev, line 56: Oh so in emacs language splits are tabs and tabs are windows.

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,320 @@
...
+app: Emacs
+-
+tag(): user.tabs
+tag(): user.splits
+tag(): user.line_commands
+
+
+# ----- GENERAL ----- #
+#suplex: key(ctrl-x)
+exchange: key(ctrl-x ctrl-x)
```
ParetoOptimalDev, line 11: I also like to have a `quit: key(ctrl-g)` here. that's probably my most used thing I noticed missing trying this config out.

gsingh93:
I'm not sure it's a good idea to bind all of these talon commands directly to key bindings, as users may have customized the bindings to perform different things. I think we should use `M-x <command-name>` where possible.

gsingh93:
I think another approach for implementing this would be:

1. Having a dictionary mapping Talon command to Emacs command, i.e.
```
command_map = {
    "file open": "find-file",
    ...
}
ctx.lists["user.emacs_commands"] = command_map
```

2. The Talon file could look like:
```
[emacs] {user.emacs_commands}: user.do_emacs_command(emacs_command)
```

3. `do_emacs_command` would do the following:
    - Look up what key to use for the given command in `settings/emacs_commands.csv`, i.e. 'find-file' would return 'ctrl-x ctrl-f'. A user could customize this CSV to change any of the bindings, and we can create an elisp script for users to generate it themselves, until we have something better like voicemacs
    - If there's no binding in the file, we fall back to `M-x`, i,e. `M-x find-file`

The benefits of this are it's more maintainable and extendable (it's easier to modify a dictionary or CSV file instead of a Python or Talon file), it's easier for users to use their own keybindings (with the elisp script), and when Voicemacs or some other server is ready for use, it'll be easier to integrate.

Let me know what you think.

rntz:
I like the idea of a CSV file to record metadata about commands (eg their keybindings; could also use this for "short forms" - M-x will accept abbreviations of commands as long as they're sufficiently unambiguous, which can be faster than typing the whole command if there's no direct keybindings). 

I'm less sure about whether spoken forms should go in that file vs just using a .talon file. Not all voice commands will correspond to a single Emacs command, for example. But I guess we can put the names for single commands into the CSV file and the rest into a talon file.

rntz:
I have updated this PR heavily, bringing it in-line with my current configuration, which is based on @gsingh93's suggestion to use a CSV file mapping commands to metadata about them, namely their key-binding and/or short-form. Almost all uses of `key(...)` have been replaced with `user.emacs(command_name)`, which performs the given command by whatever method is appropriate - by default, it uses a keybinding if known, otherwise M-x.

rntz:
@nriley if you'd be willing to give this a spin it would be greatly appreciated.

apps/emacs/emacs_commands.py
```diff
@@ -0,0 +1,104 @@
...
+    def emacs(command_name: str):
+        """
+        Runs the emacs command `command_name`. Defaults to using M-x, but may use
+        a key binding if known or rpc if available.
+        """
+
+
+@emacs_ctx.action_class("user")
+class UserActions:
+    def emacs(name):
```
nriley, line 97: Talon complains that the argument name does not match between the Context and Module.

apps/emacs/emacs.py
```diff
@@ -0,0 +1,328 @@
...
+    "emacs_meta",
+    type=str,
+    default="esc",
+    desc="""What to use for the meta key in emacs. Defaults to 'esc', since that should work everywhere. Other options are 'alt' and 'cmd'.""",
+)
+
+ctx = Context()
+ctx.matches = r"""
+app: Emacs
+app: emacs
```
nriley, line 16: Would recommend we define an 'emacs' app somewhere and then just use "app: emacs" everywhere rather than repeating this in multiple places. On Mac at least we should consider using a bundle matcher (`app.bundle: org.gnu.Emacs`)

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,316 @@
...
+    key(enter)
+describe variable <user.text>$:
+    user.emacs_help("v")
+    user.insert_formatted(text, "DASH_SEPARATED")
+    key(enter)
+
+# ----- FILES & BUFFERS -----
+# many key bindings here are often overridden, so we don't use user.emacs,
+# eg: C-x C-f, C-x k, C-x C-b, C-x right, C-x left
+file open: key(ctrl-x ctrl-f)
```
nriley, line 157: Should be using `user.emacs` here?

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,316 @@
...
+open directory: user.emacs("dired-jump")
+other open directory: user.emacs("dired-jump-other-window")
+fill paragraph: user.emacs("fill-paragraph")
+occurs: user.emacs("occur")
+other scroll [down]: user.emacs_meta("pagedown")
+other scroll up: user.emacs_meta("pageup")
+insert unicode: user.emacs("insert-char")
+abort recursive [edit]: user.emacs("abort-recursive-edit")
+save some buffers: user.emacs("save-some-buffers")
+[toggle] read only [mode] [toggle]: user.emacs("read-only-mode")
```
nriley, line 29: "toggle XXX" seems to be more common in knausj, perhaps remove the other options

apps/emacs/emacs_commands.csv
```diff
@@ -0,0 +1,171 @@
...
+kill-word, meta-d
+kmacro-end-and-call-macro, ctrl-x e
+kmacro-end-macro, ctrl-x )
+kmacro-start-macro, ctrl-x (
+list-packages
+mark-paragraph, meta-h
+mark-sexp, meta-ctrl-@
+mark-whole-buffer, ctrl-x h
+narrow-to-region, ctrl-x n n
+occur,ctrl-c o
```
nriley, line 84: M-s o is the default binding.

apps/emacs/emacs_commands.csv
```diff
@@ -0,0 +1,171 @@
+Command, Key binding, Short form, Spoken form
```
nriley, line 1: Do we want to standardize on a space after comma or not? It's inconsistent.

apps/emacs/emacs_commands.csv
```diff
@@ -0,0 +1,171 @@
...
+abort-recursive-edit, ctrl-]
+auto-fill-mode,,auto-f
+backward-paragraph, meta-{
+backward-sentence, meta-a
+backward-sexp, meta-ctrl-b
+backward-up-list, meta-ctrl-up
+backward-word, meta-b
+balance-windows, ctrl-x +
+beginning-of-buffer, meta-<
+browse-kill-ring
```
nriley, line 11: Duplicate.

apps/emacs/emacs_commands.csv
```diff
@@ -0,0 +1,171 @@
...
+occur,ctrl-c o
+open-rectangle, ctrl-x r o
+other-window, ctrl-x o
+overwrite-mode,,overwr
+package-autoremove
+pop-global-mark, ctrl-x ctrl-@
+pop-mark, ctrl-u ctrl-space
+project-find-file,ctrl-x p f
+project-find-file,ctrl-x p f
+project-find-regexp,ctrl-x p g
```
nriley, line 93: Duplicate.

apps/emacs/emacs_commands.csv
```diff
@@ -0,0 +1,171 @@
...
+mark-whole-buffer, ctrl-x h
+narrow-to-region, ctrl-x n n
+occur,ctrl-c o
+open-rectangle, ctrl-x r o
+other-window, ctrl-x o
+overwrite-mode,,overwr
+package-autoremove
+pop-global-mark, ctrl-x ctrl-@
+pop-mark, ctrl-u ctrl-space
+project-find-file,ctrl-x p f
```
nriley, line 91: Duplicate.

nriley (commented):
Things basically work in my environment. I haven't had time to review all the Talon commands but think they should be reviewed a bit more carefully for consistency/lack of duplication.

The one thing I was surprised didn't have a binding was keyboard-quit/C-g. I use "quit" in other circumstances; "cancel" would be good too except that there's "cancel cancel" already in knausj for other purposes.

apps/emacs/emacs.talon
nriley: Heh, I said the same thing today when I was reviewing. Missed this somehow :)

apps/emacs/emacs.talon
rntz: See comment immediately above

nriley: Oops, good point. Makes sense.

apps/emacs/emacs_commands.csv
rntz: fixed

apps/emacs/emacs_commands.csv
rntz: fixed

apps/emacs/emacs_commands.csv
rntz: fixed

apps/emacs/emacs_commands.csv
rntz: fixed

apps/emacs/emacs.talon
rntz: I heavily use the spoken forms without any `toggle` word, eg. just `line numbers`, `truncate lines`, `debug on error`, `read only`. I'd be okay with making the initial `toggle` mandatory though. Opinion?

apps/emacs/emacs_commands.csv
rntz: What do you mean by standardize? Do you mean, have spaces in other csv headers? We could do that, I guess, but I think it would be better done in another PR. Or do you mean spaces in the entries of the CSV - this CSV, or all of them?

apps/emacs/emacs.talon
rntz: I've made `[toggle]` only appear at the beginning, not the end. But I've kept the other alternates.

apps/emacs/emacs.talon
rntz: Actually, I think it's probably ok to use `user.emacs` for these? It'll end up pressing the keybindings in emacs_command.csv, so it'll use the remapped version. I was worried about trouble with RPC, but I can use `command-remapping` to get the remapped version, *assuming* it's remapped via a command remapping rather than a straight-up keybinding... so this'll catch things like ido (overrides `switch-to-buffer`) and visual-line-mode (overrides `kill-line` to `kill-visual-line`), but not thinks like paredit, which directly overrides `ctrl-k` instead of overriding `kill-line`. This is probably fine?

apps/emacs/emacs.talon
nriley: I’m fine with that. Thanks!

apps/emacs/emacs_commands.csv
nriley: Good point. I just meant within this csv but cross-knausj makes sense too

rntz: ok, I've consistently added a space after every comma I think.

rntz: except for double commas, `foo,, bar`; emptry entries don't get a space

apps/emacs/emacs.py
rntz: done.

rntz:
ok, I think this should be ready for merging.

rntz:
oh, wait, still don't have a `cancel`/`quit` command for `ctrl-g`/`keyboard-quit`

rntz:
ok, I've added a command `cancel` for keyboard-quit/ctrl-g. I don't have a strong opinion what this should be called; I use `cancel` for this but have it bound across multiple applications (it does `esc` by default, `ctrl-c` in terminal, and presses `ctrl-g` three times in emacs with sleeps in between), so I'm just gonna disable it. I would be concerned that `quit` is, like, super short in a chaining context? but it's probably fine I guess?

apps/emacs/emacs_commands.csv
nriley: OK, that works.

nriley:
@rntz I can't think I've ever had a misrecognition with "quit", but if others feel strongly I am happy to leave "cancel" as the default. You don't feel like there is a conflict with "cancel cancel"? I admit I never actually use that knausj feature...

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,343 @@
...
+merge last: user.emacs("smerge-prev")
+merge keep upper: user.emacs("smerge-keep-upper")
+merge keep lower: user.emacs("smerge-keep-lower")
+merge keep this: user.emacs("smerge-keep-current")
+merge refine: user.emacs("smerge-refine")
+merge split: user.emacs("smerge-resolve")
+
+# outline-minor-mode #
+# frequent: overview, show, hide, next, last, forward, backward, up
+outline minor mode: user.emacs("outline-minor-mode")
```
nriley, line 321: Duplicate.

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,343 @@
...
+project vc dir: user.emacs("project-vc-dir")
+project compile [project]: user.emacs("projectile-compile-project")
+project [run] shell command: user.emacs("projectile-run-shell-command-in-root")
+project [run] async shell command: user.emacs("projectile-run-async-shell-command-in-root")
+project (switch [to buffer] | buffer | buff): user.emacs("projectile-switch-to-buffer")
+project kill buffers: user.emacs("projectile-kill-buffers")
+project switch project: user.emacs("project-switch-project")
+
+# ----- VC/GIT SUPPORT ----- #
+vc (annotate | blame): user.emacs("vc-annotate") ;
```
nriley, line 296: Talon barfs on this spurious semicolon.

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,343 @@
...
+[toggle] truncate lines: user.emacs("toggle-truncate-lines")
+[toggle] word wrap: user.emacs("toggle-word-wrap")
+
+manual: user.emacs("man")
+manual <user.text>:
+    user.emacs("man")
+    user.insert_formatted(text, "DASH_SEPARATED")
+
+# BUFFER SWITCHING #
+# many things override these keybindings, so we don't use user.emacs
```
nriley, line 49: Should remove this comment since we do actually use user.emacs here now — and add switch-to-buffer and switch-to-buffer-other-window to emacs_commands.csv?

apps/emacs/emacs.talon
nriley: Yeah, that works for me, for consistency with the buffer commands too…

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,343 @@
...
+
+# ----- MOTION AND EDITING ----- #
+mark: user.emacs("set-mark-command")
+go back: user.emacs("pop-mark")
+global [go] back: user.emacs("pop-global-mark")
+
+auto indent: user.emacs("indent-region")
+indent <user.number_signed_small>: user.emacs("indent-rigidly", number_signed_small)
+
+(search regex | regex search): user.emacs_meta("ctrl-s")
```
nriley, line 197: Do we want to use user.emacs here too?

apps/emacs/emacs.talon
```diff
@@ -0,0 +1,343 @@
...
+merge keep upper: user.emacs("smerge-keep-upper")
+merge keep lower: user.emacs("smerge-keep-lower")
+merge keep this: user.emacs("smerge-keep-current")
+merge refine: user.emacs("smerge-refine")
+merge split: user.emacs("smerge-resolve")
+
+# outline-minor-mode #
+# frequent: overview, show, hide, next, last, forward, backward, up
+outline minor mode: user.emacs("outline-minor-mode")
+outline show all: key(ctrl-c @ ctrl-a)
```
nriley, line 322: More user.emacs?

nriley (commented):
Looking good!

apps/emacs/emacs.talon
rntz: done

apps/emacs/emacs.talon
rntz: done

apps/emacs/emacs.talon
rntz: done

apps/emacs/emacs.talon
rntz: fixed

apps/emacs/emacs.talon
rntz: fixed

rntz:
ok, hopefully this is good enough now, because I don't have much more energy to work on this!

apps/emacs/emacs.talon
```diff
@@ -161,23 +156,23 @@ describe variable <user.text>$:
...
 (file open | find file) at point: user.emacs("ffap")
-other file open: key(ctrl-x 4 ctrl-f)
-(file | buffer) close: key(ctrl-x k enter)
+other file open: user.emacs("find-file-other-window")
+(file | buffer) close:
+    user.emacs("kill-buffer")
+    key(enter)
 
-buffer kill: key(ctrl-x k)
+buffer kill: user.emacs("kill-buffer")
```
nriley, line 167: Should add to .csv.

apps/emacs/emacs.talon
```diff
@@ -161,23 +156,23 @@ describe variable <user.text>$:
...
 buffer revert | revert buffer: user.emacs("revert-buffer")
 buffer finish:
     edit.save()
     user.emacs("server-edit")
-buffer list: key(ctrl-x ctrl-b)
-buffer next: key(ctrl-x right)
-buffer last: key(ctrl-x left)
+buffer list: user.emacs("buffer-menu")
+buffer next: user.emacs("next-buffer")
+buffer last: user.emacs("previous-buffer")
```
nriley, line 175: Should add to .csv.

apps/emacs/emacs_commands.csv
```diff
@@ -55,15 +55,19 @@ global-display-line-numbers-mode,, g-d-l-n-m
...
+highlight-lines-matching-regexp, meta-s h l
+highlight-regexp, meta-s h r
 hl-line-mode,, hl-l-m
 indent-region, meta-ctrl-\
 indent-rigidly, ctrl-x tab
 insert-char, ctrl-x 8 enter
 insert-register, ctrl-x r i
 isearch-backward, ctrl-r
-isearch-backward-regexp, meta-ctrl-r
+isearch-backward-regexpr, meta-ctrl-r
```
nriley, line 66: Unintended extra "r" in here?

nriley (changes requested):
LGTM with these tiny changes. Thanks for the huge amount of work!

nriley (approved):

PR #937, opened 2022-07-23
Add new commands to Discord

alexander-clarke:
This adds some new commands to discord, and additionally fixes an existing command.

I have only implemented these commands currently on Windows, as that is the OS that I use.

This is my first contribution to this repository, so do tell me if I'm doing anything wrong.


apps/discord/win.py
```diff
@@ -82,3 +89,23 @@ def discord_answer_call():
...
+        actions.key("ctrl-k")
+        if dest_type:
+            if dest_type == "user":
+                actions.key("@")
+            if dest_type == "channel":
+                actions.key("#")
+            if dest_type == "voice":
+                actions.key("!")
+            if dest_type == "server":
+                actions.key("*")
```
auscompgeek, line 109: We could put this mapping in the list above instead.

alexander-clarke: yeah that's a good idea, it's in the list now

apps/discord/win.py
```diff
@@ -7,6 +7,13 @@
 app: discord
 """
 
+ctx.lists["user.discord_destination"] = {
```
auscompgeek, line 10: Should these be available for macOS too?

alexander-clarke: yeah that's true, for some reason I tried to do this before and it didn't work, but I got it to work this time

apps/discord/discord.talon
```diff
@@ -7,6 +7,10 @@ tag(): user.emoji
 [channel] mentions last: user.discord_mentions_last()
 [channel] mentions next: user.discord_mentions_next()
 oldest unread: user.discord_oldest_unread()
+{user.discord_destination} [<user.text>]: user.discord_quick_switcher(user.discord_destination or "", user.text or "")
```
auscompgeek, line 10:
```suggestion
{user.discord_destination} [<user.text>]: user.discord_quick_switcher(user.discord_destination, user.text or "")
```

alexander-clarke:
I just fixed the merge issues related to some of the bigger commits recently

splondike (approved):
Thanks!

PR #939, opened 2022-07-30
Use insert in talonscript instead of insert_with_history

timo:
This makes these commands transparent for the purpose of parsing
talonscript to figure out what it does


timo:
i decided not to remove the insert_with_history action, because perhaps someone somewhere is using it in their own scripts (nothing in search.talonvoice.com, at least) and we don't seem to have anything that does deprecations

rntz:
Why is add_phrase_to_history() followed by insert() more transparent than insert_with_history()?

auscompgeek:
Slack discussion from here onwards: https://talonvoice.slack.com/archives/C7ENXA7C4/p1659211727193269

splondike (approved):
Change sounds fine given the Slack context. Having the raw inserts improves the potential for automatically determining how to type a given phrase (according to aegis, sounds quite fancy to me).

rntz:
We should probably deprecate or remove insert_with_history if we want to go this route.

splondike:
Yeah, deprecating it would be a good idea. I dropped a proposal for how we might deprecate actions in here: https://github.com/knausj85/knausj_talon/issues/940 . Does that sound good for this case?

splondike:
I can make the adjustments needed here once we have #1027 merged.

splondike (approved):

PR #944, opened 2022-07-31
symbols: Use real words

auscompgeek:
This changes `dub string` and `dub quote` in symbols.talon to use real words to improve recognition.


rntz (approved):

davidsgrogan:
I patched this in to my branch. I _think_ it helped "dub quote" be recognized as such instead of as "dot quote". Thanks.

auscompgeek:
> it helped "dub quote" be recognized as such instead of as "dot quote".

Yep, that was my experience too.

splondike (approved):

splondike:
This just needs to be rebased, then we can merge it in.

PR #947, opened 2022-08-01
Add formatter for .talon files to pre-commit

wenkokke:
This PR adds `talonfmt`, as well a tabs-to-spaces conversion, to pre-commit.

The changes in the .talon should _absolutely_ be reviewed.

Note that `talonfmt` supports alignment of context headers and one-line commands, but I've not enabled those options.


rntz:
This appears to remove all empty lines from .talon files, which I'm strongly against - it should be possible to separate commands into groups by using empty lines.

It also appears to make all single-line commands into multi-line-with-indent, which I also dislike. I'd be okay with doing this for long lines (and/or collapsing commands into a single line when it would be short enough).

wenkokke:
> It also appears to make all single-line commands into multi-line-with-indent, which I also dislike. I'd be okay with doing this for long lines (and/or collapsing commands into a single line when it would be short enough).

That’s just the default.

wenkokke:
> This appears to remove all empty lines from .talon files, which I'm strongly against - it should be possible to separate commands into groups by using empty lines.

What do you feel should happen with blank lines?

pokey:
> What do you feel should happen with blank lines?

I would argue that any sequence of blank lines should be collapsed into a single blank line. And any whitespace on the blank line should be removed

pokey:
Super minor, but it would be nice if comments had a space after the `#`. Prob not worth fixing if that's going to take much time tho

apps/i3wm/i3wm.talon
```diff
@@ -2,93 +2,83 @@
...
+    user.system_command("i3-msg move container to workspace back_and_forth")
+(shuffle | move (win | window) left): user.system_command("i3-msg move left")
+(shuffle | move (win | window) right): user.system_command("i3-msg move right")
+(shuffle | move (win | window) up): user.system_command("i3-msg move up")
+(shuffle | move (win | window) down): user.system_command("i3-msg move down")
+(win | window) horizontal: user.system_command("i3-msg split h")
+(win | window) vertical: user.system_command("i3-msg split v")
 make scratch: user.system_command("i3-msg move scratchpad")
-[(show|hide)] scratch: user.system_command("i3-msg scratchpad show")
+[(show | hide)] scratch: user.system_command("i3-msg scratchpad show")
```
rntz, line 69: Would be good if talonfmt could get rid of the unnecessary parentheses here, `[show | hide] scratch: ...`

apps/generic_debugger.talon
```diff
@@ -1,73 +1,59 @@
...
-
 # Code execution
-
-## instruction level
+# instruction level
 step into: user.debugger_step_into()
 step over: user.debugger_step_over()
-
-## line level
+# line level
```
rntz, line 7: I don't like how two ##s are being replaced by one # here. The use of ## seems to have been deliberate. Comments are for expressing human intent, so I don't think auto formatters should change them.

rntz (commented):
I'd prefer if empty lines were simply left alone (with whitespace removed), but combining adjacent empty lines would also be ok.

I'm against changing comments by adding a leading space, though. Comments are for expressing human intent so I think that autoformatters should leave them alone.

The PR looks much better now that it tries to put things on a single line where possible. It is removing the nice horizontal alignment some .talon files have, though. I remember there was some discussion on the corresponding issue of the right way to do this to avoid large diffs - how does talonfmt handle alignment?

pokey:
> I'd prefer if empty lines were simply left alone (with whitespace removed), but combining adjacent empty lines would also be ok.

I think it's pretty common for autoformatters to collapse adjacent empty lines, unless the spec calls for them (eg before a function in Python)

One place I'd be tempted to remove empty lines is before / after the `-` context separator line.  I think the autoformatter should have an opinion there.

I would also be tempted to remove empty lines if they appear after the rule but before the body of a command

> I'm against changing comments by adding a leading space, though. Comments are for expressing human intent so I think that autoformatters should leave them alone.

I guess I don't feel too strongly here.  I'm pretty sure both other autoformatters that we use do this, though.  And I'm not sure I see what intent one would be expressing by not including a space between the `#` and the start of the comment.

But again, I don't feel strongly, and it's easier to leave them untouched

wenkokke:
> > What do you feel should happen with blank lines?
> 
> I would argue that any sequence of blank lines should be collapsed into a single blank line. And any whitespace on the blank line should be removed

What about "1 blank line is left alone, 2 or more blank lines become 2 blank lines"?

apps/i3wm/i3wm.talon
wenkokke: Should be possible, but I'll put that on the future work backburner, currently it preserves parentheses exactly.

apps/generic_debugger.talon
wenkokke: > I'm against changing comments by adding a leading space, though. Comments are for expressing human intent so I think that autoformatters should leave them alone.

Code formatting expresses human intent. The purpose of a formatter is to enforce some measure of consistency. However, I've made you a "--no-format-comments" flag, so we can at least not hardcode this behaviour.

wenkokke:
> I'm against changing comments by adding a leading space, though. Comments are for expressing human intent so I think that autoformatters should leave them alone.

Code formatting expresses human intent. The purpose of a formatter is to enforce some measure of consistency. However, I've made you a "--no-format-comments" flag, so we can at least not hardcode this behaviour.

wenkokke:
> I guess I don't feel too strongly here. I'm pretty sure both other autoformatters that we use do this, though. And I'm not sure I see what intent one would be expressing by not including a space between the `#` and the start of the comment.
> 
> But again, I don't feel strongly, and it's easier to leave them untouched

I think the best option would be to use, e.g., [comment_formatter](https://github.com/AlexPadron/comment_formatter) on blocks of comments. Outsource it, but ideally find something that respects indented blocks.

wenkokke:
Changes:

- set `--no-format-comments` to preserve comments as-is
- added logic to keep 1 or 2 blank lines, and condense 3 or more blank lines down to 2
- force push because the previous formatting changes were destructive

apps/generic_debugger.talon
wenkokke: > The PR looks much better now that it tries to put things on a single line where possible. It is removing the nice horizontal alignment some .talon files have, though. I remember there was some discussion on the corresponding issue of the right way to do this to avoid large diffs - how does talonfmt handle alignment?

There's automatic alignment of one-line commands with `--align-short-commands` and fixed column alignment of one-line commands with, e.g., `--align-short-commands-at=32`. I've not activated the former, due to the talk about big diffs, and not activated the latter because I think it's ugly.

The same commands exist for matches in the context header (though I don't think `--align-context-matches` actually works at the moment, since I never bothered too much with it, but it should be a single function call away).

phillco:
This looks great. I find the horizontal alignment inside of existing Talon files to be a bit distracting, so I'm fan of the way they currently look in the PR.

apps/generic_debugger.talon
wenkokke: Created #949 and #950 to showcase these variants, but I'd prefer to not use either.

wenkokke:
> This looks great. I find the horizontal alignment inside of existing Talon files to be a bit distracting, so I'm fan of the way they currently look in the PR.

@phillco do you have any ideas how to resolve the issues with pre-commit.ci? It currently doesn't have C++ build tools, which breaks `tree_sitter`.

phillco:
I'm not sure. If they don't support it out of the box I'm guessing we might have to move to Github Actions.

wenkokke:
> I'm not sure. If they don't support it out of the box I'm guessing we might have to move to Github Actions.

Works locally, it's just not working with the CI.

pokey:
Maybe you could have a branch of talonfmt where you check in the compiled code so that it doesn't need to compile?

wenkokke:
> Maybe you could have a branch of talonfmt where you check in the compiled code so that it doesn't need to compile?

Was considering that, yes.

pokey:
Hmm. Looks like OS-specific native code builds for Python are still pretty painful today

Could maybe use wasm with [wasmer](https://pypi.org/project/wasmer/)?

wenkokke:
> Hmm. Looks like OS-specific native code builds for Python are still pretty painful today
> 
> Could maybe use wasm with [wasmer](https://pypi.org/project/wasmer/)?

Kind of a no-go since tree_sitter relies on the C bindings. Probably best to move pre-commit with talonfmt to Github CI and skip it on pre-commit.ci.

wenkokke:
Set pre-commit to skip `talonfmt` on pre-commit.ci. Added pre-commit to GitHub Actions.

auscompgeek:
> It currently doesn't have C++ build tools, which breaks tree_sitter.

I took a look at [a pre-commit.ci run](https://results.pre-commit.ci/run/github/240185541/1659546282.UvCU47GrRkWKexMn5OZEPQ), the build tools run fine, but the package tries to write to the installation directory at runtime, which is a big no-no since the virtual environment is immutable.

If this will require a C++ compiler at runtime, this will be very annoying for people on Windows, as having Visual Studio installed is a lot more rare than having a C++ toolchain on UNIX systems.

wenkokke:
@auscompgeek Fixed on CI by being a little nicer about where to stash the library files. Added a `download_library` function for potential Windows users with issues, which I can link to a `talonfmt` option. Also publishing precompiled libraries on the GitHub releases.

apps/i3wm/i3wm.talon
```diff
@@ -2,35 +2,40 @@
...
 center window: user.system_command("i3-msg move position center")
-resize mode: user.system_command('i3-msg mode "resize"')
+resize mode: user.system_command("i3-msg mode \"resize\"")
 focus parent: user.system_command("i3-msg focus parent")
 focus child: user.system_command("i3-msg focus child")
 
 # resize helpers
 grow window:
-    user.system_command('i3-msg mode "resize"')
+    user.system_command("i3-msg mode \"resize\"")
```
pokey, line 38: Minor point, but some formatters will use single quote if the string contains a double quote but not single

wenkokke: Let’s not have that be blocking for the merge, but open an issue on the talonfmt repository. Shouldn’t be too hard.

.github/workflows/ci.yml
```diff
@@ -15,12 +15,16 @@ jobs:
...
-          python -m pip install --upgrade pip
-          pip install pytest
+          cache: "pip" # caching pip dependencies
+          cache-dependency-path: "requirements-dev.txt"
+      - run: pip install -r requirements-dev.txt
+      - name: Cache pre-commit
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
auscompgeek, line 28: Is this cache used?

wenkokke: It's what the pre-commit action does, so presumably pre-commit uses it.

wenkokke: Not anymore, since I removed pre-commit again. Good catch.

apps/i3wm/i3wm.talon
wenkokke: Implemented this in HEAD, will release soon.

wenkokke:
Anyone on a standard Windows machine without build tools wanna try and run this?

wenkokke:
Can anyone advise to what extent .dll, .dylib, and .so files are portable? I've currently pinned them to OS (but not OS version) and chipset.

wenkokke:
@rntz You can probably setup your git hooks locally to reformat .talon files before/after commit, so that you can have aligned files locally.

auscompgeek:
For Linux .so, you should be fine if you're building in a manylinux container and running auditwheel.

macOS .dylib can be a bit trickier, but if you set the macOS deployment target and you avoid dynamically linking to non-system libraries you should be okay.

I think cibuildwheel will automate all that for you.

wenkokke:
> For Linux .so, you should be fine if you're building in a manylinux container and running auditwheel.
> 
> macOS .dylib can be a bit trickier, but if you set the macOS deployment target and you avoid dynamically linking to non-system libraries you should be okay.
> 
> I think cibuildwheel will automate all that for you.

It's not a Python package with a native extension, unfortunately, so none of this seems to apply.

auscompgeek:
Sure, but can't you build it at package build time?

wenkokke:
> Sure, but can't you build it at package build time?

I can, but I don't know how to customize a wheel bdist action to build it and package it as an OS-specific version without using the Extension mechanism. See: [wenkokke/py-tree-sitter-talon/setup.py](https://github.com/wenkokke/py-tree-sitter-talon/blob/59ca47341a898dc67197d8ae5f84a11ed69c9c6a/setup.py)

wenkokke:
Fixed shared library problems.

Working around the way the Python `tree_sitter` package was set up a little, but I'm just shipping the grammar as a Python extension now and building with`cibuildwheel`.

apps/i3wm/i3wm.talon
wenkokke: Filed this as an issue for future work.

apps/i3wm/i3wm.talon
wenkokke: Fixed, should be working now if you look at the diffs again.

wenkokke:
If desired, I can rerun the formatter from the original state.

text/numbers.talon
```diff
@@ -1,3 +1,3 @@
-not tag: user.mouse_grid_showing
+tag: user.mouse_grid_showing
```
auscompgeek, line 1: This change seems quite incorrect 🤔

pokey: Oh wow nice catch 😅

wenkokke: Woops, lemme go and fix that. Thanks for catching that.

wenkokke: Oh, yes, I see when I broke that.

wenkokke: Fixed in 1.4.4.

wenkokke:
@auscompgeek Fixed the issue with disappearing "not"s, reran the formatter from the initial files, and migrated the changes to CI/dependabot to separate PRs.

lang/css/css.talon
```diff
@@ -23,12 +23,11 @@ rule <user.text>:
     name = user.formatted_text(text, "DASH_SEPARATED")
     insert("{name}: ")
 
-value <user.number_string> [{user.css_unit}]:
-    "{number_string}{css_unit or ''}"
+value <user.number_string> [{user.css_unit}]: "{number_string}{css_unit or ''}"
```
auscompgeek, line 26: I'm surprised this is being formatted as a single line, when [the very short `go usage` in `apps/visualstudio/visual_studio.talon` is being line wrapped](https://github.com/wenkokke/knausj_talon/commit/97913a8aa8219e02fb8f507a8986402c08e97316#diff-2f575b306ef5c73a031cfd7552f6c8637bcf25d8948f05cf36631f11a3802f45L75-R76). What's the logic here?

wenkokke: The former fits onto an 80 column line, the latter is probably a bug relating to single-line commands found by themselves. I'll file the latter as a bug.

wenkokke: Confusingly, my tests already check that `go usage` is formatted on a single line.

wenkokke:
@auscompgeek Your comments exposed a bug in the column count in `doc_printer`, which I've subsequently fixed.

text/numbers.talon
pokey: Assuming there's now a test for this case? Would probably be good to avoid this kind of regression going forward 😅

rntz:
In principle it should be possible to property-test this: check that
parsing the result of talonfmt produces the same AST as parsing its input.
In fact, could have talonfmt do this always and error if they're different.
I think black does something like this.

wenkokke:
> In principle it should be possible to property-test this: check that parsing the result of talonfmt produces the same AST as parsing its input. In fact, could have talonfmt do this always and error if they're different. I think black does something like this.
> […](#)

I already test that formatting is idempotent, which is closely related, and perhaps more relevant, since there may be reorderings, e.g., future versions of talonfmt sorting tag includes and moving tag includes and settings blocks to the top of the file.

wenkokke:
> In principle it should be possible to property-test this: check that parsing the result of talonfmt produces the same AST as parsing its input. In fact, could have talonfmt do this always and error if they're different. I think black does something like this.
> […](#)

However, we could define a less structured comparison on the AST, which we could use to implement this.

text/numbers.talon
pokey: Ah I see https://github.com/wenkokke/talonfmt/blob/dev/tests/data/golden/simple/default/knausj_talon_text_numbers.yml; looks like that should do it 👍

apps/web/github.talon
```diff
@@ -2,7 +2,6 @@
 tag: browser
 browser.host: github.com
 -
-
```
pokey, line 5: @wenkokke in grooming session we aligned on actually keeping one empty line after the `-` line, and no empty lines before.  Assuming you have no objections, I'm guessing that's an easy fix?  Sorry to go back and forth on this one 😅

misc/extensions.talon
```diff
@@ -1 +1,2 @@
+-
```
pokey, line 1: Adding the `-` when there is no context header was a bit of a sticking point in our grooming session

pokey: I'm guessing that removing the `-` when there is no context header will result in a smaller diff, so maybe worth doing that and then filing a follow-up issue to discuss always having the `-`, in the interest of getting this PR merged sooner

wenkokke: Cursorless support will be better if it’s mandatory, since with it optional the grammar is ambiguous.

apps/web/github.talon
wenkokke: Easy change, can do. Probably best to check back if the other thing (mandatory “-“) is kinda important?

wenkokke:
Added blank lines after the match context, as requested.

misc/extensions.talon
rntz: Yes, we do want to remove auto-adding the `-`.

rntz: (Tangent, but: I'm reasonably sure it's not actually ambiguous, just hard to express in an LR formalism. Can you give an example of something that has multiple parses = is ambiguous?)

wenkokke: It’s not ambiguous, but it’s not parseable but a non-backtracking parser.

wenkokke: Do you all want to keep the `-` if it's already there, even if it's not needed?

rntz:
I have no strong opinion on that case.

wenkokke:
Latest version:

- supports key-binding declarations;
- checks that formatting is idempotent and preserves the AST *at runtime* (the latter was suggested by @rntz); 
- includes fixes to the parser (suggested by @pokey and @rntz) which:
  + resolve the ambiguity issues around match contexts, using arbitrary lookahead in the lexer; and
  + simplify the parsing of matches, no longer attempting to parse match contexts into logical formulas.

.pre-commit-config.yaml
```diff
@@ -40,3 +40,15 @@ repos:
     hooks:
       - id: black
         language_version: python3.9
+  - repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.3.0
+    hooks:
+      - id: remove-tabs
+        types: [file]
+        files: \.talon$
+        args: ["--whitespaces-count=4"]
```
auscompgeek, line 49: Huh, it does seem like most of our .talon files use 4 space indents. We should fix the .editorconfig to do the same.

wenkokke:
Can we merge this? I don’t think there’s still any blocking issues.

.pre-commit-config.yaml
auscompgeek: Opened #970 for this.

rntz:
We should probably decide at this point whether we're standardizing on 2 or 4 spaces for indentation in talon files. Per #970 we mostly use 4 spaces but the editorconfig file is set up to use 2 spaces. I prefer 2 spaces but I reckon I'm probably outvoted here :P. Still:

rntz:
upvote for 2 spaces

rntz:
upvote for 4 spaces

phillco:
upvote for tabs

wenkokke:
Okay, it's been open to a vote for 9 days. Can we merge this now?

phillco (approved):

phillco:
I'd naturally wait for @rntz or @pokey, but this looks good to me.

.pre-commit-config.yaml
pokey: Would be cool if remove-tabs respected .editorconfig.  I believe prettier does

wenkokke:
🎉🎉🎉

phillco:
Awesome! Thanks a bunch for this change, wen.

Now to merge up my fork :D

PR #953, opened 2022-08-07
Setup dependabot

wenkokke:
Setup dependabot to automatically open PRs bumping the versions of GitHub actions used in the CI.

For instance, the `actions/setup-python` we use is out of date, which we'll be notified of once this is merged.

PR #954, opened 2022-08-07
Freeze and cache pip dependencies on CI

wenkokke:
Creates `requirements-dev.txt` which contains the dependencies needed for running the test suite, and sets up the CI to cache these dependencies.

Once this and #953 are both merged, I'll also setup dependabot to help maintain those dependencies.


requirements-dev.txt
```diff
@@ -0,0 +1 @@
+pytest==7.1.2
```
auscompgeek, line 1: Pinning the pytest version SGTM. What's the plan to keep this up-to-date? Are we going to set up dependabot to bump this for us?

.github/workflows/ci.yml
```diff
@@ -12,15 +12,13 @@ permissions:
...
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v3
+      - uses: actions/setup-python@v3
         with:
           python-version: "3.9"
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install pytest
+          cache: "pip" # caching pip dependencies
```
auscompgeek, line 20: nit: the comment isn't conveying any more useful information
```suggestion
          cache: "pip"
```

.github/workflows/ci.yml
```diff
@@ -12,15 +12,13 @@ permissions:
 jobs:
   test:
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v3
+      - uses: actions/setup-python@v3
```
auscompgeek, line 17: Thoughts on keeping/removing the `name`? I left it in so that it's obvious from reading the CI logs what version of Python is being used, especially if it gets bumped later, but maybe it won't be useful?

requirements-dev.txt
wenkokke: Yes.

.github/workflows/ci.yml
wenkokke: Figured it'd be a bother to maintain that info in multiple places. Could maybe use `env`/`matrix`, though?

auscompgeek: Hmm, yeah, probably a good idea to test with 3.10 before Talon starts using 3.10 as well.

auscompgeek (approved):
I'm surprised you're able to request my review despite me not being a maintainer 🤔

LGTM with my comment about the Python version addressed.

wenkokke:
Oh, weird, I don't remember doing that.

phillco (approved):

splondike:
Thanks!

PR #961, opened 2022-08-21
Bugfixes for c, rust

chrisnicollo:
Apologies, I accidentally closed the first pull request for this.

When these insert library functions were called in their respective talon files, the first argument was the name of the library, and the second argument was an empty string. As a result, it was pasting an empty string ("selection") instead of the library name ("text").

Also, one of the capture rules for c didn't match the name used for it in the associated talon file, causing the voice commands to not work.

### DIFF
--- a/lang/c/c.py
+++ b/lang/c/c.py
@@ -391,4 +391,4 @@ class UserActions:
         actions.user.code_insert_function(result, None)
 
     def code_insert_library(text: str, selection: str):
-        actions.user.paste(f"include <{selection}>")
+        actions.user.paste(f"include <{text}>")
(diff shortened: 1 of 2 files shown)


lang/c/c.py
```diff
@@ -188,7 +188,7 @@ def c_cast(m) -> str:
 
 
 @mod.capture(rule="[<self.stdint_signed>] <self.stdint_types> [<self.c_pointers>+]")
-def c_stdint_cast(m) -> str:
+def stdint_cast(m) -> str:
```
auscompgeek, line 191: I wonder whether we should change the `.talon` file to match this instead.

auscompgeek (commented):
The `code_insert_library` fixes LGTM.

lang/c/c.py
chrisnicollo: My reasoning for changing the ```.py``` file instead was that this happened to be the only place where the 'c_' prefix was used in front of a 'stdint' name. Changing the ```.talon``` file to match this one difference could potentially result in stylistic inconsistency.

splondike (approved):

splondike:
Thanks!

PR #963, opened 2022-08-28
Add support for relative home paths

SandeepJoel:
The change will allow us to have the entries like below in the settings/system_path.csv. 
`~/Downloads/screenshots, screenshots`

This can be useful if we're using this same file for multiple machines having different usernames using dotFiles approach.

### DIFF
--- a/apps/generic_terminal/generic_terminal.talon
+++ b/apps/generic_terminal/generic_terminal.talon
@@ -8,7 +8,7 @@ lisa all:
     user.terminal_list_all_directories()
 katie [<user.text>]: user.terminal_change_directory(text or "")
 katie root: user.terminal_change_directory_root()
-go <user.system_path>: insert("cd \"{system_path}\"\n")
+go <user.system_path>: insert("cd {system_path} \n")
 clear screen: user.terminal_clear_screen()
 run last: user.terminal_run_last()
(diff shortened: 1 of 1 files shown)


apps/generic_terminal/generic_terminal.talon
```diff
@@ -8,7 +8,7 @@ lisa all:
     user.terminal_list_all_directories()
 katie [<user.text>]: user.terminal_change_directory(text or "")
 katie root: user.terminal_change_directory_root()
-go <user.system_path>: insert("cd \"{system_path}\"\n")
+go <user.system_path>: insert("cd {system_path} \n")
```
rntz, line 11: This is a bad idea, as it means any path that contains spaces (or other characters with special meaning to the shell) will not work without escaping. Instead, I suggest making a capture `<user.system_path>` which calls `os.expanduser` on the path.

rntz (changes requested):

auscompgeek:
Workaround: Use environment variables such as `$HOME` or `%USERPROFILE%` in the CSV.

rntz:
this would partially address #1036

splondike:
It sounds like we'd need some changes to the current implementation to want to merge this in. Since the PR hasn't moved in a few months, wouldn't fully address #1036, and is so simple I'm just going to close this out.

PR #964, opened 2022-08-31
Abstract browser copy address to its own function with fallback

david-tejada:
Creates action user.browser_copy_address() and user.browser_address_fallback().

My reason for doing this is so I can reimplement browser_copy_address in Rango so it uses the extension to get the address.

I also changed the way the fallback works so it doesn't leave the focus on the address bar.

### DIFF
--- a/code/browser.py
+++ b/code/browser.py
@@ -18,6 +18,28 @@ def is_url(url):
         return False
 
 
+@ctx.action_class("user")
+class UserActions:
+    def browser_address_fallback() -> str:
+        try:
+            address = actions.browser.address()
+            if address:
+                return address
+        except Exception:
+            pass
+        actions.browser.focus_address()
+        actions.sleep("50ms")
+        with clip.capture() as s:
+            actions.edit.copy()
+        actions.key("escape")
+        actions.key("tab")
+        return s.text()
+
+    def browser_copy_address():
+        address = actions.user.browser_address_fallback()
+        clip.set_text(address)
+
+
 @ctx.action_class("browser")
 class BrowserActions:
     def address():
--- a/apps/generic_browser.talon
+++ b/apps/generic_browser.talon
@@ -1,10 +1,7 @@
 tag: browser
 -
 address bar | go address | go url: browser.focus_address()
-address copy | url copy | copy address | copy url:
-    browser.focus_address()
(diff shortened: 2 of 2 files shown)


code/browser.py
```diff
@@ -18,6 +18,28 @@ def is_url(url):
...
+            if address:
+                return address
+        except Exception:
+            pass
+        actions.browser.focus_address()
+        actions.sleep("50ms")
+        with clip.capture() as s:
+            actions.edit.copy()
+        actions.key("escape")
+        actions.key("tab")
```
rntz, line 35: I'm not entirely convinced by this approach. The only current use for this function is to be called by browser_copy_address, and this use of clip.capture/clip.set_text introduces an unnecessary layer of indirection compared to just calling edit.copy. If the clipboard apis are even a bit flaky, as I've observed them to be on Linux, then this is a disimprovement.

rntz, line 35: What do you expect this to do? Are you sure it will always put the focus where you want it even across browsers and platforms?

david-tejada: This implementation was suggested to me by @lunixbochs in the context of reimplementing browser.address with Rango. This can't be done at the moment because it's not safe (we have to press keys) so instead he suggested to implement browser_address_fallback, and then I can reimplement it in Rango to return the address. I was misremembering about having to reimplement browser_copy_address.  I'm not sure about the use of clip.capture and clip.set_text, I think it has to do with plugins that implement clipboard history.

david-tejada: You are right, this is a bad idea. It doesn't even return the focus to the document, at least in Firefox, which is what I would need.

splondike: To make sure I understand, the preferred thing would be to implement browser.address so it could be used in the context matchers as well as for this voice command. But because Rango needs to press some keys to get the address (and the browser.address action could be called frequently by Talon), you might get concurrency issues.

If the current voice command is reliable enough, I would suggest we don't make this change since it's not that likely to be used elsewhere. At a later point when we have the RPC thing working or cross-platform accessibility APIs we would just want to implement browser.address using those.

david-tejada: Agreed. I will close this PR and come back to the issue when we have rpc

PR #965, opened 2022-09-01
Fix formatting on readme

moowiz:
It seems the existing readme doesn't format its bullet points correctly. This should fix it so they display nicely.


rntz (approved):

auscompgeek (approved):
I just noticed this as well, was about to submit the same thing :smile:

PR #968, opened 2022-09-03
java: Add `state var`

auscompgeek:
This adds a convenient way to insert the Java 10+ keyword for local variable type inference, as in:
```java
var url = new URL("http://example.com/");
```

### DIFF
--- a/lang/java/java.talon
+++ b/lang/java/java.talon
@@ -23,6 +23,8 @@ settings():
     user.code_protected_variable_formatter = "PRIVATE_CAMEL_CASE"
     user.code_public_variable_formatter = "PRIVATE_CAMEL_CASE"
 
+state var: "var "
+
 # Types Commands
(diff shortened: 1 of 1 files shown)


rntz (approved):

rntz:
ISTR there's a talon list used for language keywords, we should probably add it there as well - @pokey you remember this?

pokey:
> ISTR there's a talon list used for language keywords, we should probably add it there as well - @pokey you remember this?

Sure, that makes sense.  I don't think Java has this functionality yet, but trivial to add.  See Python:

- https://github.com/knausj85/knausj_talon/blob/0d6bd90e5bb1a45b5cea7bcd7ccbb45973cd9b02/lang/python/python.talon#L12
- https://github.com/knausj85/knausj_talon/blob/0d6bd90e5bb1a45b5cea7bcd7ccbb45973cd9b02/lang/python/python.py#L62-L74

PR #969, opened 2022-09-04
replace dictation_peek_{left,right} with dictation_peek

rntz:
In some circumstances, it can be more efficient to merge `dictation_peek_left` and `dictation_peek_right`. It shaves 1 keystroke off of the default implementations, and shaves a round trip off of rpc-based implementations (eg. via command client). This PR merges them. Needs testing. It might break some existing configs, unfortunately. I'm not sure what the best way to handle that is.


nriley:
Do we have a list of applications or other special cases where dictation peeking has failed in the past?

rntz:
To be clear, this should largely preserve behavior for people who haven't made custom overrides for `dictation_peek_left/right` - that's the breakage I was talking about. The diff looks large, but it's basically just a refactoring. The only thing it changes is omitting a redundant `key(space backspace)`.

rntz:
I guess another possible breakage is that this cements the choice to do all peeking _before_ insertion. I recall some folks saying things worked better for them when we peeked left before insertion but peeked right afterwards. I _think_ that was probably because of the issue about copying with no text selected? But I can't be sure.

rntz:
It seems like this PR has been blocked for a while waiting for testing. Are we waiting for something specific, or just "better tests for this sort of stuff"? If there's something specific I can address I'd like to know, otherwise maybe we should merge this and treat better testing as a separate issue?

Gonna remove the "help wanted" tag so we can discuss this at next maintenance meeting.

splondike (approved):
Looks good to me :+1:

PR #970, opened 2022-09-04
editorconfig: Default to 4 spaces

auscompgeek:
As evidenced in #947 the majority of our .talon files use 4-space indents. This changes the default to 4 spaces, overriding it for Markdown and YAML files (which should always be 2 spaces).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,10 +5,14 @@ root = true
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
-indent_size = 2
+indent_size = 4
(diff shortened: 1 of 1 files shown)


rntz:
While we're doing a reformatting of all talon files, that's probably the right time to choose an indent depth, so I'm going to discuss this question in #947 if that's all right

phillco (approved):
Approving based on the current poll results in https://github.com/knausj85/knausj_talon/pull/947, but might wait for consensus to arise on that one

PR #973, opened 2022-09-05
Proof of concept: add logic to automatically format Rust generic types

kazimuth:
Examples:

```
'type vector' -> 'Vec<>'
'type vector eye thirty two' -> 'Vec<i32>'
'type hash map borrowed mute eye thirty two string' -> 'HashMap<&mut i32, String>'
'type borrowed mute hash map eye thirty two string' -> '&mut HashMap<i32, String>'
'type tuple eye thirty two character string' -> '(i32, char, String)'
'type hash map tuple boolean boolean finish string' -> 'HashMap<(boolean, boolean), String)>'
```

I'm new to Talon, so I'd appreciate advice for how to allow users to extend the list of types + their generic arg counts in an idiomatic way, without just editing this file. (A bigger project would be pulling in-scope types from VSCode & Jetbrains and automatically populating these lists, but I have no idea how possible that is.)


lang/rust/rust.py
```diff
@@ -225,11 +234,141 @@ def code_comment_documentation_block_inner():
...
+            yield end
+
+    return "".join(parse(iter(m)))
+
+
+def code_type_(*args):
+    return code_type([generic_type(arg) for arg in args])
+
+
+assert code_type_(["Vec"]) == "Vec<>"
```
auscompgeek, line 348: Ideally these asserts would go in a pytest `test_foo.py` file, so that failures here don't break a language support entirely.

auscompgeek (commented):
Oh wow this is pretty impressive.

This'd be useful for other languages too, so it'd be great if we could extract the `parse` function; it could be extracted into a separate module.

lang/rust/rust.py
kazimuth: 👍

kazimuth:
@auscompgeek How would you recommend exposing the list of types + arities for extension? Just a capture, like generic_type here? 

Also is there a good way to export helper functions throughout talon extensions lol

auscompgeek:
I don't think I was necessarily expecting adding a capture or anything, just a plain old Python module with that function.

nriley:
From the community backlog session, @FireChickenProductivity offered to discuss this with a rust-fluent friend and potentially take this on.

getynge:
I think the PR as written would be useful. In day-to-day rust a hefty chunk of generics are either user defined or defined by libraries, so it may be helpful to expand support for generics to arbitrary types in a manner akin to the tuple list support in this PR. It would also be helpful to add support for the turbofish operation as used in `HashMap::<String, String>::new()`.

PR #975, opened 2022-09-15
blame-ignore-revs: add talonfmt changes

phillco:
This omits the changes from https://github.com/knausj85/knausj_talon/pull/947 (446ec764c9caa98973eacd7f792b6a087a1b635f) from blame for people who have the configuration set, so the reformatting changes don't affect blame.


pokey:
Ah whoops; I probably should have split that PR into two commits before merging so that the configuration itself doesn't get ignored from blame. I guess it's too late now; don't want to rewrite history in public

pokey (approved):

pokey:
@rntz any objection to adding to git blame even tho the squashed commit contains the config change as well?  My bad for squashing without separating them

rntz:
sgtm

phillco:
@pokey it actually becomes much harder to do that now that we have pre-commit in CI, because as soon as you add the config change it will insist on adding the code changes as well, and blocked squash otherwise :D

pokey:
I think we could have done it backwards right? Ie code changes first then config changes?

pokey:
But yeah generally speaking I see our point

phillco:
Oh, big brain moment. Yes we could have.

PR #977, opened 2022-09-17
vscode: Implement jumping to tab 10+

auscompgeek:
This allows jumping to tabs 10 onwards with `go tab ten` and so on.


nriley (commented):
Looks good.

nriley (approved):

PR #980, opened 2022-09-20
Make settings directory configurable

wenkokke:
The current settings directory is hard-coded to be inside repository root of knausj_talon, which means it's automatically gitignored.

The PR adds the setting `community_settings_directory`, which is used to locate the settings directory anywhere, using the `get_full_path` function adapted from Cursorless.

The default setting remains the same, so this should not be a breaking change.

### DIFF
--- a/code/user_settings.py
+++ b/code/user_settings.py
@@ -2,22 +2,41 @@ import csv
 import os
 from pathlib import Path
 
-from talon import resource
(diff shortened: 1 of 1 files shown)


wenkokke:
Closed, because this requires lifting all CSV loads into `on_ready` callbacks. Will reopen if I get to that.

PR #981, opened 2022-09-20
Make settings directory configurable

wenkokke:
The current settings directory is hard-coded to be inside repository root of knausj_talon, which means it's automatically gitignored.

The PR adds the setting `community_settings_directory`, which is used to locate the settings directory anywhere, using the `get_full_path` function adapted from Cursorless. (The default setting remains the same, so this should not be a breaking change.)

As part of this PR, I've moved all uses of `get_list_from_csv` into a "ready" callback, and have them write to their respective global dictionaries on "ready".

### DIFF
--- a/code/extensions.py
+++ b/code/extensions.py
@@ -1,4 +1,4 @@
-from talon import Context, Module
+from talon import Context, Module, app
 
 from .user_settings import get_list_from_csv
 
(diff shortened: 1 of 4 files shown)


rntz:
Was this closed deliberately?

wenkokke:
> Was this closed deliberately?

Yes. Making the settings directory configurable requires knausj to take the concurrent nature of Talon scripts seriously, which is a bit more than I think I can get merged.

PR #982, opened 2022-09-20
git: use bag-of-arguments approach

rntz:
This completely overhauls git.talon to use a bag-of-arguments approach to parsing git commands, unifying them into `git {command} {argument}*`. 

**Advantages:**

- Easier to add new commands & arguments: just edit two csv files, `git_commands.csv` for git commands (eg `rm,remove`) and `git_arguments.csv` for arguments (eg `--amend, amend` or `-b, branch`).

- More flexible: we don't have to special case every combination of arguments we might want. Eg. `git diff color words cached` -> `git diff --color-words --cached`.

- More predictable: what you say is what gets typed. To type `git branch --remote` you say `git branch remote`. Fewer things to memorize this way. 

**Changes to existing behavior:** Lots.

- Many commands get "renamed" by the policy of following the order of the arguments. For instance, `git remote branches` has to become `git branch remote`; `git log all` becomes `git log enter`.

- Some commands used to allow arbitrary text afterward, like `git checkout` and `branch` and `push` etc. This has been lost. 

- No longer automatically presses enter for some commands. We were already pretty inconsistent about this, though.

**Other disadvantages:**

- Less obvious what commands are possible just by reading a .talon file.

**Questions:**

1. Should we use settings csvs for this or hardcoded csvs? I've gone with the latter for now, since git commands/arguments ought to be the same for everyone.

2. Should we allow `<user.text>` after all/some git commands? This would support existing workflows like `git push <text>`. But it might cause false positives. I think I'll probably want to experiment with this. I've currently special-cased this for `git commit -m <message>` and `git stash push -m <message>`.

3. Do we want to make some commands auto-execute (have `enter` pressed afterward)? I've done this for `git status` and `git add --patch` because they're so common, using end-anchors:

   ```
   git status$: "git status\n"
   git add patch$: "git add --patch\n"
   ```

   I've used this for a while and really like it, but it might take some getting used to.


code/git.py
```diff
@@ -0,0 +1,43 @@
...
+        if len(row) == 0:
+            continue
+        if len(row) == 1:
+            row = row[0], row[0]
+        if len(row) > 2:
+            print("{path!r}: More than two values in row: {row}. Ignoring the extras.")
+        output, spoken_form = row[:2]
+        spoken_form = spoken_form.strip()
+        mapping[spoken_form] = output
+    return mapping
```
rntz, line 33: This is a rip off of `get_list_from_csv` but without assuming that our file comes from `settings/`. Probably we should refactor `get_list_from_csv` to allow more code reuse here.

phillco:
@rntz just FYI, I merged this up with main so that I could test it. Let me know if I made a mistake

phillco (approved):

Issue #216, opened 2020-09-04
google docs, sheets, slides


dwiel:
public beta implementations are here:

- https://github.com/dwiel/talon_community/blob/master/apps/web/google_docs.py
- https://github.com/dwiel/talon_community/blob/master/apps/web/google_slides.py
- https://github.com/dwiel/talon_community/blob/master/apps/web/google_sheets.py

NOTE: this set of commands is compatible with browser plug-ins like streamingkeys

fabform:
Handy to [find google sheet id](https://fabform.io/a/how-to-find-google-sheet-id)

PR #985, opened 2022-09-25
Add support for Google Docs

crashmoney:
This PR adds support for Google Docs, including accessibility shortcuts. It follows the object-verb standard as much as possible, but also provides alternatives.

Partially resolves issue #216.


apps/web/google_docs.talon
```diff
@@ -0,0 +1,339 @@
...
+    key(cmd-ctrl-5)
+(go | move) [to] ((previous | last) heading six | heading six (previous | last)):
+    key(cmd-ctrl-p)
+    key(cmd-ctrl-6)
+(go | move) [to] (next media | media next):
+    key(cmd-ctrl-p)
+    key(cmd-ctrl-g)
+open explore tool: key(cmd-alt-shift-i)
+open dictionary: key(cmd-shift-y)
+start voice typing: key(cmd-shift-s)
```
trapier, line 327: Should this also put talon to sleep?

crashmoney: It should, to avoid conflicts. I've applied the change – Talon now goes to sleep when voice typing is activated.

pokey (changes requested):
Thanks for submitting this one!  Lots of useful stuff here.  However, a bunch of these commands already have spoken forms defined elsewhere in knausj, and will actually already work out of the box (eg "copy that", "paste that", etc).  It would be better to remove them from this file, and if you prefer a different spoken form for these commands, you may want to remap them in your own fork of knausj.  That way they'll be consistent across all applications

pokey (changes requested):
Better, but still duplicates a bunch of globally available commands, especially at the beginning of this file

Try saying "help search blah" where "blah" is anything. Eg "help search save". Can also try "help active" to see what commands are available at any moment.

crashmoney:
I've removed most of the commands that are already present in knausj 👍🏻 I kept find-and-replace, though, since `user_find.and.replace` doesn't seem to work in my testing

apps/web/google_docs.talon
```diff
@@ -0,0 +1,326 @@
...
+tag(): user.line_commands
+
+file menu | menu file: key(ctrl-alt-f)
+open: key(cmd-o)
+[open] (revision | version) history [open]: key(cmd-alt-shift-h)
+print: key(cmd-p)
+save: key(cmd-s)
+edit menu | menu edit: key(ctrl-alt-e)
+paste without formatting: key(cmd-shift-v)
+# adding keyboard shortcuts here since user.find_and_replace doesn't work
```
auscompgeek, line 13: Please implement the `find_and_replace` tag instead of redefining these commands. You can see the actions that need to be implemented in [`tags/find_and_replace/find_and_replace.py`](https://github.com/knausj85/knausj_talon/blob/main/tags/find_and_replace/find_and_replace.py).

apps/web/google_docs.talon
```diff
@@ -0,0 +1,326 @@
+browser.host: docs.google.com
+-
+
+tag(): user.line_commands
+
+file menu | menu file: key(ctrl-alt-f)
+open: key(cmd-o)
+[open] (revision | version) history [open]: key(cmd-alt-shift-h)
+print: key(cmd-p)
+save: key(cmd-s)
```
auscompgeek, line 10: We already have `file save`:

https://github.com/knausj85/knausj_talon/blob/4d39608dc9979017d7a0218b195feb54ff136ffa/core/edit/standard.talon#L15

apps/web/google_docs.talon
```diff
@@ -0,0 +1,326 @@
...
+edit menu | menu edit: key(ctrl-alt-e)
+paste without formatting: key(cmd-shift-v)
```
auscompgeek, line 12: I believe this is called `paste match` already:

https://github.com/knausj85/knausj_talon/blob/4d39608dc9979017d7a0218b195feb54ff136ffa/core/edit/standard.talon#L14

You may need to override the `edit.paste_match_style` action, like

https://github.com/knausj85/knausj_talon/blob/4d39608dc9979017d7a0218b195feb54ff136ffa/core/edit/edit_mac.py#L142-L143

auscompgeek (changes requested):
The majority of these will only work on macOS (i.e. will not work on Windows or Linux). This would be okay, but this .talon file needs to match against `os: mac` in that case.

splondike:
It sounds like there are still a reasonable number of changes that would be needed to get this ready to be merged. Are you still interested in pushing it forward @crashmoney ?

crashmoney:
Hi @splondike, sorry for the lack of inactivity. I've been a little busy lately.

I'm interested to push this PR forward in order to merge it. Right now the main blocker is that I'm not as familiar with Python as I am with YAML, but I've been keeping up with the suggestions made in this PR thread. For now, I'm doing a few more tests locally on what does and doesn't work, and once I get the changes working I'll push them up.

With regards to what @auscompgeek said about the shortcuts only working on MacOS – I don't have Windows or Linux systems to test on, so I'll only be able to contribute to MacOS reliably for now. Thanks for the suggestions!

crashmoney:
I just pushed some changes resolving most of the duplicated commands in this PR – namely, `find_and_replace` and `paste_match_style` 👍🏻  Is there anything else that needs to be done w.r.t to either duplication or organisation of the talon file?

Thanks to all the reviewers!

apps/web/google_docs.talon
```diff
@@ -0,0 +1,320 @@
+os: mac
+browser.host: docs.google.com
+-
+
+tag(): user.line_commands
+tag(): user.find_and_replace
+tag(): browser
```
auscompgeek, line 7: Only browser applications should activate the browser tag.
```suggestion
```

apps/web/google_docs.py
```diff
@@ -0,0 +1,35 @@
+from talon import Context, Module, actions
+
+ctx = Context()
```
auscompgeek, line 3: Looks like this context doesn't match anything. This file shouldn't be overriding default implementations of actions outside of Google Docs.

apps/web/google_docs.py
```diff
@@ -0,0 +1,35 @@
...
+mod = Module()
+
+apps = mod.apps
+apps.gdocs = """
```
auscompgeek, line 7: Is this app matcher being used?

auscompgeek (changes requested):

apps/web/google_docs.talon
```diff
@@ -0,0 +1,320 @@
+os: mac
```
auscompgeek, line 1: Since this is Mac-specific, the file should be named accordingly too.

apps/web/google_docs.py
```diff
@@ -0,0 +1,35 @@
...
+tag: browser
+browser.host: docs.google.com
+"""
+
+
+@ctx.action_class("user")
+class FindAndReplace:
+    """Override generic find-and-replace"""
+
+    def find(text: str):
```
splondike, line 17: I don't think you need to implement the find, find_next, find_previous, or paste_match_style actions. Those are already implemented for mac with the same shortcuts in `core/edit/edit_mac.py`.

apps/web/google_docs.py
splondike: Probably just want to use a matcher like this:
```
os: mac
browser.host: docs.google.com
```
You can then probably get rid of the app declaration as well for now and just duplicate the browser.host matcher here and in the .talon. As auscompgeek mentioned you'll need to pin them both to be mac only anyway, so adding an app may not be worth it. I don't mind if you do want to define the app though.

splondike:
@crashmoney It's been a couple of months, are you still interested in getting this change merged in?

splondike:
I'm going to close this out for now. Easy enough to recover the code if the author would like to pick it up again.

PR #988, opened 2022-09-28
Update chrome tab duplicate to include esc press

Spiteless:
Add esc press to reset address bar if it has been edited before duplicating

### DIFF
--- a/apps/chrome/chrome.py
+++ b/apps/chrome/chrome.py
@@ -60,12 +60,20 @@ class user_actions:
         actions.app.tab_close()
 
     def tab_duplicate():
-        """Limitation: this will not work if the text in your address bar has been manually edited.
-        Long-term we want a better shortcut from browsers.
-        """
         actions.browser.focus_address()
         actions.sleep("180ms")
-        actions.key("alt-enter")
+        possibly_edited_url = actions.edit.selected_text()
+        actions.key("esc")
+        actions.browser.focus_address()
+        actions.sleep("180ms")
+        url_address = actions.edit.selected_text()
+        if possibly_edited_url == url_address:
+            actions.key("alt-enter")
+        else:
+            actions.user.paste(possibly_edited_url)
+            actions.app.tab_open()
+            actions.user.paste(url_address)
+            actions.key("enter")
 
 
 @ctx.action_class("browser")


nriley:
This seems to work OK when you have modified the URL but at least on the Mac, if you focus the address bar and have not edited the URL, pressing escape will remove focus from the address bar.

Spiteless:
Makes sense, I'm a Windows user and didn't know that

Spiteless:
If this is windows only is it worth adding?

pokey:
We actually tested on Windows, and observed the same issue

Spiteless:
No, you're correct, I managed to overlook that in my own testing. I tried it with a second browser_focus and it works for me with both an edited case and an unedited case. 

    def tab_duplicate():
        actions.browser.focus_address()
        actions.sleep("180ms")
        actions.key("esc")
        actions.browser.focus_address()
        actions.key("alt-enter")

splondike (approved):
The updated version seems to work fine for me on Linux and Mac. The cost is you lose whatever edits you had going on in the address bar, but I don't think that's worth worrying about.

pokey:
> The cost is you lose whatever edits you had going on in the address bar

Maybe we could use a `actions.edit.selected_text()` right after we focus address bar the first time to get the current value, then paste it after cloning the tab?

pokey:
> The cost is you lose whatever edits you had going on in the address bar

Maybe we could use a `actions.edit.selected_text()` right after we focus address bar the first time to get the current value, then paste it after cloning the tab?

Spiteless:
I'll check on getting this working. 

1. Navigate to URL
2. Copy what's there to variable via actions.edit.selected_text()
3. Press esc
4. Navigate to URL
5. Copy what should now be the new url
    * If nav bar wasn't edited it will be the URL anyway
7. Paste in the saved variable text
9. Create a new tab
10. Paste in the url to duplicate

pokey:
How come you can't just do

```py
def tab_duplicate():
    actions.browser.focus_address()
    actions.sleep("180ms")

    // Capture URL text in case they've edited it
    possibly_edited_url = actions.edit.selected_text()

    // Temporarily discard edits so duplicated tab is current page
    actions.key("esc")

    // Refocus address bar, because if there was no edit, <escape> unfocuses
    actions.browser.focus_address()
    actions.sleep("180ms")

    // Do the actual duplication
    actions.key("alt-enter")

    // Restore possibly edited URL
    actions.user.paste(possibly_edited_url)
```

Spiteless:
Could do that too. We'd have to add a last focus_address() to get back to the nav, and the possibly_edited_url would now be on the copied tab instead of the original tab.

Otherwise, we'd have to navigate back to the old tab, paste in possibly_edited_url, and then go back to copied tab.

```py
def tab_duplicate():
    actions.browser.focus_address()
    actions.sleep("180ms")

    # Capture URL text in case they've edited it
    possibly_edited_url = actions.edit.selected_text()

    # Temporarily discard edits so duplicated tab is current page
    actions.key("esc")

    # Refocus address bar, because if there was no edit, <escape> unfocuses
    actions.browser.focus_address()
    actions.sleep("180ms")

    # Do the actual duplication
    actions.key("alt-enter")

    # Restore possibly edited URL
    actions.browser.focus_address() # refocus address bar
    actions.user.paste(possibly_edited_url)
```

splondike (changes requested):
I think pokey's version would also leave you with a tab without any page loaded on it (and your edited text in the address bar) whereas Spiteless' would have both tabs with something loaded. I think the second is better.

Spiteless' approach looks good to me, so if they fix up the PR to only affect the files we want then I think we can get this tested and merged.

Spiteless:
Kk, I'll fix up the PR in a few days and let you know

Spiteless:
@splondike updated

splondike (approved):

splondike:
Works well on Linux and OSX, thanks!

PR #990, opened 2022-10-02
talonfmt: raise maximum line length to 88

phillco:
Opening for discussion.

I didn't notice the line length parameter until after #947 merged; we didn't discuss it much in that PR.

After I started using `talonfmt` in my own repositories, I've personally found a higher maximum length to be more ergonomic, as it means more commands can fit on a single line, which in my experience makes them easier to edit (although obviously having an autoformatter helps a lot). I've [used 100 in axkit](https://github.com/phillco/talon-axkit/pull/36) and in my own private repository and personally quite like the results.

I thought I would open this PR to kick off the discussion. If 100 is too high for people, perhaps we could find a middle ground? For context, [Black uses 88](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#:~:text=Line%20length,-You%20probably%20noticed&text=Black%20defaults%20to%2088%20characters,seems%20like%20the%20wise%20choice.), so if we set 88 we will be consistent across both languages.

cc @wenkokke


wenkokke:
@phillco You could probably set up pre-commit to format with 100 on checkout and with 80 on commit?

phillco:
That would probably create a lot of spurious local changes, which would be in hopeful. I think it should probably be consistent. My intent here is to propose that we simply set a higher value globally.

rntz:
I prefer shorter lines because I like to put things side-by-side on a laptop monitor. Even on a 15" monitor with a small-to-my-eyes font, 100 is slightly too long to put two windows side-by-side (with line numbers). And with a more normal-sized font, even 90 chars is too wide, with 88 chars just scraping in. So if we want to increase it, I vote for 88. I'm perfectly happy with 80, though.

Example:

image

rntz:
If black uses 88 by default and we're already using black, then 88 has the advantage of being consistent across Python & .talon files as well :).

phillco:
I'd settle for 88 :)

pokey:
We should prob update editorconfig while we're here https://github.com/knausj85/knausj_talon/blob/4d39608dc9979017d7a0218b195feb54ff136ffa/.editorconfig#L10

phillco:
Done. I also removed the unnecessary `indent_style = space, indent_size = 4` under `.py` because that was specified by the default, let me know if it's still needed.

.editorconfig
```diff
@@ -11,12 +11,12 @@ max_line_length = 80
...
-indent_style = space
 indent_size = 2
 
 [*.py]
-indent_style = space
-indent_size = 4
+max_line_length = 88
+
+[*.talon]
 max_line_length = 88
```
rntz, line 20: Maybe we should just set `max_line_length` to 88 globally, under `[*]`?

rntz (approved):

.editorconfig
phillco: Should we keep 80 for `[*.{md,yaml,yml}]`, or just use 88 everywhere for simplicity?

rntz: I'd say everywhere for simplicity.

PR #993, opened 2022-10-07
Optimize `number_small` capture

pokey:
This optimisation reduces DFA from >1100ms to <500ms when combined with Cursorless main branch

All credit due to @AndreasArvidsson, I made only tiny tweaks to resolve conflicts with knausj

### DIFF
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -5,12 +5,12 @@ from talon import Context, Module
 mod = Module()
 ctx = Context()
 
-digits = "zero one two three four five six seven eight nine".split()
+digit_list = "zero one two three four five six seven eight nine".split()
 teens = "ten eleven twelve thirteen fourteen fifteen sixteen seventeen eighteen nineteen".split()
 tens = "twenty thirty forty fifty sixty seventy eighty ninety".split()
 scales = "hundred thousand million billion trillion quadrillion quintillion sextillion septillion octillion nonillion decillion".split()
 
-digits_map = {n: i for i, n in enumerate(digits)}
+digits_map = {n: i for i, n in enumerate(digit_list)}
 digits_map["oh"] = 0
 teens_map = {n: i + 10 for i, n in enumerate(teens)}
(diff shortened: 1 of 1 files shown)


core/numbers/numbers.py
```diff
@@ -5,12 +5,12 @@
 mod = Module()
 ctx = Context()
 
-digits = "zero one two three four five six seven eight nine".split()
+digit_list = "zero one two three four five six seven eight nine".split()
```
pokey, line 8: This variable name conflicts with the `digits` capture function name, so my linter was complaining; figured it was worth fixing

rntz (approved):
Looks good, I haven't tested it though.

pokey:
Works fine for me; I don't think it's going to vary by platform or anything so probably good to just merge?

core/numbers/numbers.py
AndreasArvidsson: Must say I'm not a big fan of captures with plural names that only return a single value. I call my alternative just `digit`

core/numbers/numbers.py
```diff
@@ -5,12 +5,12 @@
...
 teens = "ten eleven twelve thirteen fourteen fifteen sixteen seventeen eighteen nineteen".split()
 tens = "twenty thirty forty fifty sixty seventy eighty ninety".split()
 scales = "hundred thousand million billion trillion quadrillion quintillion sextillion septillion octillion nonillion decillion".split()
 
-digits_map = {n: i for i, n in enumerate(digits)}
+digits_map = {n: i for i, n in enumerate(digit_list)}
 digits_map["oh"] = 0
```
rntz, line 14: Shit, I didn't realize it, but this is a problem. Your new version uses `digits_list` rather than `digits_map` and therefore doesn't include any small numbers that use "oh" for zero! So merging this PR broke backwards compat, my bad.

The data flow through globals in this file is getting kinda hard to follow, we should probably try to confine it to just one place.

PR #999, opened 2022-10-10
Fix recognition issues with some extensions

pokey:
The versions with spaces are not friendly to the beam search, and could cause problems in future Talon versions


rntz:
Is there a general rule to follow here? I suspect we have a lot of commands and lists affected by something similar. Was this based on feedback from aegis?

rntz (approved):

pokey:
Nvm this was a misunderstanding

PR #1009, opened 2022-10-31
Implement delete line action for iterm

gsingh93:
This PR implements `delete_line` for iterm. Although, I wonder if this should be in generic_terminal.py? I think most terminals have the same keybindings.

### DIFF
--- a/apps/iterm/iterm.talon
+++ b/apps/iterm/iterm.talon
@@ -8,3 +8,4 @@ tag(): user.generic_unix_shell
 tag(): user.git
 tag(): user.kubectl
 tag(): user.tabs
+tag(): user.readline
--- a/core/tags.py
+++ b/core/tags.py
@@ -10,6 +10,7 @@ tagList = [
     "tabs",
     "generic_windows_shell",
     "generic_unix_shell",
+    "readline",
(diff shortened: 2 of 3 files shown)


splondike:
That shortcut is actually implemented by [readline](https://en.wikipedia.org/wiki/GNU_Readline) I think. So ctrl-u could be used by any program using that library, for example bash, zsh, Python REPL, Ruby REPL, fzf.

Given that I might recommend the following instead of this PR:

1. Making `/tags/terminal/readline.py` file that defines a tag and implements this action (and potentially others). You'd then add the readline tag to the iterm.talon file.
2. Change the implementation of this action; ctrl-u deletes backwards from the current position in readline. You'd probably implement it as ctrl-e (go to end of line), ctrl-u (delete backwards from cursor).

splondike:
I see you force pushed, but FYI no code got updated.

gsingh93:
With the latest push, I've created the `readline` tag, implemented just the `delete_line` action, and added the tag only to iterm.

I can move more actions into the tag and add it to more applications, but with limited time and limited typing ability, I'd like to do that in follow up PRs instead of this one if that's ok.

splondike (approved):
Thanks, looks good :+1:

Issue #942, opened 2022-07-31
Cleaner commands with a single ctrl/cmd-key action


auscompgeek:
During this week's grooming session it was brought up that the majority of the Python actions are only defined to allow for macOS support with the same `.talon` file, and most cross-platform apps only differ in their keyboard shortcuts by usage of <kbd>cmd</kbd> or <kbd>ctrl</kbd>.

We could probably delete a large chunk of code, and easily migrate a bunch of the platform-specific `.talon` files, by having a single action to press a key with <kbd>cmd</kbd> or <kbd>ctrl</kbd>.

Funnily enough, we actually have one already, added along with Thunderbird support in #892:

https://github.com/knausj85/knausj_talon/blob/bfb282178d78880fad6b69b76f7dd097208f1f65/apps/thunderbird/thunderbird.py#L84-L85

https://github.com/knausj85/knausj_talon/blob/bfb282178d78880fad6b69b76f7dd097208f1f65/apps/thunderbird/mac.py#L30-L31

The parameter name `keys` seems to be a misnomer and should probably be renamed to singular `key` to avoid confusion.

auscompgeek:
Fair enough. To me `keys` implies it will modify all keys that are passed in, which is not the case. Perhaps for a more generic action that might actually be something we want? :shrug:

auscompgeek:
Looks like we have a similar action for Chrome now too (added in #1014), but it's actually wrong - see #446.

knausj85:
It's preferable not to do this to support VMs and such.

See https://github.com/talonhub/community/issues/446

Closing.

PR #1014, opened 2022-10-31
Add profile switch command for Chrome

gsingh93:


apps/chrome/chrome.py
```diff
@@ -30,6 +30,16 @@
 """
 
 
+@mod.action_class
+class Actions:
+    def chrome_mod(key: str):
```
gsingh93, line 35: @rntz I saw you had a concern about this approach here: https://github.com/knausj85/knausj_talon/pull/892#pullrequestreview-1013594471.

It would be good to have a general way to do this, as described by https://github.com/knausj85/knausj_talon/issues/942. If you have any suggestions for this, I'm happy to implement it, otherwise I can just leave it specific to this code for now.

splondike:
The mod thing does feel like a bit of a weird indirection, but I see the rationale. I'm going to merge this in as chrome specific because I think we should let the idea play out a bit to see if it's useful enough to define a general pattern as per #942

splondike (approved):

splondike:
Thanks for the contribution!

PR #1020, opened 2022-11-05
Allow passing more args to ls

gsingh93:


tags/terminal/unix_shell.py
```diff
@@ -6,21 +6,23 @@
...
+ctx.lists["user.ls_args"] = {
+    'list': '-l',
+    'all': '-a',
+}
+
+@mod.action_class
+class JoinAction:
+    def join_strings(l: list, s: str):
+        """Joins the strings using the given separator"""
+        return s.join(l)
```
gsingh93, line 21: This probably shouldn't go in this file but I don't know where to put it right now.

splondike:
What additions did you have in mind for 'ls"? As implemented this will break functionality in some of the window's shells. See this implementation for example: https://github.com/knausj85/knausj_talon/blob/main/apps/windows_command_processor/command_processor_win.py#L86

gsingh93:
> What additions did you have in mind for 'ls'? I presume you wanted to do more than change the current implementation while providing no new features.

The new feature is now we can do `lisa list all`, i.e. combine multiple arguments together :)

Will take a look at the Windows issues soon.

splondike:
That makes sense about the extra option. Though if you're just planning to add that one, it'd probably be easier to see if you just added `lisa list all: "ls -la"`; less abstraction is easier to understand (for example, you had to explain to me what this added!).

If you'd like to add -r -h etc. then I could see this being of benefit though. I'll let you think about how to do this while keeping other shells functional.

gsingh93:
With the latest commit this should hopefully work on Windows, but I won't be able to test it.

And yea, I do plan to add more options later, like `-t`, `-s`, `-1`, etc. And with this approach others will also be able to easily add their own options without breaking anything.

splondike:
I took a look at [the talon search](https://search.talonvoice.com/search/?q=terminal_list_directories&fold_case=auto&regex=false&context=true) and it appears likely the usage of terminal_list_*directories in knausj are the only ones in the community.

Given that I think this ticket should proceed as follows:

* It looks like this will break apps/windows_command_processor/command_processor_win.py . Could you take a look at that also? I eyeballed all the others usages in knausj and it seems likely they should continue working with the new implementation.
* We deprecate the terminal_list_directories and terminal_list_all_directories actions using the system proposed in #1027 and delete the contextual implementations of those actions in knausj_talon. This will need to wait for that PR to merge.
* I'll stick on the help wanted tag since I don't have a Windows machine either. I'd like people to verify the 'lisa' and 'lisa all' commands work for: WSL, Powershell, and cmd.exe at least. maybe mintty.exe as well if they can.
* I also caused a merge conflict with this system by merging a different PR an hour or so ago :D.

So this PR is likely to take a few weeks to get merged given the above.

gsingh93:
Sure I'll look into those files, and we can wait to merge until that deprecation PR is merged and someone's been able to test on Windows.

splondike (changes requested):
The deprecation PR has been merged, so the list of TODOs in my last comment can now be looked at.

DonnerWolfBach:
I fixed the changes for windows and tested them for:
- windows terminal
- powershell
- cmd

**However** I feel like it is lacking right now (now offense, great idea!), there are two features which I would really like to see and will try to implement in my pull request #1108:
1. allow formatters, so "lisa slasher pictures old" would result in "ls pictures/old". I need this especially with the snake formatter, since many of my directories are in lower snake case and not simple words. (We might also need a backslash formatter for cmd, which cannot do the / to \ change on the fly, but I think that's a different issue)
2. allow to pipe system_paths into ls - I can see how this would be debatable though, since "core lisa path home" would have a similar result. But I think it would be more natural. Not sure wether "lisa [<user.text>]" and "lisa [<user.system_path>]" would bite each other.

DonnerWolfBach:
> The deprecation PR has been merged, so the list of TODOs in my last comment can now be looked at. See also my comment in #1051.

So we should now first put the deprecation warning in and then wait four months?

DonnerWolfBach:
@splondike Ah yeah I confused #1051 and #1020 with my pull request - #1108 is indeed for #1051. 
I agree that these two issues are conflicting. And while something like `lisa <options> path <path>` like `git commit [<user.git_arguments>] message [<user.prose>]:` could be the solution, I am actually not sure whether I would want that. Likely not.

Anyway, with #1107 we will kinda enable this with the off tour via core (e.g. you can say "core lisa help" then.

DonnerWolfBach:
I took another look at this and came to the conclusion that I don't have a need for this (#1051 is another issue) and won't work on it anymore. 

As stated before, #1107 will allow something similar with just adding "core" upfront, thus I would suggest closing this issue.

splondike:
Since the original author hasn't written back in a few months, and we don't have the interface worked out (e.g. it conflicts with #1051 by the same author), I'm going to close this out.

PR #1022, opened 2022-11-08
git ignore .DS_Store

deszoeke:
Macs litter local directories with .DS_Store. Git should ignore them.


splondike (changes requested):
.gitignore change looks good, you just included a couple of extra changes by accident. Happy to merge once you've taken them out :).

PR #1024, opened 2022-11-10
git ignore .DS_Store as #1022 but without extra commits

deszoeke:
Macs litter local directories with .DS_Store. Git should ignore them.


splondike (approved):
Thanks!

Issue #940, opened 2022-07-30
We need a mechanism for deprecating actions


timo:
I would like to deprecate the action `insert_with_history`, but there's not yet something for that.

We should probably log a warning and add "deprecated" to the description. There should probably also be a spot to put what the code should do instead to reduce friction for users.

splondike:
Here's my idea. We have a two stage deprecation process for actions. First they display a popup notification about deprecation once per Talon session but otherwise still function. Second (4 months later say) they just display the notification and don't function. 6 months after that we can completely remove them.

In addition to the above we have a file that lists deprecated actions along with the date and some instructions about how to migrate off them. The deprecation notification would say "user.insert_with_history is deprecated please see DEPRECATED_ACTIONS.md" for the first case and "...is no longer supported..." for the second.

For the second non-functional deprecation we could move the action over to a 'deprecated_actions.py' file so it stopped cluttering up the original 'functional' file.

gsingh93:
Another reason why having this deprecation code would be useful is that right now some commands are structured like verb-object while the recommendation is to structure them like object-verb. If we were to change all of these commands to match the recommendation, then that would lead to a lot of deprecated commands.

splondike:
I've pushed up a proposal for the deprecation process and system in #1027 . It's a little simpler than the one I put in my above comment. The built in notification system in the PR is also only partial in that it doesn't handle the case of lists, or of people overriding captures or actions.

For all of these harder cases I think it would be ideal to have 1st party support. I.e. we get the ability to mark those things as deprecated somehow with a Talon API. We could also do something with abstract syntax tree processing, but that sounds a bit invasive.

If the PR looks good and gets merged, maybe we should close out this issue by just raising a feature request for 1st party deprecation of these constructs on the Talon issue tracker.

splondike:
PR got merged in (see discussion on #1027 for what we eventually implemented). Closing out this issue.

PR #1025, opened 2022-11-11
Update emote to add colons and support formatters

theryangeary:
It seems to me that the `emote` slack command should format the user.text as an emoji in slack, using colons. This adds wrapping colons to the existing emote command, and adds an additional emote command supporting formatters, which leaves off the trailing comma to leave the emoji selection window open.

(if this is not what the emote command is supposed to do, please disregard)

### DIFF
--- a/apps/slack/slack_mac.talon
+++ b/apps/slack/slack_mac.talon
@@ -45,7 +45,10 @@ bold: key(cmd-b)
 (slack | lack) invite: key(a)
 # Miscellaneous
 (slack | lack) shortcuts: key(cmd-/)
-emote <user.text>: "{text}"
+emote <user.text>: ":{text}:"
+emote <user.formatters> <user.text>:
+    insert(':')
+    insert(user.formatted_text(text, formatters))
 toggle left sidebar: key(cmd-shift-d)
 toggle right sidebar: key(cmd-.)
 
--- a/apps/slack/slack_mac.talon
+++ b/apps/slack/slack_mac.talon
@@ -47,7 +47,7 @@ bold: key(cmd-b)
 (slack | lack) shortcuts: key(cmd-/)
 emote <user.text>: ":{text}:"
 emote <user.formatters> <user.text>:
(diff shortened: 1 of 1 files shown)


theryangeary:
(also, I realize insert(':') isn't top notch code, I'll fix if there's any interest in merging this)

splondike:
There are a couple of other things you might be interested in looking in to:

1. The [emoji system](https://github.com/knausj85/knausj_talon/blob/main/tags/emoji/emoji.talon) lets you say something like 'emoji shrugging' and it will insert the appropriate icon. I tested this in Slack and it actually converts the pasted thing into its native format. You could figure out how to enable this tag for Slack.
2. It might be good to have a command to bring up the emoji search pane. I wasn't able to find a shortcut for the browser based slack, but perhaps the desktop client has one. If you do find one, that would probably be a useful addition (e.g. by adding an emoji search command).

Happy to hear your thoughts about these ideas vs what you have here :slightly_smiling_face: .

theryangeary:
I did not know about the emoji system, that's super handy! the tag to enable emoji is already [applied to slack](https://github.com/theryangeary/talon/blob/main/apps/slack/slack_mac.talon#L5). 

Unfortunately, I am not able to find any way to open the emoji searcher in slack other than typing : and any two characters.

Now that I know about the emoji system, this doesn't seem like a necessary change. Perhaps this command could be removed since it really inserts the text, but maybe it's better to leave it for backwards compatibility. What do you think?

splondike:
That's a shame about the searcher, but I'm glad you found the emoji tag useful :).

I'd be in favour of removing the command, it indeed doesn't seem that useful. I'm actually working on a deprecation system for this repo at the moment (#940) . I reckon we could close out this PR if you're happy, and if you or I remember in a few weeks when that's merged we could deprecate that command using the new system. How does that sound?

theryangeary:
Sounds perfect. Putting a reminder on my calendar for one month from now.

PR #928, opened 2022-07-20
Make i3wm internally consistent with itself

EdJoJob:
- Stop vim-modeline attempted interpretation
- Remove use of literal `ten` when there is a `<number_small>` capture
- Simplify readability of multiple window close commands in i3wm
- i3wm window movements separated
- i3wm match window flipper to port flipper

PR #1027, opened 2022-11-12
feat(#940): Implement action/command/capture deprecation process

splondike:
Implementation of a variant of the action deprecation system from #940 that also covers captures and voice commands.

This system should let us take care of the breaking changes proposed in #928 and #939 smoothly for example.

### DIFF
--- a/core/deprecations.py
+++ b/core/deprecations.py
@@ -58,7 +58,7 @@ from talon import Module, actions, speech_system
 REPO_DIR = os.path.dirname(os.path.dirname(__file__))
 
 mod = Module()
-setting_deprecation_warning_interval_hours = mod.setting(
+setting_deprecate_warning_interval_hours = mod.setting(
     "deprecate_warning_interval_hours",
     type=float,
     desc="""How long, in hours, to wait before notifying the user again of a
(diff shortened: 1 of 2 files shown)


rntz:
Summary of discussion:

1. We want every deprecated voice command to come with a replacement. We also decided to throttle these per-phrase rather than every 2 hours; since there's a replacement available, the idea is that folks should immediately switch to the replacement. Throttling could also be unhelpful: if someone sees a notification but it disappears before they read it, then says the command again, they may find it frustrating if they can't get the notification to reappear.

2. We're not sure about the deprecation period of 4 months; many people don't update knausj for longer periods than that, so we probably want more like 6 months to a year.

core/deprecations.py
```diff
@@ -0,0 +1,171 @@
...
+just put the information in BREAKING_CHANGES.txt so people can look there to
+see what happened.
+
+Usages:
+
+    # myfile.talon - demonstrate voice command deprecation
+    ...
+    old legacy command:
+        user.deprecate_command("2022-11-10", "old legacy command")
+        # perform command
```
rntz, line 27: This example needs to be updated to include a replacement command.

splondike:
I made a couple more refactoring changes and adjusted the setting to have the same prefix as the actions.

I'm happy for this to be merged in.

phillco:
@pokey @rntz does this look good to merge?

rntz (approved):

PR #1028, opened 2022-11-12
Add help command to show the options for any list

gsingh93:
Adds a 'help list {user.help_lists}' command to display the available options for any list. For example, with this PR you can say 'help list window snap positions' and see the following window:

image

This is currently more of a PoC, as I'm not sure if I'm doing things correctly.

### DIFF
--- a/core/help/help.py
+++ b/core/help/help.py
@@ -9,7 +9,7 @@ from talon import Context, Module, actions, imgui, registry
 
 mod = Module()
 mod.list("help_contexts", desc="list of available contexts")
-mod.list("help_lists", desc="list of lists") # TODO: Better desc
+mod.list("help_lists", desc="list of lists")  # TODO: Better desc
 mod.mode("help", "mode for commands that are available only when help is visible")
 setting_help_max_contexts_per_page = mod.setting(
     "help_max_contexts_per_page",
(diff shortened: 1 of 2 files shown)


core/help/help.py
```diff
@@ -24,6 +25,23 @@
...
+def update_lists():
+    mapping = {}
+    for name in registry.lists.keys():
+        list_name = name.split('.')[-1]
+        display_name = list_name.replace("_", " ")
+
+        short_names = actions.user.create_spoken_forms(
+            display_name,
+            generate_subsequences=False,
+        )
```
gsingh93, line 39: I copied this from somewhere else in the file, I'm not sure exactly what it's doing, as the output is usually the same as the input, other than abbreviations.

In any case, I'm not sure exactly what's being done (if anything) to convert context names to name that can be pronounced for `user.help_contexts`, but ideally we want something like that here.

core/help/help.py
```diff
@@ -24,6 +25,23 @@
...
+        for short_name in short_names:
+            mapping[short_name] = list_name
+
+    ctx.lists["user.help_lists"] = mapping
+
+update_lists()
```
gsingh93, line 46: I'm not sure if doing this globally is the right way to do it, as I don't see that being done for `user.help_contexts`.

splondike:
There are already a [few debugging commands](https://github.com/knausj85/knausj_talon/blob/main/plugin/talon_helpers/talon_helpers.talon#L37) like this that dump out various things to the log. You need to set the `user.talon_populate_lists` but can then say things like 'talon debug list user formatters'.

I only remembered those while looking at your PR though, discoverability in knausj is a problem! Do you think there's an advantage to the approach in this PR (i.e. using a GUI) over that?

splondike:
It's been a couple of months, and since there's a command which does something like this already I'm going to close this out.

PR #1031, opened 2022-11-13
Add command for adding padding around symbols

gsingh93:
The current pad command leaves the cursor in the middle of two spaces. I think it would be much more convenient if we could say something like "pad equals" and have " = " with the cursor already at the end of the inserted text, making it easier to continue typing.

### DIFF
--- a/core/edit/standard.talon
+++ b/core/edit/standard.talon
@@ -16,7 +16,7 @@ file save: edit.save()
 file save all: edit.save_all()
 (pad | padding): user.insert_between(" ", " ")
 (pad | padding) <user.symbol_key>+:
-    insert(" {symbol_key_list} ")
+    insert(" ")
     user.insert_many(symbol_key_list)
     insert(" ")
 slap: edit.line_insert_down()
--- a/core/edit/standard.talon
+++ b/core/edit/standard.talon
@@ -15,4 +15,8 @@ paste match: edit.paste_match_style()
 file save: edit.save()
 file save all: edit.save_all()
 (pad | padding): user.insert_between(" ", " ")
+(pad | padding) <user.symbol_key>+:
+    insert(" {symbol_key_list} ")
+    user.insert_many(symbol_key_list)
+    insert(" ")
 slap: edit.line_insert_down()


gsingh93:
Ping @splondike and @rntz.

Do you think this type of change is too subjective to merge into the community repo? This has been really helpful for me, but I'm happy to just leave it in my own config if it doesn't make sense to change the behavior of this command for everyone.

rntz:
I like the idea behind this but it changes existing behavior in two ways:

1. It breaks the existing command "padding" when not followed by a symbol key, eg "pad say hello" or just "pad" on its own are now not valid commands. This needs to be fixed.

2. It changes the behavior of "pad" followed by symbol keys. For example, "pad plus plus" used to produce ` ++| ` where `|` is the cursor position; now it produces ` ++ |`. We usually try to avoid behavior changes, but I do personally think this is an improvement; maybe it's ok?

gsingh93:
> It changes the behavior of "pad" followed by symbol keys. For example, "pad plus plus" used to produce ++| where | is the cursor position; now it produces ++ |. We usually try to avoid behavior changes, but I do personally think this is an improvement; maybe it's ok?

I think this will be a tradeoff between the most common use cases and some particular edge cases. For most operators, I think the common usage is the binary operator usage, where there are spaces on both sides and the cursor follows the final space. I could try to scope this change down to just binary operators, but that will introduce a lot of complexity.

In the `pad plus plus` case, maybe I'm misunderstanding your use case, but I can't see myself using this as a way of adding a pre/post-increment operator. If I want to write `i++` I would say `sit plus plus` followed by whatever character I wanted next (semicolon, space, etc.). I'm not sure when the original functionality of `pad plus plus` would be useful?

As a side note, `++` is actually a binary operator in Haskell, which is a case where the change in this PR makes sense anyway.

splondike:
How does this new command fit with the existing 'op *' commands? For example if I have 'blah|blah' (| is cursor pos) in a properly configured editor then 'pad plus' and 'op plus' both produce the same result ('blah + |blah').

In Haskell where you want the blah ++ blah operator and there's no generic thing to map it on to you could just add an 'op concat' command to the Haskell specific .talon file.

gsingh93:
Yes, there's some redundancy here, but I think there's some significant benefits with `pad`:

1. It works with all symbols, not just ones used as operators
2. It works even if the program language isn't activated. I can use it to send simple equations in messages without needing to enable any language modes
3. It works even if the language isn't supported in knausj. I need to write code in some languages that aren't supported right now (i.e. haskell, prolog, etc.), and while adding support for them would be great, having `pad` allows me to get by more easily until I can add that support.

splondike (approved):
Cool, that makes sense. The behaviour change identified by rntz sounds fine to me also; probably it would be preferred once gotten used to.

Happy for this to be merged in.

PR #1033, opened 2022-11-14
implement splits and tabs for tmux

theryangeary:
tmux panes and windows are pretty analogous to talon splits and tabs. this implements the appropriate actions to support split and tab commands for tmux. users can add the tmux tag to their terminal/whatever to enable.

### DIFF
--- a/apps/tmux/tmux.py
+++ b/apps/tmux/tmux.py
@@ -1,0 +1,120 @@
+from talon import Context, Module, actions
+
+ctx = Context()
+mod = Module()
+
+ctx.matches = r"""
+tag: user.tmux
+"""
+
+setting_tmux_prefix_key = mod.setting(
+    "tmux_prefix_key",
+    type=str,
+    default="b",
+    desc="The key used to prefix all tmux commands",
+)
+
+
+@mod.action_class
+class TmuxActions:
+    def do_tmux_prefix():
+        """press control and the configured tmux prefix key"""
+        actions.key(f"ctrl-{setting_tmux_prefix_key.get()}")
+
+    def do_tmux_keybind(key: str):
+        """press tmux prefix followed by a key bind"""
+        actions.user.do_tmux_prefix()
+        actions.key(key)
+
+    def enter_tmux_command_mode():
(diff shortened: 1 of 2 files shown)


apps/tmux/tmux.talon
```diff
@@ -0,0 +1,9 @@
+tag: user.tmux
+-
+tag(): user.splits
+
+# pane management - these commands use the word split to match with the splits
+# tag defined in tags/splits/splits.talon
+split move <user.arrow_key>: user.do_tmux_keybind(arrow_key)
```
theryangeary, line 7: this could also be `go split <user.arrow_keys>`, which might be more consistent. wdyt?

apps/tmux/tmux.py
```diff
@@ -0,0 +1,120 @@
...
+mod = Module()
+
+ctx.matches = r"""
+tag: user.tmux
+"""
+
+setting_tmux_prefix_key = mod.setting(
+    "tmux_prefix_key",
+    type=str,
+    default="b",
```
gsingh93, line 13: It would be neat if we could parse the user's default prefix key from their config automatically (but doesn't need to be done in this PR).

apps/tmux/tmux.talon
```diff
@@ -0,0 +1,9 @@
...
++#Say a number after this command to switch to pane
```
gsingh93, line 8: Seems to be an extra + on these two lines?

theryangeary: thanks for pointing that out 😅 fixed

apps/tmux/tmux.py
theryangeary: I guess I can work on that if anyone is willing to approve this pr 😅

gsingh93:
I tested this and it works for me. But I think if we're going to merge this we should remove https://github.com/knausj85/knausj_talon/blob/main/apps/tmux/tmux_linux.talon (and make sure we haven't lost any functionality).

Also, I wonder if it's possible to automatically enable the tmux tag if tmux is running?

rntz (approved):

PR #1034, opened 2022-11-17
Add more commands to readline tag

gsingh93:
In order to implement this, I had to move some of the cut and copy actions into their own functions (this is related to https://github.com/knausj85/knausj_talon/issues/1015).

### DIFF
--- a/core/edit/edit.py
+++ b/core/edit/edit.py
@@ -39,3 +39,28 @@ class Actions:
         """Moves right by n words."""
         for _ in range(n):
             actions.edit.word_right()
+
+    def cut_word_left():
+        """Cuts the word to the left."""
+        actions.edit.extend_word_left()
+        actions.edit.cut()
+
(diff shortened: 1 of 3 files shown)


rntz (approved):

rntz:
looks good to me, however, on testing it looks like the readline context is not enabled by default in the apple terminal - we should probably fix that (in a future PR).

gsingh93:
> looks good to me, however, on testing it looks like the readline context is not enabled by default in the apple terminal - we should probably fix that (in a future PR)
I saw you did that, thanks!

> the copy commands don't copy to the system clipboard
Yeah, I wasn't quite sure what to do here as on one hand it makes sense to use readline functionality for the readline tag, but then what do we do when the user actually wants to copy to the system clipboard? For a similar reason, I didn't override the paste command, but we should figure out how to support both versions of paste.

PR #1046, opened 2022-11-20
Add homophones for 'align'

gsingh93:


rntz:
My only concern here is that I don't think we have any multi-word homonyms currently (like "align/a line") and I'm not sure whether anything might go wrong.

Also, is "aline" a word in the talon lexicon? A quick googling suggests it's not really a word...

gsingh93:
Yea, I saw it was the first multi word in the list. So far I haven't had any issues, I feel like it's safe enough to try out and we can revert if something goes wrong.

> Also, is "aline" a word in the talon lexicon? A quick googling suggests it's not really a word...

I'm not sure what it is but if I do "word align", it will come up instead of "a line" or "align". I could always add that to words to replace, but other people will probably run into the same issue

rntz:
You should probably report the presence of "aline" as an issue on the Talon lexicon repository, or make a PR to it: https://github.com/talonvoice/lexicon

rntz (approved):

PR #1047, opened 2022-11-21
Add the ability to save macros and copy them as commands

gsingh93:
With this PR, you can now record multiple macros by saving them and then play them by referencing their name. You can list all of these saved macros with 'macro list'. The saved macros are not persisted across Talon sessions (this may cause confusion as 'save' is in the name of the command, open to suggestions). To save a macro persistently, you can copy the last macro or copy a specific macro by name and then paste that in a talon file.


splondike (approved):
Works well, thanks! I just added in a close button to the GUI to fit the pattern we've been using in other imgui dialogues.

PR #1048, opened 2022-11-21
Move abbreviations to a CSV file

gsingh93:
Move abbreviations to a CSV file to make it easier to manage. I needed to modify `get_list_from_csv` in order to be able to reuse it for CSV files not in the settings folder. In a follow up PR, I plan to add support for a CSV file in the settings folder that a user can customize which will be merged with the default one. Alternatively, I could just move this one to the settings folder, but I followed the advice in the comment that this should be like homophones.csv.


splondike:
To me it makes more sense to put this in the settings folder. Can you let me know where you got the idea to put it next to the code? Would be good to discuss with that person to see what the reasoning was.

gsingh93:
There is a comment in `core/abbreviate/abbreviate.py` that says this:
> TODO: Make this a csv file. Not necessarily a settings/ csv file, it might be better to be like homophones.csv.

`homophones.csv` is located at `core/homophones/homophones.csv`. So I figured the comment was suggesting to do the same.

I would prefer to just have one file in the settings folder, let me know if I should go with that.

splondike:
Personally I think homophones.csv should be in settings.csv as well.

That said, let's give @rntz a week or so to weigh in on this since they made the original comment and might have considerations I hadn't thought of.

gsingh93:
Hmm, I see now why there's some benefit of having this be like `homophones.csv` and not be in the settings folder.

Everything in the `settings/` folder is excluded from `git`, so we can't ship the CSV file directly. So we'll essentially need to keep this massive Python dictionary and pass it to `get_list_from_csv` as the default list of abbreviations, which would generate `abbreviations.csv` in `settings/`, but then the process of contributing abbreviations upstream has actually gotten worse. We would still need to modify this massive Python dictionary to add abbreviations, and when users updated, they wouldn't get the new abbreviations as they'd already have a `abbreviations.csv` file created.

Editing and updating `homophones.csv` is a good experience, I can just say `customize homophones`, a text editor opens up, I add the homophones, and if I think others will benefit I commit that and make a PR. I'd like the process to be just as seamless for abbreviations.

Given this, I think it makes sense to go with the approach in this PR: keep the core list of abbreviations in `core/abbreviate`, allow users to easily modify that CSV and upstream changes, and still have a `settings/abbreviations.csv` for personal customizations that you don't want to upstream (which will be added in a future PR).

Thoughts?

splondike:
OK, let's see if we can work out a general set of rules for these kinds of things and then apply them to this case. Here are the considerations you've listed and some I can think of:

1. The reason we have the CSV files system is it's easier to edit, especially for non-programmers. Otherwise we could just get people to use the [captures trick](https://talon.wiki/unofficial_talon_docs/#captures-vs-lists-in-voice-commands) to augment our built in list or entirely replace it as needed.
2. Dealing with merge conflicts is annoying, particularly for non-programmers. Editing a committed .csv might mean you run in to these.
3. Some (few) lists that people often want to edit are not likely to ever change upstream (e.g. the alphabets.csv we added yesterday).
4. Some lists that people often want to edit are added to occasionally upstream (most of them, e.g. homophones, abbreviations, websites etc.). Some potentially also have entries removed.
5. The existing CSV system snapshots the list when the user first gets the feature. After that it doesn't change unless the user edits it. So we might add some more entries, and the user won't get them.
6. People might want to automatically get extra entries as we add them to upstream. For example somebody want new homophones added in automatically as people notice them.
7. People might want their list to stay the way it is once they've edited it. For example they might get misrecognitions with some of the upstream entries in 'additional_words.csv' and want to permanently remove them.

So I think the rule for 3 is clear enough; the existing settings/.csv thing is ideal.

I can think of a few ways to deal with the 'varying list' case (6 and 7). There's classifying each use case into 'I'll want the new entries' vs 'I want stability' (this is what we have now, cost is non-programmers need to deal with git). There's making the settings/.csv file act as a 'patch' on the list so you'd say whether you wanted to add or subtract an entry (maybe complex for non-programmers). There's putting everything in settings/.csv (default to stability) and then have a wizard that lets you merge in upstream changes.

I'm not immediately seeing a clear leader with any of those options. Do you have any further thoughts?

gsingh93:
Thanks for laying out those considerations! I think your suggestions for cases 6 and 7 could work. Let me summarize some of this in my own words to help me get my thoughts straight:

* If we want a list that is easily editable by non-programmers, has a good default setting where any customizations don't make sense to upstream, and we don't want to worry about merge conflicts or modifications breaking things, `settings/*.csv` makes the most sense, like you said. `additional_words.csv` and `alphabets.csv` are good examples of this, we essentially always want the "snapshot" functionality here.
* Now consider we have a list that we want to be easily editable by non-programmers, but having community contributions to this list may be beneficial. `homophones.csv` is a good example: while it's true that different accents will result in variations about what's considered a homophone, within the same accent there's a very minor amount of variation, so if someone upstreams homophones they've added to their config, then that can benefit me as long as our accents are relatively similar. If there are homophones that I know will only be useful to myself, I should be able to add them to a `settings/homophones.csv` file which will get "merged" (not literally merging the files, just merging the Python dictionaries) with the upstream one (with the settings one having priority, and the ability to override the builtin one. So in this case, `core/*.csv` with a `settings/*.csv` will work well.
    * The downside here is stability, as you mentioned. For each file like this, we can give users one of two options (I have no preference which one should be default):
    	1. Keep the behavior described above: `core/*.csv` and `settings/*.csv` will be "merged" (in terms of Python dictionaries)
    	2. If `settings/*.csv` doesn't exist, create it from `core/*.csv`, and after that only use `settings/*.csv`. If the user wants the latest changes, we have a simple script that will walk through each one and assist with merging (similar to `git add --patch`). This may be a little complicated for non-technical users, but it's no requirement they do this; if they never attempt to merge they would still have more options and more functionality available to them than they have today.
    	
    	
If these points match up with your thoughts, I think it just comes down to deciding which lists we think having community contributions for would be beneficial. I think `homophones.csv` and `abbreviations.csv` fit this category, which would mean going with the combined `core/*.csv` and `settings/*.csv`, with a setting to choose between combining the lists or always preferring `settings/*.csv`.

Of course, I can totally see how reasonable people would disagree with this classification of `homophones.csv` and `abbreviations.csv`.

splondike:
Your idea sounds good to me. So we'd change `get_list_from_csv` to `get_list_from_csv(filename, headers, default, append=False)` then change `append` to True for e.g. homophones.csv. When True this parameter would merge the lists. False would be the current behaviour of only sourcing from the settings/.csv one.

To cover the original intent of moving the long homophone list to a CSV this PR perhaps we could lean in to Python's dynamic typing and make `default: Union[dict[str, str], str]`. The string version would be a pathname to a CSV to use instead of a dictionary directly.

Does that sound good?

rntz:
This looks interesting. It seems like we're proposing having two csvs, a git-tracked one (in `core/` or wherever) and a non-tracked one in `settings/`, and either (a) merging them or (b) populating the `settings/` from the `core/`.

Option (a), **auto-merging**, is more convenient if users want to automatically receive upstream changes. This is useful for eg `homophones.csv`.

Option (b), **populating defaults**, is basically what `get_list_from_csv` already does, except it populates from a Python dictionary rather than another csv. Using a separate "template" csv would make it easier to have tooling that compares the upstream & local versions.

Honestly either option seems fine for `abbreviations.csv`, I don't think we update it particularly often or that it's very important users get updates to it.

Another option is to allow *either* auto-merging or populating-defaults, depending on the list. One issue is that this difference in behavior may not be obvious to the users.

Potential problems:

1. It may confuse users if there are two files named `foo.csv`. Right now we tell people to edit `additional_words.csv`, for instance. Possible solution: name the git-tracked versions something like `foo.template.csv`?

2. What if users want to _remove_ a template entry from an auto-merged list?

rntz:
If we're going to standardize the locations of the template csvs, I suggest having them in their own directory rather than putting them in core. Maybe `template/` or `data/`?

rntz:
For future reference:

# Settings csvs

| CSV | Description |
| --- | --- |
| additional_words.csv | Additional words to recognize |
| alphabet.csv | The spoken alphabet |
| search_engines.csv | Common search engines |
| system_paths.csv | Commonly used file paths |
| unix_utilities.csv | Unix command-line programs |
| websites.csv | Common web sites |
| words_to_replace.csv | Phrases to replace in spoken text |

# Other csvs

| CSV | Description |
| --- | ----------- |
| apps/git/git_arguments.csv | Common arguments/options to git commands |
| apps/git/git_commands.csv | Git subcommands |
| core/app_switcher/app_name_overrides.*.csv | Alternative names for common applications, for `focus` commands |
| core/homophones/homophones.csv | Common homophones for `phones` commands |
| tags/emoji/*.csv | Emoji, emoticon, and kaomoji names |

splondike:
Referring back to our actual set of CSVs is useful, thanks for compiling those lists. You've pushed me back towards the current system rather than a merging one given it's more simple and the fact that maybe only homophones and abbreviations might benefit from merging.

We might want to encourage people to contribute their emoji, app_name_overrides, homophones, and git* changes upstream. They're probably unlikely to have idiosyncratic changes that would not be useful to others (or be private info). Having these lists where they are (in git) encourages that.

It looks like the last change to the set of items in the abbreviate list prior to this PR was in October 2020; so as you say it doesn't change often. I could imagine people putting in their own industry specific abbreviations though. So this would be good to have in settings/.csv.

So I'm now thinking all we should change for the moment is to move abbreviations into the existing settings/.csv system. Let's see what gsingh93 thinks.

splondike:
I've implemented the above in #1141, closing out this PR as superceeded.

PR #1049, opened 2022-11-23
implement "clone that", "clone line" in more places

rntz:
This:

1. Makes `clone that` and `clone line` globally available rather than only when the line_commands tag is active.
2. Makes `clone that` duplicate the current selection, in line with our normal naming conventions. (Previously it duplicated the current line, a synonym for `clone line`.)
3. Provides default implementations of `edit.selection_clone`, `edit.line_clone`, and `user.line_clone`.

### DIFF
--- a/core/edit/edit.py
+++ b/core/edit/edit.py
@@ -18,6 +18,20 @@ class EditActions:
         actions.edit.line_end()
         actions.key("enter")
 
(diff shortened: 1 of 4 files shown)


splondike (approved):
Seems to work fine from my testing, thanks!

PR #1050, opened 2022-11-23
Add Gnome-terminal support to app_name_overrides.linux.csv

maxkwallace:
I don't have the link (it's older than 90 days, no longer searchable) but at least one other person ran into this issue/confusion, I found this fix via the Talon Slack.

### DIFF
--- a/core/app_switcher/app_name_overrides.linux.csv
+++ b/core/app_switcher/app_name_overrides.linux.csv
@@ -1,3 +1,3 @@
 grip, DataGrip
 py, jetbrains-pycharm-ce
-term, Gnome-terminal
+terminal, Gnome-terminal
--- a/core/app_switcher/app_name_overrides.linux.csv
+++ b/core/app_switcher/app_name_overrides.linux.csv
@@ -1,2 +1,3 @@
 grip, DataGrip
 py, jetbrains-pycharm-ce
+term, Gnome-terminal


splondike:
What's the issue that this fixes? On my Fedora machine I can say 'focus gnome terminal' or 'focus gnome' and it focusses the Gnome terminal. Mabye you don't get those aliases? In this case what do you get when you run `registry.lists['user.running']` in the REPL?

```
>>> registry.lists['user.running']
[{'evolution': 'Evolution', 'kitty': 'kitty', 'firefox': 'firefox', 'gnome': 'Gnome-terminal', 'gnome terminal': 'Gnome-terminal', 'gnome dash terminal': 'Gnome-terminal'}]
```

maxkwallace:
@splondike this fixes the issue of someone new to talon expecting to be able to say "focus terminal" and have it work as expected. If you don't agree that this is a worthwhile behavior have baked in by default, that's fine, feel free to close the PR.

I am on Ubuntu 20.04 and "focus gnome" and "focus gnome terminal" both work out-of-the-box but "focus terminal" and "focus term" do not.

Personally, I am up and running with talon now, but frankly I had a really hard time using talon at first. I realize this makes me sound like an idiot, in fact I am a quite competent programmer, but I found learning talon initially very overwhelming, including finding the info I needed in the documentation. I don't remember how I ended up figuring it out, but it was unexpectedly difficult to figure out how to get the application names needed for the app_name_overrides file. I don't think this is discussed anywhere in the docs? Yes, in theory, anyone should "just" be able to figure it out for themselves, and then add their desired alias to the file, but from my own experience, I strongly feel like there is a basic level of "least surprise" ergonomics that is worth aspiring to, when there are users who have recently experienced a major upper extremity injury, and are trying to figure out voice coding for the first time (a major mental shift), including hardware set up, learning the alphabet, navigating talon's help menus, etc etc. -- each of these things is individually manageable, but when you add them all together it becomes overwhelming.

For the terminal in particular, on Ubuntu with Gnome, it just says "Terminal" everywhere in the UI. So the user may only know they need to say "gnome" if they already have a terminal open, read the docs and learn they can say "running list" to get these names. But "focus gnome terminal" is too long to say every time, and "focus gnome" is not intuitive, since Gnome is a desktop environment, not the terminal. Moreover, the names displayed in "running list" do not work in the app_name_overrides file.

For all these reasons, I feel like it is not asking a lot to add one line of code and have "focus terminal" work out of the box, and in fact this would be beneficial for many users in the future.

splondike:
Talon is indeed quite difficult to get your head around initially. For example we have a discoverability problem with commands (i.e. how do I know what's possible and find out how to do X) that I don't know we have a good solution for yet. There's a lot of functionality you might want to cover and the README is already pretty long.

I just like to get it clear what it is we're solving with these changes before we put them in. For example the issue might have been that gnome terminal was getting some strange alias on your machine which would suggest we might want to generally fix the automatic alias generator.

So the idea here is to make it easier to focus gnome terminal specifically because as you say it's often called 'Terminal' in the window title and application menus. That seems reasonable to me, what do you think about changing the alias to 'terminal' instead of 'term' in the .csv file in that case? That way the alias would match that title.

Incidentally that will also fix the breakage this change would introduce with the 'xterm' terminal client. The alias for that is currently 'term' and this change would stop that working. I'm not too concerned about this though either way since I don't expect many people would still be using that client.

maxkwallace:
@splondike got it, thanks for the clarification.

100% good with "terminal" instead of "term", just made the change in the branch.

splondike (approved):

splondike:
Thanks!

PR #1051, opened 2022-11-27
Allow specifying a path to ls

gsingh93:
Allow specifying a path to `ls` for unix terminals. Because I can't test on Windows, I only fixed the function signature without making use of the path in these cases. I still made the change for unix terminals on Windows like WSL, but I did not test these. If you think we need to test these, I can just make the change for unix terminals on Linux/macOS.

### DIFF
--- a/apps/iterm/iterm.py
+++ b/apps/iterm/iterm.py
@@ -54,12 +54,12 @@ class user_actions:
     #     """selects the file"""
     #     actions.insert(path)
 
-    def terminal_list_directories():
-        actions.insert("ls")
+    def terminal_list_directories(path: str):
+        actions.insert(f"ls {path}")
         actions.key("enter")
 
-    def terminal_list_all_directories():
-        actions.insert("ls -a")
+    def terminal_list_all_directories(path: str):
+        actions.insert(f"ls -a {path}")
         actions.key("enter")
 
     def terminal_change_directory(path: str):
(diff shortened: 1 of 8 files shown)


splondike:
I don't mind the change here given `lisa` presses enter currently. My personal ideal UX would probably just be lisa and your one that takes arguments just printing out the `ls -al ` type thing without enter, but with a space. You could then say 'lisa list all word code enter' to get the behaviour in this PR, or 'lisa list all snake noise macros enter' to get that. Happy to put in something like this change though. 

One problem though is this change seems to conflict a bit with the proposed solution to #1020 . In that we were going to deprecate the `terminal_list_directories` and `terminal_list_all_directories` actions which are being modified here. We'd want to work out a unified solution I think.

DonnerWolfBach:
I will continue working on this and improve #1108, but it will likely take some time because I want to think about what makes sense there and wether and how it should be consistent with how cd works.

DonnerWolfBach:
I tried a bit around with more complex "Placeholders" like the one suggested from @splondike `[(<user.word>|<user.format_text>+)]` but either the implementation would be much to complex to be fitting to such a minor use case or lacking it feature richness.
So as in #1020 I don't see this being worth it and won't continue working on it. However, I might just not have enough experience with knausj's inner workings to find a clean solution.

splondike:
Since the original author hasn't written back in a few months, and we don't have the interface worked out (e.g. it conflicts with #1020  by the same author), I'm going to close this out.

PR #1054, opened 2022-11-28
make window management on mac work a bit better

rntz:
At least for me on a MacBook pro on ventura, the current implementations of `window_move_desktop_{left,right}` produce an error, and even if they didn't they would use keybindings that don't work by default. This fixes that error and uses the default keybindings of ctrl-left/right.

### DIFF
--- a/plugin/desktops/desktops_mac.py
+++ b/plugin/desktops/desktops_mac.py
@@ -14,14 +14,16 @@ def _drag_window_mac(win=None):
     if win is None:
         win = ui.active_window()
     fs = win.children.find(AXSubrole="AXFullScreenButton")[0]
-    rect = fs.AXFrame["$rect2d"]
-    x = rect["x"] + rect["width"] + 5
-    y = rect["y"] + rect["height"] / 2
+    rect = fs.AXFrame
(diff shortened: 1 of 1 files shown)


phillco:
Is this the error that you get?

```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "talon/scripting/actions.py", line 88, in __call__
  File "/Users/phillco/.talon/user/knausj_talon/plugin/desktops/desktops_mac.py", line 43, in window_move_desktop_left
    with _drag_window_mac():
  File "lib/python3.9/contextlib.py", line 119, in __enter__
  File "/Users/phillco/.talon/user/knausj_talon/plugin/desktops/desktops_mac.py", line 17, in _drag_window_mac
    rect = fs.AXFrame["$rect2d"]
TypeError: 'Rect' object is not subscriptable
```

Just confirming that I see the same if so on Monterey. I guess @nriley @pokey and I don't use those commands for moving between desktops. :)

The `$rect2d` syntax was removed last year, so any other locations should be fixed up in the same way.

phillco (approved):

rntz:
Yep that's the error that I get.

PR #1056, opened 2022-12-03
turn alphabet into a list grabbed from a csv file

timo:
thanks a bunch to @gsingh93 for the inspiration (from #1048)


splondike (approved):

xavier630:
It could be worth creating a default CSV that gets shipped so users know the expected header structure, folder to put the CSV in etc.

It also would mean that you could remove the default alphabet from the code.

splondike:
A `settings/alphabet.csv` should be created automatically the first time you boot Talon with this code (which has a header). Hopefully people will either find that or get directed to it one way or another.

You may also be interested in this discussion about settings/.csv vs included .csv files starting here: https://github.com/knausj85/knausj_talon/pull/1048#issuecomment-1336317288

PR #1057, opened 2022-12-07
edit_mac.py: omit cmd-left for file_start/file_end

rntz:
The default bindings for `edit.file_start` on mac presses `cmd-up cmd-left` (and `edit.file_end` is `cmd-down cmd-left`). I find this `cmd-left` to be redundant (at least in TextEdit and in Emacs, `cmd-up` goes to the start of the document, including the beginning of the line) and to produce bad behavior in, eg, web browsers, where `cmd-up` successfully scrolls to the top but then `cmd-left` goes back in history.

Are there some applications where this `cmd-left` is necessary?

### DIFF
--- a/core/edit/edit_mac.py
+++ b/core/edit/edit_mac.py
@@ -84,10 +84,10 @@ class EditActions:
         actions.key("shift-alt-right")
 
     def file_end():
-        actions.key("cmd-down cmd-left")
+        actions.key("cmd-down")
 
     def file_start():
-        actions.key("cmd-up cmd-left")
+        actions.key("cmd-up")
 
     def find(text: str = None):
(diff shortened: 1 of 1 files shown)


auscompgeek (approved):
I don't believe there are actually any apps that require cmd-left. VSCode certainly doesn't.

Given the bad interaction with browsers, I'd be inclined to get this merged and deal with any other consequences later.

splondike (approved):
Tried this out on Chrome, Firefox, VSCode, Notes app, XCode, MS Word. It worked fine on all of those except MS Word (where neither the current version or the new one do the right thing).

Since this improves the situation in the browser I agree that we should merge it in.

PR #1058, opened 2022-12-07
remove bogus comments created by migration script

rntz:
Removes some comments created by the action migration script. Perhaps we should have TODOs to implement `app.tab_detach` on various platforms instead? I'm not sure there _is_ a standard way to accomplish that action, though.


splondike (approved):

splondike:
Yeah, I'm not aware of any standard shortcuts for doing tab_detach.

PR #1059, opened 2022-12-07
windows_and_tabs_mac.py: use accessibility to implement app.window_close

rntz:
This works more reliably than `cmd-w` because that really tries to close the current tab. But we can't use `cmd-shift-w` because that shortcut doesn't exist unless there's more than one tab.

However, the mac accessibility API might be experimental, so I'd like signoff from @lunixbochs on this one.


rntz:
Ok, apparently aegis doesn't want accessibility code in knausj until at least two OSes have accessibility support so we have some idea what the stable API will look like. Closing this.

PR #1060, opened 2022-12-10
simplify create_spoken_forms and avoid empty spoken forms

rntz:
Currently `create_spoken_forms` can generate empty spoken forms if passed certain inputs (eg. the empty input). This prevents that, and also cleans up some code along the way.

### DIFF
--- a/core/create_spoken_forms.py
+++ b/core/create_spoken_forms.py
@@ -272,24 +272,28 @@ def generate_string_subsequences(
     words_to_exclude: list[str],
     minimum_term_length=DEFAULT_MINIMUM_TERM_LENGTH,
 ):
+    # Includes (lower-cased):
+    # 1. Each word in source, eg "foo bar baz" -> "foo", "bar", "baz".
+    # 2. Each leading subsequence of words from source,
+    #    eg "foo bar baz" -> "foo", "foo bar", "foo bar baz"
+    #    (but not "bar baz" - TODO: is this intentional?)
+    #
+    # Except for:
+    # 3. strings shorter than minimum_term_length
+    # 4. strings in words_to_exclude.
     term_sequence = source.split(" ")
-    terms = list(
-        {
-            term.lower().strip()
-            for term in (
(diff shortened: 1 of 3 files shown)


splondike (approved):
This code would probably benefit from some unit tests. The change looks good and seems to work fine though. Happy for it to be merged.

core/create_spoken_forms.py
```diff
@@ -321,28 +322,22 @@ def create_spoken_forms(
...
-            for form in full_spoken_forms:
-                if form not in terms:
-                    terms.append(form)
-
-        else:
-            terms = full_spoken_forms
-
-        return terms
+        # Avoid empty spoken forms.
+        return list(x for x in spoken_forms if x)
```
pokey, line 340: Nit: 
```suggestion
        return [x for x in spoken_forms if x]
```

pokey (approved):
Looks good to me. I haven't tested it though

I do think a unit test or two would be nice. What is the current recommended way to do that in knausj?

core/create_spoken_forms.py
```diff
@@ -270,27 +270,31 @@ def create_spoken_forms_from_regex(source: str, pattern: re.Pattern):
...
     source: str,
     words_to_exclude: list[str],
-    minimum_term_length=DEFAULT_MINIMUM_TERM_LENGTH,
+    minimum_term_length: int,
 ):
+    # Includes (lower-cased):
+    # 1. Each word in source, eg "foo bar baz" -> "foo", "bar", "baz".
+    # 2. Each leading subsequence of words from source,
+    #    eg "foo bar baz" -> "foo", "foo bar", "foo bar baz"
+    #    (but not "bar baz" - TODO: is this intentional?)
```
pokey, line 279: I believe this was to avoid quadratic explosion back when DFA was more of an issue. Might still be a problem for dragon?

splondike:
Dropped in some tests for create_spoken_forms. Will let rntz merge when they're ready.

PR #1071, opened 2022-12-24
add more extensions

rntz:
Adds more extensions to the defaults in `file_extensions.py`.


phillco (approved):
It'd be nice to sort these. I guess we'd lose blame.

core/file_extension/file_extension.py
```diff
@@ -32,6 +33,22 @@
...
     "dot text": ".txt",
+    "dot julia": ".jl",
+    "dot J L": ".jl",
+    "dot html": ".html",
+    "dot css": ".css",
+    "dot sass": ".sass",
+    "dot svg": ".svg",
+    "dot png": ".png",
+    "dot wave": ".wav",
+    "dot flack": ".flack",
```
auscompgeek, line 44: Was this mean to be `.flac`?

phillco: Whoops, missed that too :-(

phillco: https://github.com/knausj85/knausj_talon/pull/1076

PR #1075, opened 2022-12-27
Add some more abbreviations

auscompgeek:
This adds a mixbag of abbreviations, mostly tech related, some plural versions of those that already exist.


core/abbreviate/abbreviate.py
```diff
@@ -72,22 +82,26 @@
...
     "environment": "env",
+    "error": "err",
     "escape": "esc",
     "etcetera": "etc",
     "example": "ex",
     "exception": "exc",
     "execute": "exec",
-    "expression": "exp",
+    "exponent": "exp",
+    "expression": "expr",
```
auscompgeek, line 100: This is technically a breaking change, but I much more commonly see `expr` as an abbreviation for "expression", presumably thanks to [`expr(1)`].

[`expr(1)`]: https://en.wikipedia.org/wiki/expr

rntz (approved):
I think the breaking change is probably OK. If there's no opposing views I suggest we merge in a week. I guess *maybe* this should go in BREAKING_CHANGES.txt?

splondike:
Yeah, I don't think this would be a big deal for anybody; it'd be quite noticeable if somebody found it was a problem and would easy for them to fix. Merging.

PR #1078, opened 2023-01-01
Change actions shortcuts in go_back and go_forward in Firefox mac

david-tejada:
When editing text `cmd-left` and `cmd-right` just move the cursor to the start or end of the line. `cmd-[` and `cmd-]` don't have that issue.


splondike (approved):
Thanks, works well!

PR #1081, opened 2023-01-04
Deprecate `murder`, add `shuffle flipper` to i3wm

splondike:
Also remove redundant 'port ten' command.

Replaces #928


EdJoJob (approved):

apps/i3wm/i3wm.talon
```diff
@@ -66,6 +65,8 @@ vertical (shell | terminal):
     user.system_command("i3-msg move container to workspace number 10")
```
EdJoJob, line 65: This `10` special can be removed as well

rntz (approved):

PR #1083, opened 2023-01-04
Improve `edit.select_word()`

xavier630:
`select_word()` has several issues as per https://github.com/knausj85/knausj_talon/issues/1084.

Notably accidental selection of symbols and word selection breaking if the caret is before a new line.

This pull request aims to resolve these.

Note: This is only tested on MacOS, though it uses higher level functions for all operations, so I don't foresee any major issues on other operating systems.

### DIFF
--- a/core/edit/edit.py
+++ b/core/edit/edit.py
@@ -3,6 +3,10 @@ from talon import Context, Module, actions, clip
 ctx = Context()
 mod = Module()
 
+PUNCTUATION_SYMBOLS_WHICH_SIGNIFY_THE_END_OF_A_WORD = \
+    ['.', '!', '?', ';', ':', '—', '_', '/', '\\', '|',
+     '@', '#', '$', '%', '^', '&', '*', '(', ')', '[', ']',
+     '{', '}', '<', '>', '=', '+', '-', '~', '`']
 
 @ctx.action_class("edit")
 class EditActions:
@@ -31,6 +35,18 @@ class EditActions:
         actions.edit.right()
         actions.key("enter")
         actions.edit.paste()
+
+    def select_word():
+        actions.edit.extend_right()
+        character_to_right_of_initial_caret_position = actions.edit.selected_text()
+        actions.edit.left()  # in order to return to original caret position
+
+        if (character_to_right_of_initial_caret_position in
+                PUNCTUATION_SYMBOLS_WHICH_SIGNIFY_THE_END_OF_A_WORD):
+            actions.edit.extend_word_left()
+        else:
+            actions.edit.word_right()
+            actions.edit.extend_word_left()
 
 
 @mod.action_class
--- a/core/edit/edit.py
+++ b/core/edit/edit.py
@@ -39,7 +39,7 @@ class EditActions:
     def select_word():
         actions.edit.extend_right()
         character_to_right_of_initial_caret_position = actions.edit.selected_text()
-        actions.edit.left()  # in order to return to original caret position
+        actions.edit.left()
 
         if (character_to_right_of_initial_caret_position in
                 PUNCTUATION_SYMBOLS_WHICH_SIGNIFY_THE_END_OF_A_WORD):
--- a/core/edit/edit.py
+++ b/core/edit/edit.py
@@ -77,11 +77,11 @@ class EditActions:
         did_select_text = character_to_right_of_initial_caret_position != ""
 
         if did_select_text:
-        
+
             # .strip() is to handle newline characters which become an empty string.
             if (
-                (character_to_right_of_initial_caret_position.strip()
-                    in SYMBOLS_WHICH_SIGNIFY_THE_END_OF_A_WORD)
+                character_to_right_of_initial_caret_position.strip()
+                in SYMBOLS_WHICH_SIGNIFY_THE_END_OF_A_WORD
             ):
                 # Come out of the highlight in the initial position.
                 actions.edit.left()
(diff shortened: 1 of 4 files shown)


core/edit/edit.py
```diff
@@ -3,6 +3,41 @@
...
+    "<",
+    ">",
+    "=",
+    "+",
+    "-",
+    "~",
+    "`",
+    " ",
+    "",
+]
```
xavier630, line 39: Quotation marks are left out of this list - they didn't seem to cause issues in my brief testing of them.

splondike (approved):
This is a pareto improvement on the current situation from my testing on Linux; it still gets things wrong, but there are no regressions on the current situation that I can see, and it gets some more things right.

I actually tested this using the following script. It registers a new action that lets you perform some tests against the currently selected textbox. It should work on OSX as well if you want to use it to do any tests:
```python
import time


mod = Module()

test_cases = [
    ("one +|two three\nfour", "one +two| three\nfour"),
    ("one two+| three\nfour", "one +two| three\nfour"),
    ("one +t|wo three\nfour", "one +two| three\nfour"),
    ("one tw+o| three\nfour", "one +two| three\nfour"),
    ("one tw|o+ three\nfour", "one +two| three\nfour"),
    ("one |t+wo three\nfour", "one +two| three\nfour"),
    ("one two three+|\nfour", "one two +three|\nfour"),
    ("one two three\nfour+|", "one two three\n+four|"),
    ("one two three\nfou+r|", "one two three\n+four|"),
    ("+|one two three\nfour", "+one| two three\nfour"),
    ("one (tw+|o) three\nfour", "one (+two|) three\nfour"),
    ("one (two+|) three\nfour", "one (+two|) three\nfour"),
    ("one (two+)| three\nfour", "one (+two|) three\nfour"),
    ("one |(+two) three\nfour", "one (+two|) three\nfour"),
    ("one two.+| three\nfour", "one +two|. three\nfour"),
    ("one two+|-hyphen three\nfour", "one +two|-hyphen three\nfour"),
]

original_title = None


def calculate_raw_text(before):
    return before.replace("+", "").replace("|", "")


def calculate_expected_text(after):
    one = after.index("+")
    two = after.index("|")

    return after[(min(one, two)+1):max(one, two)]

def set_selection(before):
    start_pos = before.index("+")
    end_pos = before.index("|")

    if start_pos < end_pos:
        end_pos -= 1
    else:
        start_pos -= 1

    for _ in range(start_pos):
        actions.edit.right()

    if end_pos < start_pos:
        for _ in range(start_pos - end_pos):
            actions.edit.extend_left()
    elif end_pos > start_pos:
        for _ in range(end_pos - start_pos):
            actions.edit.extend_right()


@mod.action_class
class Actions:
    def select_word_test():
        """Test select_word"""
        global original_title
        original_title = actions.win.title()

        for before, after in test_cases:
            if original_title != actions.win.title():
                # We've changed window focus, stop the test
                break

            before_raw = calculate_raw_text(before)

            # Put in the text and position the cursor at the start
            actions.edit.select_all()
            actions.edit.delete()
            actions.insert(before_raw)
            actions.edit.file_start()

            # Position the selection where we want it to be
            set_selection(before)

            # Run select word
            actions.edit.select_word()

            # Confirm we're selecting the expected text
            actual_text = actions.edit.selected_text()
            expected_text = calculate_expected_text(after)

            before_ = before.replace("\n", "\\n")
            actual_text_ = actual_text.replace("\n", "\\n")
            if actual_text == expected_text:
                print("OK", before_, "  ->  ", actual_text_)
            else:
                print("KO", before_, "  ->  ", actual_text_)

            time.sleep(0.5)


# Run something like this in the REPL to give you a chance to swap
# to the relevant textbox:

# import time;time.sleep(2);actions.user.select_word_test()

# Results with patch

# ## Geany
# KO one |(+two) three\nfour   ->   one

# ## Github editor
# KO one (tw+|o) three\nfour   ->   two)
# KO one (two+|) three\nfour   ->   two)
# KO one (two+)| three\nfour   ->   two)
# KO one |(+two) three\nfour   ->   one
# KO one two.+| three\nfour   ->   two.
# KO one two+|-hyphen three\nfour   ->   two-

# ## VSCode
# Perfect

# ## Gedit
# KO one |(+two) three\nfour   ->   (
# KO one two.+| three\nfour   ->   .
```

In any case I'm happy to merge this in. I'll do so around the 25th if nobody else does before then.

rntz:
I haven't looked at this closely but it seems kinda complicated for such a simple feature. I'd appreciate the chance to give it a think before merging - I might have time this weekend or next, so could you gimme a couple of weeks to look at it? If I haven't gotten round to it by then go ahead and merge.

splondike:
I'm happy to wait :+1: .

xavier630:
Please do take a look 👍  

> it seems kinda complicated for such a simple feature

One thing worth noting is that it does the same thing as the existing implementation (go one character to the left right, go an entire word to the left, then highlight the word to the right) except that the initial step to the right is conditional on what the character to the right is.

rntz:
After brainstorming on a maintenance session we were unable to come up with any simpler way to do this. I've simplified the code slightly by turning that list into a string, but otherwise lgtm.

My only note is that this won't work in editors like emacs where pressing left/right with a selection active doesn't always put you at the left/right end of the selection. But, emacs already needs tons of special cases; chrome is far more common than emacs.

rntz (approved):

rntz:
I'll let someone else take a look at my changes and merge this, or merge it myself in a week if nothing happens.

splondike:
Looks good to me, thank you both :).

Issue #1088, opened 2023-01-13
add command for retrieving user.system_path itself in a terminal


DonnerWolfBach:
I really like the system to store the paths via a csv introduced in #802 as a result of #785 , but I would like to have a way to get these paths without the "cd" in front. Thus I propose adding this in terminal.talon under `go <user.system_path>: insert('cd "{system_path}"\n ls\n')`:
```
path <user.system_path>: insert('"{system_path}"')
# or
<user.system_path> path: insert('"{system_path}"')
```
Is any of these better or worse for working with talon (like impacting performance)?

DonnerWolfBach:
Btw, I am already working on the pull request, but I'll have to cleanup some other things first

DonnerWolfBach:
Pull request is done + removed "<> path" as suggested by splondike

splondike:
PR has been merged, closing this out.

PR #1089, opened 2023-01-13
added way of retrieving user.system_path

DonnerWolfBach:
meant to address #1088


splondike:
I think we should just have the 'path ...' variant, we more often have commands with a 'scoping' prefix then whatever arguments are needed.

DonnerWolfBach:
@splondike Ok, I'll fix that. Wanted to present it though because I personally prefer the "<>path" over "path <>".

splondike (approved):

splondike:
Thanks for the change. No worries about providing the suggestion, good to discuss :).

PR #1091, opened 2023-01-14
alternate capitalization for windows_explorer app.exe

tpensyl:
Issue [called out in slack](https://talonvoice.slack.com/archives/CJ2EE3LEB/p1673705822738799) by user Gabriele. 

On their Windows 10 machine and my Windows 11 machine, app.exe for windows explorer is cased as "Explorer.EXE".

On non english machines, the title may not match the English "Windows Explorer", and so having an accurate exe matcher is essential.

Testing: I confirmed on my Windows 11 machine that (if I remove the app.name matchers), the context matches with the new capitalization, but not with the old capitalization.

### DIFF
--- a/apps/windows_explorer/windows_explorer.py
+++ b/apps/windows_explorer/windows_explorer.py
@@ -12,6 +12,8 @@ os: windows
 and app.name: Windows-Explorer
 os: windows
 and app.exe: explorer.exe
+os: windows
+and app.exe: Explorer.EXE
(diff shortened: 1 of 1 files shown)


timo:
maybe use a case insensitive regex match instead like `app.exe: /^explorer.exe$/i` to match any combination of cases

splondike (approved):
Thanks!

Issue #919, opened 2022-07-14
Updates for Talon 0.3


lunixbochs:
knausj_talon should stop importing from `talon_plugins`. See `actions.list('tracking')` for an official list of actions to interact with the mouse control systems. I've added compatibility shims for now that throw a warning, but they will be removed eventually.

This is primarily done in:
- mouse_grid/mouse_grid.py
- code/mouse.py

(Note that Talon 0.3 hasn't been released to Windows yet, so any code for this shouldn't be merged until Talon Windows 0.3 is out with support for the new actions)

lunixbochs:
Talon 0.3.1 was released to Windows, so it's a good time to move on this issue.

lunixbochs:
I think you can pass a bool into the toggle functions rather than checking first

lunixbochs:
@knausj85 thanks for the changes!

I think you can pass a bool directly into the toggle functions rather than checking if they're enabled first.

Also, you shouldn't be using control1_* in new code. I'm pretty sure the compat shim points users at the new control mouse already, so you should use the new version here.

PR #1092, opened 2023-01-14
Address #919

knausj85:
Address #919: Remove imports of talon_plugin; use new tracking actions.

### DIFF
--- a/plugin/mouse/mouse.py
+++ b/plugin/mouse/mouse.py
@@ -109,7 +109,7 @@ class Actions:
 
     def mouse_wake():
         """Enable control mouse, zoom mouse, and disables cursor"""
-        actions.tracking.control_zoom_toggle() 
+        actions.tracking.control_zoom_toggle()
         # eye_mouse.control_mouse.enable()
         if setting_mouse_wake_hides_cursor.get() >= 1:
             show_cursor_helper(False)
@@ -335,7 +335,7 @@ def stop_scroll():
     if control_mouse_forced:
         if actions.tracking.control1_enabled():
             actions.tracking.control1_toggle()
-        
+
         control_mouse_forced = False
 
     scroll_job = None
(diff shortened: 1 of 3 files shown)


plugin/mouse/mouse.py
```diff
@@ -111,43 +109,11 @@ def mouse_hide_cursor():
 
     def mouse_wake():
         """Enable control mouse, zoom mouse, and disables cursor"""
-        eye_zoom_mouse.toggle_zoom_mouse(True)
+        actions.tracking.control_zoom_toggle()
```
splondike, line 112: Might want to make this:
```python
actions.tracking.control_zoom_toggle(True)
```

At present it looks like it would turn off zoom mouse if invoked twice.

plugin/mouse/mouse.py
```diff
@@ -164,8 +130,12 @@ def mouse_drag_end():
 
     def mouse_sleep():
         """Disables control mouse, zoom mouse, and re-enables cursor"""
-        eye_zoom_mouse.toggle_zoom_mouse(False)
-        toggle_control(False)
+        if actions.tracking.control_zoom_enabled():
+            actions.tracking.control_zoom_toggle()
```
splondike, line 134: Could use the explicit state form here as well if you wanted:
```python
actions.tracking.control_zoom_toggle(False)
```

splondike (approved):
Looks fine (though don't have an eye tracker to test). Couple of minor suggestions.

plugin/mouse/mouse.py
knausj85: > Looks fine (though don't have an eye tracker to test). Couple of minor suggestions.

I've addressed your feedback, thanks.  I have an eye tracker and have tested these changes.

splondike (approved):
Looks good to me!

PR #1097, opened 2023-01-23
Support parentheses commands in dictation mode as well as command mode.

bjaspan:
Parentheses are common in dictated prose.


rntz:
I suggest not including "paren" in here, but only the disambiguated left/right versions, since "paren" is pretty close to "parent", so including it raises the chance of a misrecognition.

splondike (approved):

splondike:
Thanks!

PR #1100, opened 2023-01-28
Add command to require a particular host in talon files

knausj85:
- Add command to require a particular host in talon files
- Add new action talon_get_hostname() to retrieve hostname


splondike (approved):
Looks good!

PR #1102, opened 2023-01-30
Add dictation auto formatting rules for time of day, currency, and percentages.

bjaspan:


bjaspan:
The pre-commit build failure should be fixed by https://github.com/knausj85/knausj_talon/pull/1101.

phillco:
Merging in main to try to pull in the pre commit fixes

phillco:
We should add a quick description of the added functionality in the PR description.

bjaspan:
Added a description.

phillco:
Thanks!

rntz:
Thanks, this looks really useful! I've made one change in functionality: I've require "hours" in times using "hundred", eg. "twelve hundred hours" will produce 12:00 but "twelve hundred" will not. Eventually we might want bare numbers in dictation mode to get turned into numerals, eg. "twelve hundred" -> 1200, assuming they're not too small (eg "one" -> 1 is not always desirable). So I'd like to find some way to make it clear whether you're speaking a time or a number. I'm not sure that I've made the best choice here, however; feel free to suggest another approach, especially if you feel strongly one way or the other.

I've also made some small changes to the code to clean & simplify it. It looks good to me. I wouldn't mind more eyes on this, so I'll give it another week or so for people to look at it and merge it if nobody has objections or suggestions.

rntz (approved):

phillco:
We reviewed this in the knausj grooming session on Discord, by the way; scheduling is a little random at the moment but keep an eye on #community if you'd like to hop into future sessions :)

core/text/text_and_dictation.py
```diff
@@ -85,24 +85,15 @@ def prose_percent(m) -> str:
...
-
-@mod.capture(
-    rule="<user.number_string> {user.currency_denomination} [and] <user.number_string> [cents|pence]"
-)
-def prose_money_with_cents(m) -> str:
-    return m.currency_denomination + m.number_string_1 + "." + m.number_string_2
-
-
-@mod.capture(rule="<user.prose_money_with_cents> | <user.prose_simple_money>")
+@mod.capture(rule="<user.number_string> {user.currency_denomination} [[and] <user.number_string> [cents|pence]]")
```
bjaspan: Huh. I tried exactly this rule and it didn't seem to work, I thought maybe nested []'s were not supported.  I'm assuming you tested it, so I must have done something else wrong.

core/text/text_and_dictation.py
```diff
@@ -85,24 +85,15 @@ def prose_percent(m) -> str:
...
 def prose_money(m) -> str:
-    return str(m)
+    s = m.currency_denomination + m.number_string_1
+    if hasattr(m, "number_string_2"):
+        s += "." + m.number_string_2
+    return s
 
 
-hours_twelve = {
+hours_12 = {
```
bjaspan, line 98: I also started with hours_12 and swear it caused parsing errors, which confused me since it obviously is a valid identifier, but when I changed to hours_twelve it worked.

rntz: Note that I didn't change the name of the list talon-side, just this Python variable. Talon may not allow numbers in list names? Honestly not sure, I didn't try.

core/text/text_and_dictation.py
rntz: It seemed to work in my testing, yeah. Feel free to test it on your end, maybe I missed something?

Nested [] should definitely be supported, though.

bjaspan:
Your changes seem to work fine.

bjaspan:
Regarding @rntz's comment, "So I'd like to find some way to make it clear whether you're speaking a time or a number": 

First, your change to require "hours" after "hundred" is definitely a good one, thanks.

Second, I don't think there actually is a serious conflict between time parsing and number parsing. If I say "twelve thirty seven," I could mean 12:37 or 1237, but in the latter case I think a native speaker is much more likely to say "twelve hundred thirty seven" (or of course "one thousand hundred thirty seven").  I believe the same pattern applies to all time phrases versus comparable number phrases.  The one exception would have been interpreting "Twelve hundred" as a time phrase; it should definitely mean 1200 and not 12:00, which is why your change was necessary.

I have noticed one thing that concerns me.  With these changes applied, I discovered that when dictating the above paragraph:

* when I say "numeral one thousand two hundred thirty seven" I immediately get 1237
* when I say "one thousand two hundred thirty seven"  I get nothing.  

If I check out origin/main, however, then when I say "one thousand two hundred thirty seven" I immediately get that text.   Thus it looks like somehow  this change is breaking the dictation of number words.

splondike (approved):
This looks pretty good to me. I'm also not able to reproduce the issue with saying 'one thousand two hundred thirty seven', it seems to work correctly in all of command, dictation, and 'say'.

The thing about 20xx and 19xx years being formatted as times is a bit annoying, but as you say the current situation isn't great for those anyway. Probably ideally we'd want to exclude those from the hour formatter, but I think that's something that can be done later.

So if you're not still running into that 'one thousand two hundred thirty seven' issue I'm happy for us to merge this in.

brief:
I'm running into the same problem as @bjaspan on this one in dictation / say mode. Believe the issue is tied to how the percent and money rules begin by capturing `user.number_string` since prefacing each with a word (e.g. `currency <user.number_string>`) fixes it for me.

In my case, it affects several combinations of numbers:

- "one hundred and one" → nothing
- "one thousand two hundred" → nothing
- "one million twenty one" → nothing

Yet the following all work:

- Any numbers up to "one hundred"
- "one hundred one" → "one hundred one"
- "one thousand one" → "one thousand one"
- "one million twenty"  → "one million twenty"

splondike:
Hmm, I removed all my other user scripts and checked out a7d67f5ec39cb8475f7af24b5cce93b17a4a0ade on knausj_talon. I wasn't able to reproduce your issues brief. In the following the utterance is on the left, and the output is on the right:

* 'one hundred and one' -> 101
* 'say one hundred and one' -> one hundred and one
* 'one thousand two hundred' -> 1200
* 'say one thousand two hundred' -> 1200
* 'one thousand two hundred thirty seven' -> 1237
* 'say one thousand two hundred thirty seven' -> 1237
* 'say numeral one thousand two hundred thirty seven' -> 1237

Could you try the same approach of making sure there's no other potentially conflicting user scripts in play?

brief:
No luck, I'm afraid. I have the same issue. The final numeral works as it should as do the numbers in command mode, but every other example with 'say' does not.

rntz:
I've not tried to replicate the issues discussed above, but one remaining concern is that eg "nineteen fourteen" could mean 19:14 (ie 7:14pm) or 1914 (the year). Similarly with "twenty twenty three" for 20:23 vs 2023.

Solutions:

1. Always require a disambiguating "o'clock"/"hours"/"am"/"pm", eg "nineteen fourteen hours" for 19:14.

2. Default to time formatting. If later we introduce bare numbers in dictation mode, years will have to be disambiguated as "two thousand twenty three" (unidiomatic for me) or "numb twenty twenty three".

3. As @splondike suggests, specifically exclude 19xx & 20xx dates from the auto-time formatting. This might "do the right thing" the most often, but I would prefer consistency.

We can always change this later, so I'm inclined not to hold up on merging this, but now seems like a good time to start gathering feedback.

I prefer (1) and dislike (3), but not strongly, because I prefer consistency and avoiding ambiguity over doing the right thing automatically. Opinions?

rntz (changes requested):
Changing my review to "request changes" until we can figure out what's up with these reported bugs from @brief and @bjaspan . Feel free to override if bugs are found & fixed.

rntz:
I was able to replicate the issue on a7d67f5e with Talon 0.3.1-63-g48c8. 

In fact, it may be a Talon issue? "say one hundred and one" straight-up *fails to be recognized* (no entry in command history, no subtitle, nothing in log, events.tail() shows nothing except sometimes times for grammar compilation - indicating it's hearing that I said something, but rejecting it at some point). 

`mimic("one hundred and one")` has correct behavior, however.

rntz:
@lunixbochs it appears we have a reproducible issue with command recognition here. (see previous comment). want to take a look?

rntz:
I tested on Talon 0.3.1-63-g48c8. The rust beta is now out, so this needs testing on:
1. the most recent rust beta (as of writing, 0.3.1-546-gc4cd)
2. talon public

brief:
On Talon public here, so 0.3.1.

splondike:
Oops, forgot to test on public. Picking out 'say one thousand two hundred thirty seven' as the test case, I can reproduce the issue.

That is, on 0.3.1 (public with Conformer B) I get no output. On 0.3.1 (public with Conformer D, lol) I get '1237'. On 0.3.1-546-gc4cd (latest beta) with Conformer B or D I get 'one thousand two hundred thirty seven' (I guess an upgrade changed the behaviour here for me since yesterday).

bjaspan:
I've been absent from this PR  for a while but hopefully should be able to stay engaged again.

I've brought this PR  up to date with upstream.   I am now using  0.3.1-546-gc4cd with Conformer D. I can no longer reproduce the issue I originally reported on Feb 7:

*  When I say "numeral one thousand two hundred thirty seven"  I get "1237"
*  When I say "one thousand two hundred thirty seven"  I get "one thousand two hundred thirty seven"
*  kept just to confirm I'm using the correct code, when I say "twelve thirty seven" I get "12:37".

 I then tried saying each of these test cases that @brief  posted a few days ago:

* "one hundred and one" →  one hundred and one
* "one thousand two hundred" →  one thousand two hundred
* "one million twenty one" →  one million twenty one

In short,  my commands seem to be working and number word strings also seem to be working with 0.3.1-546-gc4cd with Conformer D.   I realize that does not help with the public version, but it does suggest that there is not something wrong with the code changes themselves.

Regarding @rntz's  comments from a couple days ago about times  versus years (e.g. "twenty twenty three"),  I honestly do not think it will be possible to construct  a high quality dictation system for things like this using static command grammars. It is going to require some kind of statistical/machine learning model to decide when "twenty twenty three" should be formatted as 20:23 or 2023. I currently have only the  barest understanding of how machine learning works but I suspect that in 2023, this level of functionality should be quite simple achieve;  after all,  Dragon NaturallySpeaking had it over 20 years ago. My hope would be that it could be implemented as a secondary  system after the recognition engine generates the sequence of words so that it does not require  waiting for an entirely new  recognition model.

core/text/text_and_dictation.py
```diff
@@ -35,6 +35,18 @@
...
 
+mod.list(
+    "currency_denomination", desc="Currency denominations that can be used within prose"
+)
+ctx.lists["user.currency_denomination"] = {
+    "dollar": "$",
+    "dollars": "$",
+    "euro": "€",
+    "euros": "€",
+    "pound": "£",
```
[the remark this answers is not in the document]
bjaspan: "I am going to pound in that nail"  seems to work.

 Of course, "I'd like to lose £15" is probably not  what most people saying that sentence mean. As I posted elsewhere, I do not think this problem is really solvable with a strict grammar. It is going to require machine learning.

rntz:
Yeah, since this is working on latest beta with conformer D, I'm even more inclined to think this is a talon issue. Unfortunate we can't merge it until it works on talon public. It would also be good if someone could test it on a Dragon+Talon setup, since there may be something about the complexity of the grammar here that's causing the issue?

I would really appreciate lunixboch (the Talon developer)'s opinion on this, I might try to ping him on Slack.

brief:
One to revisit now that Conformer D is officially out?

nriley:
Can anyone confirm that these issues are fixed with the current public version of Talon? On the community meetup we were not able to reproduce this either with conformer b or d.

core/text/text_and_dictation.py
```diff
@@ -36,6 +36,18 @@
     "frowny": ":-(",
 }
 
+mod.list(
+    "currency_denomination", desc="Currency denominations that can be used within prose"
+)
+ctx.lists["user.currency_denomination"] = {
```
knausj85, line 42: The default list should be implemented via a .talon-list these days

core/text/text_and_dictation.py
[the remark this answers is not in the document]
knausj85: This seems like an ambiguity pretty specific to those in the UK. For those in the US, I imagine most people would simply remove pound from the currency list :-p

knausj85:
I took a few minutes to bring this up to current standards, and reimplement the relevant lists using functionality from numbers. This allows us to define the numbers in one place. 

At the moment, I'm not aware of any reason we can't merge this. I'll leave it open for others to review however.

core/numbers/numbers.py
```diff
@@ -17,12 +17,47 @@
...
 
+def get_spoken_form_under_one_hundred(
+    start,
+    end,
+    include_oh_variant_for_single_digits,
+    include_default_variant_for_single_digits,
+):
+    """Helper function to get dictionary of spoken forms for non-negative numbers in the range [start, end] under 100"""
+
+    import math
```
AndreasArvidsson, line 35: Move to the top of the file

core/text/text_and_dictation.py
```diff
@@ -36,6 +45,10 @@
     "frowny": ":-(",
 }
 
+ctx.lists["user.hours_twelve"] = get_spoken_form_under_one_hundred(1, 12, True, True)
```
AndreasArvidsson, line 48: Use named arguments for the booleans

core/text/text_and_dictation.py
```diff
@@ -64,6 +77,56 @@ def prose_number(m) -> str:
...
+    s = m.number_string
+    if hasattr(m, "digit_string"):
+        s += "." + m.digit_string
+    return s + "%"
+
+
+@mod.capture(
+    rule="<user.number_string> {user.currency_denomination} [[and] <user.number_string> [cents|pence]]"
+)
+def prose_money(m) -> str:
```
AndreasArvidsson, line 93: Should we be consistent in using money or currency?

knausj85: I settled on currency, fixed

bjaspan:
Yippee!

Issue #1106, opened 2023-02-10
Add commands to easily add arguments to terminal commands


DonnerWolfBach:
The goal is to add commands which make it easier to add arguments to terminal commands, like "ls --all", "mkdir --help" etc as discussed with @2d6 on slack. 

Currently there are two approaches to be tried and discussed:
# 1. generic formatinng
E.g. `flag [<user.letter>]: "-{letter or ''}"` or `param [<user.text>]: " --{text or ''}"`.

# 2. curated list of options
Like arguments_git.csv a "bag-of-arguments approach" (see #982 ), but it needs to be discussed wether to put this in the settings folder (like unix_utilities) or somewhere tracked by git (like arguments_git.csv).

DonnerWolfBach:
@rntz I guess you are going to see this anyway, but since you did #982 I want to ensure you see this

DonnerWolfBach:
An idea - that is maybe to big of an refactor for this issue - splitting up settings in a part that is tracked by git and one that is not (where private stuff like specific system_paths go).

DonnerWolfBach:
#1107 contains my current draft for that, try out:
- "core makedir help"
- "core makedir"
- "option help"
- "param help"
- "param absolutely everything"
- "flag air"
- "flag shift air"

2d6:
> Another idea - that is maybe to big of an refactor for this issue - splitting up settings in a part that is tracked by git and one that is not (where private stuff like specific system_paths go).

After our discussion, I also thought about this. I think this would combine the best of both worlds in that it supports both user-specific as well as system-specific configuration. For me, the amount of refactoring isn't the issue here. My main reservation would be that this structure is going to be less clear to new users of knausj. It also breaks with other features implemented previously. None of those arguments are a deal breaker to me, but it would be good to have some input from one of the maintainers.

I'm going to take a look at the merge request now :)

PR #1107, opened 2023-02-10
1106 unix argument commands

DonnerWolfBach:
a suggestion for #1106


apps/windows_terminal/windows_terminal.talon
```diff
@@ -16,6 +16,9 @@ tag(): user.tabs
 # TODO: file_manager
 tag(): user.splits
 
+# Uncomment the following line to enable common unix utilities from unix_utilities.py
+tag(): user.unix_utilities
```
2d6, line 20: I'm not sure whether this should be enabled by default. In my initial pull request, we decided on having this commented in this source code by default. We should stay consistent here, IMO.

tags/terminal/unix_utilities.py
```diff
@@ -83,3 +85,22 @@
 
 mod.list("unix_utility", desc="A common utility command")
 ctx.lists["self.unix_utility"] = unix_utilities
+
+# 2. arguments
+
+default_unix_arguments = {"--help": "help"}
```
2d6, line 91: This needs to be swapped. 
```suggestion
default_unix_arguments = {"help": "--help"}
```
Unintuitively, `get_list_from_csv` swaps the order of the columns.

2d6: I took all the commands currently within my `unix_utilities.csv` and looked at the output of their `--help` option:
```
cat unix_utilities.csv | grep  -v command | cut -f1 -d, | cut -f1 -d' ' | sort -u | xargs -I% bash -c "% --help || true | sort -u" | grep -oE -- '--[a-zA-Z_-]+' | sort | uniq -c | sort -n
```
I think this is going to give a good starting point of often-used parameters for the initial `default_unix_arguments`.
For example:
```
--all,all
--auth,auth
--backup,backup
--binary,binary
--bytes,bytes
--check,check
--context,context
--debug,debug
--file,file
--files,files
--filter,filter
--follow,follow
--force,force
--form,form
--help,help
--ignore-case,ignore-case
--include,include
--json,json
--lines,lines
--null,null
--output,output
--pretty,pretty
--print,print
--quiet,quiet
--raw,raw
--recursive,recursive
--relative,relative
--reverse,reverse
--silent,silent
--tag,tag
--terse,terse
--text,text
--timeout,timeout
--times,times
--time,time
--user,user
--verbose,verbose
--verify,verify
--version,version
--zero-terminated,zero-terminated
--zero,zero
```

2d6 (commented):
I think that this pull request is potentially a great addition to the unix utilities. I tried this out and think it works pretty well. There are just two things I am somewhat worried about:

* How would we deal with commands which do not use the convention of using two dashes for long form parameters? Also, some commands to have subcommands. Example (which is already covered by `knausj_talon`, incidentally): `terraform init -upgrade`. Do you have an idea how we could improve this generically?
* Having Talon listen for trailing options might impede fluent dictation of commands if there is ambiguity between the options defined in the csv file and the directory. Contrived example, if you have would like to `echo force yourself to use the keyboard`, you could previously say `core echo force`. Now, this would result in `echo --force yourself to use the keyboard`, which is often not what you intended. I am not sure if this is really a problem in practice, though.

apps/windows_terminal/windows_terminal.talon
DonnerWolfBach: Good that you pointed that out. I'll fix it once this is final.

DonnerWolfBach:
>     * How would we deal with commands which do not use the convention of using two dashes for long form parameters? Also, some commands to have subcommands. Example (which is already covered by `knausj_talon`, incidentally): `terraform init -upgrade`. Do you have an idea how we could improve this generically?

Personally I would not put a huge emphasis on these because [they are kinda breaking POSIX conventions](https://softwareengineering.stackexchange.com/questions/70357/command-line-options-style-posix-or-what). However, since they are quite a lot of programs/commands which have these, I would guess we could add `single dash [<user.text>]: user.insert_formatted(" -{text or ''}", "DASH_SEPARATED")` for the dynamic generation.
If we want to support this for the curated list, I would say we would change how the list is done: instead of storing the argument with "--", we could change it to something like:
```
argument,spoken alias
all
ff-only, fast forward only
```
That way we could add one or two dashes based on the command and while we are at it making the spoken alias optional (since at least in the current file it is the same 90% of the time).

unix_utilities.csv does this optional thing anyway. Is this complex to do or has any disadvantages?

DonnerWolfBach:
>     * Having Talon listen for trailing options might impede fluent dictation of commands if there is ambiguity between the options defined in the csv file and the directory. Contrived example, if you have would like to `echo force yourself to use the keyboard`, you could previously say `core echo force`. Now, this would result in `echo --force yourself to use the keyboard`, which is often not what you intended. I am not sure if this is really a problem in practice, though.

Spontaneously these two options to handle this come to my mind:
1. add an optional "over" at the end of the command like in say or Sentence: `core {user.unix_utility} [<user.unix_arguments>] [over]:`
2. Ignore it and tell people to define an explizit echo command for that like it is done with ls (or will be done in #1020)

DonnerWolfBach:
About the subcommands: I'll do that like git_commands.csv:
```
(...)
remote set-url, remote set url
(...)
worktree move
(...)
```

DonnerWolfBach:
Made this PR a draft again because I got new feedback on how to improve things.

nriley:
@DonnerWolfBach apologies we did not get a chance to look at this at the community backlog session for… nearly two years. We are actually getting caught up and wondering if you would like to take another whack at this, potentially using a Talon list file? Thanks!

DonnerWolfBach:
Hi @nriley,
Unfortunately my new semester just started, so I likely  will not have time until summer.

PR #1108, opened 2023-02-10
List path

DonnerWolfBach:
Addresses #1020 with windows fixes

### DIFF
--- a/apps/iterm/iterm.py
+++ b/apps/iterm/iterm.py
@@ -54,12 +54,12 @@ class user_actions:
     #     """selects the file"""
     #     actions.insert(path)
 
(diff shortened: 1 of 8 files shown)


splondike (commented):
I think you're addressing #1051 not #1020 with this PR.

I made a comment in the former to gsinghe93 about working out how to merge the changes in both into a single system since the implementations conflicted. I'm not sure they are an active user of Talon anymore though, so I'm happy if we want to abandon #1020 and just proceed with this change to allow paths.

DonnerWolfBach:
See https://github.com/knausj85/knausj_talon/pull/1051#issuecomment-1437255863

PR #1109, opened 2023-02-10
additional messaging and slack commands

ghouston:
* Messaging:
  * add `channel previous`
  * add `channel next`

* Slack:
  * make the toggle commands shorter by not requiring the ending `(sidebar | bar)`
  * add `bar switch [left]` and `bar switch right` alternative commands for toggling (similar to vscode sidebar toggling)


splondike:
I personally prefer the pattern of having a single command to accomplish something rather than having alternatives. The main reason is reducing misrecognition chance by binding fewer commands. For alternative prefixes it also 'uses up' prefixes that may want to be used for other purposes in the future.

So if it was me I'd pick a single command for each action, choosing the one that was most consistent with other commands in knausj_talon. I'd mark the other as deprecated using the [deprecation system](https://github.com/knausj85/knausj_talon/blob/main/core/deprecations.py).

I'm not sure what other maintainers think though, so I suggest we leave this comment open for a week or two to see if anybody else has thoughts.

knausj85:
> I personally prefer the pattern of having a single command to accomplish something rather than having alternatives. The main reason is reducing misrecognition chance by binding fewer commands. For alternative prefixes it also 'uses up' prefixes that may want to be used for other purposes in the future.
> 
> So if it was me I'd pick a single command for each action, choosing the one that was most consistent with other commands in knausj_talon. I'd mark the other as deprecated using the [deprecation system](https://github.com/knausj85/knausj_talon/blob/main/core/deprecations.py).
> 
> I'm not sure what other maintainers think though, so I suggest we leave this comment open for a week or two to see if anybody else has thoughts.

I tend to agree with this philosophy. I've hesitated to add variants without a really good reason (i.e. beyond personal preference/needs), as it can really cause other people headaches. I think adding additional alternatives should be _mostly_ left to people's forks, for better or worse

fidgetingbits:
I agree. I believe we (knausj85 maintainers) had agreed on this a long time ago.

splondike:
OK, I'm going to close this out based on the above.

PR #1110, opened 2023-02-10
add google_meet.talon

ghouston:
adds support for using Google Meet as a Progressive Web App

### DIFF
--- a/apps/google_meet/google_meet.talon
+++ b/apps/google_meet/google_meet.talon
@@ -1,0 +1,17 @@
+app: Google Meet
+# note: unable to use mac bundle id since its randomly generated when installing app
+# for example com.google.Chrome.app.kjgfgldnnfoeklkmfkjfagphfepbbdan
+-
+
+# Google Meet is a Desktop version of Meet.
+# Its a "Progressice Web App" which is installable on the desktop.
+# Installation instructions are found at https://support.google.com/meet/answer/10708569?hl=en
+
+tag(): browser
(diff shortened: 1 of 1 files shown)


apps/google_meet/google_meet.talon
```diff
@@ -0,0 +1,17 @@
+app: Google Meet
```
splondike, line 1: You probably want `app.name:` instead of `app:` here.

apps/google_meet/google_meet.talon
```diff
@@ -0,0 +1,17 @@
...
+# note: unable to use mac bundle id since its randomly generated when installing app
+# for example com.google.Chrome.app.kjgfgldnnfoeklkmfkjfagphfepbbdan
+-
+
+# Google Meet is a Desktop version of Meet.
+# Its a "Progressice Web App" which is installable on the desktop.
+# Installation instructions are found at https://support.google.com/meet/answer/10708569?hl=en
+
+tag(): browser
```
splondike, line 10: What fraction of the commands in https://github.com/knausj85/knausj_talon/blob/main/tags/browser/browser.talon would work for this PWA?

If it's your typical chromeless type situation (i.e. no URL bar), then I'm guessing you probably don't get anything by enabling this tag.

splondike:
I might close this out around the 25th given we haven't heard back from the author.

splondike:
Closing this out.

PR #1111, opened 2023-02-10
Add Google Chrome Canary support on Mac

ghouston:
Adds `app: com.google.Chrome.canary`

### DIFF
--- a/apps/chrome/chrome_mac.py
+++ b/apps/chrome/chrome_mac.py
@@ -5,6 +5,7 @@ ctx = Context()
 ctx.matches = r"""
 os: mac
 app: chrome
+app: com.google.Chrome.canary
 """
 ctx.tags = ["browser", "user.tabs"]
 


apps/chrome/chrome_mac.py
```diff
@@ -5,6 +5,7 @@
 ctx.matches = r"""
 os: mac
 app: chrome
+app: com.google.Chrome.canary
```
splondike, line 8: Should be `app.bundle: ` instead of `app:`. See also comment on #1113 .

splondike:
@ghouston Still interested in merging this in?

splondike:
I might close this out around the 25th given we haven't heard back from the author.

splondike (approved):
Actually, since the existing matcher is using `app:`, it's OK using it for this one for now also I guess. Merging this in.

splondike:
But there are merge conflicts... So I'm just going to close it, easy enough to add back in if somebody wants to.

PR #1115, opened 2023-02-12
Add additional terminal path traversal commands

2d6:
This pull request is intended to enhance directory traversal in terminals. It adds the capability to:

* Traverse not only to the parent directory (`katie up`), but an arbitrary number of directories upwards by appending an ordinal: `katie up third` results in `cd ../../..` in POSIX shells. This behavior differs a little from the usual use of ordinal suffixes. The commands in `repeater.talon` usually repeat the last command N times, but in the context of changing directories in a terminal, this repetition is more sensibly applied to the directory traversal itself. I've been using this extensively on my machines and found it pretty ergonomic to use.
* Toggle-traverse between the most recent directories (`cd -`)
* Navigate back to previous directories, similar to a browser back button. This is not possible in POSIX shells, but the shells usually used in developer terminals like `bash` or `zsh` provide this capability. This is why I introduced an additional tag.

I don't have any experience with windows terminals and no access to OSX. The latter should default to `zsh` shells, so we should be good there. I could use some contributions regarding implementations for other shells/OSes.


DonnerWolfBach:
Wouldn't `katie up third` have the same effect without your modification? (It would result in `cd ..\n cd ..\n cd ..\n`)

2d6:
@DonnerWolfBach : You know what, in vanilla knausj_talon, you are perfectly right. I didn't even realize that. In my local setup, I don't have `key(enter)` in the implementation. In this case, you can't really repeat the command with ordinals. While my implementation here still leads to a cleaner bash command history, its value is definitely debatable in vanilla. Would you say that this pull request still has merit this particular command is removed? So only `cd -` and `popd`?

DonnerWolfBach:
Yeah, I think having `cd -` and `popd` in knausj still makes sence. We can even copy and paste the code for powershell since it's (rougly) the same there: https://superuser.com/questions/593987/change-directory-to-previous-directory-in-powershell

Unfortunately, there is no equivalent in windows command prompt to `cd -` - but that's not a big deal, while not officially deprecated it kinda is.

PR #1116, opened 2023-02-13
Prefer tmux splits and tabs over containing terminal

EdJoJob:
Done by having the context for matching tmux be a more restrictive match than that of the terminals.

This was modeled after the matches for the in-browser Slack and Teams apps.

### DIFF
--- a/apps/tmux/tmux.py
+++ b/apps/tmux/tmux.py
@@ -1,10 +1,10 @@
 from talon import Context, Module, actions
 
-ctx = Context()
 mod = Module()
 
-ctx.matches = r"""
-tag: user.tmux
+mod.apps.tmux = """
+tag: terminal
+and tag: user.tmux
 """
 
 setting_tmux_prefix_key = mod.setting(
@@ -43,6 +43,10 @@ class TmuxActions:
             f'confirm-before -p "{confirmation_prompt} (y/n)" {command}'
         )
         actions.key("\n")
+
+
+ctx = Context()
+ctx.matches = "app: tmux"
 
 
 @ctx.action_class("app")
--- a/apps/tmux/tmux.talon
+++ b/apps/tmux/tmux.talon
@@ -1,6 +1,17 @@
-tag: user.tmux
+app: tmux
 -
 tag(): user.splits
+tag(): user.tabs
+
+# Note that you will need to add something to match the tmux app in your configuration
+# This is not active by default
+# Adding a file with a matcher for detecting tmux active in your terminal and activating
+# the tmux tag is required
+# Something like:
+#
+# title: /^tmux/
+# -
+# tag(): user.tmux
 
 # pane management - these commands use the word split to match with the splits
 # tag defined in tags/splits/splits.talon


apps/tmux/tmux.py
```diff
@@ -3,8 +3,17 @@
 ctx = Context()
 mod = Module()
 
+apps = mod.apps
+
+apps.tmux = "app.name: tmux"
+
+mod.apps.tmux = """
+tag: terminal
+and win.title: /tmux/
```
splondike, line 12: I don't get tmux in the title when I run it on my machine. I'm also not sure the app.name thing will work because Talon will probably just pick up the name of the terminal.

Probably just want to do:
```
mod.apps.tmux = """
tag: terminal
and tag: user.tmux
"""

**Review by splondike — CHANGES_REQUESTED** [review]
Don't think this will work in general as implemented.

**Comment by splondike** [comment]
@EdJoJob Are you still interested in updating this and merging it in?

**Comment by EdJoJob** [comment]
I'll work on getting it fixed.  Been off GH for a while. Thanks for the patience.

File: apps/tmux/tmux.py (hunk 1 above)
Comment (EdJoJob) [reply apps/tmux/tmux.py:12]: I have the below in my `tmux` configuration which allowed me to see it in my window title.

```tmux.conf
# Window manager title settings
set-option -g set-titles on
set-option -g set-titles-string "tmux - #{session_name} / #{window_name}: #{pane_title}"
```

apps/tmux/tmux.py
```diff
@@ -3,8 +3,17 @@
 ctx = Context()
 mod = Module()
 
+apps = mod.apps
+
+apps.tmux = "app.name: tmux"
```
splondike, line 8: This line doesn't work for me using your tmux config on Linux (I do get the updated window title, but app.name doesn't match it). I'd actually be inclined to just leave it out completely since it only works if you've got Tmux set up a particular way.

I reckon the best way of enabling this is just for people to drop a file like this somewhere in their Talon user files:

`tmux_enabler.talon`
```
title: /^tmux/
-
tag(): user.tmux
```

EdJoJob: What would be the best way to communicate that? A comment or doc-block in this file, or an extension of the README? or something else?

splondike: I reckon just put a short comment up near the top of the tmux.talon file to say the tag won't be enabled by default and suggesting how you might do it.

splondike:
Looks good. I reckon just remove the `apps.tmux = "app.name: tmux"` code block on line 8 of tmux.py since I don't think it's doing anything.

splondike:
Not sure if you intended to push a fix for my previous comment in the last force push, but it hasn't changed the thing I was pointing to. (Just want to make sure we're not both waiting on the other person!).

EdJoJob:
I'm in the habit of force-pushing a rebase, and then force-pushing the new changes.  Clearly this time I just missed doing the second push.  Should be up-to-date now

splondike:
One more thing I noticed while testing, the context used by the tab commands doesn't have any matchers on it, so it's not very specific. You probably want to add in a line like this so that Tmux's tab implementation overrides the default one:

```python
ctx.matches = "app: tmux"
```

splondike (approved):
Looks good, thanks for your persistence!

PR #1118, opened 2023-02-18
Detect Chromium as a browser under Manjaro Linux

2d6:
On my Manjaro Linux machine, the browser Chromium appears with a executable name which is not yet recognized by `chrome.py`. This pull request makes chromium detect correctly for me.

As a side note: since multiple tags are connected by `or`, shouldn't the entire context match block for Linux be something like
```
os: linux
and app.exe: chrome
os: linux
and app.exe: chromium-browser
os: linux
and app.exe: chromium
```
Also, IMO the subsequent block could also be included (lines 25-28) in this, thus joining all `os: linux` matchers. 

If you agree, I'll gladly amend this in this PR.

### DIFF
--- a/apps/chrome/chrome.py
+++ b/apps/chrome/chrome.py
@@ -20,6 +20,7 @@ mod.apps.chrome = """
 os: linux
 app.exe: chrome
 app.exe: chromium-browser
+app.exe: chromium
 """
 mod.apps.chrome = """
 os: linux


splondike:
Yeah, you could certainly join all the linux blocks together as:
```
os: linux
app.exe: chrome
app.exe: chromium-browser
app.exe: chromium
```

That will be interpreted as `(is_linux and (exe_chrome or exe_chromium or exe_chromium-browser)`.

rntz (approved):

PR #1121, opened 2023-02-20
add numb as number prefix for prose formatting

DonnerWolfBach:
make it consistent with the formatter_immune capture in formatters.py, which has both numb and numeral

### DIFF
--- a/core/text/text_and_dictation.py
+++ b/core/text/text_and_dictation.py
@@ -41,17 +41,17 @@ def prose_modifier(m) -> Callable:
     return getattr(DictationFormat, m.prose_modifiers)
 
 
-@mod.capture(rule="numeral <user.number_string>")
+@mod.capture(rule="(numb | numeral) <user.number_string>")
 def prose_simple_number(m) -> str:
     return m.number_string
 
 
-@mod.capture(rule="numeral <user.number_string> (dot | point) <digit_string>")
+@mod.capture(rule="(numb | numeral) <user.number_string> (dot | point) <digit_string>")
 def prose_number_with_dot(m) -> str:
     return m.number_string + "." + m.digit_string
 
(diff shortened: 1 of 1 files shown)


rntz:
We deliberately didn't do this when adding this feature to avoid misrecognitions: https://github.com/knausj85/knausj_talon/pull/737#issuecomment-1013947323

eg. "he was numb to the pain" --> "he was 2 the pain"

I'm not super firm on this, if you want to argue that it's worth it or that the misrecognitions won't be super common, go ahead.

DonnerWolfBach:
I see. Then I don't see why the formatters allow numb - of course it is less likely to have "numb" in a variable name I guess, but it still feels unclean and possibly confusing to me. 
Like that "scratch that" and "nope that" both work for prose/text.talon, but only the first works in dictation mode. 

But I can see if these little inconsitences are not such a big of a deal, but I feel like they are making it more difficult than it has to be.

splondike:
I'd be inclined to make this change (so numb two/to/too -> 2), I reckon the sub-phrase is uncommon enough that the benefits of consistency outweigh the potential costs.

I just tried this branch with Conformer D (beta), and Conformer B (public) and both of them actually continue to output 'He has become numb to it' (and some other variants) for some reason. For other numbers it outputs the digits as you'd want.

So empirically it sounds like this isn't an issue as well (at least with my accent).

DonnerWolfBach:
Ok, I guess if somebody has issues on a regular bases with that they can still remove it, but adding it is imho always more difficult because you first have to know about it.

splondike:
I'm going to leave this open for another week to see if anybody else wants to weigh in, and then plan to merge it.

knausj85 (approved):

PR #1123, opened 2023-02-27
Add tab search command similar to firefox.talon

Spiteless:
Tab search command via ctrl/cmd + shift + A

There's already an implementation for firefox that uses the address bar, this pull request mimics that for Chrome

### DIFF
--- a/apps/chrome/chrome.talon
+++ b/apps/chrome/chrome.talon
@@ -1,9 +1,7 @@
 app: chrome
 -
-tag(): browser
 profile switch: user.chrome_mod("shift-m")
 
-tag(): user.tabs
 tab search: user.chrome_mod("shift-a")
 
 tab search <user.text>$:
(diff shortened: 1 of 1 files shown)


apps/chrome/chrome.talon
```diff
@@ -1 +1,13 @@
+app: chrome
+-
+tag(): browser
 profile switch: user.chrome_mod("shift-m")
+
+tag(): user.tabs
```
rntz, line 6: I don't think we need this or `tag(): browser`, that should be handled by eg. `chrome_linux.py`, `chrome_win.py`, etc.

rntz (changes requested):

rntz (approved):

apps/chrome/chrome.talon
Spiteless: Thanks for your help!

PR #1127, opened 2023-03-02
on_pop requires mouse control to be active

jmegner:
Before this fix, pop to click would work even if you were not controlling the mouse.  I don't understand why `not actions.tracking.control_zoom_enabled()` was there before, so it just kept it.

### DIFF
--- a/plugin/mouse/mouse.py
+++ b/plugin/mouse/mouse.py
@@ -250,7 +250,10 @@ def show_cursor_helper(show):
 def on_pop(active):
     if setting_mouse_enable_pop_stops_scroll.get() >= 1 and (gaze_job or scroll_job):
         stop_scroll()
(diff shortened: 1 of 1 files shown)


splondike:
This looks to be a duplicate of #1119 . In that one I also include the legacy control mouse as a possible way you could be doing mouse control. Suggest we close this one out.

splondike:
Although, if you have an eye tracker to test with and implement #1119 in this PR we can actually get the fixed merged. I'm a maintainer, but I can't merge my own PR :/.

knausj85:
I've tested and merged #1119, which accounts for both control mouse modes. Closing this as duplicate-thanks!

The zoom mouse check is there to prevent double clicking when it is enabled, for reference

PR #1134, opened 2023-03-04
added "nope that" etc. to dictation mode

DonnerWolfBach:
Making it consistent with text.talon, where "nope that" is an alternative for "scratch that".


DonnerWolfBach:
Not sure wether we should go with `nope that | scratch that:` or `(nope| scratch) that:`, but we should keep it the same for text and dictation

splondike (approved):
Looks good, thanks! I find 'nope that | scratch that' marginally more clear, but it doesn't matter too much I think.

PR #1138, opened 2023-03-11
add 'proud' command for inserting capitalized word

jmegner:
I have found this command very useful even though we already have 'title' command.  Just, like how the 'word' command is useful in addition to the 'say' command.

### DIFF
--- a/core/text/text.talon
+++ b/core/text/text.talon
@@ -13,7 +13,7 @@ phrase <user.text> over:
 word <user.word>:
     user.add_phrase_to_history(word)
     insert(word)
-grand <user.word>: user.insert_formatted(word, "CAPITALIZE_FIRST_WORD")
+proud <user.word>: user.insert_formatted(word, "CAPITALIZE_FIRST_WORD")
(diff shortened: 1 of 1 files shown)


pokey:
Yeah super useful. I use "proud" fwiw. We may start using "grand" in cursorless so might be marginally better to steer clear of it, all else being equal. See https://github.com/cursorless-dev/cursorless/issues/124

jmegner:
@pokey awesome, thanks, I have made the change from grand to proud.

knausj85 (dismissed):
I believe we'll still want to call
user.add_phrase_to_history(word)
as in the command above for improved undo support

otherwise, looks great

rntz:
It appears that `user.insert_formatted` actually adds the formatted phrase to history already, so I don't think we need that.

rntz:
Arguably, we shouldn't be doing that; it's inconsistent with the request from aegis to avoid an `insert_with_history` command. But that's out of scope for this PR, I think.

rntz (approved):

PR #1139, opened 2023-03-12
feat(#755) Pop to click setting for all users

splondike:
This PR addresses #755 and replaces #858 . It does two things:

1. The pop noise triggers an action `user.noise_trigger_pop`.
2. The existing setting `user.mouse_enable_pop_click` gets a third value '2' which says always pop to click.

The first point lets people 'take over' the pop noise so they don't have to worry about whatever knausj talon is doing with it. It also provides a nice scaffold in case you want it to do different things in different contexts.

The second point differs a bit from the pop_click = yes/no boolean discussed in #755 . This is for backwards compatibility. With a yes/no setting I'd either have to turn it on by default for people without eye trackers (i.e. undo https://github.com/knausj85/knausj_talon/pull/1119) or turn it off and break the expected behaviour for most eyetracker users. If I figured out a way to conditionally enable the setting when an eye tracker was on (not trivial given how you have to check for this) then this would break for eyetracker users who had turned the setting off because they didn't want pop to click.

Hopefully the addition of a new value is simple enough for people to use if they want pop to click in a non-eyetracker context.

### DIFF
--- a/plugin/mouse/mouse.py
+++ b/plugin/mouse/mouse.py
@@ -50,7 +50,7 @@ setting_mouse_enable_pop_click = mod.setting(
     "mouse_enable_pop_click",
     type=int,
     default=0,
-    desc="Pop noise clicks left mouse button. 0 = off, 1 = on with eyetracker but not zoom mode, 2 = on",
+    desc="Pop noise clicks left mouse button. 0 = off, 1 = on with eyetracker but not with zoom mouse mode, 2 = on but not with zoom mouse mode",
 )
 setting_mouse_enable_pop_stops_scroll = mod.setting(
     "mouse_enable_pop_stops_scroll",
--- a/plugin/mouse/mouse.py
+++ b/plugin/mouse/mouse.py
@@ -1,6 +1,6 @@
 import os
 
-from talon import Module, actions, app, clip, cron, ctrl, imgui, noise, ui
+from talon import Module, Context, actions, app, clip, cron, ctrl, imgui, ui
 from talon_plugins import eye_zoom_mouse
 
 key = actions.key
@@ -38,6 +38,8 @@ hidden_cursor = os.path.join(
 )
 
 mod = Module()
+ctx = Context()
+
 mod.list(
     "mouse_button", desc="List of mouse button words to mouse_click index parameter"
 )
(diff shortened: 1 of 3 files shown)


plugin/mouse/mouse.py
```diff
@@ -247,24 +249,29 @@ def show_cursor_helper(show):
...
+    else:
+        # Otherwise respect the mouse_enable_pop_click setting
+        setting_val = setting_mouse_enable_pop_click.get()
+
+        is_using_eye_tracker = (
+            actions.tracking.control_zoom_enabled()
+            or actions.tracking.control_enabled()
+            or actions.tracking.control1_enabled()
+        )
+        should_click = setting_val == 2 or (
```
knausj85, line 266: I think mode two (2) should still exclude the zoom mouse, as it will cause nonobvious problems like double clicking or dismissing the zoom mouse. maybe add a warning?

otherwise, looks good to me

knausj85 (changes requested):

settings.talon
```diff
@@ -18,8 +18,11 @@ settings():
     # Stop continuous scroll/gaze scroll with a pop
     user.mouse_enable_pop_stops_scroll = 1
 
-    # Enable pop click with 'control mouse' mode
-    user.mouse_enable_pop_click = 1
+    # Enable pop click with 'control mouse' mode.
+    # 0 = off
+    # 1 = on with eyetracker but not zoom mode
+    # 2 = on
+    user.mouse_enable_pop_click = 2
```
splondike, line 25: Oops, I should change this this value to 1.

plugin/mouse/mouse.py
splondike: I've not used an eye tracker. Are they something like this?

* Control mouse: just move cursor with eyes (and head).
* Zoom mouse: move cursor with eyes (and head). Pop to zoom in, then pop to click while zoomed in. After clicking return to unzoomed state (like https://www.youtube.com/watch?v=VMNsU7rrjRI)

For the zoom mouse case, shouldn't people be using this setting with value = 1 (or indeed 0)? I'm thinking option 2 is mostly for people who want to position the cursor with the mouse or something like that, but don't want to click with their finger.

knausj85: Yeah, those descriptions are accurate. 

The issue is zoom mouse registers its own pop handler that the user doesn't currently control.

As such, I would personally vote to err on the side of making it hard to misconfigure.

splondike: Fair enough, I'll add an exclusion for zoom mouse to option 2.

splondike:
I've addressed the code review comments, ready for another look.

knausj85 (approved):

Issue #945, opened 2022-07-31
replacements for system_command and system_command_nb


lunixbochs:
Right now `system_command` and `system_command_nb` use `os.system` and `subprocess.Popen(shell=True)` respectively.

These both can have some undesirable behavior:
- `os.system` and `Popen(shell=True)` launch an OS-native shell, which will behave differently between windows and posix, and it's difficult to safely escape shell input.
- `os.system` leaks Talon's file handles to the child, which may cause unexpected behavior in both Talon and the launched app.
- `Popen(shell=True)` may close the launched process when Talon itself closes, which is likely undesirable. This behavior also depends on your platform.
- `Popen(shell=True)` can leak resources inside Talon with each subprocess launched.

I propose improving these in a couple of ways:
- Use `shlex.split` instead of `shell=True`. `shlex` will behave the same across platforms, and doesn't do shell expansion of variables and subcommands, so is generally safer.
- Use `subprocess.run(check=True)` instead of `os.system()`. `subprocess.run` doesn't leak Talon's file handles, doesn't invoke an extra `/bin/sh` process, and `check=True` means you'll get an exception if the subprocess fails, which is usually a good thing.
- Use talon's `system.launch` instead of `subprocess.Popen(shell=True)`. This spawns the process separately from Talon, so it won't close when Talon exits, and won't leak resources.

```python
from talon import Module, system
import shlex
import subprocess

mod = Module()
@mod.action_class
class Actions:
    def system_command(cmd: str) -> None:
        """Run a command and wait for it to finish"""
        args = shlex.split(cmd)
        subprocess.run(args, check=True)

    def system_command_nb(cmd: str) -> None:
        """Run a command without waiting for it"""
        args = shlex.split(cmd)
        cmd, args = args[0], args[1:]
        system.launch(path=cmd, args=args)
```

splondike:
Had a go at this on the branch mentioned above. The implementation is the same as lunixbochs' with an attempt to allow `~` to expand to the user's home directory. There are some issues with it though relative to the current implementation:

* `shlex.split` doesn't work properly with Windows it seems. `system_command("WScript C:\\Users\\Administrator\\Desktop\\test.vbs hello")` came out of `shlex.split` as something like `["WScript", "C:UsersAdministratorDesktoptest.vbs", "hello"]`.
[...]

PR #1140, opened 2023-03-13
feat(#945): Reimplement system_command(_nb) actions

splondike:
Addresses the issues with the current implementation mentioned by lunixbochs in #945 and will additionally fix the following:

* system_command now throws an error when the command exits unsuccessfully.
* system_command_nb now leaves the child process running when Talon exits.

I've tested this implementation on Linux, OSX, and Windows against variants of the test cases listed here: https://github.com/knausj85/knausj_talon/issues/945#issuecomment-1464885283 and it works OK.

# Technical notes

shlex.split doesn't do what you want with Windows' backslash based file paths: "C:\\Windows\\Path" gets converted to "C:WindowsPath". The lexer I implemented here lets you escape quotation characters, but backslash plus a non quote character comes out unchanged. This lexer doesn't support inline comments or newlines.


splondike:
@fidgetingbits FYI, if this gets merged your use of shell constructs will no longer work (e.g. https://github.com/fidgetingbits/knausj_talon/blob/master/apps/i3wm/i3wm.talon#L28).

You could fix that line by wrapping the whole command in a bash shell like this:
```sh
    user.system_command("bash -c \"/home/aa/scripts/i3/i3-nth_window_in_workspace.py $(i3-msg -t get_workspaces | jq -r '.[] | select(.focused==true).name') {number_small-1}\"")
```

fidgetingbits:
> @fidgetingbits FYI, if this gets merged your use of shell constructs will no longer work (e.g. https://github.com/fidgetingbits/knausj_talon/blob/master/apps/i3wm/i3wm.talon#L28).
> 
> You could fix that line by wrapping the whole command in a bash shell like this:
> 
> ```shell
>     user.system_command("bash -c \"/home/aa/scripts/i3/i3-nth_window_in_workspace.py $(i3-msg -t get_workspaces | jq -r '.[] | select(.focused==true).name') {number_small-1}\"")
> ```

Thanks for the heads up

rntz:
I haven't looked at this deeply but I would really rather we not have to maintain a custom lexer, is there an alternative?

splondike:
I've put my thoughts on how to proceed in #945 and have marked this as a draft until we work that out.

splondike:
Sounds like we won't be merging this in any time soon. I'm going to close this out in order to stop cluttering the PR list. It's linked in the issue so we can recover the code pretty easily if we want to go this way.

PR #1141, opened 2023-03-13
feat: Allow abbreviation customisation via settings/abbreviations.csv

splondike:
This PR implements the results of the discussion in #1048 (and replaces that PR); it seems worthwhile to have abbreviations in the settings/*.csv system.

### DIFF
--- a/core/abbreviate/abbreviate.py
+++ b/core/abbreviate/abbreviate.py
@@ -250,14 +250,12 @@ abbreviations = {
     "work in progress": "wip",
 }
 
-
-def get_abbreviations_list():
-    return get_list_from_csv(
-        "abbreviations.csv",
-        headers=("Abbreviation", "Spoken Form"),
-        default=abbreviations,
-    )
-
+# This variable is also considered exported for the create_spoken_forms module
+abbreviations_list = get_list_from_csv(
+    "abbreviations.csv",
+    headers=("Abbreviation", "Spoken Form"),
+    default=abbreviations,
+)
 
 ctx = Context()
-ctx.lists["user.abbreviation"] = get_abbreviations_list()
+ctx.lists["user.abbreviation"] = abbreviations_list
--- a/core/abbreviate/abbreviate.py
+++ b/core/abbreviate/abbreviate.py
@@ -1,12 +1,11 @@
 # XXX - would be nice to be able pipe these through formatters
 
 from talon import Context, Module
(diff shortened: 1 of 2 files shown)


core/abbreviate/abbreviate.py
```diff
@@ -2,11 +2,11 @@
...
 from talon import Context, Module
 
+from ..user_settings import get_list_from_csv
+
 mod = Module()
 mod.list("abbreviation", desc="Common abbreviation")
 
-# TODO: Make this a csv file. Not necessarily a settings/ csv file, it might be
-# better to be like homophones.csv.
 abbreviations = {
```
knausj85, line 10: since create spoken forms imports this dictionary, I think this might cause problems as-is
https://github.com/knausj85/knausj_talon/blob/b11b773a46f4f299ddcce4f9f84a566e78d0bb14/core/create_spoken_forms.py#L9
namely, don't think it would be in sync with the users file as written?

splondike: Thanks for the pickup! I think I'll expose a function from abbreviate.py that gives you the list and use that in create_spoken_forms.py.

splondike:
I think the usage in create_spoken_forms might just be around abbreviating numbers, but in any case I've fixed the integration.

core/create_spoken_forms.py
```diff
@@ -39,7 +39,7 @@
 )
 
 REVERSE_PRONUNCIATION_MAP = {
-    **{value: key for key, value in abbreviations.items()},
+    **{value: key for key, value in get_abbreviations_list().items()},
```
knausj85, line 42: won't this cause us to parse the abbreviations csv twice, both on initial load and when it's updated by the user? any reason not just to return ctx.lists["user.abbreviation"] for use externally?

I'm not entirely sure it matters, but figured I'd bring it up.

knausj85 (changes requested):

knausj85:
> I think the usage in create_spoken_forms might just be around abbreviating numbers, but in any case I've fixed the integration.

Just FYI, the main use case: applying abbreviations when navigating in file managers. So e.g. apps -> applications and such. This should perhaps be a setting in the future, but I don't think it's feasible as-is

splondike:
I've addressed the above feedback.

knausj85 (approved):

PR #1143, opened 2023-03-16
Fix issue with alt-d not always focusing address bar in Firefox Windows

david-tejada:
If the user uses German menus that shortcut doesn't work and opens the file "Datei" menu instead.

Pressing F6 before pressing ctrl-l also ensures that the command will work if the user is focused in the devtools (instead of clearing the console).

### DIFF
--- a/apps/firefox/firefox.py
+++ b/apps/firefox/firefox.py
@@ -18,24 +18,21 @@ and app.bundle: org.mozilla.firefox
 """
 
 ctx.matches = r"""
+os: windows
+os: linux
+os: mac
+tag: browser
 app: firefox
 """
-
-cmd_ctrl = "cmd" if app.platform == "mac" else "ctrl"
 
 
 @mod.action_class
 class Actions:
     def firefox_bookmarks_sidebar():
         """Toggles the Firefox bookmark sidebar"""
-        actions.key(f"{cmd_ctrl}-b")
 
     def firefox_history_sidebar():
         """Toggles the Firefox history sidebar"""
-        if app.platform == "mac":
-            actions.key("cmd-shift-h")
-        else:
-            actions.key("ctrl-h")
 
 
 @ctx.action_class("user")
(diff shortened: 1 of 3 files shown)


apps/firefox/firefox.py
```diff
@@ -50,6 +50,14 @@ class BrowserActions:
...
         actions.key(f"{cmd_ctrl}-shift-o")
 
+    def focus_address():
+        if app.platform == "mac":
+            actions.key("cmd-l")
+        else:
+            # Only using "ctrl-l" might fail and clear the console if the user
+            # is focused in the devtools
+            actions.key("f6 ctrl-l")
+
```
rntz: I'm not sure about this - this is bound in a context matching `app: firefox`, but the default behaviors are bound in contexts matching `tag: browser` (or `tag: browser\nos: mac`). Neither of these contexts is strictly more specific than the other, so which one matches is kinda dependent on talon implementation details, which I'd rather not rely on.

rntz (changes requested):
I think we need to use talon's context-matching mechanism to implement this rather than using `app.platform`, in order to ensure that this is guaranteed to override the default implementations. Make sure you match on a context that is strictly more specific than the defaults.

david-tejada:
I have to agree with you. I wasn't entirely convinced with the way I did it. I think I have to do something like I did for Opera. Although even in Opera I think I duplicated the tab_duplicate action unnecessarily. What do you think if I make another PR fixing all that and then we deal with this one?

I also wanted to suggest to move all the browsers to its own folder, what do you think about that?

Another thing I wasn't entirely convinced was having the default implementation for Windows and Linux instead of in its own file. I'll explain, for example, in browser.py I have all the implementations of actions that are either system independant and those for windows and linux. Then in browser_mac.py those that are mac exclusive. The alternative is to have another file for windows and linux implementations. I guess a set of guidelines like @splondike mentioned in slack would be handy.

splondike:
Responding to david-tejada's above comment:

* Happy for you to fix up Opera first if that makes sense to you.
* I probably wouldn't move them to their own folder if it's just because they're conceptually related. At the moment the apps folder is only one level deep, so you can quickly scan all the supported applications in knausj.
* I don't mind too much either way about splitting up Linux/Windows into separate files. I guess splitting them up is more explicit with what's going on.

david-tejada:
I went ahead and fixed this first. I split the python file in three, one for generic, other for mac and another for windows and linux. I like the result, I think it's much easier to see what's going on. I think I will do the same for browser.py and review the other browser apps. Let me know what you think.

rntz (approved):

rntz:
I added a comment explaining why the os matches were necessary in `firefox.py`, otherwise lgtm, merging.

david-tejada:
Nice, thanks!

PR #1148, opened 2023-03-21
Include safari technology preview in safari.py

davidsgrogan:


apps/safari/safari.py
```diff
@@ -6,7 +6,8 @@
 apps = mod.apps
 mod.apps.safari = """
 os: mac
-and app.bundle: com.apple.Safari
+app.bundle: com.apple.Safari
+app.bundle: com.apple.SafariTechnologyPreview
 """
 
 ctx.matches = r"""
```
davidsgrogan, line 13: Line 18/19 below might need to be updated. It's only called on line 31/32.

I can't figure out where `address()` is actually used. E.g. when I do "copy url" when from STP when Safari is also running and on a different page, I get the right address. @nriley, do you have ideas how I can exercise the functionality of address()? Sorry for dumb question.

nriley: Sure, it gets used when matching a website.

davidsgrogan: Got it, thanks. Updated this PR to make that work in Safari or STP.

splondike (approved):

apps/safari/safari.py
```diff
@@ -28,7 +25,8 @@ def browser_open_address_in_new_tab():
 class BrowserActions:
     def address() -> str:
         try:
-            window = safari_app().windows()[0]
+            mac_app = ui.apps(bundle=actions.app.bundle())[0]
+            window = mac_app.windows()[0]
```
rntz, line 29: I think maybe we could just use ui.active_app() here?

davidsgrogan: SGTM. Done. Seems to work with local testing.

PR #1149, opened 2023-03-24
Make repetition work when repeating a command more than 99 times

xavier630:
**Problem:**
Currently, you can only use repetition eg `touch 75 times` up to 99 times. If you try to say `touch 100 times`, the command doesn't work properly.

This is useful in various contexts. An example where you might want to repeat commands more than 99 times is when playing a single-player game, or when creating a template. 

This cap has affected me many times, so it likely affects others too.

**Solution Applied:**
This edits the repetition logic to allow repetition any number of times rather than applying a cap of 99 times.

### DIFF
--- a/plugin/repeater/repeater.talon
+++ b/plugin/repeater/repeater.talon
@@ -1,5 +1,5 @@
 # -1 because we are repeating, so the initial command counts as one
 <user.ordinals>: core.repeat_command(ordinals - 1)
-<number_small> times: core.repeat_command(number_small - 1)
+<number> times: core.repeat_command(number - 1)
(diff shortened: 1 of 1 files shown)


pokey (changes requested):
The current command is intentionally limited. Allowing any number for repeater is too dangerous to have in knausj. It could result in talon trying to repeat a command literally billions of times. I would recommend making a command in your own fork if you're interested in this functionality. I guess we could add a tag, but not sure it's worth it for a one-line command

xavier630:
Happy to keep it local then 👍

PR #1152, opened 2023-03-27
Add `tex` to language modes

pokey:


knausj85 (approved):

rntz:
uh, do we have a tex language mode? it doesn't look like there's a `lang/tex` directory?

pokey:
No it's more for mathfly

rntz:
Hm. We [specifically have a comment](https://github.com/knausj85/knausj_talon/pull/1152/files#diff-0f717cf8e4a3f504aa3233919d4d4e94bdeb02d950a57cce803ae8ce54bd4f12R6-R8) saying that we only put things in that list which have a supported language mode. I don't remember what the reason was for that, though. Do you?

I guess it makes sense that we don't want all mathfly users to have to modify knausj; it would be nice to figure out what our approach here should be in general.

pokey:
That's probably old. We now automatically generate language modes from that list. Guessing the comment predates that machinery

knausj85:
As I recall, languages without implementations were removed since (1) newbies were confused  and (2) it adds global force language commands that don't actually do anything

pokey:
That's fair. So should we roll it back? Alternately we could upstream some of the mathfly stuff

knausj85:
I wonder if this is as big of a deal as it used to be with .talon-lists coming; I'm tempted to leave it as is

Also open to upstreaming some of the math fly stuff, though there may be advantages to keeping it separate?

PR #1158, opened 2023-04-02
Abbreviations

fidgetingbits:
A bunch of abbreviations I've added over the years. I think the only thing that's not backwards compatible is this change:

```
-    "authentication": "auth",
+    "authentication": "authn",
```

Some of these are pretty specific and may be not useful to most people but I'm just leaving most of them in, because they probably don't hurt. But I don't care about any of them getting dropped if we don't want them in there.


splondike:
I think this is mostly fine. There are a handful of terminal commands in there (systemctl, rm, journalctl). Are those terms you type out outside of the terminal context as well (i.e. in prose)?

If it's just for the terminal, then perhaps [this PR](https://github.com/knausj85/knausj_talon/pull/1107/files) is a better approach and we should remove them here.

fidgetingbits:
Ya I write a lot of documentation in markdown, which is why they are in there.

knausj85 (approved):
This seems fine to me too. I'll leave it open for a bit for others to ponder.

splondike (approved):

PR #1159, opened 2023-04-02
Modified create spoken form to fix todos and add some other features

fidgetingbits:
This is a modified version of create_spoken_form.py that I worked on a while ago, but am finally finding time to post it now. I've been using it for quite a while and it seems to be fine, but it would be good for other people to test. I left a couple  commented out code in just to explain why I changed it, but we could remove it from the final version if folks are ok with the changes otherwise.

The main differences are: 
- When abbreviated versions are expanded, the original version is also kept. 
- When uppercase words are exploded, the original upper case and a new lower case version are kept (ex: `LICENSE` becomes speakable as `license` instead of only becoming `L I C E N S E`).
- Small lower case words (2-3 characters) are exploded into upper case versions, under the assumption that they might be dictated by character (ex: `usb` becomes `U S B`).
- Some words that are transformed are re-passed through other transformation passes, this way if we do something like have an upper case version of a word that's exploded, we may end up also passing the lower case version through an reverse abbreviation pass.

The main thing I use this for is filesystem stuff in my shell, because I have a zsh script that auto updates talon with files and folders in the current directory. So I wanted ways of quickly being able to dictate `vm`, `usb`, `README.md`, etc and it wasn't working well with the existing version.

I didn't run the existing create_spoken_form tests that exist yet, and I didn't add new tests. If someone wouldn't mind to quickly tell me what the best way to run the tests are, I'm happy to do it and I can add new ones for the stuff I added.


core/create_spoken_forms.py
```diff
@@ -187,23 +182,130 @@ def create_single_spoken_form(source: str):
...
+                file_extension_forms.append(substring)
+                dotted_extension_form.append(substring)
+                truncated_forms.append(substring)
+        # print(file_extension_forms)
+        if have_file_extension:
+            new_spoken_forms.append(" ".join(file_extension_forms))
+            new_spoken_forms.append(" ".join(dotted_extension_form))
+        new_spoken_forms.append(" ".join(truncated_forms))
+
+    return list(dict.fromkeys(new_spoken_forms))
```
auscompgeek, line 257: Is this intended to just be a `set`?

core/create_spoken_forms.py
```diff
@@ -14,11 +14,8 @@
 mod = Module()
 
 
-# TODO: 'Whats application': 'WhatsApp' (Should keep "whats app" as well?)
-# TODO: 'V O X': 'VOX' (should keep "VOX" as well?)
-# Could handle by handling all alternatives for these, or by having hardcoded list of things that we want to handle specially
-# TODO: Tests
-DEFAULT_MINIMUM_TERM_LENGTH = 3
+DEFAULT_MINIMUM_TERM_LENGTH = 2
```
knausj85, line 17: just curious, why did you drop this to 2? I recall settling on three for _*waves hands*_ reasons lost to time.

knausj85 (changes requested):
This looks good to me, and is working well in my tests thus far. 

I do think it would be wise to add some tests before merging; from the readme:
Automated tests
There are a number of automated unit tests in the repository. These are all run outside of the Talon environment (e.g. we don't have access to Talon's window management APIs). These make use of a set of stubbed out Talon APIs in test/stubs/ and a bit of class loader trickery in conftest.py.

To run the test suite you just need to install the pytest python package in to a non-Talon Python runtime you want to use for tests (i.e. don't install in the ~/.talon/.venv directory). You can then just run the pytest command from the repository root to execute all the tests.

core/create_spoken_forms.py
fidgetingbits: I believe I dropped it to 2 because otherwise if you have a file like say vm.md or a folder vm/ a value of 3 wouldn't allow you to access it by saying "V M". I'll double check though when I implement three tests.

core/create_spoken_forms.py
fidgetingbits: Yep, will update when I implement the tests. Thanks.

core/create_spoken_forms.py
knausj85: That makes sense, thanks

core/create_spoken_forms.py
fidgetingbits: I updated them all to sets now

fidgetingbits:
This should be good now. All tests pass and I added a few tests for the new stuff.

One thing I thought about while I was writing the tests is that longer term we probably want to move away from only expanding a single short word like `vm`, and break out the exploded function into two separate functions one that deals with small words and one that deals with packed words, but I don't have time to do it at the moment. The original case I had written it for was just I had a bunch of directories lying around like `vm` which is enough for this to address.

knausj85 (approved):

knausj85:
Looks good, thank you!

PR #1160, opened 2023-04-02
Switch screenshot_rect to become a user action

fidgetingbits:
As part of writing a screenshot tool a long time ago, I changed `screenshot_rect` to a user action. As I'm hoping to eventually push or share the other tool for others to use, I figured I'd push this change.

For reference the other tool is called shotbox https://github.com/fidgetingbits/knausj_talon/tree/master/plugin/shotbox, and the usage of this action is specifically https://github.com/fidgetingbits/knausj_talon/blob/master/plugin/shotbox/shotbox.py#L659

I vaguely recall when I wrote this the general consensus was that we shouldn't be importing other local python files (I think because of potential negative affects on talon dynamic module reloading), which is why I did this as a user action instead of just `from .. import screenshot_rect`. If this is no longer the case than it may be not necessary and I can just change my shotbox code, so I'll leave it up to you folks to decide.

### DIFF
--- a/plugin/screenshot/screenshot.py
+++ b/plugin/screenshot/screenshot.py
@@ -2,7 +2,7 @@ import os
 from datetime import datetime
 from typing import Optional
 
-from talon import Context, Module, actions, app, clip, cron, screen, ui
+from talon import Context, Module, actions, app, clip, cron, registry, screen, ui
 from talon.canvas import Canvas
 
 mod = Module()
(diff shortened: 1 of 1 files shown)


plugin/screenshot/screenshot.py
```diff
@@ -26,13 +26,13 @@ class Actions:
...
-        screen = get_screen(screen_number)
-        screenshot_rect(screen.rect)
+        selected_screen = get_screen(screen_number)
+        actions.user.screenshot_rect(selected_screen.rect)
 
     def screenshot_window():
         """Takes a screenshot of the active window and saves it to the pictures folder"""
         win = ui.active_window()
-        screenshot_rect(win.rect, win.app.name)
+        actions.user.screenshot_rect(win.rect, win.app.name)
```
splondike, line 35: Need to put in a screen_num arg here (currently crashes).

splondike (commented):
I like this having an action as the interface rather than the import stuff. Tests out okay for me except for that minor issue that I identified.

plugin/screenshot/screenshot.py
fidgetingbits: Easier to fix by adding keyword assignment `title=win.app.name`. I did that and confirmed it works.

plugin/screenshot/screenshot.py
```diff
@@ -52,20 +52,23 @@ def screenshot_settings():
...
         clipboard_rect(win.rect)
 
-
-def screenshot_rect(rect: ui.Rect, title: str = ""):
-    flash_rect(rect)
-    img = screen.capture_rect(rect)
-    path = get_screenshot_path(title)
-    img.write_file(path)
+    def screenshot_rect(
+        rect: ui.Rect, screen_num: Optional[int] = None, title: str = ""
```
auscompgeek, line 64: Should `screen_num` and `title` be swapped around? If for some reason someone wanted to call this from TalonScript it wouldn't be possible to pass `title` without passing a non-None `screen_num`.

knausj85 (approved):

plugin/screenshot/screenshot.py
splondike: This seems like a worthwhile change. Otherwise this all tests out OK for me now.

fidgetingbits: Sure I'll change that

PR #1161, opened 2023-04-02
Add lua

fidgetingbits:
Basic lua support

### DIFF
--- a/lang/lua/lua.py
+++ b/lang/lua/lua.py
@@ -161,10 +161,10 @@ class UserActions:
         actions.insert("nil")
 
     def code_insert_is_null():
-        actions.insert("== nil")
+        actions.insert(" == nil")
 
     def code_insert_is_not_null():
-        actions.insert("~= nil")
+        actions.insert(" ~= nil")
 
     ##
     # code_functions
--- a/lang/lua/lua.py
+++ b/lang/lua/lua.py
@@ -14,8 +14,6 @@ mod.setting(
 )
 mod.tag("stylua", desc="Tag for stylua linting commands")
 
-# XXX - this should be updatable with other modules like nse, neovim, base
-# off some tags
 ctx.lists["user.code_common_function"] = {
     "to number": "tonumber",
     "eye pairs": "ipairs",
@@ -61,9 +59,6 @@ ctx.lists["user.code_common_function"] = {
     "O S execute": "os.execute",
 }
 
-# XXX - it would be nice to correlate what we import these library is as to the
-# function calls. Ex: if someone does `local j = require("json")` then when we
-# call in too json libraries, we would prefix it with `j.parse()`
 ctx.lists["user.code_libraries"] = {
     "bit": "bit",
     "I O": "io",
@@ -78,9 +73,6 @@ ctx.lists["user.code_libraries"] = {
     "H T T P": "http",
     "web": "http",
     "jason": "json",
-    # Nmap Scripting Engine
-    # XXX - They should go in its own extension for .nse files
-    "short port": "shortport",
 }
 
 
@@ -249,7 +241,6 @@ class UserActions:
     def code_operator_multiplication():
         actions.insert(" * ")
 
-    # action(user.code_operator_exponent): " ** "
     def code_operator_division():
         actions.insert(" / ")
 
--- a/lang/lua/lua.py
+++ b/lang/lua/lua.py
@@ -1,0 +1,321 @@
+from talon import Context, Module, actions, settings
+
+mod = Module()
+ctx = Context()
+ctx.matches = r"""
+tag: user.lua
+"""
+
+mod.setting(
+    "lua_version",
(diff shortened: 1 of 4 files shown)


rntz:
probably want to uncomment https://github.com/knausj85/knausj_talon/blob/main/core/modes/language_modes.py#L27

lang/lua/lua.py
```diff
@@ -0,0 +1,312 @@
...
+    "I O": "io",
+    "string": "string",
+    "U T F eight": "utf8",
+    "table": "table",
+    "math": "math",
+    "O S": "os",
+    "debug": "debug",
+    "L F S": "lfs",
+    "socket": "socket",
+    "H T T P": "http",
```
rntz, line 73: conformer should be able to handle just "io", "utf eight", "os", "lfs", "http" - are you using Dragon / does Dragon have issues with acronyms?

lang/lua/lua.py
```diff
@@ -0,0 +1,312 @@
...
+    "table unpack": "table.unpack",
+    "table insert": "table.insert",
+    "tabel get N": "table.getn",
+    "tabel sort": "table.sort",
+    # math
+    "math max": "math.max",
+    # json
+    "jason parse": "json.parse",
+    # http
+    "web get": "http.get",
```
rntz, line 51: might be more consistent to use "http get" here, or at least have it as an alternative

rntz (approved):
I have a few optional suggestions but this looks fine. Haven't tested it. Feel free to merge with/without taking suggestions.

lang/lua/lua.py
fidgetingbits: No I use conformer, just an old pre-conformer habit from not wanting to assume stuff isn't known and then not working. Generally (to me) it seems sane to just not assume or spend time verifying what it knows vs not and just spell it out though. Curious if there is an advantage, outside I guess readability, to having utf8 vs U T F 8?

lang/lua/lua.py
```diff
@@ -0,0 +1,312 @@
...
+class UserActions:
+    # tag-related actions listed first, indicated by comment. corresponds to
+    # the tag(): user.code_imperative style declaration in the language .talon
+    # file
+
+    ##
+    # code_imperative
+    ##
+    def code_state_if():
+        actions.insert("if  then\n\nend")
```
splondike, line 95: See comment here about not including end block delimiters: https://github.com/knausj85/knausj_talon/pull/1162#discussion_r1160410725 (let's discuss there if needed). This looks like it applies to all your control blocks and public/private function actions.

lang/lua/lua.py
```diff
@@ -0,0 +1,312 @@
...
+        actions.insert("false")
+
+    ##
+    # code_data_null
+    ##
+    def code_insert_null():
+        actions.insert("nil")
+
+    def code_insert_is_null():
+        actions.insert("== nil")
```
splondike, line 168: Space before the operator in these for consistency with e.g. Python.

splondike (commented):
A few comments about consistency with other languages in knausj_talon.

fidgetingbits:
I think this is fine to merge now but I'll wait a day or two just to make sure there are no other comments

lang/lua/lua.py
rntz: IME with conformer splitting words into separate characters actually harms recognition; so "utf eight" will actually recognize more accurately than "u t f eight". But I think it's fine to just have "u t f eight", it may well be more future-proof against different/future models.

knausj85: As is, this will work better for for those that still use dragon too

nriley:
@fidgetingbits A couple errors I noticed with this PR:

```
WARNING actions: skipped because they have no matching declaration: (user.code_state_continue)
WARNING tags: skipped because they have no matching declaration: (user.nvim_lua)
```

Looks like you are missing a declaration for this action (missing file?) and may want to remove the tag since there is no other mention of nvim_lua in knausj currently.

lang/lua/lua.py
```diff
@@ -0,0 +1,308 @@
...
+        actions.user.insert_between("while ", " do")
+
+    def code_state_return():
+        actions.insert("return ")
+
+    def code_break():
+        actions.insert("break ")
+
+    # Assumes a ::continue:: label
+    def code_state_continue():
```
auscompgeek, line 123: Is there a reason for defining a new action for this instead of reusing `code_state_next()`?

fidgetingbits: Nope, just missed that code_next is what I needed when I implemented it I guess. Can change it to next instead.

auscompgeek: Probably appropriate. I had `state (next | continue): user.code_next()` in my local changes and ended up with the conflict.

fidgetingbits: Sounds good. I'll have time tomorrow to fix it.

PR #1162, opened 2023-04-02
basic Julia support

rntz:


lang/julia/julia.py
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+@ctx.action_class("user")
+class UserActions:
+    def code_operator_subscript():
+        actions.user.insert_between("[", "]")
+
+    def code_insert_null():
+        actions.auto_insert("nothing")
```
splondike, line 24: It's better to just use insert for code stuff because it doesn't go through `auto_format`.

lang/julia/julia.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def code_state_if():
+        actions.auto_insert("if\nend")
```
splondike, line 27: Other languages with non-brace block delimiters (e.g. ruby, vimscript) seem to not include the closing delimiter (i.e. end) in their control block commands. So I reckon we shouldn't here either.

lang/julia/julia.py
```diff
@@ -0,0 +1,115 @@
...
+    def code_insert_function(text: str, selection: str):
+        text += f"({selection or ''})"
+        actions.user.paste(text)
+        actions.edit.left()
+
+    def code_default_function(text: str):
+        actions.user.code_public_function(text)
+
+    def code_public_function(text: str):
+        result = "function {}()\nend".format(
```
splondike, line 58: Remove this closing delimiter as well I think.

lang/julia/julia.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def code_operator_in():
+        actions.auto_insert(" ∈ ")
+
+    def code_operator_not_in():
+        actions.auto_insert(" ∉ ")
+
+
+def julia_end():
+    actions.insert("\nend")
```
splondike, line 107: I think these are unused functions.

lang/julia/julia.talon
```diff
@@ -0,0 +1,30 @@
...
+tag(): user.code_operators_math
+
+settings():
+    user.code_private_function_formatter = "SNAKE_CASE"
+    user.code_public_function_formatter = "SNAKE_CASE"
+    user.code_private_variable_formatter = "SNAKE_CASE"
+    user.code_protected_variable_formatter = "SNAKE_CASE"
+    user.code_public_variable_formatter = "SNAKE_CASE"
+
+[state] [at] assert: "@assert "
```
splondike, line 22: It's probably worth making the state prefix mandatory for consistency with other languages.

Also are the dassert things standard Julia? I couldn't find them with a web search.

lang/julia/julia.py
fidgetingbits: I think generally I'm okay with removing them, and I'll remove it from the lua pull request. But it does make me wonder if there should be a separate action that corresponds to the full set of block delimiters? It seems like there's enough languages structured that way and a lot of the time I know I'm going to want both anyways so it would reduce the requirement to rely on snippets in the editor or having to chain commands.

rntz:
I don't think I have the bandwidth to work on this so I'm going to close; anyone else should feel free to reopen or complete.

PR #1167, opened 2023-04-09
i3wm: Call i3-msg using subprocess instead of os.system

auscompgeek:
Need to test this first but opening to get the ball rolling.

Starts on #945

### DIFF
--- a/apps/i3wm/i3wm.py
+++ b/apps/i3wm/i3wm.py
@@ -1,9 +1,10 @@
 import subprocess
 from typing import Optional, Union
 
-from talon import Module, actions, settings
+from talon import Context, Module, actions, settings
 
 mod = Module()
+ctx = Context()
 
 mod.tag("i3wm", desc="tag for loading i3wm related files")
 mod.setting(
@@ -18,6 +19,16 @@ mod.setting(
     default="super",
     desc="The default key to use for i3wm commands",
 )
+
+ctx.matches = """
+tag: user.i3wm
+"""
+
+
+@ctx.action_class("app")
+class AppActions:
+    def window_close():
+        subprocess.check_call(("i3-msg", "kill"))
 
 
 @mod.action_class
--- a/apps/i3wm/i3wm.py
+++ b/apps/i3wm/i3wm.py
@@ -85,9 +85,13 @@ class Actions:
     def i3wm_move_to_workspace(which: Union[str, int]):
         """Move the focused container to the specified workspace"""
         if isinstance(which, int):
-            subprocess.check_call(("i3-msg", "move", "container", "to", "workspace", "number", str(which)))
+            subprocess.check_call(
+                ("i3-msg", "move", "container", "to", "workspace", "number", str(which))
+            )
         else:
-            subprocess.check_call(("i3-msg", "move", "container", "to", "workspace", which))
+            subprocess.check_call(
+                ("i3-msg", "move", "container", "to", "workspace", which)
+            )
 
     def i3wm_move_to_output(which: str):
         """Move the focused container to the specified output."""
--- a/apps/i3wm/i3wm.py
+++ b/apps/i3wm/i3wm.py
@@ -1,3 +1,6 @@
+import subprocess
+from typing import Optional, Union
+
 from talon import Module, actions, settings
 
 mod = Module()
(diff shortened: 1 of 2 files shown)


apps/i3wm/i3wm.talon
```diff
@@ -2,82 +2,81 @@
...
 
-(win | window) left: user.system_command("i3-msg focus left")
-(win | window) right: user.system_command("i3-msg focus right")
-(win | window) up: user.system_command("i3-msg focus up")
-(win | window) down: user.system_command("i3-msg focus down")
+(win | window) left: user.i3wm_focus("left")
+(win | window) right: user.i3wm_focus("right")
+(win | window) up: user.i3wm_focus("up")
+(win | window) down: user.i3wm_focus("down")
 (win | window) kill: user.system_command("i3-msg kill")
```
auscompgeek, line 14: I think `window close` should also work, right?

apps/i3wm/i3wm.py
```diff
@@ -29,6 +63,44 @@ def i3wm_shell():
...
         actions.key(f"{key}-enter")
 
+    def i3wm_focus(what: str):
+        """Move focus"""
+        subprocess.check_call(("i3-msg", "focus", what))
+
+    def i3wm_switch_to_workspace(which: Union[str, int]):
+        """Focus the specified workspace"""
+        if isinstance(which, int):
+            subprocess.check_call(("i3-msg", "workspace", "number", str(which)))
```
auscompgeek, line 73: Should we just override `user.desktop(number: int)`?

Noteworthy is that the desktop stuff is in `plugin/desktops/`. If this starts overriding those actions, should it live in `core/`?

apps/i3wm/i3wm.py
```diff
@@ -29,6 +63,44 @@ def i3wm_shell():
...
+
+    def i3wm_move(to: str):
+        """Move the focused container"""
+        subprocess.check_call(("i3-msg", "move", to))
+
+    def i3wm_move_to_workspace(which: Union[str, int]):
+        """Move the focused container to the specified workspace"""
+        if isinstance(which, int):
+            subprocess.check_call(
+                ("i3-msg", "move", "container", "to", "workspace", "number", str(which))
```
auscompgeek, line 89: If we decide to override `user.desktop`, we should also override `user.window_move_desktop(number: int)`, and `window_move_desktop_{left,right}` too.

apps/i3wm/i3wm.talon
```diff
@@ -2,82 +2,81 @@
...
 tag: user.i3wm
 -
-port <number_small>: user.system_command("i3-msg workspace number {number_small}")
-(port flip | flipper): user.system_command("i3-msg workspace back_and_forth")
-port right: user.system_command("i3-msg workspace next")
-port left: user.system_command("i3-msg workspace prev")
+port <number_small>: user.i3wm_switch_to_workspace(number_small)
+(port flip | flipper): user.i3wm_switch_to_workspace("back_and_forth")
+port right: user.i3wm_switch_to_workspace("next")
+port left: user.i3wm_switch_to_workspace("prev")
```
auscompgeek, line 8: I guess if we override `user.desktop` we should override `user.desktop_next()` and `user.desktop_last()` too.

auscompgeek:
cc @fidgetingbits @splondike

apps/i3wm/i3wm.py
auscompgeek: I suppose the other alternative to overriding is to check whether we're in i3 in the Linux implementation, but that seems a bit awkward to muddy the common implementation with.

apps/i3wm/i3wm.talon
```diff
@@ -2,82 +2,81 @@
...
 
-make scratch: user.system_command("i3-msg move scratchpad")
-[(show | hide)] scratch: user.system_command("i3-msg scratchpad show")
+make scratch: user.i3wm_move("scratchpad")
+[(show | hide)] scratch: user.i3wm_show_scratchpad()
 next scratch:
-    user.system_command("i3-msg scratchpad show")
-    user.system_command("i3-msg scratchpad show")
+    user.i3wm_show_scratchpad()
+    user.i3wm_show_scratchpad()
```
splondike, line 79: `next scratch` doesn't do the right thing on my computer (it just swaps in and out of scratch mode). But I think we can just keep this PR as about refactoring the code, not fixing bugs.

apps/i3wm/i3wm.talon
splondike: Yeah, the app.window_close action could be implemented to enable that command. The default of alt-f4 isn't in the default i3 config I think. Happy if you want to implement that action to replace this system_command usage rather than creating an i3 specific one.

apps/i3wm/i3wm.py
splondike: I reckon leave off this kind of thing for another PR to keep this one focused on just removing the system_command usage with no usage changes.

splondike (approved):
Tested this out and it seems to work fine. Happy to have this merged in after we reimplement the 'win kill' command.

apps/i3wm/i3wm.talon
auscompgeek: For some reason I thought that action sent an X11 signal directly rather than pressing keys. I'll override it and call it here.

We can discuss whether `win kill` should be deprecated in a follow-up PR.

splondike:
Looks good, thanks!

PR #1168, opened 2023-04-09
Mode cleanup

knausj85:
- Switch gdb and windbg to tags

Addresses the remainder of https://github.com/knausj85/knausj_talon/issues/603

Probably most relevant to @fidgetingbits


fidgetingbits:
Looks good to me, though didn't test it

core/modes/modes.py
```diff
@@ -8,7 +8,6 @@
     "gdb": "a way to force gdb commands to be loaded",
```
splondike, line 8: I think we want to delete this gdb line as well.

splondike (approved):
Gave it a bit of a test (just checked that when the tag was enabled the right commands ran), looks good! Probably just want to remove that leftover gdb mode.

core/modes/modes.py
knausj85: Fixed, thanks

core/modes/modes.py
```diff
@@ -5,10 +5,8 @@
 modes = {
     "admin": "enable extra administration commands terminal (docker, etc)",
     "debug": "a way to force debugger commands to be loaded",
-    "gdb": "a way to force gdb commands to be loaded",
```
auscompgeek, line 8: I think we forgot to remove the voice commands activating this mode

https://github.com/knausj85/knausj_talon/blob/99869cbaebe68ec43f4ed992e0317deb9bd01070/core/modes/language_modes.talon#L3-L4

apps/gdb/gdb.py
```diff
@@ -1,13 +1,30 @@
-from talon import Context, actions
+from talon import Context, Module, actions
 
-ctx = Context()
+mod = Module()
+mod.tag("gdb", "Tag to enabled gdb-related functionality")
```
auscompgeek, line 4: Looking around I think this conflicts with the `gdb` language tag 😬

PR #1169, opened 2023-04-10
browser: Simplify title address extraction

auscompgeek:
This refactors the `browser.address` implementation to be clearer, making the following changes along the way:

- The same word is no longer checked to be a valid URL twice.
- Check the first word is a URL before checking the rest of the words in reverse. The suggested Firefox extension prepends the URL to the window title (and does not wrap it in brackets).
- Avoid checking the empty string, in case a webpage has consecutive spaces in its title.

### DIFF
--- a/tags/browser/browser.py
+++ b/tags/browser/browser.py
@@ -9,12 +9,12 @@ tag: browser
 """
 
 
-def is_url(url):
+def is_url(url: str) -> bool:
     try:
         # Valid if url successfully parsed
         result = urlparse(url)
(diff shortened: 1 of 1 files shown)


splondike (approved):
Rango and the recommended Chrome extension both stick the address towards the end. This code works with both from my testing.

This wouldn't be a bad candidate for some automated tests if you feel like writing them. If not I'll merge this in a week or so.

splondike:
Thanks!

PR #1171, opened 2023-04-13
emacs: support shell-script-mode/sh-mode

rntz:
also sorts the mode commands in emacs.talon alphabetically


apps/emacs/emacs.talon
```diff
@@ -87,6 +87,7 @@ overwrite mode: user.emacs("overwrite-mode")
 paredit mode: user.emacs("paredit-mode")
 rainbow mode: user.emacs("rainbow-mode")
 read only mode: user.emacs("read-only-mode")
+shell script mode: user.emacs("sh-mode")
```
rntz, line 90: sh-mode is the same thing as shell-script-mode, fyi

PR #1172, opened 2023-04-15
Initial meeting tag support

pokey:
- Partially addresses https://github.com/knausj85/knausj_talon/issues/835
- Blocked by https://github.com/talonvoice/talon/issues/606


apps/zoom/zoom.py
```diff
@@ -0,0 +1,45 @@
...
+@ctx.action_class("user")
+class UserActions:
+    def meeting_mute():
+        app.notify("Mute meeting")
+
+    def meeting_unmute():
+        app.notify("Unmute meeting")
+
+    def meeting_exit():
+        app.notify("Exit meeting")
```
pokey, line 41: @nriley here's where you'd put the implementations of mute / unmute / exit.  Need to decide whether we're comfortable using accessibility here.  Also prob need to make the tag mac-specific if we use accessibility

apps/zoom/zoom.py
```diff
@@ -0,0 +1,45 @@
...
+global_ctx = Context()
+
+ctx = Context()
+ctx.matches = r"""
+tag: user.meeting_zoom
+"""
+
+
+def is_zoom(app):
+    return app.bundle == "us.zoom.xos"
```
pokey, line 15: Todo: make this work on Windows

apps/zoom/zoom.talon
pokey, line 1: Maybe we should rename this file?  Because this file name should really be for when Zoom is focused

pokey:
@nriley feel free to push directly to this branch, or open a PR against it if you can't for whatever reason

nriley:
Instead of looking at the menu (with a different title per-language) I can recognize the shape of the "you are muted" icon and determine the mute status this way; it worked across the localizations and meeting window layouts I tried. Still might be some issues (even ignoring the Talon bug) in recognizing the meeting window title.

If we just wanted to support toggling mute, we could do it without accessibility.

I put in a partial workaround for the bug by focusing another app and back to Zoom. I don't think actually checking the 2nd time should help in most cases but it should help Talon to notice the window for **next** time. The problem is still though that the tag won't activate, so the voice commands don't do anything (other than "media mute" as @pokey experienced yesterday :).

One thing to note while I was testing, which in part inspired my two-microphone solution: it seems that if you mute audio in Zoom, you can't talk to Talon either **on the same microphone** (works fine with another microphone).

Another action that might be helpful would be to allow you to focus the active meeting window since I've certainly had cases where it gets "lost". This wouldn't require accessibility either (at least for Zoom), since my window filtering just uses the title.

nriley:
Here's a thought of a slightly more robust workaround for the problem of meeting commands not being active since Talon hasn't "found" the meeting window yet.

We could have the meeting commands active if *any* of the meeting apps are active, using "user.running". Then in the default implementation for "meeting mute", etc., we could "poke" each of the apps (via accessibility to find the window) to see if we're failing to notice an active meeting window.

Another option is to poll windows in active meeting apps by accessibility (which I may end up doing in my own repo, so I can accurately switch microphones) and "force" Talon to notice any it hasn't found. I do that in one of my private repos, but I'd be a bit worried about doing that by default.

nriley:
Added Teams support — due to its use of Electron this is more complex/asynchronous. Given Teams is pending a rewrite I'm disinclined to spend much more time on this particular implementation, but it at least works for me. Talon seems to be more reliably "see" the meeting window in Teams, so I didn't include the same workaround I did for Zoom.

nriley:
Refactored a bit to put the `user.meeting_appname` tag context in one place rather than creating a context per meeting app. At this point we could also just set user.meeting in the module implementation rather than using .talon files to do so, but I left that part alone for now as I don't feel strongly one way or the other.

Here's where I override the `user.meeting_started`/`user.meeting_ended` in my `suspend` script to turn off or switch the microphone during a meeting: https://github.com/nriley/knausj_talon/commit/26ac780f5cc0191de97f7eb3c648c01d07f7643a

auscompgeek:
> I can recognize the shape of the "you are muted" icon and determine the mute status this way

Does this work even when the Zoom meeting controls are set to auto-hide?

nriley:
> > I can recognize the shape of the "you are muted" icon and determine the mute status this way
> 
> Does this work even when the Zoom meeting controls are set to auto-hide?

Yes, I think so — the icon I am matching shows up even if the controls are hidden.

image

If you can find cases where the above mute icon does not appear, then I can implement checking the menus in addition (like I do on Webex) — unfortunately this will not be robust to non-English UI languages.

nriley:
OK, I've moved off my slick-but-broken Zoom mute detection.

More testing appreciated!

auscompgeek:
Does the Zoom menu bar item need to be enabled for that to work, or does it use the menus for the meeting window?

nriley:
Yes, it does not use the menu in the top right corner of the screen but instead uses the application menus even if the application is not active.

nriley:
Since I upgraded either to a recent Talon beta or macOS (or both), this has become incredibly unreliable. (It's blocked on the window activation callback issue in core Talon, but it was at least usable previously.) Will work on fixing it as I have time.

nriley:
https://github.com/talonvoice/talon/issues/606 is fixed in the latest beta! I tested Zoom and it seems reliable.

freeload101:
> Instead of looking at the menu (with a different title per-language) I can recognize the shape of the "you are muted" icon and determine the mute status this way; it worked across the localizations and meeting window layouts I tried. Still might be some issues (even ignoring the Talon bug) in recognizing the meeting window title.
> 
> If we just wanted to support toggling mute, we could do it without accessibility.
> 
> I put in a partial workaround for the bug by focusing another app and back to Zoom. I don't think actually checking the 2nd time should help in most cases but it should help Talon to notice the window for **next** time. The problem is still though that the tag won't activate, so the voice commands don't do anything (other than "media mute" as @pokey experienced yesterday :).
> 
> One thing to note while I was testing, which in part inspired my two-microphone solution: it seems that if you mute audio in Zoom, you can't talk to Talon either **on the same microphone** (works fine with another microphone).
> 
> Another action that might be helpful would be to allow you to focus the active meeting window since I've certainly had cases where it gets "lost". This wouldn't require accessibility either (at least for Zoom), since my window filtering just uses the title.

Oh geez can't you just look at the DOM or is it wrecked because of  different languages ?  say a lot of people have their own hotkeys or special things toggling MIC because every platform is completely different...

nriley:
Yeah, unfortunately not every app that uses web technologies actually exposes the DOM via accessibility. Teams on macOS went backwards in this area.

Some new features coming in the beta/experimental version of Talon may make picking up this PR a bit more realistic but right now with public Talon it would be quite painful to do well. Putting it on hold…

PR #1173, opened 2023-04-15
fix buffer next/buffer last

okonomichiyaki:
I think these are supposed to match `emacs.talon`?
```
buffer next: user.emacs("next-buffer")
buffer last: user.emacs("previous-buffer")
```

### DIFF
--- a/apps/emacs/emacs_commands.csv
+++ b/apps/emacs/emacs_commands.csv
@@ -10,8 +10,8 @@ balance-windows, ctrl-x +
 beginning-of-buffer, meta-<
 browse-kill-ring,, b-k-r
 buffer-menu, ctrl-x ctrl-b
-buffer-next, ctrl-x right
-buffer-previous, ctrl-x left
+next-buffer, ctrl-x right
+previous-buffer, ctrl-x left
 bury-buffer,, bur
 clear-rectangle, ctrl-x r c
 clone-indirect-buffer,, clo-i


apps/emacs/emacs_commands.csv
```diff
@@ -10,8 +10,8 @@ balance-windows, ctrl-x +
 beginning-of-buffer, meta-<
 browse-kill-ring,, b-k-r
 buffer-menu, ctrl-x ctrl-b
-buffer-next, ctrl-x right
-buffer-previous, ctrl-x left
+next-buffer, ctrl-x right
+previous-buffer, ctrl-x left
```
rntz, line 14: You're right, of course, but also these keybindings should be swapped I think? Don't have emacs handy now but http://gnu.ist.utl.pt/software/emacs/manual/html_node/Select-Buffer.html says that `C-x left` is `next-buffer` and `C-x right` is previous-buffer. Could you check this and swap them if so?

okonomichiyaki: tbh I have never thought about what "next" and "previous" mean in this context and I'm a bit confused after reading that link and experimenting. I'm also actually using Spacemacs and I'm never sure if there's any behavior that has been changed

that being said, `next-buffer` and `C-x right` do the same thing for me. is it possible that linked documentation is actually wrong?? it appears to contradict itself if you compare the list of commands with the paragraph of text

this doc page is different https://emacsdocs.org/docs/emacs/Select-Buffer

rntz: Ok, I'm betting those docs are just mixed up then.

rntz (approved):

PR #1174, opened 2023-04-18
match emacs on windows

okonomichiyaki:
without this the new code doesn't activate on Windows. tested on Mac to make sure the matchers are not not conflicting

### DIFF
--- a/apps/emacs/emacs.py
+++ b/apps/emacs/emacs.py
@@ -12,9 +12,14 @@ setting_meta = mod.setting(
 
 mod.apps.emacs = "app.name: Emacs"
 mod.apps.emacs = "app.name: emacs"
+mod.apps.emacs = "app.name: /GNU Emacs/"
 mod.apps.emacs = """
 os: mac
 app.bundle: org.gnu.Emacs
+"""
+mod.apps.emacs = """
+os: windows
+app.exe: emacs.exe
 """
 
 ctx = Context()


apps/emacs/emacs.py
```diff
@@ -12,10 +12,15 @@
 
 mod.apps.emacs = "app.name: Emacs"
 mod.apps.emacs = "app.name: emacs"
+mod.apps.emacs = "app.name: /GNU Emacs/"
```
rntz, line 15: Why the slashes? How would this differ from `app.name: "GNU Emacs"`?

rntz (approved):
Lgtm. I would like to know what the slashes in `/GNU Emacs/` are for - would this match any app with `GNU Emacs` anywhere in its name? Is that necessary?

apps/emacs/emacs.py
okonomichiyaki: I had asked about this on Slack, b/c I'm definitely no expert on how the context matchers work so defer to someone else on how to do this... but slashes make it match anything with "GNU Emacs" in the name. I also considered just changing `app.name: Emacs` to `app.name: /Emacs/`

actually, no changes to `app.name` are needed if we have the `emacs.exe` matcher and maybe this line should be omitted. here is the output from `copy context` with Emacs on Windows 10 for me:

```
os: windows
and app.name: GNU Emacs: The extensible self-documenting text editor
os: windows
and app.exe: emacs.exe
```

rntz: Thanks! I think this is probably fine, especially since relying exclusively on .exe names on Windows seems a bit flaky to me.

auscompgeek: Matching it anywhere could be a bit overboard, perhaps the regex should be anchored instead? i.e.
```
app.name: /^GNU Emacs/
```

PR #1175, opened 2023-04-22
Delete the pytest path loader hook hack

auscompgeek:
We had a path loader hack in place to test various parts of the code:

https://github.com/knausj85/knausj_talon/blob/624e6f21c7d24191236e61c278a7a9856576ca0d/conftest.py#L12-L21

We've since renamed the `code` directory, so this is no longer necessary.

Instead we can use the pythonpath config: https://docs.pytest.org/en/stable/reference/reference.html#confval-pythonpath


splondike (approved):
Works fine, nice one!

PR #1181, opened 2023-05-05
Allows the abbreviated/short form to be used as spoken phrase

AndreasArvidsson:
`"brief application` => `app`
`"brief app"` => `app`

### DIFF
--- a/core/abbreviate/abbreviate.py
+++ b/core/abbreviate/abbreviate.py
@@ -449,7 +449,7 @@ abbreviations_list = get_list_from_csv(
     default=abbreviations,
 )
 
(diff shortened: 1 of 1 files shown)


fidgetingbits:
Didn't test this yet as afk, but off the top of my head I'd be curious how it handles some existing entries like for instance "brief configuration" returns "config" but "brief config" returns "cfg". Seems like that could break things it as will create duplicate entries? 

Could always check and not add the additional entries if one already exists.

fidgetingbits (approved):
Looks good to me given you tested the concern I had. But still didn't test it myself.

PR #1182, opened 2023-05-05
Added additional edit commands

AndreasArvidsson:
`"copy start of line"`
`"clear end of line"`

`"paste line"`
`"paste word"`

`"drink line"`
`"pour line"`

Made sure all operations(select, copy, cut, paste, delete) had implementations for all, line, line start, line end, and word. Grouped a few commands to make finding then easier and added actions for reusable logic.

### DIFF
--- a/core/edit/edit.talon
+++ b/core/edit/edit.talon
@@ -150,5 +150,5 @@ clone that: edit.selection_clone()
 clone line: edit.line_clone()
 
 # Insert new line
-drink line: edit.line_insert_up()
-pour line: edit.line_insert_down()
+new line above: edit.line_insert_up()
+new line below: edit.line_insert_down()
--- a/core/edit/edit.talon
+++ b/core/edit/edit.talon
@@ -1,79 +1,59 @@
+# Searching
 find it: edit.find()
-
 next one: edit.find_next()
 
 go word left: edit.word_left()
-
 go word right: edit.word_right()
 
 go left: edit.left()
-
 go right: edit.right()
-
 go up: edit.up()
-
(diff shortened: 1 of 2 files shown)


rntz:
Drink/pour are cursorless-isms, not part of knausj. Not super self explanatory commands.

AndreasArvidsson:
@rntz Updated

core/edit/edit.talon
```diff
@@ -161,20 +138,17 @@ cut all:
...
+(pace | paste) word: user.paste_word()
 
-# duplication
+# Duplication
 clone that: edit.selection_clone()
 clone line: edit.line_clone()
+
+# Insert new line
+new line above: edit.line_insert_up()
+new line below: edit.line_insert_down()
```
auscompgeek, line 154: We already have `slap` for this. I don't think we need another one.

https://github.com/knausj85/knausj_talon/blob/2e9e03ddb559c782e65650120a9d91079b6dc643/core/edit/standard.talon#L22

rntz: I think it's fine to have a duplicate since otherwise we'll have `new line above` without `new line below` which would be very confusing.

rntz (approved):

core/edit/edit.talon
auscompgeek: Fair. Maybe worth considering what should go in `standard.talon` as opposed to `edit.talon` then.

AndreasArvidsson: I think we should remove `standard.talon` and move the content to `edit.talon`.

PR #1183, opened 2023-05-06
Improve Discord support

nriley:
Some fixes from #937:

- Remove redundant "channel" command (also in messaging.talon)
- Implement commands on macOS
- Match Discord on macOS by bundle ID in case it's not named "Discord"

Also add command for sticker picker.

Needs testing on Windows.


knausj85 (approved):
It looks good to me, works fine on windows too.

PR #1184, opened 2023-05-06
more mode cleanup

knausj85:
- help_scope and help modes converted to tags
- Removed old commands for enabling gdb mode

### DIFF
--- a/core/help/help_scope.py
+++ b/core/help/help_scope.py
@@ -1,7 +1,7 @@
-from talon import Module, actions, imgui, scope, ui
-
+from talon import Module, Context, actions, imgui, scope, ui
+ctx = Context()
 mod = Module()
-mod.mode("help_scope", "Mode for showing the scope help gui")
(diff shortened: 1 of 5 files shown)


core/help/help_scope.py
```diff
@@ -1,7 +1,8 @@
-from talon import Module, actions, imgui, scope, ui
+from talon import Context, Module, actions, imgui, scope, ui
 
+ctx = Context()
 mod = Module()
-mod.mode("help_scope", "Mode for showing the scope help gui")
+mod.tag("help_scope", "tag for showing the scope help gui")
```
AndreasArvidsson, line 5: Should we named this tag `helps_scope_open` to be consistent with `help_open`?

AndreasArvidsson (approved):

core/help/help_scope.py
knausj85: Make sense, done

PR #1187, opened 2023-05-07
repeater: Add repeat phrase commands

auscompgeek:
`repeat phrase <N> times` initially from https://talonvoice.slack.com/archives/C9MHQ4AGP/p1679478692465999

I'm not entirely sure if we should have `again` on its own as a command, but would love to know everyone's thoughts.

### DIFF
--- a/plugin/repeater/repeater.talon
+++ b/plugin/repeater/repeater.talon
@@ -3,4 +3,6 @@
 <number_small> times: core.repeat_command(number_small - 1)
 (repeat that | twice): core.repeat_command(1)
 repeat that <number_small> [times]: core.repeat_command(number_small)
+
 again: core.repeat_phrase()
+^repeat phrase <number_small> times$: core.repeat_phrase(number_small - 1)
--- a/plugin/repeater/repeater.talon
+++ b/plugin/repeater/repeater.talon
@@ -4,5 +4,5 @@
 (repeat that | twice): core.repeat_command(1)
 repeat that <number_small> [times]: core.repeat_command(number_small)
 
-again: core.repeat_phrase()
-^repeat phrase <number_small> times$: core.repeat_phrase(number_small - 1)
+(repeat phrase | again) [<number_small> times]:
+    core.repeat_partial_phrase(number_small or 1)
--- a/plugin/repeater/repeater.talon
+++ b/plugin/repeater/repeater.talon
@@ -3,3 +3,4 @@
 <number_small> times: core.repeat_command(number_small - 1)
 (repeat that | twice): core.repeat_command(1)
 repeat that <number_small> [times]: core.repeat_command(number_small)
+again: core.repeat_phrase()


knausj85 (approved):
This looks good to me. I'll leave it open for a bit for others to evaluate.

plugin/repeater/repeater.talon
```diff
@@ -3,3 +3,6 @@
 <number_small> times: core.repeat_command(number_small - 1)
 (repeat that | twice): core.repeat_command(1)
 repeat that <number_small> [times]: core.repeat_command(number_small)
+
+again: core.repeat_phrase()
+^repeat phrase <number_small> times$: core.repeat_phrase(number_small - 1)
```
AndreasArvidsson, line 8: Do we need an anchor here?

plugin/repeater/repeater.talon
```diff
@@ -3,3 +3,6 @@
 <number_small> times: core.repeat_command(number_small - 1)
```
AndreasArvidsson, line 3: Should we have the phrase form of this one?
`phrase  <number_small> times`

plugin/repeater/repeater.talon
auscompgeek: You'd have to ask Ryan 😄

I presume it might behave in a confusing manner if it wasn't anchored. I actually haven't tested what would happen if you used this in the middle of a phrase.

I suppose the other one should be anchored too if this one is.

plugin/repeater/repeater.talon
auscompgeek: That would conflict with

https://github.com/knausj85/knausj_talon/blob/main/core/text/text.talon#L2-L7

rntz:
`again` should be, at minimum, start-anchored. And the anchoring should be consistent across all repeat phrase commands.

For consistency with `repeat that`, there should definitely be a numberless `repeat phrase`, eg:

```
^repeat phrase$: core.repeat_phrase()
```

Personally I'd ditch `again` and just have `repeat phrase`, but I'm not opposed to it as long as it's anchored like the rest of the commands.

plugin/repeater/repeater.talon
rntz: I don't think the `- 1` should be here; `repeat phrase 1 time` shouldn't be a noop. The minus one is for ordinals.

rntz:
> If we use `core.repeat_partial_phrase` without a front anchor you can end the phrase with that command the same way the command repeater works to get the entire phrase you repeated. `"air bat cap again"` => `abcabc`

Oh, interesting. I assumed it would repeat the _previous_ phrase, not the _current_ one. What does it do if you use it in the middle of a phrase, `air bat cap again drum`? `abcabcd`?

rntz:
Ok, looks like `core.repeat_partial_phrase` is basically just an improved version of `core.repeat_phrase`. If we use that, I don't think we need the anchors. Still kinda on the fence about `again` as a command, but it's probably fine?

rntz:
Assuming we want to keep `again`, I suggest:

```talon
again: core.repeat_partial_phrase()
again <number_small> times: core.repeat_partial_phrase(number_small)
repeat phrase: core.repeat_partial_phrase()
repeat phrase <number_small> times: core.repeat_partial_phrase(number_small)
```

auscompgeek:
Does `core.repeat_partial_phrase` repeat the previous phrase if it's at the beginning? From the name it sounds like it only does the current phrase.

rntz:
It does.

AndreasArvidsson:
@rntz looks good to me
@auscompgeek  `core.repeat_partial_phrase` yes it will work both as a full and a partial repeater.

AndreasArvidsson:
@rntz looks good to me
@auscompgeek  `core.repeat_partial_phrase` yes it will work both as a full and a partial repeater.

auscompgeek:
Ok, I've changed it to use `repeat_partial_phrase` and made the two consistent.

auscompgeek:
An aside, I've noticed the phrase repeaters have slightly quirky behaviour if the phrase being repeated has a repeater too – it seems the repeater will do nothing, so something like `tab new 5 times` then `again` will result in 6 new tabs, not 10. I presume this is intentional Talon behaviour to avoid infinite loops, but initially surprised me.

For any future travellers, the `repeat_command` action won't repeat the `repeat_partial_phrase`, only the last command of the repeated phrase, so this command does need to handle `<N> times` itself to get the desired behaviour.

AndreasArvidsson (approved):

rntz (approved):

PR #1189, opened 2023-05-10
use pop-to-mark-command not pop-mark

rntz:
pop-mark isn't an emacs command (only a function) and doesn't do what we want anyway. The keybinding it was listed with (ctrl-u ctrl-@) has the intended behavior, which is to invoke pop-to-mark-command. This PR means that we'll now do the right thing if `user.emacs` has to fall back to `M-x` or if it gets overridden to use RPC instead of keybindings.

### DIFF
--- a/apps/emacs/emacs.py
+++ b/apps/emacs/emacs.py
@@ -153,11 +153,11 @@ class UserActions:
         if left:
             actions.edit.extend_word_left()
             before = actions.edit.selected_text()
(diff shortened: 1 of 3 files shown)


nriley (approved):
Works for me. I do have an issue with context-sensitive dictation in Emacs in which there's no space after a period — does this happen for you? Behavior doesn't change pre or post this change, so it shouldn't block merging it...

PR #1190, opened 2023-05-10
Refactor emacs_commands.py

rntz:
Move most functionality into `emacs.py`; expose access to the command-name-to-info mapping via talon actions. Simplify emacs_commands.py.

Replace `user.emacs("universal-argument")` with `user.emacs_prefix()`. This now has correct behavior if we need to fall-back to M-x or use RPC.

Allow `emacs_commands.csv` to specify the keybindings used for `user.emacs_prefix()` and for `execute-extended-command` (used by `user.emacs()` if no keybinding is found). This reduces the number of places a user needs to customize to use nonstandard keybindings.


apps/emacs/emacs.talon
```diff
@@ -14,7 +14,7 @@ execute <user.text>$:
     user.emacs_meta_x()
```
nriley, line 14: There's still a few uses of `emacs_meta_x` in emacs.talon that need updating.

nriley (changes requested):
Looks good other than one nit!

apps/emacs/emacs.talon
rntz: Ok, should have fixed this now.

rntz:
I don't understand why the CI tests are failing. @phillco any idea what's up?

rntz:
I don't understand why the CI tests are failing. @phillco any idea what's up?

nriley (approved):

PR #1192, opened 2023-05-10
add "copy way {left,right,up,down}"

rntz:
Adds "copy way {left,right,up,down}" commands. Also moves the scroll commands in `edit.talon` closer together and clarifies a few doc strings.

See https://github.com/knausj85/knausj_talon/pull/1182

### DIFF
--- a/core/edit/edit.talon
+++ b/core/edit/edit.talon
@@ -8,11 +8,6 @@ find it: edit.find()
 next one: edit.find_next()
 
 # Navigation
-
-# The reason for these spoken forms is that "page up" and "page down" are globally defined as keys.
-scroll up: edit.page_up()
-scroll down: edit.page_down()
-
 go word left: edit.word_left()
 go word right: edit.word_right()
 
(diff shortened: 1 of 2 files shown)


AndreasArvidsson (approved):

core/edit/edit.talon
```diff
@@ -113,6 +111,19 @@ copy word: user.copy_word()
...
+copy way right:
+    edit.extend_line_end()
+    edit.copy()
+copy way up:
+    edit.extend_file_start()
+    edit.copy()
+copy way down:
+    edit.extend_file_end()
+    edit.copy()
+
```
nriley, line 126: On the knausj grooming session we had agreed that it may make more sense to make these individual actions rather than compound actions.

nriley (changes requested):

splondike (approved):

rntz:
It's a bit confusing that `<action> way left` and `<action> line start` do different things with this PR. Eg `select`, `copy`, etc.

This follows the differences introduced in #1182, I think.

The difference is: `select line start` tries to cancel the current selection if there is one before going to head of line, while `select way left` tries to extend the current selection.

We should discuss this further.

knausj85:
At this point, I think this is probably best reimplemented via the compound edit commands implemented in https://github.com/talonhub/community/pull/1498. Closing this one out.

PR #1194, opened 2023-05-13
Added mode indicator

AndreasArvidsson:
Screenshot 2023-05-13T18-04-21

### DIFF
--- a/core/modes/mode_indicator.py
+++ b/core/modes/mode_indicator.py
@@ -1,0 +1,172 @@
+from talon import app, registry, scope, ui, Module
+from talon.canvas import Canvas
+from talon.skia.canvas import Canvas as SkiaCanvas
+from talon.skia.imagefilter import ImageFilter
+from talon.screen import Screen
+from talon.ui import Rect
+
+prefix = "mode_indicator"
+canvas: Canvas = None
+current_mode = ""
+mod = Module()
+
+setting_show = mod.setting(
+    f"{prefix}_show",
+    bool,
+    False,
+    "If true the mode indicator is shown",
+)
+setting_size = mod.setting(
+    f"{prefix}_size",
+    float,
+    30,
+    "Mode indicator diameter in pixels",
+)
+setting_x = mod.setting(
+    f"{prefix}_x",
+    float,
+    None,
+    "Mode indicator center X-position in percentages(0-1)",
+)
+setting_y = mod.setting(
+    f"{prefix}_y",
+    float,
+    None,
+    "Mode indicator center Y-position in percentages(0-1)",
+)
+setting_color_sleep = mod.setting(
+    f"{prefix}_color_sleep",
+    str,
+    "808080",  # Grey
+)
+setting_color_dictation = mod.setting(
+    f"{prefix}_color_dictation",
+    str,
+    "da70d6",  # Orchid
+)
+setting_color_mixed = mod.setting(
+    f"{prefix}_color_mixed",
+    str,
+    "6b8e23",  # OliveDrab
+)
+setting_color_other = mod.setting(
+    f"{prefix}_color_other",
+    str,
+    "f8f8ff",  # GhostWhite
+)
+setting_color_alpha = mod.setting(
+    f"{prefix}_color_alpha",
+    float,
+    0.5,
+    "Mode indicator alpha/opacity in percentages(0-1)",
+)
+
+setting_paths = {
+    s.path
+    for s in [
+        setting_show,
+        setting_size,
+        setting_x,
+        setting_y,
+        setting_color_sleep,
+        setting_color_dictation,
+        setting_color_mixed,
+        setting_color_other,
+        setting_color_alpha,
+    ]
+}
+
+
+def get_color() -> str:
+    if current_mode == "sleep":
+        color = setting_color_sleep.get()
+    elif current_mode == "dictation":
+        color = setting_color_dictation.get()
+    elif current_mode == "mixed":
+        color = setting_color_mixed.get()
+    else:
+        color = setting_color_other.get()
+    alpha = f"{int(setting_color_alpha.get() * 255):02x}"
+    return color + alpha
+
+
+def on_draw(c: SkiaCanvas):
+    c.paint.style = c.paint.Style.FILL
+    c.paint.color = get_color()
+    c.paint.imagefilter = ImageFilter.drop_shadow(1, 1, 1, 1, "000000")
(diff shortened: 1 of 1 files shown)


core/modes/mode_indicator.py
```diff
@@ -0,0 +1,172 @@
...
+    if current_mode == "sleep":
+        color = setting_color_sleep.get()
+    elif current_mode == "dictation":
+        color = setting_color_dictation.get()
+    elif current_mode == "mixed":
+        color = setting_color_mixed.get()
+    else:
+        color = setting_color_other.get()
+    alpha = f"{int(setting_color_alpha.get() * 255):02x}"
+    return color + alpha
```
auscompgeek, line 90: nit:
```suggestion
    return f"{color}{int(setting_color_alpha.get() * 255):02x}"
```

plugin/mode_indicator/mode_indicator.py
```diff
@@ -0,0 +1,149 @@
...
+
+canvas: Canvas = None
+current_mode = ""
+mod = Module()
+
+setting_show = mod.setting(
+    "mode_indicator_show",
+    bool,
+    desc="If true the mode indicator is shown",
+)
```
rntz, line 16:
```suggestion
setting_show = mod.setting(
    "mode_indicator_show",
    bool,
    desc="If true the mode indicator is shown",
    default=False,
)
```

rntz (commented):
Looks sensible, but I need to test it and I'm away from computer atm. Probably want a default=False for the activation setting so that if the .talon file gets deleted (a thing users are supposed to be able to do) the Python file doesn't start erroring.

plugin/mode_indicator/mode_indicator.py
```diff
@@ -0,0 +1,149 @@
...
+def on_draw(c: SkiaCanvas):
+    c.paint.style = c.paint.Style.FILL
+    c.paint.color = get_color()
+    c.paint.imagefilter = ImageFilter.drop_shadow(1, 1, 1, 1, "000000")
+    c.draw_circle(c.rect.center.x, c.rect.center.y, c.rect.height / 2 - 2)
+
+
+def move_indicator():
+    screen: Screen = ui.main_screen()
+    rect = screen.rect
```
pokey, line 79: I'd call this `screen_rect`

plugin/mode_indicator/mode_indicator.py
```diff
@@ -0,0 +1,149 @@
...
+)
+setting_color_alpha = mod.setting(
+    "mode_indicator_color_alpha",
+    float,
+    desc="Mode indicator alpha/opacity in percentages(0-1). 0=fully transparent, 1=fully opaque",
+)
+setting_color_sleep = mod.setting("mode_indicator_color_sleep", str)
+setting_color_dictation = mod.setting("mode_indicator_color_dictation", str)
+setting_color_mixed = mod.setting("mode_indicator_color_mixed", str)
+setting_color_other = mod.setting("mode_indicator_color_other", str)
```
pokey, line 40: Was kind of expecting this to be called "command", but I guess your name is more accurate?

pokey (approved):
Haven't tested but looks good to me. Gradient would be nice but prob don't need it to be that fancy

pokey:
> Probably want a default=False for the activation setting so that if the .talon file gets deleted (a thing users are supposed to be able to do) the Python file doesn't start erroring.

Yeah good point. I think the standard thing here would be to use defaults for everything and then have the settings commented out in the settings file, but I think just doing that for activation setting is reasonable. Then can have a line that's commented out in the settings file that will activate it when I commented

plugin/mode_indicator/mode_indicator.py
AndreasArvidsson: Yeah the logic is implemented that this is the else condition. For example if you have a game mode or something else. Without that piece of logic the mode indicator would have an undefined behavior for those kind of scenarios. The comment in the Talon file above that setting mentions command mode.

AndreasArvidsson:
> > Probably want a default=False for the activation setting so that if the .talon file gets deleted (a thing users are supposed to be able to do) the Python file doesn't start erroring.
> 
> Yeah good point. I think the standard thing here would be to use defaults for everything and then have the settings commented out in the settings file, but I think just doing that for activation setting is reasonable. Then can have a line that's commented out in the settings file that will activate it when uncommented

Initially I had all the defaults in the python file but after a lengthy discussion with @phillco we decided that the Talon file and the python file is part of the plugin. If you don't want this you can just delete the entire folder.

AndreasArvidsson:
@pokey Gradient added as per your request :D

plugin/mode_indicator/mode_indicator.talon
```diff
@@ -0,0 +1,23 @@
+settings():
+    # Don't show mode indicator by default
+    user.mode_indicator_show = 0
+    # 30pixels diameter
```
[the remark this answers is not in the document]
AndreasArvidsson: I'm still not sure that point is more clear to people.

AndreasArvidsson:
@lunixbochs Looks like screen scaling is applied automatically on mac, but not on windows. Is that a bug or something that we should work with?

nriley:
Another Mac versus Windows difference that results in inconsistent behavior in the mode indicator appears to be the corner of the window that stays fixed when resizing a canvas:

Mac:
```py
>>> from talon.canvas import Canvas; canvas = Canvas.from_rect(Rect(0,0,0,0)); canvas.resize(20,20); canvas.rect
Rect(0.0, -20.0, 20.0, 20.0)
```

Windows:
```py
>>> from talon.canvas import Canvas; canvas = Canvas.from_rect(Rect(0,0,0,0)); canvas.resize(20,20); canvas.rect
Rect(0.0, 0.0, 20.0, 20.0)
```

@lunixbochs — should this behavior be consistent cross-platform in core Talon or should we be accommodating it in our Python code?

plugin/mode_indicator/mode_indicator.talon
```diff
@@ -0,0 +1,23 @@
+settings():
+    # Don't show mode indicator by default
+    user.mode_indicator_show = false
+    # 30pixels diameter
+    user.mode_indicator_size = 30
```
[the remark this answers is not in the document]
AndreasArvidsson: If you try it out now I think this should be fixed

AndreasArvidsson:
@nriley I have now updated it so it resizes before move. Please try it out again.

PR #1200, opened 2023-05-25
command to allow focusing on previously focused app

saidelike:


JWaldenback:
It would be cool to make this more like how the browser remembers what pages a particular tab has visited. And when you go back two pages by clicking on the back arrow then you can still go forth two pages to return to the page you were on (Windows Explorer works the same). Not sure how they technically implement it, maybe it's way more difficult to achieve, thus not viable for this function in Talon?

AndreasArvidsson:
We had a look at this on today's meet up with @phillco and @nriley 
This won't really work if the user uses a mouse, a eye tracker, keyboard commands or any other way to focus windows.
Personally my implementation is just `actions.key("alt-tab")`. Left operating system take care of which window was last focused. That only needs a mac specific implementation and we are done.

JWaldenback:
@AndreasArvidsson agree. Better to let the operating system take care of this

saidelike:
@AndreasArvidsson I agree with you actually, thanks for the feedback. I guess the proposed PR would work even if the OS does not support hotkeys but I know at least Ubuntu and Windows support ALT+TAB and ya the fact this PR would not work if we click with eye tracker is not great indeed. So I guess we close this and never merge.

saidelike:
I've pushed a new version that works 100% on Windows. I don''t have a mac so can't really test. We could merge now the code as-is and add support for mac later.

AndreasArvidsson:
Much better, but `if app.platform == "windows":` it's not a good pattern. We should have context specific actions for different operating systems.

saidelike:
I agree having code based on different context like in https://github.com/knausj85/knausj_talon/blob/main/apps/vscode/vscode.py would make sense but it is a different problem that what this PR is about imho. We should probably track it as a different issue which is to refactor app_switcher.py?

saidelike:
One thing to consider if we want to split in 2 contexts is we would have to probably duplicate some logic for some of the user actions where most of the code is the same for the different operating systems except some special cases which are currently being treated with just an "if".

Also, I haven't tested and can't without a Mac but I have been told this shortcut should work on Mac: `actions.key("cmd-h")`

core/app_switcher/app_switcher.py
```diff
@@ -254,6 +254,13 @@ def get_running_app(name: str) -> ui.App:
                 return application
         raise RuntimeError(f'App not running: "{name}"')
 
+    def switch_last_focused():
```
splondike, line 257: Having an if statement here instead of a new context seems OK to me for now.

I'd just suggest we change the action name to switcher_focus_last_window() to better match with the existing action names.

knausj85:
Oops, looks like a very similar pull request was opened and already merged. https://github.com/talonhub/community/commit/d260879def9b0c8437cfc965b260e0966176e252

As such, closing this one out. Sorry about that

saidelike:
Awesome, the way it is done is better!

PR #1204, opened 2023-05-27
simple_keys, add {super, alt}

jmegner:
It is useful to simulate a super key press to toggle the start menu. Likewise, it is useful to simulate an alt key press to toggle showing of keyboard shortcuts or alt modes in apps.

### DIFF
--- a/core/keys/keys.py
+++ b/core/keys/keys.py
@@ -244,6 +244,8 @@ simple_keys = [
     "pageup",
     "space",
     "tab",
+    "super",
+    "alt",
 ]
 
 alternate_keys = {


splondike (approved):
I was a little concerned this might interfere with the "<modifier> <key>" command, but it doesn't seem to on my testing. Looks good to me. If you bring it up to date I'm happy to merge it after leaving another week for other folks to comment.

rntz:
I'm nervous about `alt` as a bare command word. Too close to `odd`?

jmegner:
>Too close to odd in some accents

You're right.  I get false positives for alt more often than is like.  I think the existing "press alt" is good enough.

I didn't know about the "press" command when I originally made these changes for my personal repo.  I now think this PR should be rejected.

PR #1205, opened 2023-05-28
add "rounded" formatter for paren surrounded string

jmegner:
I am not married to the "rounded" formatter and surround2 names.  I could even change `surround` to take an optional second argument.

### DIFF
--- a/core/text/formatters.py
+++ b/core/text/formatters.py
@@ -71,6 +71,10 @@ def surround(by):
     return lambda i, word, last: (by if i == 0 else "") + word + (by if last else "")
 
 
+def surround2(prefix, postfix):
+    return lambda i, word, last: (prefix if i == 0 else "") + word + (postfix if last else "")
+
+
 def words_with_joiner(joiner):
     """Pass through words unchanged, but add a separator between them."""
     return (NOSEP, lambda i, word, _: ("" if i == 0 else joiner) + word)
@@ -132,6 +136,7 @@ formatters_dict = {
         if i == 0 or word not in words_to_keep_lowercase
         else word,
     ),
+    "PAREN_SURROUNDED_STRING": (SEP, surround2("(", ")")),
 }
 
 # This is the mapping from spoken phrases to formatters
@@ -151,6 +156,7 @@ formatters_words = {
     "snake": formatters_dict["SNAKE_CASE"],
     "string": formatters_dict["SINGLE_QUOTED_STRING"],
     "title": formatters_dict["CAPITALIZE_ALL_WORDS"],
+    "rounded": formatters_dict["PAREN_SURROUNDED_STRING"],
 }
 
 all_formatters = {}
--- a/core/text/formatters.py
+++ b/core/text/formatters.py
@@ -72,7 +72,11 @@ def surround(by):
 
 
 def surround2(prefix, postfix):
-    return lambda i, word, last: (prefix if i == 0 else "") + word + (postfix if last else "")
+    return (
+        lambda i, word, last: (prefix if i == 0 else "")
+        + word
+        + (postfix if last else "")
+    )
 
 
 def words_with_joiner(joiner):


splondike:
I'm a bit disinclined to add this formatter since it doesn't do something you can't do pretty straightforwardly with the existing commands and adds a new global prefix. Currently you'd have to say 'args say in parentheses', with this you'd instead say 'rounded in parentheses'. Do you think the slightly shorter prefix is worthwhile?

splondike:
I'll let you take a look into the chaining.

I did some archaeology, and it turns out nobody's added a new formatter since knausj_talon was created back in early 2020 :D. I view adding new commands (particularly global ones that add a new prefix) as a tradeoff between usefulness and discoverability/misrecognitions/prefix takeover. You can see another tradeoff case [here](https://github.com/knausj85/knausj_talon/pull/881#issuecomment-1172806188). I probably wouldn't have had several of the formatters (e.g. padded, slasher, dunder).

We haven't discussed whether others feel the same way though to my knowledge. When we do have these conversations on general topics we're hoping to put the results here: https://github.com/knausj85/knausj_talon/blob/main/CONTRIBUTING.md .

In any case if you think this addition is significantly valuable vs the alternative we can put it in.

auscompgeek:
FWIW there was [a recent thread where someone had trouble using the formatters to insert `StringValue`](https://talonvoice.slack.com/archives/C7ENXA7C4/p1688824521475389?thread_ts=1688823777.628919&cid=C7ENXA7C4), since `string` is also a formatter that can be chained.

jmegner:
I'm back from traveling and food poisoning, and I still have to do research so I can understand the pros and cons of [surround-based-formatters](https://github.com/talonhub/community/blob/7525147851e159ddb6b3d40da7f8ffce5a6e6deb/core/text/formatters.py#L122-L124) vs [empty/inside commands](https://github.com/talonhub/community/blob/7525147851e159ddb6b3d40da7f8ffce5a6e6deb/plugin/symbols/symbols.talon#L10-L20).

Existing surround-based-formatters:
- padded, spaces
- string, single quotes
- dub string, double quotes 

Existing1 empty/inside commands:
```
empty dub string: user.insert_between('"', '"')
empty escaped (dub string | dub quotes): user.insert_between('\\"', '\\"')
empty string: user.insert_between("'", "'")
empty escaped string: user.insert_between("\\'", "\\'")
(inside parens | args): user.insert_between("(", ")")
inside (squares | square brackets | list): user.insert_between("[", "]")
inside (bracket | braces): user.insert_between("{", "}")
inside percent: user.insert_between("%", "%")
inside (quotes | string): user.insert_between("'", "'")
inside (double quotes | dub quotes): user.insert_between('"', '"')
inside (graves | back ticks): user.insert_between("`", "`")
```

Empty/inside commands are far more numerous than surround formatters.  So, internal consistency is in favor of inside commands. (Though really the empty commands should use the inside terminology.)

For simple usage, we have "rounded some words here" vs "inside parens say some words here".  The inside way is more verbose.  The formatter way uses up valuable first-two-syllables recognition space.

For usage that is equivalent to multiple formatters, we have "rounded dub string kebab some words here" vs "inside parens inside dub quotes kebab some words here" to get `("some-words-here")`.  Both seem to stack well.  The inside way is more verbose.  The inside way is more friendly to doing things and pieces rather than racing against the timeout clock so much.

I am very open to the idea that I am missing some trickier consequence of formatter vs inside command, but as far as I can tell, the inside command way is fine other than being a bit more verbose and requiring some getting used to.  Perhaps this issue should be closed and I should make other issues for...
- Getting rid of or changing empty commands. Ex: "empty string" and "inside string" do the same thing.  Probably add "string" as alternative for "quotes" in "insert dub quotes" command.
- Change "string" formatter to "sig string" formatter because it screws up formatters when you want to format something like "StringValue" (from "hammer string value").

splondike:
The above proposal to close out this PR and make a couple of issues sounds reasonable. Just an FYI we have a fairly long deprecation period when removing commands: https://github.com/talonhub/community/blob/main/core/deprecations.py#L6 .

jmegner:
I just thought of another difference between inside commands and surround formatters.  Inside commands are more awkward to continue typing stuff afterwards because you have to move over the surrounding characters.

If you say "rounded dub string hello there over, space, say apple banana", you get `("hello there") apple banana`.  If you want the same thing with inside commands, you would say "inside parens, inside dub quotes, say hello there over, go right second, space, say apple banana".

jmegner:
Perhaps an even better way is a series of "after" commands.  They are like the inside commands, but leave your cursor after stuff.

ex: `after parens [<user.text>] [over]: '({text or ""}) '`

One problem with the above is that it is inconsistent with the inside commands.  To get `(stuff) more words`, you do "inside parens, **say stuff**, tail, say more words" and using my example command would be "after parens **stuff over**, say more words".

jmegner:
I previously forgot to bring up a big advantage of formatters: there are a lot of commands that use formatters, like formatting selected text, reformatting recently inserted text, formatting a homophone choice, and applying a format in cursorless on a target.

Adding another formatter gives you all of that other functionality for free.  If you add an after/inside command, you only get that exact functionality.  To achieve functionality parity with a formatter, you need the following flavors:
- inside
- after
- surround selection
- surround recently inserted text
- surround homophone choice
- surround cursorless target

Maybe we should have a `<user.surround>` capture so that we can have generic `inside/after/etc <user.surround>: ...` commands rather than each surround character set needing more commands.

Also, if we don't like that each formatter takes up precious beginning-syllables-of-command space, then the commands should start with `form <user.formatters>` rather that starting with the formatter.

splondike:
Those are a few extra features you get with formatters for sure. I think I can see the rounded formatter being more useful for prose than programming.

In prose you'd probably use them like `say I can jump tall buildings over args say in my dreams over go right dot`. Versus `say I can jump tall buildings over rounded in my dreams over dot`. So that is better. Do you imagine you'd want to go back and parenthesise some text you'd spelt out previously using e.g. the reformatter?

In programming you probably want to stay within the parentheses in order to put together multiple arguments. So the args form would probably be more useful. You might want to use a cursorless target to wrap arithmetic operators to deal with precedence issues (e.g. `1 + 2 * 3 vs (1 + 2) * 3`), but I think it already has that built in: https://www.cursorless.org/docs/#wraprewrap

PR #1206, opened 2023-05-28
add `control off` command for mouse

jmegner:


splondike (approved):
If I understand this correctly, this command would turn off eye tracking and unhide the cursor if the user is doing that. This sounds useful.

I don't have an eye tracker however, so I might leave this another week for somebody who does to comment before I merge it.

splondike (approved):
Thanks!

PR #1215, opened 2023-06-17
modes: simplify modes a little bit

phillco:
https://github.com/knausj85/knausj_talon/issues/1122

We want to reduce the use of `mode:all` and instead have files that specifically target specific modes. This breaks up the existing sleep files to better capture their semantics and adds some comments.


AndreasArvidsson (approved):

rntz:
work work work work

phillco:
Working hard!

splondike (approved):
This looks good and appears to work (at least for the non-Dragon case). Happy to merge it in if you're ready.

knausj85 (approved):

PR #1216, opened 2023-06-17
Improve capitalization and other prose formatters

nriley:
- Use Python's `.title()` to better handle hyphenated words.
- Make `CAPITALIZE_ALL_WORDS` (should really be named something like `TITLE_CASE_ALL_WORDS`) a prose formatter and extend it to handle punctuation.
- Don't change case of words that already include capital letters.
- Add tests.


brief:
I like this!

The challenge with `title()` is [possessives and contractions](https://docs.python.org/3/library/stdtypes.html?highlight=title#str.title):

> they're bill's friends from the UK

Becomes:

> They'Re Bill'S Friends From The UK

Looks like this also removes `title` from `formatters_words`, which means `formatted title <text>` no longer works in dictation mode. Is there another way to call it in that mode as a prose formatter?

brief:
A few other things I've been mulling over for my own repo:

- According to all major style guides, the last word of a title should be capitalized, even if it is typically lowercased. So `There is Nothing to Be Afraid of` would be `There is Nothing to Be Afraid Of`. Worth doing here? Or is it more likely to confuse?
- The community small words list includes `is` and `up`, which really shouldn't be on the list. It's missing `en`, `if`, `per`, and `via`, which typically are. Plus `v` and `vs` from a legal context (though I expect most people would only use those if they're overriding `versus` in `words_to_replace.csv`).
- I like what this PR does with hyphenated words. Can we be even smarter? `man-in-the-middle` should be capitalized as `Man-in-the-Middle` but `stand-by` as `Stand-By`.

nriley:
@brief Thanks for the feedback.

Think I have fixed everything except for the issue with the prose formatter not being applicable in dictation mode. I also noticed that seemed to be a problem when applying a formatter from Cursorless or using "format help" — they completely ignore the prose formatters.

Can anybody think of a reason why prose formatters should not be available everywhere (code) formatters are?

brief:
@nriley `up` is still included in the list of small words. Looks great otherwise.

Assuming prose formatters should be available everywhere, which I agree with, `capitalize()` introduces an edge case with quoted strings (or any word that's prefixed by a non-alpha character).

> 'hello world' and "hello world"

Become:

> '**h**ello World' and "**h**ello World"

Instead of:

> '**H**ello World' and "**H**ello World"

Only other character that might matter in prose is a left parenthesis (e.g. `(Hello World)`)?

nriley:
Updated; it should handle punctuation at the beginning of words.

There's a bunch of nuances with prose formatters I still need to sort out — for example, should you be able to combine code and prose formatters (e.g., "kebab title")?

brief:
Thanks. This looks good to me.

Re: combining code and prose formatters, I'm not entirely sure how they differ. Why are there two different types?

nriley:
Thanks! There are three main differences between prose and code formatters:

1) When dictating with a prose formatter, punctuation and symbols and will work just like you are in dictation mode, including automatic spacing and capitalization. For example, if I am in command mode and say:

    - "phrase this is comma a test" → "this is comma a test" [no formatter with user.text]
    - "all down this is comma a test" → "this is,a test" [code formatter]
    - "kebab this is comma a test" → "this-is,a-test" [code formatter]
    - "say this is comma a test" → "this is, a test" [prose formatter]

2) You can compose code formatters (order is significant).

    - "kebab hammer this is a test" → "This-Is-A-Test" [2 code formatters]
    - "hammer kebab this is a test" → "This-is-a-test" [2 code formatters]

3) You can use code formatters in other places (e.g., reformatting with `<user.formatters> that`)

Most of the changes I made to "title" will be of limited use if it remains a code formatter, **unless** you're using it for reformatting of text you might have inserted with a prose formatter.

My main concern would be if people are composing "title" with other code formatters, e.g. using "kebab title this is a test", which in stock knausj returns "This-is-a-Test", but won't work with this PR is currently.

If that is not a major concern, what I think will work is replacing the definition of `user.format_text` so it only contains code formatters then expanding `user.formatters` to include prose formatters.

brief:
Wow, thank you for the detailed explanation!

Curious to know what the feedback will be on composing "title" with other code formatters. No concerns here, but I'm the wrong person to ask. I only ever use it in the context of prose and prefer "hammer" when composing since I typically want to cap the start of every word.

nriley:
OK, I think I've got a reasonable solution here. You can compose prose and code formatters, but only when reformatting. Feedback appreciated!

brief:
Looks good on my end.

AndreasArvidsson (approved):

PR #1218, opened 2023-06-21
Make the tmux modifier key configurable

antonio:
Make it possible to configure the modifier key for the people that use alternative keybindings (like I do with `alt-a` :smile:).

This pull request also updates how windows are switched to use the keybind and numbers from 0 to 9 if possible, for a slightly faster user experience.


splondike (approved):
Seems to work for me, thanks!

PR #1219, opened 2023-06-23
Allow overriding the <phrase> and <word> used in dictation and text

tbodt:


rntz:
Could you explain your use-case for this?

This modifies a pretty core part of knausj so we probably want to be a little cautious about it, just to be sure that it's useful, worth the added layer of indirection, and doesn't cause any performance or correctness issues. I think we'd also want aegis/Ryan to sign off on this.

tbodt:
This is to convince English conformer to recognize other languages. I found that overriding these captures with a list of toki pona words is enough to get usable output.

lunixbochs:
this definitely has a chance to cause correctness or quadratic parsing issues, I think it's a small enough change that you can probably just make it in a fork rather than upstreaming it

knausj85:
It sounds like this it's best to keep this one in a fork for now. Closing.

PR #1225, opened 2023-07-08
typescript: Add type operation commands

auscompgeek:
This adds a few commands for some TypeScript type operations:

- `type union/intersect` after a type to union or intersect it with another type
  - example: `type union undefined` will make a type possibly undefined ` | undefined`
- `state type` to start a `type` statement
- `as const` to insert `as const` at the end of an expression


splondike (approved):
Thanks!

PR #1197, opened 2023-05-23
add support for launching Linux apps

gband85:
Added ability to launch Linux apps. Parses .desktop shortcut files for app launch commands. Correctly parses arguments and passes them to ui.launch.

PR #1235, opened 2023-07-16
fix: Handle spaces for 'launch <appname>' in OSX

splondike:
Regression from https://github.com/talonhub/community/pull/1197 . Issue is the code will populate the self.launch list with something like:
```
{
  'my app': '/Applications/My App.app'
}
```
Then later in `actions.user.switcher_launch` `shlex.split` would turn that into `cmd = '/Applications/My', args='App.app'`. Quoting the path makes it treat the whole path as one unit.

### DIFF
--- a/core/app_switcher/app_switcher.py
+++ b/core/app_switcher/app_switcher.py
@@ -327,12 +327,15 @@ class Actions:
 
     def switcher_launch(path: str):
         """Launch a new application by path (all OSes), or AppUserModel_ID path on Windows"""
-        if app.platform != "windows":
-            # separate command and arguments
+        if app.platform == "mac":
+            ui.launch(path=path)
+        elif app.platform == "linux":
+            # Could potentially be merged with OSX code. Done in this explicit
+            # way for expediency around the 0.4 release.
             cmd = shlex.split(path)[0]
             args = shlex.split(path)[1:]
             ui.launch(path=cmd, args=args)
-        else:
+        elif app.platform == "windows":
             is_valid_path = False
             try:
                 current_path = Path(path)
@@ -344,6 +347,8 @@ class Actions:
             else:
                 cmd = f"explorer.exe shell:AppsFolder\\{path}"
                 subprocess.Popen(cmd, shell=False)
+        else:
+            print("Unhandled platform in switcher_launch: " + app.platform)
 
     def switcher_menu():
         """Open a menu of running apps to switch to"""
@@ -384,7 +389,7 @@ def update_launch_list():
                 for name in os.listdir(base):
                     path = os.path.join(base, name)
                     name = name.rsplit(".", 1)[0].lower()
-                    launch[name] = f"'{path}'"
+                    launch[name] = path
 
     elif app.platform == "windows":
         launch = get_windows_apps()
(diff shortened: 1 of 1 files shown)


lunixbochs:
I'm not sure you should quote in the list rather than on use, unless you're putting cli args in the list

knausj85:
I wonder if it would be more robust to create an overload of launch that explicitly takes args, and call it only for Linux in the existing action.

It's pretty weird for the action to try to extract args, but indicate it takes only a path. At a minimum, this should probably be reflected in the doc string. 

It's also creating a discrepancy in how it functions across OSes as-is, which is a little unfortunate.

splondike:
Yeah, the quoting in the list is because Linux .desktop files have args for the command in them sometimes.

This action is only used for this launch functionality. You could be right that being more explicit about the arguments would be better. Lists only take strings as values though, so we'd have to pack all that in using JSON or something.

So I think we could lean in to making this a single use action and just rename path to something like 'command'. Or we could try to make the thing generic and have a JSON string get passed in. In the second case maybe also have the switcher_launch function call out to a function like 'open_gui_app(path: str, args: List[str])'. I'm inclined toward just renaming path to command.

See also discussion here about the API for an action for launching background commands: https://github.com/talonhub/community/issues/945

knausj85:
Simply renaming to command makes sense to me 👍

rntz:
Shouldn't you be using `shlex.escape` for this rather than just throwing single quotes on? The latter is rather fragile.

lunixbochs:
Using shlex is outright incorrect on windows. This whole thing also makes me slightly uneasy security wise fwiw.

phillco:
If there isn't an a obvious consensus on this, should we just quickly revert https://github.com/talonhub/community/pull/1197 for now? Not being able to launch applications with spaces on macOS is a pretty decent regression.

splondike:
Windows doesn't use the shlex.split branch: https://github.com/talonhub/community/blob/207c6a132a885fbd82fd513a993626b6af9b4a25/core/app_switcher/app_switcher.py#L330

It sounds like folks might be most comfortable with having the list population code do the path/arg extraction in an OS specific fashion and putting the executable and its arguments in as a JSON string. This way the necessarily OS specific part of the code could work out the finer details and let the `switcher_launch` action be more generic. So you'd have something like this in the list:

```
{
  'my app': '{"path": "/my/thing", "args": ["arg1", "arg2"]}'
}
```
The action on OSX and Linux would just decode that and pass the two parts directly into the arguments of `ui.launch`. Windows would keep its current behaviour of just giving the path to explorer.exe.

nriley:
Particularly with the impending release of Talon 0.4, I think it makes sense to revert #1197 until we can get OS-specific versions of this nailed down.

phillco:
Ah, #1197 doesn't revert cleanly because there have been subsequent changes.

splondike:
Made a new implementation and tested it on OSX, Linux, and Windows. It is probably not the most elegant, but it does the job alright. I didn't end up going with the JSON idea because the Windows code path is doing something a bit odd in this regard.

knausj85 (approved):

PR #1234, opened 2023-07-15
Talon v0.4 support + some related improvements

knausj85:
- Convert many lists to .talon-list to jumpstart the process / discussion. We may need some sort of migration plan.
- Support for common commands when composing talon-list, .talon, and talon .py

Note: does not port to resource.watch (see https://github.com/talonhub/community/pull/1199)

Questions:
I believe we'll want a migration script for CSVs that are moved to .talon-list: reads the CSV, spits out a talon-list. 

- Should CSVs migrated to talon-lists migrated continue to live in Settings?
- do we want to auto regenerate a talon list if it's deleted as was done for the CSVs? 
- is there a 'good' way to migrate targeted lists that exist only in python without breaking every customization?

PR #1199, opened 2023-05-23
port from resource.open to resource.watch (Talon v0.4)

lunixbochs:
This change for Talon v0.4 should substantially improve the stability of knausj_talon around first install and CSV loading.

PR #1239, opened 2023-07-24
Talon v0.4 support: port from resource.open to resource.watch & initial talon-list conversions

knausj85:
Splitting https://github.com/talonhub/community/pull/1234 into a few pull requests for convenience. This is not fully vetted yet, opening a draft for convenience and initial discussion. 

- port from resource.open to resource.watch (Talon v0.4) (merge in https://github.com/talonhub/community/pull/1199)
- Convert many lists to .talon-list to jumpstart the process / discussion. I did not attempt to convert them all, just the highest priority targets and a few interesting ones for discussion. We need some sort of migration plan, as I would like to move away from CSVs wherever possible.  
- Also adds support for keypad keys, as it wasn't worth the effort to separate into a separate pull request
 
todo:
- [ ] Fix unit tests for create spoken forms
- [ ] Create a CSV to .talon-list converter for targeted conversions
- [ ] Verify emacs functionality

Questions:
1. Should CSVs migrated to talon-lists migrated continue to live in Settings?
2. do we want to auto regenerate a talon list if it's deleted as was done for the CSVs?
3. is there a 'good' way to migrate targeted lists that exist only in python without breaking every customization?


core/keys/punctuation.talon-list
```diff
@@ -0,0 +1,29 @@
+list: user.punctuation
+-
+`: `
```
knausj85, line 3: This can be moved to a dragon specific punctuation list.

core/keys/function_key.talon-list
```diff
@@ -0,0 +1,26 @@
...
+f fifteen: f15
+f sixteen: f16
+f seventeen: f17
+f eighteen: f18
+f nineteen: f19
+f twenty: f20
+f twenty one: f21
+f twenty two: f22
+f twenty three: f23
+f twenty four: f24
```
knausj85, line 26: I don't know if anyone overrides numbers, but would be a little unfortunate have to find all instances of them in .talon-lists if so. I left this here for discussion.

core/keys/keys.py
```diff
@@ -2,28 +2,17 @@
...
-
-    return initial_default_alphabet_dict
-
-
-alphabet_list = get_list_from_csv(
-    "alphabet.csv", ("Letter", "Spoken Form"), setup_default_alphabet()
+default_digits = "zero one two three four five six seven eight nine".split(" ")
+numbers = [str(i) for i in range(10)]
+default_f_digits = (
+    "one two three four five six seven eight nine ten eleven twelve".split(" ")
```
knausj85, line 8: Merge error, can be removed.

plugin/text_navigation/navigation_target_name.talon-list
```diff
@@ -0,0 +1,16 @@
+#todo: if we wanted to convert this
+# we'd probably want to rework the list to have hard-coded strings
+# like e.g. BRACES on the right side.
+# list: user.navigation_target_name
+# -
```
knausj85, line 5: I'm leaving a note here for discussion purposes. I don't think this should be converted as is.

core/keys/arrow_key.talon-list
```diff
@@ -0,0 +1,6 @@
+list: user.arrow_key
+-
+down: down
+left: left
+right: right
+up: up
```
lunixbochs, line 6: You can just do `up` instead of `up: up` in all of these identity mapped lists. With up: up it's actually kind of confusing which side to edit too.

core/keys/function_key.talon-list
lunixbochs: I think it's over abstraction to parameterize the number part of your f-keys like that.

plugin/talon_helpers/talon_helpers.talon
```diff
@@ -36,6 +37,7 @@ talon dump context:
     user.talon_action_find("{user.talon_actions}")
 ^talon debug list {user.talon_lists}$: user.talon_debug_list(talon_lists)
 ^talon copy list {user.talon_lists}$: user.talon_copy_list(talon_lists)
+^talon convert list {user.talon_lists}$: user.talon_convert_list(user.talon_lists)
```
knausj85, line 40: Probably remove this

core/keys/function_key.talon-list
knausj85: I tend to agree.

core/keys/arrow_key.talon-list
knausj85: It's arguably a bit more convenient for those that may customize to keep up: up though.

This is a similar problem to eg https://github.com/talonhub/community/blob/7b09a4a789ed1dc380865d2840a5abda9f65af25/core/windows_and_tabs/window_snap_positions.talon-list

where we may be better off with something more obviously not a spoken form on the right side.

lunixbochs: I think it's not more convenient, because it's not obvious which side to edit when the sides are symmetrical.

core/text/prose_snippets.talon-list
```diff
@@ -0,0 +1,15 @@
...
+    """
+new paragraph":
+    """
+
+    """
+open quote: “
+close quote: ”
+smiley": :-)
+winky": ;-)
+frowny": :-(
```
auscompgeek, line 15: Typo?

```suggestion
smiley: :-)
winky: ;-)
frowny: :-(
```

core/user_settings.py
```diff
@@ -60,6 +46,48 @@ def get_list_from_csv(
...
+    def decorator(fn: CallbackT) -> CallbackT:
+        @resource.watch(str(path))
+        def on_update(f):
+            data = read_csv_list(f, headers)
+            fn(data)
+
+    return decorator
+
+
+# NOTE: this is deprecated, use @track_csv_list instead.
```
auscompgeek, line 76: TODO: raise a Deprecation/FutureWarning?

test/stubs/talon/__init__.py
```diff
@@ -184,6 +184,9 @@ class Resource:
     def open(self, path: str, mode: str = "r"):
         return open(path, mode, encoding="utf-8")
 
+    def watch(self, path: str):
+        pass
```
auscompgeek, line 188: Shouldn't this return an identity function?

```suggestion
        return lambda f: f
```

PR #1240, opened 2023-07-24
Talon v0.4 support: common commands for talon python / talon files / .talon-list files

knausj85:
Splitting https://github.com/talonhub/community/pull/1234 into a few pull requests for convenience. This is not fully vetted yet, opening a draft for convenience and initial discussion.

- Support for common commands when composing talon-list, .talon, and talon .py
- Support for the talon debug window, with commands consistent with other existing commands


lang/talon/talon-common.talon
```diff
@@ -0,0 +1,26 @@
...
+key <user.keys> over: "{keys}"
+key <user.modifiers> over: "{modifiers}"
+
+# all actions (requires uncommenting user.talon_populate_lists tag above)
+
+funk cell <number>: user.talon_code_insert_function(number - 1, "")
+funk wrap <user.code_common_function>:
+    user.code_insert_function(code_common_function, edit.selected_text(), False)
+funk wrap <number>:
+    user.talon_code_insert_function(number - 1, edit.selected_text(), False)
```
knausj85, line 26: Need to verify that these work as expected

lang/talon/talon-python-activator.talon
```diff
@@ -0,0 +1,7 @@
+app: vscode
+win.title: /— user/
+win.title: /— knausj_talon/
+win.title: / - user - Visual Studio Code/
+tag: user.python
+-
+tag(): user.talon_python
```
knausj85, line 7: tag to restrict talon-related python commands to relevant projects only. This is provided as an example for discussion.

lang/talon/talon-code_common_function.talon-list
```diff
@@ -0,0 +1,8 @@
+list: user.code_common_function
+tag: user.talon
+tag: user.talon_python
+-
+insert: insert
+key: key
+print: print
+repeat: repeat
```
knausj85, line 8: This is a useful example of how we might move all code_common_function lists to the new .talon-list format.

lang/talon/talon-context.talon
```diff
@@ -0,0 +1,21 @@
...
+    user.paste("app: {app}")
+mode require [{user.talon_modes}]:
+    mode = talon_modes or ""
+    user.paste("mode: {mode}")
+tag require [{user.talon_tags}]:
+    tag = talon_tags or ""
+    user.paste("tag: {tag}")
+host require:
+    hostname = user.talon_get_hostname()
+    user.paste("hostname: {hostname}\n")
```
knausj85, line 21: This file defines all context related commands.

lang/talon/talon-common.talon
knausj85: These appear to be unnecessary, removed.

lang/talon/talon.py
```diff
@@ -30,16 +29,20 @@
 mod.list("talon_scopes")
 mod.list("talon_modes")
 
-ctx.matches = r"""
+ctx_talon.matches = r"""
 tag: user.talon
 """
-ctx.lists["user.code_common_function"] = {
+ctx_talon.lists["user.code_common_function"] = {
```
auscompgeek, line 35: We don't want to set this list in Python next to the .talon-list right?

lang/talon/talon-context.talon
```diff
@@ -0,0 +1,21 @@
+tag: user.talonlist
+tag: user.talon
+tag: user.talon_python
+-
+#context requirements
+win require: user.paste("os: windows\n")
+mac require: user.paste("os: mac\n")
+linux require: user.paste("os: linux\n")
+title require: user.paste("win.title: ")
```
auscompgeek, line 9: Is there a reason these are `user.paste`?

lang/talon/talon.py
knausj85: I would personally be inclined to remove all default list implementations from python where possible. 

the .talon-list is more specific and will take precedence though

apps/talon/talon_debug/talon_debug.py
```diff
@@ -0,0 +1,14 @@
...
+mod.apps.talon_debug = """
+os: mac
+and app.bundle: com.talonvoice.Talon
+win.title: Talon Debug
+"""
+mod.apps.talon_debug = """
+os: windows
+and app.name: Talon
+os: windows
+and app.exe: talon.exe
```
AndreasArvidsson, line 13: We should probably add a title here as well. Otherwise this will just match to the update window and so on.

apps/talon/talon_debug/talon_debug.talon
```diff
@@ -0,0 +1,16 @@
...
+key <user.modifiers> over: "{modifiers}"
+
+action {user.talon_actions}: insert(talon_actions)
+# requires user.talon_populate_lists tag. do not use with dragon
+list {user.talon_lists}: "{talon_lists}"
+
+# requires user.talon_populate_lists tag. do not use with dragon
+capture {user.talon_captures}: "{talon_captures}"
+set {user.talon_settings}: user.paste("{talon_settings}")
+application {user.talon_apps}: user.paste("{talon_apps}")
```
AndreasArvidsson, line 16: We will probably recommend to remove user.paste and insert and just use the literal for all insertions. It's a bit confusing that we are inserting text in three different ways now.

lang/talon/talon-code_common_function.talon-list
```diff
@@ -0,0 +1,7 @@
+list: user.code_common_function
+tag: user.talon
+-
+insert: insert
+key: key
+print: print
+repeat: repeat
```
AndreasArvidsson, line 7: Talon list actually supports a single item that will be interpreted as both key and value
`insert: insert` => `insert` are equivalent

lang/talon/talon.py
```diff
@@ -73,7 +76,57 @@ def on_ready():
...
-@ctx.action_class("user")
+@mod.action_class
+class Actions:
+    def talon_code_insert_function(text: str, selection: str):
+        """inserts talon-specific action call"""
+
+    def talon_code_assert_tag(tag: str):
+        """enables tag in either python or talon files"""
+
+    def talon_code_assert_setting(setting: str):
```
AndreasArvidsson, line 87: Should we replace assert with enable in both of these actions?

lang/talon/talon-python-activator.talon
nriley: My suggestion would be to just leave this active for every python file by default, and either comment out or provide as an example your lines 1 through 4. That way beginning users will just have this active where necessary and more advanced users can adjust the targeting if they feel it is not working for them.

apps/talon/talon_debug/talon_debug.py
```diff
@@ -0,0 +1,14 @@
+from talon import Module
```
nriley, line 1: Potentially consider naming this talon_debug_window? It is a little confusing that it is an application that is not actually an application.

nriley (changes requested):

apps/talon/talon_debug/talon_debug.py
knausj85: Good catch.

I just realized this is a beta-only thing though. What's the thinking on committing this?

I'm inclined not to commit, or perhaps at least disable the 'talon debug' with a comment to that affect.

lang/talon/talon-code_common_function.talon-list
knausj85: Do we want to standardize on this approach?

apps/talon/talon_debug/talon_debug.py
nriley: Maybe we can just display a message if you try to use it? The mode doesn't do any harm if you don't have the beta.

knausj85: I generally don't like adding global commands that simply don't work. I've removed the relevant command and added a note to each file that it's beta only

nriley: Sounds good too!

lang/talon/talon-context.talon
```diff
@@ -0,0 +1,21 @@
...
+tag: user.talon_python
+-
+#context requirements
+win require: insert("os: windows\n")
+mac require: insert("os: mac\n")
+linux require: insert("os: linux\n")
+title require: insert("win.title: ")
+application [require] [{user.talon_apps}]:
+    app = talon_apps or ""
+    insert("app: {app}")
```
nriley, line 12: Consider adding `\n` here for consistency — I know you're just moving things around…

nriley (commented):
Think we're getting really close! This change is optional — could also move it to another PR.

lang/talon/talon-context.talon
knausj85: I believe we'd want to leave the cursor at the correct position when talon_apps is specified. Updated accordingly

lang/talon/talon_common.talon
```diff
@@ -0,0 +1,22 @@
...
+capture {user.talon_captures}: "<{talon_captures}>"
+
+setting {user.talon_settings}: user.talon_code_enable_setting(talon_settings)
+
+#commands for dictating key combos
+key <user.keys> over: "{keys}"
+key <user.modifiers> over: "{modifiers}"
+
+action {user.talon_actions}:
+    user.talon_code_insert_function(talon_actions, edit.selected_text())
```
nriley, line 22: This implementation is the same as code_insert_function so there is no need for an additional action to be defined here.

lang/talon/talon.py
```diff
@@ -73,7 +70,57 @@ def on_ready():
 app.register("ready", on_ready)
 
 
-@ctx.action_class("user")
+@mod.action_class
+class Actions:
+    def talon_code_insert_function(text: str, selection: str):
```
nriley, line 75: This action is unneeded — please see the review comment in the .talon file.

lang/talon/talon.py
```diff
@@ -73,7 +70,57 @@ def on_ready():
...
+        if not tag:
+            actions.edit.left()
+            actions.edit.left()
+
+    def talon_code_enable_setting(setting: str):
+        """asserts setting in either python or talon files"""
+        actions.user.paste(f'ctx.settings["{setting}"] = ')
+        if not setting:
+            actions.edit.left()
+            actions.edit.left()
```
nriley, line 120: This does not move the cursor far enough left. Consider using `actions.user.insert_between` instead.

nriley (changes requested):
This looks really good. Please let us know if you want some help with this

lang/talon/talon_common.talon
knausj85: I don't think this is correct.

The intent here is to able to use "action" command for talon-specific action calls and append "actions." as required in python and omit in .talon files.

The existing code_insert_function can then can be used for regular python functions even when the talon_python tag is active. 

I think this necessitates a separate action unless I'm missing something

nriley: Gotcha, I see now — @AndreasArvidsson and I didn't realize the the same action would be used across .talon/.py files. We had only thought of the first one. However it should still be possible to implement the action in terms of the existing user.code_insert_function, right? Just trying to eliminate duplicate implementations.

knausj85: As I recall, reusing code_insert_function in the python case by inserting "actions." before calling down creates an additional entry on the undo stack, which I felt is undesirable. I didn't see a better way.

In the .talon file case, I suppose we can do so without issue.

knausj85:
This is ready for a final review.

lang/talon/talon.py
```diff
@@ -73,7 +70,53 @@ def on_ready():
 app.register("ready", on_ready)
 
 
-@ctx.action_class("user")
+@mod.action_class
+class Actions:
+    def talon_code_insert_function(text: str, selection: str):
```
pokey, line 75: Do we want to call this `talon_code_insert_action_call`?

apps/talon/talon_debug_window/talon_debug_window.talon
pokey, line 1: we should prob paste the usual snippet here

```

# uncomment user.talon_populate_lists tag to activate talon-specific lists of actions, scopes, modes etcetera.
# Do not enable this tag with dragon, as it will be unusable.
# with conformer, the latency increase may also be unacceptable depending on your cpu
# see https://github.com/talonhub/community/issues/600
# tag(): user.talon_populate_lists
```

pokey, line 1: I couldn't get any of these commands to work. Talon recognized them once I had the `talon_populate_lists` tag on, but the window just de-focused. Typing with a keyboard worked, though

pokey, line 1: Is the idea to use these for typing into the search box?

lang/talon/talon_common.talon
pokey, line 1: do we want to paste the usual snippet here too?

```talon
# uncomment user.talon_populate_lists tag to activate talon-specific lists of actions, scopes, modes etcetera.
# Do not enable this tag with dragon, as it will be unusable.
# with conformer, the latency increase may also be unacceptable depending on your cpu
# see https://github.com/talonhub/community/issues/600
# tag(): user.talon_populate_lists
```

I wonder if we should just have one place that they activate the tag and it works in all the places?

Also, I wonder if DFA is even an issue on this one anymore

pokey (commented):
Looks good! Left a few minor comments

apps/talon/talon_debug_window/talon_debug_window.talon
knausj85: Yeah, these are only useful when in the search box at the moment. I should probably add a comment 😀

lang/talon/talon_common.talon
knausj85: I think we originally had this in the settings.talon file so it was in one place. 

I believe the intent behind the current approach was to restrict the DFA hit to more specific contexts (repl, debug window, talon files), rather than globally. It's been a minute though.

apps/talon/talon_debug_window/talon_debug_window.talon
knausj85: These seemed to work fine for me. Was this with the search box focused?

lang/talon/talon_common.talon
knausj85: > I wonder if we should just have one place that they activate the tag and it works in all the places?

I would vote to handle this bit in a separate PR if we want to revisit.

knausj85:
@nriley or @pokey - if either of you have some time for a final review for this one, I'd greatly appreciate it! I believe it's done-done.

apps/talon/talon_debug_window/talon_debug_window.talon
```diff
@@ -0,0 +1,24 @@
...
+-
+# uncomment user.talon_populate_lists tag to activate talon-specific lists of actions, scopes, modes etcetera.
+# Do not enable this tag with dragon, as it will be unusable.
+# with conformer, the latency increase may also be unacceptable depending on your cpu
+# see https://github.com/talonhub/community/issues/600
+# tag(): user.talon_populate_lists
+
+tag {user.talon_tags}: "{talon_tags}"
+
+#commands for dictating key combos
```
nriley, line 13: Would be helpful if there is a comment that explains how to use these in the context of the Talon debug window.

nriley (approved):

PR #1242, opened 2023-08-04
Added voice command to create Talon application context

AndreasArvidsson:
`"talon create app context"` will generate a new folder in `apps` with a Talon and a python file. 

Example from visual studio code: 

`apps/visual_studio_code/visual_studio_code.talon`
```talon
app: visual_studio_code
-

```

`apps/visual_studio_code/visual_studio_code.py`
```py
from talon import Module, Context, actions

mod = Module()
ctx = Context()

mod.apps.visual_studio_code = r"""
os: windows
and app.exe: Code.exe
"""

ctx.matches = r"""
os: windows
app: visual_studio_code
"""

# @mod.action_class
# class Actions:
```

### DIFF
--- a/plugin/talon_helpers/create_app_context.py
+++ b/plugin/talon_helpers/create_app_context.py
@@ -18,7 +18,7 @@ class Actions:
         app_name = get_app_name(active_app.name)
         app_dir = APPS_DIR / app_name
         filename = get_filename(app_name)
-        talon_file = app_dir / f"{filename}.talon"
+        talon_file = app_dir / f"{app_name}.talon"
         python_file = app_dir / f"{filename}.py"
 
         talon_context = get_talon_context(app_name)
--- a/plugin/talon_helpers/create_app_context.py
+++ b/plugin/talon_helpers/create_app_context.py
@@ -26,8 +26,11 @@ class Actions:
         if not app_dir.is_dir():
             os.mkdir(app_dir)
 
-        create_file(talon_file, talon_context)
-        create_file(python_file, python_context)
+        talon_file_created = create_file(talon_file, talon_context)
+        python_file_created = create_file(python_file, python_context)
+
+        if talon_file_created or python_file_created:
+            actions.user.file_manager_open_directory(str(app_dir))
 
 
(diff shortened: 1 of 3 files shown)


nriley:
Would recommend you create platform-specific .py/.talon files by default (e.g., visual_studio_code_win.py/talon). Some of us use Talon on more than one platform…

AndreasArvidsson:
Is that standard for community now? When I look in the applications folder it's quite rare.

nriley:
It's a standard we discussed when we did our last rearrangement. Since keyboard commands usually differ between platforms, and my feeling is that if you're using this feature, you may not be an advanced user enough to be able to build a cross-platform implementation, so it's better for the defaults to be "safe".

AndreasArvidsson:
Sounds reasonable

AndreasArvidsson:
@nriley Better?

nriley:
I'd recommend adding the platform suffix to the .talon file and the app: as well.

Also — if you're not watching the log, it doesn't seem like anything happens when I use this command. Perhaps consider opening the directory in the user's file browser, or at the least popping up a notification?

AndreasArvidsson:
I don't think it makes sense to have a talon file per os for a single app. Maybe We can talk about this in tomorrow's meetup?

AndreasArvidsson:
@nriley In today's meetup we decided to make the platform name an optional word in the spoken phrase. The current implementation just repeats the command I would prefer to have a list(`user.platform`) instead. What do people think?

plugin/talon_helpers/create_app_context.py
```diff
@@ -0,0 +1,98 @@
...
+
+APPS_DIR = Path(__file__).parent.parent.parent / "apps"
+
+mod = Module()
+
+
+@mod.action_class
+class Actions:
+    def talon_create_app_context(platform_suffix: str = None):
+        """Create a new python context file for the current application"""
```
nriley, line 16: Maybe include something about the directory here

nriley (changes requested):

nriley (approved):

PR #1244, opened 2023-08-05
Added hiss noise to scroll down

AndreasArvidsson:
Added Talon action `def noise_trigger_hiss(active: bool)` each call to this action is debounced to avoid it triggering during speech. By default there is a disabled mouse continuous scroll down implementation that you can enable with the following setting: `user.mouse_enable_hiss_scroll = 1`


vincvincvinc:
thanks, this will be amazingly helpful!

settings.talon
```diff
@@ -24,6 +24,9 @@ settings():
     # 2 = on but not with zoom mouse mode
     user.mouse_enable_pop_click = 1
 
+    # Uncomment the below too enabled mouse scroll down using hiss noise
```
auscompgeek, line 27:
```suggestion
    # Uncomment the below to enable mouse scroll down using hiss noise
```

knausj85:
There's a version of this that enables you to scroll up and down using eye tracking. Is there interest in integrating that as well?

AndreasArvidsson:
> There's a version of this that enables you to scroll up and down using eye tracking. Is there interest in integrating that as well?

That sounds interesting. Do you mean that you still make the hiss noise and depending on where you are looking you scroll up or down? If that's the case how about we merge this as is then in a follow up we enable `user.mouse_enable_hiss_scroll = 2` for the eye tracker based scrolling? If it has nothing to do with the hiss noise I even more think it should be a separate pr.

vincvincvinc:
I wonder if there are clever ways in which we can get scroll up as part of the same hissing UX without using an eye tracker.

Would it be hard to program a function that distinguishes between single hissing and double hissing ( Sorry, these are terms I'm making up on the spot),  By recognizing a pattern where the user first hisses  for less than a quarter of a second. And then pauses another quarter of a second and then starts a longer hiss.

 So that if I want to scroll down, I just hiss,  But if I want to scroll up, I kind of stutter before I start hissing:

hsssssssssss : scrolls down

hs  hssssssss : scrolls up

not sure how hard to program this would be, Or how hard to fine tune to make it work well.

knausj85:
> > There's a version of this that enables you to scroll up and down using eye tracking. Is there interest in integrating that as well?
> 
> 
> 
> That sounds interesting. Do you mean that you still make the hiss noise and depending on where you are looking you scroll up or down? If that's the case how about we merge this as is then in a follow up we enable `user.mouse_enable_hiss_scroll = 2` for the eye tracker based scrolling? If it has nothing to do with the hiss noise I even more think it should be a separate pr.

Exactly, scrolls via hiss up or down depending on gaze location. Though I remembered why we didn't merge it: it imports eyemouse and such, which aren't public APIs. We've been working to get rid of those.

Regardless, merging this one as-is makes sense to me.

AndreasArvidsson:
> I wonder if there are clever ways in which we can get scroll up as part of the same hissing UX without using an eye tracker.
> 
> Would it be hard to program a function that distinguishes between single hissing and double hissing ( Sorry, these are terms I'm making up on the spot), By recognizing a pattern where the user first hisses for less than a quarter of a second. And then pauses another quarter of a second and then starts a longer hiss.
> 
> So that if I want to scroll down, I just hiss, But if I want to scroll up, I kind of stutter before I start hissing:
> 
> hsssssssssss : scrolls down
> 
> hs hssssssss : scrolls up
> 
> not sure how hard to program this would be, Or how hard to fine tune to make it work well.

Probably not that hard code wise, but would probably require a lot of testing to get all the different thresholds correct. Since I have a separate noise to scroll up I am probably not the right person to build something that would require you to use it yourself for an extended amount of time.

AndreasArvidsson:
> > > There's a version of this that enables you to scroll up and down using eye tracking. Is there interest in integrating that as well?
> > 
> > 
> > That sounds interesting. Do you mean that you still make the hiss noise and depending on where you are looking you scroll up or down? If that's the case how about we merge this as is then in a follow up we enable `user.mouse_enable_hiss_scroll = 2` for the eye tracker based scrolling? If it has nothing to do with the hiss noise I even more think it should be a separate pr.
> 
> Exactly, scrolls via hiss up or down depending on gaze location. Though I remembered why we didn't merge it: it imports eyemouse and such, which aren't public APIs. We've been working to get rid of those.
> 
> Regardless, merging this one as-is makes sense to me.

Got it. Internal apis is probably something we should be hesitant about using.

splondike (approved):
Looks good to me and works well. Potentially somebody will want to customise that threshold in the future, but that can be another MR.

Thanks!

PR #1249, opened 2023-08-19
Added head and tail commands

AndreasArvidsson:


core/edit/edit.talon
```diff
@@ -24,10 +24,11 @@ go down: edit.down()
 go line start: edit.line_start()
```
nriley, line 24: head/tail fit a bit better here (which is where I have them defined…)

AndreasArvidsson: better?

nriley (approved):
Looks good to me.

PR #1251, opened 2023-08-21
libraries GUI common selection

frosk1:
There is a mechanics for using a GUI to import modules for programming languages.
Unfortunately a selection method (like for common functions) has been missing in the .talon files.
And the actual Tag has been missing for some languages like python in python.talon.

Therefore I have added a generic command into the common library_gui_open.talon file. 
And the tag into the language specific python.talon file.

If this features should be part of a common grammar, this would need to be handled in every defined language.
And additionally the user action function  "code_insert_library" would need to be defined in e.g. python.py

### DIFF
--- a/lang/python/python.talon
+++ b/lang/python/python.talon
@@ -12,6 +12,7 @@ tag(): user.code_functions_common
 tag(): user.code_keywords
 tag(): user.code_libraries
 tag(): user.code_libraries_gui
+tag(): user.code_libraries_gui_showing
 tag(): user.code_operators_array
 tag(): user.code_operators_assignment
 tag(): user.code_operators_bitwise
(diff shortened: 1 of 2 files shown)


auscompgeek:
Judging by the comment in the .talon file here I think this is intentional.

lang/python/python.talon
```diff
@@ -12,6 +12,7 @@ tag(): user.code_functions_common
 tag(): user.code_keywords
 tag(): user.code_libraries
 tag(): user.code_libraries_gui
+tag(): user.code_libraries_gui_showing
```
auscompgeek, line 15: This tag should not be permanently activated. It's used to enable commands that should only be available whilst the relevant window is open.


```suggestion
```

auscompgeek (changes requested):

PR #1252, opened 2023-09-03
javascript: Add `is [not] nullish` for loose null check

auscompgeek:
These check whether a LHS is nullish, i.e. null or undefined.

This is intended to complement #1255.

### DIFF
--- a/lang/javascript/javascript.talon
+++ b/lang/javascript/javascript.talon
@@ -27,6 +27,8 @@ settings():
 
 (op | is) strict equal: " === "
 (op | is) strict not equal: " !== "
+is nullish: " == null"
+is not nullish: " != null"
 op null else: " ?? "
 
 state const: "const "


AndreasArvidsson:
I wonder if we just want to update these actions instead?
https://github.com/talonhub/community/blob/8aeed406b6539e88f3aa010ff0684dfaf7381b35/lang/javascript/javascript.py#L76

The number of times she want to check just for null, but not undefined is it very low so I think that ``

auscompgeek:
I do agree it's pretty rare for me to check for specifically null or undefined. With #1255 it'll likely be confusing though.

AndreasArvidsson:
Notes from community session: We do have plans of redoing all the language modes in the near future. We gonna do list for operators.
https://github.com/talonhub/community/issues/1597

Issue #664, opened 2021-11-20
allow 'this dot', etc. as synonyms for 'self dot'


rntz:
would be more convenient for javascript. should either make these synonyms globally, or just in javascript.talon by adding commands.

eg, globally:
```
(self | this) dot: ...
```

auscompgeek:
My idea, which I've been running with for a bit: define a `code_self` list. That way different objects can be referred to without hardcoding them (e.g. Go method receivers don't have a standard parameter name).

```talon
{user.code_self} dot: ...
```

Bonus points for turning `code_operator_object_accessor` into a list too (to accommodate languages that use `->` instead):

```talon
{user.code_self} {user.code_operator_object_accessor}:
	insert("{code_self}{code_operator_object_accessor}")
```

PR #1253, opened 2023-09-03
object_oriented: Use a list for `code_self`

auscompgeek:
This allows using `this dot` or `state this` instead of `self dot` or `state self` in languages that don't use `self` as the name for the method receiver.

Ref: #664


AndreasArvidsson:
Notes from community session: We do have plans of redoing all the language modes in the near future. We gonna do list for operators more reliance on snippets and so on. We're not against the direction this pull request is going, but we don't really know right now it with will be a good fit. Just keep it open for now.

nriley:
From the community backlog session — we did not feel that introducing `this` would be of benefit. Please see a longer explanation here: https://github.com/talonhub/community/pull/1254#issuecomment-3418655843

PR #1254, opened 2023-09-03
Turn code_object_oriented_accessor into list

auscompgeek:
This turns the "dot" in `self dot` into a list. This allows for a different phrase to be used to accomodate languages, such as PHP, which don't use `.` to access fields and methods.

This could then be further used to define `code_common_member_function` (and the corresponding `dot <method>` command) in a more appropriate language-agnostic place.

Depends on: #1253
Ref: #664

### DIFF
--- a/lang/csharp/csharp.py
+++ b/lang/csharp/csharp.py
@@ -9,6 +9,7 @@ ctx.lists["user.code_common_function"] = {
     "print": "Console.WriteLine",
     "string": ".ToString",
 }
+ctx.lists["user.code_self"] = ["this"]
 
 
 @ctx.action_class("user")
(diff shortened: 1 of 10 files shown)


auscompgeek:
I'll try to reorder this PR stack at some point. I think this is more obviously desirable whereas #1253 is less clear as to whether it's a good approach.

nriley:
From the community backlog session — sorry we took so incredibly long to get to this. On the basis of our recent work in implementing a C++ language mode, we would prefer to implement this in a language-specific fashion. Also, using "this" as a spoken form is a challenge given Cursorless and other voice commands that use this to mean something else. "self" seems like the only safe word, that is both used to refer to the current object in multiple languages and does not conflict with other voice commands.

PR #1255, opened 2023-09-03
data_null: Turn null into a list

auscompgeek:
The intention here is to accomodate languages that have multiple different null-like values. For example:

- Rust
  - `None` (i.e. the `Option::None` enum)
  - `core::ptr::null()`
- JavaScript
  - `null`
  - `undefined`
- C++
  - `NULL` (from C)
  - `nullptr` (i.e. [`nullptr_t`](https://en.cppreference.com/w/cpp/types/nullptr_t))
  - `std::nullopt` (i.e. the [`std::optional<T>`](https://en.cppreference.com/w/cpp/utility/optional) template)

---

TODO:

- [ ] Make `code_insert_is_(not_)null` accept the null kind (required for Rust to check non-comparable `Option` vs pointer)

Note: This may conflict with #1253/#1254 due to different language implementations defining the actions next to each other.

### DIFF
--- a/lang/sql/sql.py
+++ b/lang/sql/sql.py
@@ -5,6 +5,10 @@ ctx.matches = r"""
 tag: user.sql
 """
 
+ctx.lists["user.code_data_null"] = {
+    "null": "NULL",
+}
+
 # these vary by dialect
 ctx.lists["user.code_common_function"] = {"count": "Count", "min": "Min", "max": "Max"}
 
@@ -52,9 +56,6 @@ class UserActions:
 
     def code_operator_or():
         actions.auto_insert("OR ")
-
-    def code_insert_null():
-        actions.auto_insert("NULL")
 
     def code_insert_is_null():
         actions.auto_insert(" IS NULL")
--- a/lang/tags/data_null.py
+++ b/lang/tags/data_null.py
@@ -4,13 +4,18 @@ ctx = Context()
 mod = Module()
 
 mod.tag("code_data_null", desc="Tag for enabling commands relating to null")
+mod.list("code_data_null", desc="null-like value (e.g. Python `None`, C++ `nullptr`)")
+
+ctx.lists["self.code_data_null"] = {
+    "null": "null",
+    # backwards compatibility
+    "nil": "null",
+    "none": "null",
+}
 
 
 @mod.action_class
 class Actions:
-    def code_insert_null():
-        """Inserts null"""
-
     def code_insert_is_null():
         """Inserts check for null"""
 
--- a/lang/c/c.py
+++ b/lang/c/c.py
@@ -131,6 +131,10 @@ ctx.lists["user.code_common_function"] = {
     "free": "free",
 }
 
+ctx.lists["user.code_data_null"] = {
+    "null": "NULL",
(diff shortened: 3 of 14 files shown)


lang/tags/data_null.py
```diff
@@ -4,13 +4,18 @@
 mod = Module()
 
 mod.tag("code_data_null", desc="Tag for enabling commands relating to null")
+mod.list("code_data_null", desc="null-like value (e.g. Python `None`, C++ `nullptr`)")
+
+ctx.lists["self.code_data_null"] = {
```
pokey, line 9: Looks like yiu dropped "no". Was that intentional?

lang/sql/sql.py
```diff
@@ -5,6 +5,10 @@
 tag: user.sql
 """
 
+ctx.lists["user.code_data_null"] = {
+    "null": "NULL",
```
pokey, line 9: This drops "nil", "none", and "no". I do find it useful not to need to think about whether a language uses "none" or "null" when switching back and forth. I wonder if there's a way to recover that behaviour without a bunch of duplication 🤔

pokey (commented):
Interesting idea. Left a couple comments

lang/tags/data_null.py
auscompgeek: I'm actually surprised `state no` was a command for null in the first place. It's inconsistent with the `is not (null|none)` command.

I figure this space could be useful for someone that might want to use `state yes/no` for booleans instead.

pokey: My guess is that was there for recognition accuracy, as "null" can be challenging for talon to recognise. I wouldn't say I'm strictly opposed to removing it

lang/sql/sql.py
AndreasArvidsson: Then what you want is probably an action? Because then you could have a bunch of different spoken phrases that would just insert a correct thing for the language.

auscompgeek: Are you suggesting an action `code_insert_null(kind: str)`? I suppose that could work.

AndreasArvidsson: I was thinking more that was pokey liked was the action you just removed, but maybe a middle ground can be reached?

auscompgeek: The entire premise of this PR is being able to get _different_ things for `none` and `null` in languages where the two exist, are similar, but different concepts. I don't see how we can achieve that without a list.

AndreasArvidsson: Me neither.

auscompgeek: I just came up with a wild idea as an alternative to adding a `code_insert_null` action, which is potentially overengineering it, but here it is:

- We wrap the list in a capture.
- To simplify implementing languages with a single null-like value, we add a str setting to allow a language to indicate it has a single null-like value. The capture would check this setting, and if unset it passes through the list value as is.

I'm not sure whether I like wrapping it in an action, or a capture more.

nriley:
From the community backlog session — wanted to add a +1 that this seems like a good idea to implement. Let us know if you have any questions.

PR #1256, opened 2023-09-03
Remove language tags and instead utilize `code.language`

AndreasArvidsson:
Removed our hack with language tags and instead rely on the Talon defined `code.language`

From
```
tag: user.c
-
```

to
```
code.language: c
-
```

### DIFF
--- a/lang/lua/stylua.talon
+++ b/lang/lua/stylua.talon
@@ -1,4 +1,4 @@
-code.language: stylua
+tag: user.stylua
 -
 
 lint ignore: "-- stylua: ignore"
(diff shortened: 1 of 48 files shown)


AndreasArvidsson:
@nriley I think this turned out quite nicely

lang/lua/stylua.talon
```diff
@@ -1,4 +1,4 @@
-tag: user.stylua
+code.language: stylua
```
auscompgeek, line 1: AFAICT stylua isn't a language but rather a linter/formatter. It shouldn't be turned into a language.

nriley (approved):
This looks good.

PR #1260, opened 2023-09-11
add 1password chrome extension commands

adam-edison:
Just a nice convenience for those with the extension instead of the desktop app.


knausj85:
Going thru old PRs:

I'm not sure this is a good fit for community for a couple reasons:

- The various sleeps between app switching and such are going to be system dependent and potentially a bit fragile.

- Modern MacOS has its own password app. We may be better served with this for accessibility purposes.

Leaving this open for further discussion in subsequent backlog sesssions

nriley:
On the community backlog session we discussed this and decided that the implementation was too fragile to be usable across platforms. In general 1Password's user interface is very difficult to automate, so this is not really an issue with the PR!

PR #1261, opened 2023-09-12
Add optional homophone code for cancel.py

rokubop:


plugin/cancel/cancel.py
```diff
@@ -1,20 +1,29 @@
...
-            return
-        # cancel the command
-        d["parsed"]._sequence = []
-        actions.app.notify(f"Command canceled: {' '.join(before)!r}")
+        text_str = " ".join(d["text"])
+        for phrase in cancel_phrases:
+            if text_str.endswith(phrase):
+                # cancel the command
+                d["parsed"]._sequence = []
+                actions.app.notify(f"Command canceled: {text_str!r}")
```
nriley, line 25: It looks like you are now including the cancellation words at the end of the command that was canceled whereas the previous version did not do this. Is this intentional?

nriley (changes requested):
Just a tiny nit.

plugin/cancel/cancel.py
rokubop: Good catch!

plugin/cancel/cancel.py
```diff
@@ -1,20 +1,32 @@
...
     if "text" in d and "parsed" in d:
-        before, after = d["text"][:-n], d["text"][-n:]
-        if after != cancel_phrase:
-            return
-        # cancel the command
-        d["parsed"]._sequence = []
-        actions.app.notify(f"Command canceled: {' '.join(before)!r}")
+        text_str = " ".join(d["text"])
+        for phrase in cancel_phrases:
+            if text_str.endswith(phrase):
```
auscompgeek, line 22: This now matches words that end in the first word of the phrase. This won't be a problem for `cancel` since I can't come up with any English words that end in `cancel`, but could very much be a problem for any synonyms that anyone adds to the list.

PR #1262, opened 2023-09-15
Add capability to focus window by title

pturpin:
# Problem
When an app has multiple windows focus just picks the first one it fights, which may not be the desired one, and there is not a great way to focus just the window you want.

# Solution
Add a new method, switcher_focus_app_title, to app_switcher.py that will focus the window whose title matches a given app name and a given regular expression. Also added a command to that will focus a window with a given spoken phrase.

# Testing
Verified working in current windows eleven environment.

### DIFF
--- a/core/app_switcher/app_switcher.py
+++ b/core/app_switcher/app_switcher.py
@@ -3,6 +3,7 @@ import shlex
 import subprocess
 import time
 from pathlib import Path
+import re
 
 import talon
 from talon import Context, Module, actions, app, fs, imgui, ui
@@ -315,6 +316,16 @@ class Actions:
             if time.perf_counter() - t1 > 1:
                 raise RuntimeError(f"Can't focus app: {app.name}")
             actions.sleep(0.1)
+
+    def switcher_focus_app_title(app: str, regex: str):
+        """Focus the window whose app name constains app and title matches regex"""
+        for window in ui.windows():
+            if not(window.hidden) and (app in window.app.name or app == "*") and window.title != "":
+                # logging.warn(f'Checking Window: "{window.app.name}" window:"{window.title}" hidden: "{window.hidden}"')
+                if (regex is None or re.search(regex, window.title)):
+                    window.focus()
+                    return
+        logging.error(f'(switcher_focus_app_title) Window not found: "{app}" "{regex}"')
 
     def switcher_focus_window(window: ui.Window):
         """Focus window and wait until switch is made"""
--- a/core/app_switcher/app_switcher.py
+++ b/core/app_switcher/app_switcher.py
@@ -1,9 +1,9 @@
 import os
+import re
 import shlex
 import subprocess
 import time
 from pathlib import Path
-import re
 
 import talon
 from talon import Context, Module, actions, app, fs, imgui, ui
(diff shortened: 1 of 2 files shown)


core/app_switcher/app_switcher.py
```diff
@@ -316,6 +317,20 @@ def switcher_focus_app(app: ui.App):
...
+    def switcher_focus_app_title(app: str, regex: str):
+        """Focus the window whose app name constains app and title matches regex"""
+        for window in ui.windows():
+            if (
+                not (window.hidden)
+                and (app in window.app.name or app == "*")
+                and window.title != ""
+            ):
+                # logging.warn(f'Checking Window: "{window.app.name}" window:"{window.title}" hidden: "{window.hidden}"')
+                if regex is None or re.search(regex, window.title):
```
auscompgeek, line 329: What's your usecase for passing a regex here?

pturpin: I currently have two commands that utilize an actual regex vs a plain string.

`focus outlook: user.switcher_focus_app_title("Outlook", "(Inbox|Calendar)")
focus studio: user.switcher_focus_app_title("Microsoft Visual Studio", "^(?:(?!Application Management).)*$")`

fidgetingbits: I think this seems reasonable given the use cases. @auscompgeek any objection or you think it's ok?

auscompgeek: I suppose the use-case makes sense. I'm not sure we should make this the primary way, or indeed the only way, of focusing by title, as it'd be prone to user error – it's unclear from the action name that it takes a regex.

knausj85:
Continuing to go through old pull requests:
I think we need to take a moment to reevaluate this one in an upcoming community backlog session. 
If we do move forward with is functionality, I do think it should be reimplemented using dynamic lists
https://github.com/talonhub/community/pull/1621

nriley:
From the community backlog session — we agreed that this seems like a useful feature to add to community but we would prefer that it's implemented in terms of dynamic lists as @knausj85 recommended.

fidgetingbits:
Pretty sure aegis recently(ish) said in slack he doesn't want community using dynamic lists atm as they are in beta only? Discussion was related to someone wanting my [justfile stuff](https://github.com/fidgetingbits/fidgetingbits-talon/blob/70e640b6d29a17b92829a00b435dec492ffa3b97/apps/j/justfile/justfile.py#L17) added to community.

I'd personally be okay to see community use dynamic lists gated behind a beta flag, but will let someone else have that debate. I believe in that justfile discussion (or around the same time) it was noted you can detect beta programmatically, but then of course still need to maintain two versions of the logic until it hits public.

nriley:
AFAIK, dynamic lists are in public; [dynamic] selection lists are only in beta.

lunixbochs:
you shouldn't detect beta, because there are more branches than just beta and public

fidgetingbits:
> AFAIK, dynamic lists are in public; [dynamic] selection lists are only in beta.

Could be I'm confused, but my justfile doesn't use selection list (?) and was the topic of the discussion, and iirc no public release since, so maybe @lunixbochs can confirm.

PR #1265, opened 2023-09-17
Added Talon based implementation for `app.window_previous/next`

AndreasArvidsson:
The current implementation of "window next/last" doesn't really work. This is an implementation using the Talon ui/window api similar to how focusing applications works.

### DIFF
--- a/core/windows_and_tabs/windows_and_tabs.py
+++ b/core/windows_and_tabs/windows_and_tabs.py
@@ -1,0 +1,48 @@
+from talon import Context, actions, ui
+
+ctx = Context()
+
+
+@ctx.action_class("app")
+class AppActions:
+    def window_previous():
+        cycle_windows(ui.active_app(), -1)
+
+    def window_next():
+        cycle_windows(ui.active_app(), 1)
+
+
+def cycle_windows(app: ui.App, diff: int):
+    """Cycle windows backwards or forwards for the given application"""
+    active = app.active_window
+    windows = [w for w in app.windows() if w == active or is_window_valid(w)]
+    windows.sort(key=lambda w: w.id)
+    current = windows.index(active)
+    max = len(windows) - 1
+    i = cycle(current + diff, 0, max)
+
+    while i != current:
+        try:
+            actions.user.switcher_focus_window(windows[i])
+            break
+        except Exception:
+            i = cycle(i + diff, 0, max)
+
+
+def is_window_valid(window: ui.Window) -> bool:
+    """Returns true if this window is valid for focusing"""
+    return (
+        not window.hidden
+        and window.title != ""
+        and window.rect.width > window.screen.dpi
+        and window.rect.height > window.screen.dpi
+    )
+
+
+def cycle(value: int, min: int, max: int) -> int:
+    """Cycle value between minimum and maximum"""
+    if value < min:
+        return max
+    if value > max:
+        return min
+    return value
--- a/core/windows_and_tabs/windows_and_tabs.py
+++ b/core/windows_and_tabs/windows_and_tabs.py
@@ -18,15 +18,14 @@ def cycle_windows(app: ui.App, diff: int):
     windows = [w for w in app.windows() if w == active or is_window_valid(w)]
     windows.sort(key=lambda w: w.id)
     current = windows.index(active)
-    max = len(windows) - 1
-    i = cycle(current + diff, 0, max)
+    i = (current + diff) % len(windows)
 
     while i != current:
         try:
             actions.user.switcher_focus_window(windows[i])
             break
         except Exception:
-            i = cycle(i + diff, 0, max)
+            i = (i + diff) % len(windows)
 
 
 def is_window_valid(window: ui.Window) -> bool:
@@ -39,12 +38,3 @@ def is_window_valid(window: ui.Window) -> bool:
         and window.rect.width > window.screen.dpi
         and window.rect.height > window.screen.dpi
     )
-
-
-def cycle(value: int, min: int, max: int) -> int:
-    """Cycle value between minimum and maximum"""
-    if value < min:
-        return max
-    if value > max:
-        return min
-    return value
--- a/core/windows_and_tabs/windows_and_tabs_linux.py
+++ b/core/windows_and_tabs/windows_and_tabs_linux.py
@@ -35,7 +35,6 @@ class AppActions:
 
     def window_hide_others():
         actions.key("win-d alt-tab")
-        # requires easy window switcher or equivalent (built into most Linux)
 
     def window_open():
         actions.key("ctrl-n")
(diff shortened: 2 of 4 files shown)


auscompgeek:
Any reason to not keep using the macOS support? (I have no objections to this on Windows/Linux.)

AndreasArvidsson:
> Any reason to not keep using the macOS support? (I have no objections to this on Windows/Linux.)

I wasn't sure if it worked to be honest. I guess one could argue that if this implementation works on all platform it doesn't gain us much, but I don't mind either way. 
Do mac users prefer to keep the old implementation? @pokey @phillco @nriley

nriley:
I think it makes sense to keep the existing implementation on Mac because it behaves identically to the way users expect. "Next/previous window in this app" is not a standard OS feature on Windows so we can do whatever we want :)

phillco:
I'm actually not opposed to implementing it with Talon APIs if it works the same way as the system keyboard shortcut. The advantage would be that it would work in applications like Screen Sharing (built in VNC client), properly rotating its windows. Whereas if you use the keyboard shortcut, that would get sent to the remote computer.

I use a similar implementation for quitting applications -- `app.quit` -- and it has the same benefit.

nriley:
But what if you want to switch windows on the remote machine? :) I think that may be a bigger discussion…

phillco:
"super backtic" :)

core/windows_and_tabs/windows_and_tabs.py
```diff
@@ -0,0 +1,48 @@
...
+            i = cycle(i + diff, 0, max)
+
+
+def is_window_valid(window: ui.Window) -> bool:
+    """Returns true if this window is valid for focusing"""
+    return (
+        not window.hidden
+        and window.title != ""
+        and window.rect.width > window.screen.dpi
+        and window.rect.height > window.screen.dpi
```
phillco, line 38:
```suggestion
        # This excludes many tiny windows that are not actual windows, and is a rough heuristic.
        and window.rect.width > window.screen.dpi
        and window.rect.height > window.screen.dpi
```

core/windows_and_tabs/windows_and_tabs.py
```diff
@@ -0,0 +1,48 @@
...
+            break
+        except Exception:
+            i = cycle(i + diff, 0, max)
+
+
+def is_window_valid(window: ui.Window) -> bool:
+    """Returns true if this window is valid for focusing"""
+    return (
+        not window.hidden
+        and window.title != ""
```
phillco, line 36:
```suggestion
        # On Windows, there are many fake windows with empty titles -- this excludes them.
        and window.title != ""
```

phillco: We _might_ just want to scope this check to only windows, since that's the only platform where we've seen this issue. I guess it's a mix of "specific workaround for a known problem that plagues windows" versus "general useful heuristic across platforms". I'm not sure if it would be normal for a macOS window to not have a title. @nriley thoughts?

core/windows_and_tabs/windows_and_tabs.py
phillco: I think this is another case where on macOS, window widths are reported in resolution-independent points, so comparing it against dots per inch doesn't make sense.

Also @nriley -- do you know if there's some property on the window in macOS that we could be checking, to exclude it from this kind of logic? This just made me remember that I already run to this problem in other cases where I use the Talon window snapping code.

For example, the find dialog in chrome counts as a "window":

image

... but macOS knows to exclude it from cmd-`. If there's an OS property we could be checking we could exclude it from window snapping logic in other places, subject to the user's preferences, naturally.

AndreasArvidsson:
Me and Phil talked about this and I have re enabled the mac specific implementation.

nriley:
OK, I agree in theory, just need to make sure there aren't any annoying issues when compared with the OS implementation, for example in apps that have floating panels.

phillco:
yup, think we're on the same page :)

AndreasArvidsson:
Okay great. Could any of you try this new implementation on Mac and see if it works? I just want to make sure that I haven't made some windows only assumptions.

core/windows_and_tabs/windows_and_tabs_win.py
```diff
@@ -37,12 +37,6 @@ def window_hide_others():
         actions.key("win-d alt-tab")
         # requires easy window switcher or equivalent (built into most Linux)
```
auscompgeek, line 38: I think these comments are referring to the window_next/previous actions

```suggestion
```

core/windows_and_tabs/windows_and_tabs_win.py
```diff
@@ -37,12 +37,6 @@ def window_hide_others():
...
 
-    def window_next():
-        actions.key("alt-`")
-
     def window_open():
         actions.key("ctrl-n")
         # requires easy window switcher or equivalent (built into most Linux)
```
auscompgeek, line 42:
```suggestion
```

core/windows_and_tabs/windows_and_tabs_linux.py
```diff
@@ -37,12 +37,6 @@ def window_hide_others():
         actions.key("win-d alt-tab")
         # requires easy window switcher or equivalent (built into most Linux)
```
auscompgeek, line 38: per above review
```suggestion
```

core/windows_and_tabs/windows_and_tabs_linux.py
```diff
@@ -37,12 +37,6 @@ def window_hide_others():
...
 
-    def window_next():
-        actions.key("alt-`")
-
     def window_open():
         actions.key("ctrl-n")
         # requires easy window switcher or equivalent (built into most Linux)
```
auscompgeek, line 42:
```suggestion
```

fidgetingbits:
Works fine on gnome fwiw. Also since tiling wm on linux don't usually use alt+` afaik, so this has the added benefit of working on those now by default.

core/windows_and_tabs/windows_and_tabs.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+
+def cycle(value: int, min: int, max: int) -> int:
+    """Cycle value between minimum and maximum"""
+    if value < min:
+        return max
+    if value > max:
+        return min
+    return value
```
nriley, line 50: Consider using the modulo operator (%) so this works with diff > 1.

nriley (commented):
Looking good!

core/windows_and_tabs/windows_and_tabs.py
AndreasArvidsson: done

knausj85 (approved):
This looks like it's working well. Think this is ready to be merged.

PR #1266, opened 2023-09-18
Added voice commands to change hiss scrolling direction

AndreasArvidsson:


nriley:
The main challenge here is that macOS has an option to invert the scrolling direction globally. It would be helpful if we had an additional option in the hiss scrolling setting which would invert the scrolling direction and the up and down toggles would respect it.

AndreasArvidsson:
I agree with @phillco. Would probably be better to do a separate pr where we fixed the scrolling directions for all scrolling actions and not just this one.

knausj85 (approved):

PR #1271, opened 2023-10-15
Call action window next when refocusing current application

AndreasArvidsson:
`"focus firefox"` => focuses first window of firefox
`"second"` => cycles to next window

This is even more beneficial if you have a repeater noise


knausj85 (approved):
Nice enhancement, thanks

PR #1273, opened 2023-10-18
add "pop to wake" function

gpunktschmitz:
I've added this functionality to be able to pop to wake talon - added a setting to the `sleep_mode.talon` which might be better to set to 0?


plugin/mouse/mouse.py
```diff
@@ -258,29 +264,34 @@ def show_cursor_helper(show):
 @ctx.action_class("user")
 class UserActions:
     def noise_trigger_pop():
-        if setting_mouse_enable_pop_stops_scroll.get() >= 1 and (
-            gaze_job or scroll_job
-        ):
-            # Allow pop to stop scroll
-            stop_scroll()
+        if actions.speech.enabled():
```
phillco, line 267: This breaks using pop-to-click in sleep mode (adding `setting_mouse_enable_pop_click = 2` in `core/modes/sleep_mode.talon`), because of the new if statement. If speech is not enabled, your new else block only checks for pop to wake, not any of the other behavior, which could be valid in sleep mode.

You probably want to change this to:

```suggestion
        # Pop can be configured to wake Talon, in which case we don't want to process it as a click.
        if not actions.speech.enabled() and setting_mouse_enable_pop_wake.get() == 1:
            actions.speech.enable()
            return
```

The early return also reduces the complexity of the function by not adding a new level of indentation / nested ifs.

phillco: As an aside, @nriley debugged this for a little bit and we were a bit confused by the initial behavior of the existing `user.mouse_enable_pop_click = 1` definition in `settings.talon`, because changing it to `2` doesn't enable pop to click in sleep mode. But that's because the settings block in that file, absent a `mode:` header, is implicitly scoped to non-sleep modes, so you have to add it to something like `core/modes/sleep_mode.talon`.

That probably reduces the number of people using pop in sleep mode, and we should probably address it separately, but in the meantime, we shouldn't break it if people have enabled it specifically.

phillco (changes requested):

plugin/mouse/mouse.py
```diff
@@ -52,6 +52,12 @@
     default=0,
     desc="Pop noise clicks left mouse button. 0 = off, 1 = on with eyetracker but not with zoom mouse mode, 2 = on but not with zoom mouse mode",
 )
+setting_mouse_enable_pop_wake = mod.setting(
+    "mouse_enable_pop_wake",
+    type=int,
```
phillco, line 57:
```suggestion
    type=bool,
```

Made it a suggestion for easier application :)

phillco:
Thanks for the PR! I think one change will be necessary to make this not break existing behavior, but otherwise, this will be nice functionality to have in stock config.

core/modes/sleep_mode.talon
```diff
@@ -7,3 +7,5 @@ settings():
     user.mouse_enable_pop_click = 0
     # Stop mouse scroll down using hiss noise
     user.mouse_enable_hiss_scroll = 0
+    # Enable pop wake
+    user.mouse_enable_pop_wake = 1
```
phillco, line 11: Just noticed this. This should probably default to false; popping can occasionally have false positives so for something that can have as many side effects as waking Talon it should probably be aped in, but we can advertise it in the README as a setting that people can enable.

Also added a note that this will intrinsically disable popping to click, since you can't have both.

```suggestion
    # Enable pop wake. Note that if this is enabled, it will disable popping to click in sleep mode, if you set `mouse_enable_pop_wake = 2`.
    user.mouse_enable_pop_wake = 0
```

pokey:
I wonder if double pop would be better given high false positive rate. That's what @AndreasArvidsson used to run before parrot and I think it worked well for him

phillco:
Update: I had an incorrect statement about why simply changing `user.mode_indicator_show = 2` in `settings.talon` wasn't taking effect, which I've updated.

My inference about modeless files was wrong; it behaves that way simply because there's an existing override for sleep mode for that setting specifically (in `core/modes/sleep_mode.talon`) setting it to 0, and that context is more specific (since it sets `mode: sleep`), which I missed.

Quoting @lunixbochs for posterity:

> With no mode header, commands, noises, hotkeys, deck, face, gamepad are partitioned into their respective modes. Other bindings like settings and tags are modeless.

phillco:
Update: I had an incorrect statement about why simply changing `user.mode_indicator_show = 2` in `settings.talon` wasn't taking effect, which I've updated.

My inference about modeless files was wrong; it behaves that way simply because there's an existing override for sleep mode for that setting specifically (in `core/modes/sleep_mode.talon`) setting it to 0, and that context is more specific (since it sets `mode: sleep`), which I missed.

gpunktschmitz (commented):
I am fine with every comment/edit. I personally use this method for quite some time and never noted any false-positives that would make a double-pop necessary but would mind that either. I just don't know how to implement this.

core/modes/sleep_mode.talon
```diff
@@ -7,6 +7,8 @@ settings():
     user.mouse_enable_pop_click = 0
     # Stop mouse scroll down using hiss noise
     user.mouse_enable_hiss_scroll = 0
+    # Enable pop wake. Note that if this is enabled, it will disable popping to click in sleep mode, if you set `mouse_enable_pop_wake = 2`.
```
phillco, line 10: I think this was a typo on my part?
```suggestion
    # Enable pop wake. Note that if this is enabled, it will disable popping to click in sleep mode, if you set `mouse_enable_pop_click = 2`.
```

plugin/mouse/mouse.py
```diff
@@ -269,29 +275,37 @@ def show_cursor_helper(show):
...
+            if setting_mouse_enable_pop_stops_scroll.get() >= 1 and (
+                gaze_job or scroll_job
+            ):
+                # Allow pop to stop scroll
+                stop_scroll()
+            else:
+                # Otherwise respect the mouse_enable_pop_click setting
+                setting_val = setting_mouse_enable_pop_click.get()
+
+                is_using_eye_tracker = (
```
phillco, line 291: This shouldn't be indented after the early return above.

plugin/mouse/mouse.py
```diff
@@ -269,29 +275,37 @@ def show_cursor_helper(show):
...
-            ) or (
-                setting_val == 1
-                and is_using_eye_tracker
-                and not actions.tracking.control_zoom_enabled()
-            )
-            if should_click:
-                ctrl.mouse_click(button=0, hold=16000)
+            # test mouse_enable_pop_wake setting
+            if setting_mouse_enable_pop_wake.get() == 1:
+                actions.speech.enable()
```
phillco, line 308: Is this supposed to be here?

gpunktschmitz:
fixed dedent and typo and stuff - tested locally and confirmed working

plugin/mouse/mouse.py
```diff
@@ -269,6 +275,10 @@ def show_cursor_helper(show):
 @ctx.action_class("user")
 class UserActions:
     def noise_trigger_pop():
+        # Pop can be configured to wake Talon, in which case we don't want to process it as a click.
```
AndreasArvidsson, line 278: I think that doing stuff like this with if statements is an anti pattern in Talon. Preferably you should add a new context that matches sleep mode and add the action there.

plugin/mouse/mouse.py
```diff
@@ -269,6 +275,10 @@ def show_cursor_helper(show):
...
+        if not actions.speech.enabled() and setting_mouse_enable_pop_wake.get() == 1:
```
AndreasArvidsson, line 279: The setting is a boolean and you should treat it as such.

alan-barzilay:
> I wonder if double pop would be better given high false positive rate. That's what @AndreasArvidsson used to run before parrot and I think it worked well for him

In my config files I deleted everything mouse related (I have no real need for it at the moment) and mapped the pop sound to toggle between wake and sleep states but also suffer from false positives so I like this idea.

If this happen to be the direction followed by this PR, I would like to suggest a change from "pop to wake" to pop to toggle between sleep and wake states.

plugin/mouse/mouse.py
gpunktschmitz: Sorry but I don't know how I should do that. Could you point me to where this has been done so I may adopt from there?

plugin/mouse/mouse.py
gpunktschmitz: fixed - also added boolean value to sleep_model.talon

plugin/mouse/mouse.py
AndreasArvidsson: Something like this
```python
ctx_sleep = Context()
ctx_sleep.matches = r"""
mode: sleep
"""

@ctx_sleep.action_class("user")
class UserActions:
    def noise_trigger_pop():
        # Pop can be configured to wake Talon, in which case we don't want to process it as a click.
        if setting_mouse_enable_pop_wake.get():
            actions.speech.enable()
```

plugin/mouse/mouse.py
AndreasArvidsson: Personally I would not use `== True`. It's enough to do `if setting_mouse_enable_pop_wake.get():`

pokey:
> If this happen to be the direction followed by this PR, I would like to suggest a change from "pop to wake" to pop to toggle between sleep and wake states.

If there's a highly reliable noise (eg double pop), my leaning is to use it for wake and repeat. A reliable repeater noise is a game changer. Wrt double pop specifically tho, I worry that might mess with double clicking if people use pop to click?

For putting talon to sleep, I find that a chainable sleep command works fine for me, and has the added benefit of not having timing issues where you say a command and then pop, but because of speech timeout the pop gets registered first so cancels your command

pokey:
> I am fine with every comment/edit. I personally use this method for quite some time and never noted any false-positives that would make a double-pop necessary but would mind that either. I just don't know how to implement this. 

Best place to look for impl is prob @AndreasArvidsson's commit history. He had a solid one before he switched to parrot noises

AndreasArvidsson:
You find my implementations here

https://github.com/AndreasArvidsson/andreas-talon/blob/ef049e9cf50b2694ee1b2f039fc102bd488ca1ae/misc/on_pop.py
https://github.com/AndreasArvidsson/andreas-talon/blob/ef049e9cf50b2694ee1b2f039fc102bd488ca1ae/misc/sleep/sleep.py#L23-L29

plugin/mouse/mouse.py
gpunktschmitz: removed `== True`

plugin/mouse/mouse.py
gpunktschmitz: replaced if with this snipped - unsure where to put it though - added it to the very end of the file

plugin/mouse/mouse.py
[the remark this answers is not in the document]
gpunktschmitz: changed to `type=bool`

alan-barzilay:
> For putting talon to sleep, I find that a chainable sleep command works fine for me, 

@pokey could you provide an example of such chainable command? I liked the idea and want to steal it haha

pokey:
> > For putting talon to sleep, I find that a chainable sleep command works fine for me,
> 
> @pokey could you provide an example of such chainable command? I liked the idea and want to steal it haha

you'd remove the `^` from the start of the commands in https://github.com/talonhub/community/blob/main/core/modes/to_sleep_mode.talon

jaresty:
I love this concept. Thanks for putting it together!

gpunktschmitz:
merging is blocked and I have no clue how to resolve it - also tried rebasing but this also fails. I will close this PR and create a new one *sigh*

gpunktschmitz:
created #1363

PR #1277, opened 2023-10-24
Java define return types

alexp141:
Added ability to define return types in Java more easily when creating a function. Previously, the return types would always default to void. Now, you can (optionally) decide to specify the return type of a function during creation. The previous way of defining functions still works too.

Example: saying "pub static String funky my function" will return 
`public static String myFunction()`

This works for all of the code types defined in java.py


alexp141:
Just realized that this breaks every language that uses the code_functions tag that isn't Java. Will update if I come up with another solution.

alexp141:
I have finished testing this and it looks like it works well. 

Some final thoughts: There are two main ways I thought of doing this. The first way is the way that I did it in my first commit which involved changing all the code_* functions in functions.py to allow them to accept return values as arguments and the way that I did it in my last commit where the correct function is chosen based on whether the user dictates a return type or not. Both ways have pros and cons.

Any feedback is appreciated!

nriley:
From the community backlog session — we apologize for taking so long to review this! @AndreasArvidsson plans on some pretty major refactoring of the language modes (you can see some of the discussion in #1597) which would use a Talon list for modifiers such as `private`, `static`, etc. rather than many separate Talon actions for each combination of modifiers. So unfortunately we did not feel it was helpful to merge at this time.

PR #1305, opened 2023-11-03
Add wrapper snippets

AndreasArvidsson:


pokey:
Hmm I'm struggling to see how this belongs in community. This code is all very Cursorless-specific

knausj85:
> Hmm I'm struggling to see how this belongs in community. This code is all very Cursorless-specific

Since it's guarded by the cursorless tag, I'm not opposed to including this functionality. I think the power of the feature is significant enough to justify including it.

Is there any alternative -- support within cursorless itself?

AndreasArvidsson:
This feature requires 1 list that only exist in community, it requires 2 actions that only exists community and it requires 1 dataclass that only exists in community. On the other side the Cursorless snippet api and the Cursorless target capture feels much more public and stable.

I've already added the Cursorless tag to community so we shouldn't have any warnings in the log. Defining a tag twice also doesn't seem to be any problem in Talon.

lunixbochs:
Defining something on two mods may become an error, the current behavior of ignoring one of them at random is incidental and not stable

AndreasArvidsson:
Okay then we'll rethink this. 
My current solution is to add things in both community and Cursorless to minimize the dependencies between them

https://github.com/cursorless-dev/cursorless/pull/1998

pokey (commented):
Yeah these seem like more modest modifications to community. See comments below

core/snippets/snippets_insert.py
```diff
@@ -47,3 +47,28 @@ def insert_snippet_by_name_with_phrase(name: str, phrase: str):
...
+        snippet: Snippet = actions.user.get_snippet(snippet_name)
+        variable = snippet.get_variable_strict(variable_name)
+        reg = re.compile(rf"\${variable_name}|\$\{{{variable_name}\}}")
+
+        if not reg.search(snippet.body):
+            raise ValueError(
+                f"Can't wrap non existing variable '{variable_name}' in snippet body '{snippet.body}'"
+            )
+
+        body = reg.sub("$TM_SELECTED_TEXT", snippet.body)
```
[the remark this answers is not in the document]
AndreasArvidsson: 2 reasons:
1. Cursorless doesn't support wrapping numbered tab stops which I use in all my snippets.
2. For future ide support we probably want as much of the behavior to live in community as possible.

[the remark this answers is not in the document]
pokey: We could support 1 fairly easily; I'd be happy to open a PR

Re 2, ok I can see that, but I don't think it should come at the expense of robbing semantics from the payload we send to cursorless. As we've rediscovered many times, it tends to be better if we send Cursorless a payload that is as close as possible to what the user said. If we don't tell Cursorless the variable that's being wrapped, I'd argue we're losing semantics

AndreasArvidsson:
I was hoping to add support to wrap selection without cursorless, but if cursorless can support numbered holes then I'll be ok with letting cursorless deal with it.

core/snippets/snippets_insert.py
```diff
@@ -47,3 +47,28 @@ def insert_snippet_by_name_with_phrase(name: str, phrase: str):
             )
 
         actions.user.insert_snippet_by_name(name, substitutions)
+
+    def insert_wrapper_snippet(
+        body: str,
+        target: Any,
```
knausj85, line 53: I don't see the relevant import for Any fyi

knausj85:
It seems like this should be a draft pull request until the cursorless side lands?

core/snippets/snippets_insert.py
AndreasArvidsson: Done!

AndreasArvidsson:
> It seems like this should be a draft pull request until the cursorless side lands?

Sounds good

AndreasArvidsson:
@pokey Updated: We now send the wrapper variable

core/snippets/snippets_insert.py
```diff
@@ -47,3 +48,23 @@ def insert_snippet_by_name_with_phrase(name: str, phrase: str):
...
+        variable_name: str,
+        scope: str = None,
+    ):
+        """Wrap the target with snippet"""
+
+    def insert_wrapper_snippet_by_name(name: str, target: Any):
```
pokey, line 60: Any objection to calling this `wrap_with_snippet_by_name`, and the above `wrap_with_snippet`? cc/ @knausj85

pokey (commented):
Looks good with minor comment

core/snippets/snippets_insert.py
AndreasArvidsson: I could go either way. @knausj85 any thoughts?

knausj85: I kinda-sorta prefer the more namespace-y names as-is

AndreasArvidsson: Action names updated
@pokey better?

pokey (dismissed):

pokey:
We had an idea for a better api for this, where community just has actions

```py
@dataclass
class InsertionSnippet:
    body: str;
    scopes: str[];

@dataclass
class WrapperSnippet:
    body: str;
    variableName: str[];
    scope: str;

@mod.actions
class Snippet:
    get_insertion_snippet(name: str) -> InsertionSnippet:
        ...
    get_wrapper_snippet(name: str) -> WrapperSnippet:
        ...
```

note that the `name` will be the value from the `user.snippet` and `user.snippet_wrapper` lists, respectively, for the two above actions

pokey:
We could add a voice command "<snippet_wrapper> wrap" (ie with no target), which will wrap the selection. GLobally, it will use a hack of sorts, but in vscode it can replace variable with `TM_SELECTED_TEXT`

AndreasArvidsson:
Closing this and will open a new one with the new direction

PR #1312, opened 2023-11-11
Added setting for additional user snippets

AndreasArvidsson:
Enables setting
`user.snippets_dir = "snippets"`


core/snippets/snippets.py
```diff
@@ -29,6 +36,12 @@
     context_map[lang] = ctx
 
 
+def get_setting_dir():
+    if not setting_dir.get():
+        return None
+    return Path(actions.path.talon_user()) / setting_dir.get()
```
pokey, line 42: I think the idea was to do what we do with Cursorless, where we only prepend talon user if it's not an absolute path, right?

core/snippets/snippets.py
```diff
@@ -158,6 +175,10 @@ def create_lists(
 
 def on_ready():
     fs.watch(str(SNIPPETS_DIR), lambda _1, _2: update_snippets())
+
+    if get_setting_dir():
```
knausj85, line 179: since this uses a setting, wouldn't the user have to restart talon after changing it as-is? I'm not sure if there's a way around this at the moment.  

It's also a little unique in that I don't _think_ it makes sense to change the setting per context, so we may wish to add a note to the readme/code.

core/snippets/snippets.py
AndreasArvidsson: done

core/snippets/snippets.py
AndreasArvidsson: Correct

I added a comment to the settings file

knausj85 (approved):

AndrewDant:
@AndreasArvidsson I've tested this and confirmed it is working as intended for me. I'll be dropping my local hack and using this instead going forward.

Issue #197, opened 2020-08-19
talon.exe uses 20% of CPU and makes system slow to respond, while throwing errors continuously


jeheydorn:
I'm using talon beta on Windows 10 with Dragon 15.3. When I start up talon, after it initializes and connects to Dragon, it uses about 15 – 23% of my CPU (about one for core), and causes the computer to become slow to respond. The mouse becomes shaky, and applications can take several seconds to respond sometimes.

I ran talon_console.exe, and it constantly gives these errors which might be related (or they might not be, see my later discussion):

2020-08-19 11:59:31 WARNING actions: implementations skipped because they have no matching declaration: (user.password_show, user.password_duplicate, user.password_new, user.password_delete, user.password_edit, user.password_fill)
2020-08-19 11:59:31 WARNING actions: implementations skipped because they have no matching declaration: (user.password_show, user.password_duplicate, user.password_new, user.password_delete, user.password_edit, user.password_fill)
2020-08-19 11:59:31 ERROR cb error topic="win_focus" cb=<function update_maps at 0x0000000009ECEAF8>
   18:                  <stdlib>/threading.py:890|
   17:                  <stdlib>/threading.py:926|
   16:                  <stdlib>/threading.py:870|
   15:             talon\windows\winevents.py:172|
   14:            talon\scripting\dispatch.py:98 |
   13:            talon\scripting\dispatch.py:133|
   12:            talon\scripting\dispatch.py:124|
   11:                talon\scripting\rctx.py:199|
   10: ------------------------------------------# 'win_focus' main:_on_event()
    9:                    talon\windows\ui.py:79 |
    8:            talon\scripting\dispatch.py:98 |
    7:            talon\scripting\dispatch.py:133|
    6:            talon\scripting\dispatch.py:126|
    5:                talon\scripting\rctx.py:201|
    4: ------------------------------------------# 'win_focus' user.knausj_talon.code.file_manager:update_maps()
    3: ------------------------------------------# stack splice
    2:                talon\scripting\rctx.py:199|
    1: user\knausj_talon\code\file_manager.py:486| ctx.lists.update(lists)
AttributeError: 'DictionaryMeta' object has no attribute 'update'
2020-08-19 11:59:31 WARNING actions: implementations skipped because they have no matching declaration: (user.password_show, user.password_duplicate, user.password_new, user.password_delete, user.password_edit, user.password_fill)
2

[...]

PR #96, opened 2020-05-17
modified code.py file suffix identification regex

fidgetingbits:
this adds a few additional languages, a few minor pep8-related tweaks, but most importantly I modified the regex because a file naming convention I was using was causing problems. the modified regex seems to work fine for me across the board so far (been a couple weeks?), but I suspect it might run into problems on mac or windows, so I'm just posting a pull request so you can test it out on your end

PR #1313, opened 2023-11-12
lang/go: redo from scratch

Xe:
The former Go bindings do have some useful things, but overall they create invalid syntax more than they should (never). This brave new syntax binding set will do this a lot better.

Replaces #1308

### DIFF
--- a/lang/go/code-common-functions.talon-list
+++ b/lang/go/code-common-functions.talon-list
@@ -1,0 +1,20 @@
+list: user.code_common_function
+code.language: go
+-
+
+length: len
+append: append
+close: close
+capacity: cap
+clear: clear
+complex: complex
+delete: delete
+imaginary: imag
+max: max
+min: min
+new: new
+panic: panic
+print: print
+print line: println
+real: real
+recover: recover
--- a/lang/go/code-types.talon-list
+++ b/lang/go/code-types.talon-list
@@ -1,0 +1,36 @@
+list: user.code_type
+code.language: go
+-
+
+bool: bool
+boolean: bool
+
+byte: byte
+bite: byte
+
+complex sixty four: complex64
+complex six four: complex64
+complex one twenty eight: complex128
+complex one two eight: complex128
+
+number: int
+integer: int
+int: int
+integer eight: int8
+integer sixteen: int16
+integer thirty two: int32
+integer sixty four: int64
+
+float: float64
+float thirty two: float32
+float sixty four: float64
+
+character: rune
+rune: rune
+
+string: string
+
+machine word: uintptr
+raw pointer: uintptr
+
+comparable: comparable
--- a/lang/go/go.py
+++ b/lang/go/go.py
@@ -1,0 +1,215 @@
+from talon import Context, Module, actions
+
+mod = Module()
+ctx = Context()
+ctx.matches = r"""
+code.language: go
+"""
+
+
+ctx.lists["user.code_type"] = {}
+
+
+def base_function(text: str, visibility: str):
+    """Inserts a public function definition, this assumes a lot about how your editor works"""
+    result = "func {}() {{\n\n}}".format(
+        actions.user.formatted_text(
+            text, visibility
+        )
+    )
+
+    actions.user.insert_between(f"func {text}() {{\n", "")
+
+
+@ctx.action_class("user")
+class UserActions:
+    def code_insert_true():
+        actions.auto_insert(" true ")
+
+    def code_insert_false():
+        actions.auto_insert(" false ")
+
+    def code_insert_null():
+        actions.auto_insert(" nil ")
+
+    def code_insert_is_null():
+        actions.auto_insert(" == nil ")
+
+    def code_insert_is_not_null():
+        actions.auto_insert(" != nil ")
+
+    def code_comment_documentation():
+        """inserts godoc syntax"""
+        actions.key("up")
+        actions.auto_insert("// ")
+
+    def code_default_function(text: str):
+        if text == "main" or text == "mane":
+            actions.user.code_private_function("main")
+        else:
+            actions.user.code_public_function(text)
+
+    def code_public_function(text: str):
+        base_function(text, "PUBLIC_CAMEL_CASE")
+
+    def code_private_function(text: str):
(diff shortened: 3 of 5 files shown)


lang/go/go.py
```diff
@@ -0,0 +1,211 @@
+from talon import Context, Module, actions
+
+mod = Module()
+ctx = Context()
+ctx.matches = r"""
+code.language: go
+"""
+
+
+ctx.lists["user.code_type"] = {}
```
knausj85, line 10: This line may override the .talon-list and probably should be removed. At least on my end, code_type is empty without this change since they are contexts are equally specific.

knausj85:
This looks great, thanks. sadly, I don't know go. As such, I will leave this open so any go experts have the opportunity to take a peak.

pokey:
Cc/ @josharian

josharian:
Will look early next week. Sorry about the delay.

lang/go/code-common-functions.talon-list
```diff
@@ -0,0 +1,20 @@
+list: user.code_common_function
```
josharian, line 1: This it the list of predeclared functions, but many of them aren't very common. (Also: `make`?)

These are the top ten identifiers across a pretty large Go corpus:

```
  4.62%   4.62% 10517712 err
  3.14%   7.75% 7149343 nil
  2.64%  10.40% 6019367 string
  2.57%  12.97% 5859233 t
  1.13%  14.09% 2564886 error
  0.99%  15.09% 2267076 c
  0.91%  16.00% 2081426 name
  0.87%  16.87% 1982235 s
  0.83%  17.70% 1881744 _
  0.81%  18.50% 1835017 ctx
```

len is #16 in the list. append is #35. close is #96. println is #197 (and note that that includes uses of fmt.Println!). clear is #1725. imag doesn't even make [the top 5000](https://gist.github.com/josharian/593ff6dc72e249d838c20484c5b61550).

I'd rather not worry about the built-ins and choose based on use.

My 2c: I'd keep len, append, close, cap, clear, delete, max, min, new, panic, and recover. And ditch the rest.

And consider maybe adding important std calls, such as `errors.New`, `fmt.Errorf`, `fmt.Printf`, and the like. If it'd be helpful, I can pull usage stats on those.

FWIW, here's what I have for package fmt (some of which are pretty metal):

```
# fmt
fmt dot print: fmt.Print
thumped up print: fmt.Print

fmt dot print line: fmt.Println
thumped up print line: fmt.Println

fmt dot print ef: fmt.Printf
thumped up print ef: fmt.Printf
fmt dot print dev: fmt.Printf
thumped up print dev: fmt.Printf
fmt dot print death: fmt.Printf
thumped up print death: fmt.Printf

# # this is how conformer D spells/hears "error eff"
fmt dot erf: fmt.Errorf
thumped up erf: fmt.Errorf

fmt dot sprintf: fmt.Sprintf
thumped up sprintf: fmt.Sprintf
fmt dot sprint dev: fmt.Sprintf
thumped up sprint dev: fmt.Sprintf
fmt dot sprint death: fmt.Sprintf
thumped up sprint death: fmt.Sprintf
```

Also, there are common identifiers that are hard to pronounce but aren't function calls, like `req`, `resp`, `cmd`. How do those fit in here?

lang/go/code-types.talon-list
```diff
@@ -0,0 +1,36 @@
...
+complex one twenty eight: complex128
+complex one two eight: complex128
+
+number: int
+integer: int
+int: int
+integer eight: int8
+integer sixteen: int16
+integer thirty two: int32
+integer sixty four: int64
```
josharian, line 22: I believe these are more commonly pronounced `int sixty four`. Here's most of my types talon-list:

```
int sixty four: int64
int sixty four: int64
you int sixty four: uint64
int thirty two: int32
you int thirty two: uint32
int sixteen: int16
you int sixteen: uint16
int eight: int8
you int eight: uint8
float thirty two: float32
float sixty four: float64
complex sixty four: complex64
complex one twenty eight: complex128
you int pointer: uintptr
you int: uint
```

lang/go/go.talon
```diff
@@ -1,195 +1,37 @@
...
-    insert(user.formatted_text(text, "PUBLIC_CAMEL_CASE"))
-
-length <user.text> [over]:
-    insert("len(")
-    insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
-
-append <user.text> [over]:
-    insert("append(")
-    insert(user.formatted_text(text, "PRIVATE_CAMEL_CASE"))
+[state] context: " ctx "
```
josharian, line 32: FWIW, I pronounced `ctx` as `cats`, to disambiguate with the full word `context`.

lang/go/go.talon
```diff
@@ -1,195 +1,37 @@
...
+[state] context (are you | argue): " ctx context.Context "
+state any: " any "
 
 state (air | err): "err"
```
josharian, line 36: I find `err` and `error` to both have high misrecognition rates. This is more idiosyncratic, but I use `oops` for `err` and `boom` for `error`. Not sure I can fairly foist those on others.

josharian (commented):
Some initial thoughts, sorry again for the delay, I'm drowning in other stuff.

lang/go/code-common-functions.talon-list
poetworrier: apologies if i'm misunderstanding the context here

'make' is necessary to initialize maps, slices, chans. its pretty essential

what is the Go corpus you looked at?

josharian: Yes, my parenthetical was saying we need `make`. :)

The corpus: https://github.com/mvdan/corpus

poetworrier: oh man, i worried about that, sorry!
kinda interesting make doesnt show up that often then?

josharian: Yeah. In particular, I'm surprised that `new` shows up more than `make`.

Of course, that's a list of _all_ identifiers, not just functions.

knausj85:
I'm going to convert this one to draft and tell someone has time to pick it up.

PR #1315, opened 2023-11-16
Get snippets actions

AndreasArvidsson:
Added actions to get insertion and wrapper snippet content(body, scope, etc). This will be used by example cursorless to leverage the community snippets

Follow up from:
https://github.com/talonhub/community/pull/1305


core/snippets/snippet_types.py
```diff
@@ -30,3 +30,16 @@ def get_variable_strict(self, name: str):
...
+@dataclass
+class InsertionSnippet:
+    body: str
+    scopes: list[str] = None
+
+
+@dataclass
+class WrapperSnippet:
+    body: str
+    variableName: str
```
AndreasArvidsson, line 44: @pokey In your example code in the previous poll request this was a list. Was that just a typo?

core/snippets/snippets_parser.py
```diff
@@ -54,6 +54,7 @@ def create_snippet(
         name=document.name or default_context.name,
         languages=document.languages or default_context.languages,
         phrases=document.phrases or default_context.phrases,
+        insertion_scopes=document.insertionScopes or default_context.insertionScopes,
```
AndreasArvidsson, line 57: Missed this line in the original pull request. This is just a bugfix.

core/snippets/snippet_types.py
pokey: Yep

core/snippets/snippet_types.py
```diff
@@ -30,3 +30,16 @@ def get_variable_strict(self, name: str):
...
+    scope: list = None
```
pokey, line 45: should be `str` no?

AndreasArvidsson: Fixed

pokey:
Ok another thought: what if instead of having a separate list and action, we just had a capture that returned the result of running the action on the list value?

AndreasArvidsson:
> Ok another thought: what if instead of having a separate list and action, we just had a capture that returned the result of running the action on the list value?

Doable, but feels a bit weird to throw exceptions in a capture?

pokey:
Why would we throw exceptions?

AndreasArvidsson:
> Why would we throw exceptions?

The code is guarded against invalid identifiers. Trying to use insertion snippet to wrap and so on

pokey:
But that's not possible if it's coming from a capture, because in the rule for the capture you use the right list, no?

AndreasArvidsson:
> But that's not possible if it's coming from a capture, because in the rule for the capture you use the right list, no?

If someone puts incorrect stuff into the list that the capture is consuming

pokey:
> If someone puts incorrect stuff into the list that the capture is consuming

Hmm I guess if they override the list in their own context? I guess it's possible. Feels like they've kind of done it to themselves at that point, but yeah I don't feel strongly; I think the version we have here is like one extra line to call the action so 🤷

AndreasArvidsson:
I thought about it and I think I prefer the actions. You can just call them from your own custom context which makes them a bit more versatile than just a capture.

pokey (commented):
made some minor tweaks and left minor comment

core/snippets/snippet_types.py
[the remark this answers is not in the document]
AndreasArvidsson: In the original it's just a representation of what they are called in the snippet file. I'm fine with changing it here.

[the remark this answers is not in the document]
AndreasArvidsson: done

knausj85 (approved):
This looks reasonable to me.

core/snippets/README.md
```diff
@@ -7,7 +7,15 @@ Custom format to represent snippets.
 - Custom file ending `.snippet`.
 - Supports syntax highlighting in VSCode via an [extension](https://marketplace.visualstudio.com/items?itemName=AndreasArvidsson.andreas-talon)
 - Supports auto-formatting in VSCode via an [extension](https://marketplace.visualstudio.com/items?itemName=AndreasArvidsson.andreas-talon)
-- Support for insertion and wrapper snippets. Note that while the snippet file syntax here supports wrapper snippets, we still need to add the proper voice commands; stay tuned.
+- Support for insertion and wrapper snippets. Note that while the snippet file syntax here supports wrapper snippets, you will need to install [Cursorless](https://www.cursorless.org/) for wrapper snippets to work. You'll also need to add the following line to your `settings.talon` file:
+
+  ```talon
+  tag(): user.cursorless_use_community_snippets
+  ```
```
pokey, line 14: This isn't really a problem for the average user, but for Cursorless maintainers, they'll need to enable this tag in a `mode: command` context. Otherwise the local Cursorless Talon spoken form tests will break, because it's not possible to deactivate a tag. Any thoughts on how to handle that @AndreasArvidsson?

AndreasArvidsson: Where we're actually using this we could just put mode command?

pokey: update from meet-up:

- let's just leave this as is, and when Cursorless developers run into local talon test failures, we tell them to move the `tag(): user.cursorless_use_community_snippets` to a file with `mode: command` at the top

pokey:
update from meet-up:

- run one last local test and then ship this alongside https://github.com/cursorless-dev/cursorless/pull/1998

PR #1318, opened 2023-11-20
Replace snippet variables with numbered holes

AndreasArvidsson:


pokey:
I thought we supported named snippet placeholders?

AndreasArvidsson:
As substitutions yes, but that requires you to actually call the python api with those substitutions. When you just insert a snippet we don't do any transformation like Cursorless. The default text made behavior is to insert the variable name.

pokey:
Maybe we should use cursorless snippet insertion when it's available? The behaviour of inserting the name isn't wrong, and it leaves it selected, I believe, so i think it's still usable without cursorless

Not opposed to also making these numbers; I personally prefer names but given all the others are numbers I guess more consistent for the time being

AndreasArvidsson:
All other snippets in community use numbers. I think we should make this change and then we can have a discussion later about how to handle variables?

core/snippets/snippets/constructorDeclaration.snippet
```diff
@@ -12,7 +12,7 @@ constructor($1) {
 
 language: java
 -
-$modifiers $1($2) {
```
knausj85, line 15: Did we intentionally lose a parameter here? The spot for modifiers appears to be gone

AndreasArvidsson: Yes. Having two snippet holes only separated by space isn't really that practical. When I use these snippets I substituted the modifiers before inserting the snippet.

knausj85 (approved):

PR #1320, opened 2023-11-24
Turning off listening mode after a timeout

mik-menshikov:
Added a setting that allows to put Talon into Sleep mode automatically after a certain timeout expressed in minutes. 
On timeout the large notification "sleep mode" is shown at the center of the main screen to indicate mode change.


fidgetingbits (changes requested):
This seems like a nice idea after a quick review. 
 
One of talon strong points is it that shouldn't need to restarted almost ever, so this should avoid restart being required imo. 

To do this you could instead read the timeout from the settings from inside the called functions, rather than once at startup.

Also, you could register the post_phrase hook regardless of the timeout value, and then inside the callback check the setting value to decide if it a timeout handler should be used. The perf hit for flag check should be negligible, so I think it's OK.

knausj85:
I really like this functionality, thanks @mik-menshikov. I made some changes after testing and experimenting with it a bit. This is ready for review again.

knausj85 (approved):

knausj85 (approved):

PR #1323, opened 2023-12-01
add code_insert_if_let_okay to rust Actions module

jmegner:
without this, if you are in rust language mode, you will get a bunch of `WARNING actions: skipped because they have no matching declaration: (user.code_insert_if_let_okay)` (from rust.talon) and some other talon stuff won't work

### DIFF
--- a/lang/rust/rust.py
+++ b/lang/rust/rust.py
@@ -20,6 +20,9 @@ class Actions:
     def code_insert_if_let_error():
         """Inserts if let error block, positioning the cursor appropriately"""
 
+    def code_insert_if_let_okay():
+        """Inserts if let okay block, positioning the cursor appropriately"""
+
     def code_insert_trait_annotation(type: str):
         """Inserts type annotation for implementor of trait"""
 


auscompgeek:
… I don't understand why this is its own action in the first place. I don't see why this should ever be overridden.

jmegner:
Maybe @RyanRuane can explain.

RyanRuane:
As for why the function is missing from the parental actions class, just a sloppy oversight. 

As for why there is an actions class, this is just inexperience and copying examples I had seen elsewhere in the codebase, such as in `lang/tags/comment_block.py`. I haven't done much work on the repository since but I was under the impression that defining the actions class was talon idiomatic and where people were storing docstrings. It seems to be pretty prevalent in some other areas of the code base, but I do see that all of the other language definitions simply have a `UserActions` class. My advice would be to simply remove the parent `Actions` class and be in keeping with the other languages. I don't suppose there's much need for the docstrings either considering the obviousness of what the code is doing, but if one wanted to preserve the docstrings, they could always move them down to `UserActions`, provided that does not violate some community specific convention of which I am unaware.

PR #1326, opened 2023-12-07
Adds basic support for the Arc web browser

jaresty:
- this is heavily borrowed from the Chrome support


apps/arc/arc.talon
```diff
@@ -0,0 +1,14 @@
+app: arc
+-
+tag(): browser
+tag(): user.tabs
+
+profile switch: user.arc_mod("shift-m")
```
david-tejada, line 6: This doesn't work in Arc. There is the possibility of switching spaces with a shortcut, maybe you want to use that. Although the shortcuts for changing to a specific space (like ctrl-2) aren't working for me. Maybe something is interfering.

apps/arc/arc.talon
```diff
@@ -0,0 +1,14 @@
...
+
+tab search: user.arc_mod("shift-a")
```
david-tejada, line 8: This also doesn't work in Arc. You can achieve something similar using cmd-t, although that is more similar to the command palette in vscode. Maybe you can replace this with a similar command to `please` in vscode.

apps/arc/arc.talon
jaresty: You are right about the profile switch not working. For me the ctrl-<number> shortcuts are working. I'm not sure why they're not working for you. For now I'll just remove this command.

apps/arc/arc.talon
jaresty: This is a great idea!! I've added a please command and I like it a lot. Thank you so much for the suggestion.

apps/arc/little-arc.talon
```diff
@@ -0,0 +1,6 @@
+user.running: arc
+-
+little arc [<user.text>]:
```
david-tejada, line 3: Maybe add a comment here that this needs to be enabled in the Arc settings to work globally.

david-tejada:
I added a comment. By the way, I don't have write access so my review won't be enough for the "review required" criteria. But don't worry, someone will review it and merge it when they have time.

apps/arc/little-arc.talon
jaresty: Thank you! I'm not sure I understood the feedback - is there some setting that you're aware of? I thought this was just enabled by default.

david-tejada: Maybe it's enabled by default and I turned it off and forgot.

Screenshot 2023-12-13 at 09 57 56

jaresty: Got it! Added a comment to clarify that requirement.

fidgetingbits (approved):
Looks good to me. I haven't tested, but it looks simple enough.

auscompgeek (changes requested):
The keys in the Talonscript appear to be Mac-specific. Given Arc is now available for Windows, could you please either implement the commands for Windows too, or move the commands to Mac-specific files?

jaresty:
Sure thing. I don't have a windows box to test so I'll rename them.

apps/arc/little-arc_mac.talon
```diff
@@ -0,0 +1,7 @@
+user.running: arc
+-
```
auscompgeek, line 2: These need to match `os: mac` to be Mac-specific.

auscompgeek (changes requested):

apps/arc/little-arc_mac.talon
jaresty: Thanks - changed.

fidgetingbits:
I think this one is ok to merge now. The additional changes to mac only lgtm.

PR #1327, opened 2023-12-07
Catch redundant sleep/wake commands, and notify the user that Talon is already in the requested mode

SkepticMystic:
This commit adds some commands to catch redundant sleep/wake commands, and notifies the user.
The main example is, if Talon is already awake, the "talon wake" and "wake up" commands aren't active. So if the user is unsure of their current mode and says "wake up", Talon interprets the command as something else ("page up", for example).
These new commands are active in the mode the user is trying to activate. They catch the command, and trigger an `app.notify` to let the user know they're already in that mode.

Before https://github.com/talonhub/community/pull/1215, these were bound with `mode: all`, so using them in the wrong mode would still be recognized, but just no-op.


knausj85 (approved):

core/modes/modes.talon
```diff
@@ -10,3 +10,9 @@ not mode: sleep
     mode.disable("sleep")
     mode.disable("dictation")
     mode.enable("command")
+
+# Without this command, talon will interpret "wake up" as another command ("page up", for example)
+# This happens because the _real_ "wake up" is only active in mode:sleep
+# These commands catch the redundant "wake up" command and notify the user that talon is already awake
+^(wake up)+$: app.notify("talon is already awake")
+^(talon wake)+$: app.notify("talon is already awake")
```
phillco, line 18:
```suggestion
^(wake up|talon wake)+$: app.notify("talon is already awake")
```

core/modes/sleep_mode.talon
```diff
@@ -36,3 +36,8 @@ settings():
     user.talon_mode()
 
 ^(talon wake)+$: speech.enable()
+
+# Catch the redundant sleep command, and notify the user that Talon is already asleep.
+# For a note about the optional <phrase>, see to_sleep_mode.talon.
+^talon sleep [<phrase>]$: app.notify("talon is already asleep")
+^go to sleep [<phrase>]$: app.notify("talon is already asleep")
```
phillco, line 43:
```suggestion
^(talon|go to) sleep [<phrase>]$: app.notify("talon is already asleep")
```

core/modes/modes.talon
```diff
@@ -10,3 +10,9 @@ not mode: sleep
     mode.disable("sleep")
     mode.disable("dictation")
     mode.enable("command")
+
+# Without this command, talon will interpret "wake up" as another command ("page up", for example)
+# This happens because the _real_ "wake up" is only active in mode:sleep
+# These commands catch the redundant "wake up" command and notify the user that talon is already awake
```
phillco, line 16:
```suggestion
# This happens because the wake commands are only bound in sleep mode.
# These commands catch the redundant wake commands in other modes, and notifies the user that talon is already awake.
```

SkepticMystic:
Thanks @phillco. I agree with your suggestions. I've commited the changes :)

core/modes/modes.talon
```diff
@@ -10,3 +10,8 @@ not mode: sleep
     mode.disable("sleep")
     mode.disable("dictation")
     mode.enable("command")
+
+# Without this command, talon will interpret "wake up" as another command ("page up", for example)
+# This happens because the wake commands are only bound in sleep mode.
+# These commands catch the redundant wake commands in other modes, and notifies the user that talon is already awake.
```
phillco, line 16:
```suggestion
# Normally, the wake commands are only bound in sleep mode.
# However, sometimes people accidentally use them if Talon is already active, and if they're not bound, they'll likely be recognized as a different command (like "page up").
# So we bind them here, and show a notification instead to guide the user.
```

core/modes/modes.talon
```diff
@@ -10,3 +10,8 @@ not mode: sleep
...
+^(wake up | talon wake)+$: app.notify("talon is already awake")
```
phillco, line 17:
```suggestion
^(wake up | talon wake)+$: app.notify("Talon is already awake")
```

core/modes/sleep_mode.talon
```diff
@@ -36,3 +36,7 @@ settings():
     user.talon_mode()
 
 ^(talon wake)+$: speech.enable()
+
+# Catch the redundant sleep command, and notify the user that Talon is already asleep.
+# For a note about the optional <phrase>, see to_sleep_mode.talon.
+^(talon | go to) sleep [<phrase>]$: app.notify("talon is already asleep")
```
phillco, line 42:
```suggestion
^(talon | go to) sleep [<phrase>]$: app.notify("Talon is already asleep")
```

phillco (approved):
Sorry, one more round of suggestions 😅 

Overall, I think this is a nice change to help with the UX migration from the old behavior.

Simply re-adding these commands in the other modes would be fine too, but I actually like showing notification slightly better, since it tells the user that they're doing something they don't need to.

core/modes/sleep_mode.talon
```diff
@@ -36,3 +36,7 @@ settings():
     user.talon_mode()
 
 ^(talon wake)+$: speech.enable()
+
+# Catch the redundant sleep command, and notify the user that Talon is already asleep.
+# For a note about the optional <phrase>, see to_sleep_mode.talon.
+^(talon | go to) sleep [<phrase>]$: app.notify("Talon is already asleep")
```
knausj85, line 42: After taking a fresh look at this, I wonder if it's worth taking some time to either wrap these behavior in an action that we can override or override talon's actions based on the active mode. This would have the advantage that we could maybe define the commands in one place, potentially making it easier for folks to customize the commands and not miss that they exist in additional files.

pokey: Yeah I agree. Would be better for spoken form to be defined in one place

SkepticMystic: Sure yeah, I agree that sounds useful.
I don't really know how to do that though :sweat_smile:

knausj85: I took a quick crack at making this change. Please review/test when you can

SkepticMystic: Huh ok, I think I understand. So you're overriding it closer to the source. It's easier to reason about.
I'm going to try test this now

SkepticMystic: Yup, works fine for me. I don't use Dragon, so can't confirm those changes. But the regular sleep/wake works fine, and I get notified if Talon is already in the target mode.

core/modes/modes.py
```diff
@@ -12,6 +14,27 @@
 for key, value in modes.items():
     mod.mode(key, value)
 
+ctx_sleep.matches = r"""
+mode: sleep
+"""
+
+ctx_awake.matches = r"""
+mode: command
+mode: dictation
```
pokey, line 23: not sure it's better, but could we do `not mode: sleep` instead? Also, I wonder if there's a context that more directly corresponds to speech being enabled / disabled 🤔

core/modes/modes_not_dragon.talon
```diff
@@ -0,0 +1,32 @@
...
+# engine, this controls Talon.
+^(wake up)+$: speech.enable()
+
+# We define this *only* if the speech engine isn't Dragon, because if you're using Dragon,
+# "go to sleep" is used to specifically control Dragon, and not affect Talon.
+#
+# It's a useful and well known command, though, so if you're using any other speech
+# engine, this controls Talon.
+#
+# For a note about the optional <phrase>, see to_sleep_mode.talon.
```
pokey, line 21: `to_sleep_mode.talon` no longer exists. I wonder if we should move those comments here?

core/modes/modes_not_dragon.talon
```diff
@@ -0,0 +1,32 @@
...
+not speech.engine: dragon
+-
+^talon wake [<phrase>]$: speech.enable()
+
+# We define this *only* if the speech engine isn't Dragon, because if you're using Dragon,
+# "wake up" is used to specifically control Dragon, and not affect Talon.
+#
+# It's a useful and well known command, though, so if you're using any other speech
+# engine, this controls Talon.
+^(wake up)+$: speech.enable()
```
pokey, line 13: how come this one doesn't have an optional phrase like the others? too common an utterance?

pokey (commented):
Looks good! Left some comments

core/modes/modes.py
knausj85: We could add a custom scope that simply indicates whether or not speech is enabled?

core/modes/modes_not_dragon.talon
knausj85: couldn't tell ya, I only toggle speech via a button these days

core/modes/modes.py
pokey: Is that possible?

knausj85: Looking into this, it appears it is not possible to do a custom scope for this. In particular, I'm not sure when to call scope.update.

It probably makes sense as a future request to add to the existing speech scope. 

As long as there are no objections to not mode: sleep, I'll move forward with that

core/modes/modes_not_dragon.talon
knausj85: Yeah, I didn't change this:
https://github.com/talonhub/community/blob/main/core/modes/sleep_mode_not_dragon.talon

I don't really see the use case for phrase with wake up tbh

pokey: oh yeah good point

core/modes/modes_not_dragon.talon
```diff
@@ -0,0 +1,41 @@
+mode: sleep
+not mode: sleep
```
pokey, line 2: Is this saying "mode sleep or not mode sleep"? That is always true, right? Can't we just remove both of these lines?

pokey (commented):
Looks good with one final comment

core/modes/modes_not_dragon.talon
knausj85: Without any modes specified, a talon file will be implicitly in command mode only. I don't think there's another way to accomplish this other than mode: all, which would become in full circle

knausj85 (approved):

PR #1328, opened 2023-12-07
Azure Data Studio Enabled for Cursorless

Mark-Phillipson:
Just need to add the following four lines of code at line 43 to apps\vscode\vscode.py

```py
os: windows
and app.name: Azure Data Studio
os: windows
and app.exe: azuredatastudio.exe
```

The other changes can be ignored as they are probably no good for anybody else.


Mark-Phillipson (commented):
seems reasonable. Just removing a space.

pokey:
How different are Azure Data Studio and VSCode? Are the commands in [`vscode.talon`](https://github.com/talonhub/community/blob/main/apps/vscode/vscode.talon) generally applicable to Azure Data Studio? If so, then this PR makes sense. If not, then we should use a different approach, eg creating a new Azure Data Studio app definition and then adding that app to the list of apps supported by Cursorless

This PR makes it so it's hard to distinguish between whether you have Azure Data Studio or VSCode focused

If it makes sense to move forward, it would also be good to remove the extraneous commits from this PR; you'll want to use a separate branch in your repo rather than your main branch

Mark-Phillipson:
Yes Azure data studio seems to behave exactly the same as Visual Studio code as far as Cursorless is concerned.

However a lot of the commands in vscode.talon will not apply for example jupiter notebooks.

I will work on creating a separate azure_data_studio.talon file and removing the superfluous stuff.

PR #1331, opened 2023-12-10
lang: imperative: Remove unused code_state_continue action

auscompgeek:
No language implements this, and there's no voice command for this action either.


fidgetingbits (approved):
Ah ya, that's a leftover from when I added lua and didn't realized next existed.

PR #1332, opened 2023-12-10
lang: Add tag for brace-delimited code blocks

auscompgeek:
## Problem

Several languages had the same, or similar, copy-pasted implementations of `code_block` to insert braces and leave the cursor on a blank line between them. These all assume the editor will appropriately expand <kbd>enter</kbd> between brackets, which isn't true for all editors.

The C# implementation implemented the same idea, but appears to assume <kbd>enter</kbd> _isn't_ expanded.

Furthermore, some C-like languages were missing a `code_block` impl entirely.

## Solution

Similar to #720, this adds a common implementation for braced code blocks, which can be overridden on a per-editor basis, and adds the tag to the appropriate languages.

### DIFF
--- a/lang/c/c.talon
+++ b/lang/c/c.talon
@@ -2,6 +2,7 @@ code.language: c
 -
 tag(): user.code_imperative
 
+tag(): user.code_block_c_like
 tag(): user.code_comment_line
 tag(): user.code_comment_block_c_like
 tag(): user.code_data_bool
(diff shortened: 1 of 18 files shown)


fidgetingbits (approved):
Looks sane to me, good idea. I won't merge this one right away only because it touches lots of languages and maybe someone has opinions about something I didnt think of.

knausj85:
This looks good to me too.

PR #1333, opened 2023-12-15
Add recommended snip shortcuts

jaresty:
- These shortcuts are unambiguous for navigating snippets in VSCode

Thanks to @AndreasArvidsson for [the example](https://github.com/AndreasArvidsson/andreas-talon/blob/59cfecd26c3a01e72847364f6cafb5cdb8202b1b/apps/vscode/vscode.talon#L239)

### DIFF
--- a/apps/vscode/vscode.talon
+++ b/apps/vscode/vscode.talon
@@ -55,8 +55,8 @@ show shortcuts json: user.vscode("workbench.action.openGlobalKeybindingsFile")
 show snippets: user.vscode("workbench.action.openSnippets")
 
 # VSCode Snippets
-snip last:                  user.vscode("jumpToPrevSnippetPlaceholder")
-[snip] next:                user.vscode("jumpToNextSnippetPlaceholder")
+snip last: user.vscode("jumpToPrevSnippetPlaceholder")
+[snip] next: user.vscode("jumpToNextSnippetPlaceholder")
 
 # Display
 centered switch: user.vscode("workbench.action.toggleCenteredLayout")
--- a/apps/vscode/vscode.talon
+++ b/apps/vscode/vscode.talon
@@ -55,7 +55,7 @@ show shortcuts json: user.vscode("workbench.action.openGlobalKeybindingsFile")
 show snippets: user.vscode("workbench.action.openSnippets")
 
 # VSCode Snippets
-snip (last|previous): user.vscode("jumpToPrevSnippetPlaceholder")
+snip (last | previous): user.vscode("jumpToPrevSnippetPlaceholder")
 snip next: user.vscode("jumpToNextSnippetPlaceholder")
 
 # Display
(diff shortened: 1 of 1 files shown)


apps/vscode/vscode.talon
```diff
@@ -54,6 +54,10 @@ show shortcuts: user.vscode("workbench.action.openGlobalKeybindings")
 show shortcuts json: user.vscode("workbench.action.openGlobalKeybindingsFile")
 show snippets: user.vscode("workbench.action.openSnippets")
 
+# VSCode Snippets
+snip last: user.vscode("jumpToPrevSnippetPlaceholder")
+[snip] next: user.vscode("jumpToNextSnippetPlaceholder")
```
pokey, line 59: This is what I use and it's great. I do wonder if it's too short to have by default? I also wonder if there should be a generic action for moving placeholders that we implement this way in VSCode cc/ @knausj85 @AndreasArvidsson

AndreasArvidsson: Maybe remove the brackets?
An action is probably better as a general rule.

AndrewDant: I would support removing the brackets. 'Next' maybe fine for your personal setup but would conflict with mine and probably some other peoples'.

jaresty: Removed the brackets! Thanks for the feedback.

nriley:
This made me realize we have no consistency between "last" and "previous" in community. "Last" is definitely used across VSCode but there are a few examples of previous (e.g., "hunt previous").

What do you think about supporting both? This use of "last" is not actually correct based on the definition of the word — i.e., the definitions are "coming after all others in time or order", "most recent in time" and "only remaining" but this is using it to mean "most recent in order".  Whereas "previous" means "existing or occurring before in time or order", so still ambiguous (I think we usually use back/forward for navigating in time) but at least covers what we're doing.

jaresty:
> This made me realize we have no consistency between "last" and "previous" in community. "Last" is definitely used across VSCode but there are a few examples of previous (e.g., "hunt previous").
> 
> What do you think about supporting both? This use of "last" is not actually correct based on the definition of the word — the valid definitions are "coming after all others in time or order", "most recent in time" and "only remaining" but this is using it to mean "most recent in order". "Previous" means "existing or occurring before in time or order", so still ambiguous (I think we usually use back/forward for navigating in time) but at least covers what we're doing.

Make sense to me! I added the requested alias.

apps/vscode/vscode.talon
```diff
@@ -56,6 +56,7 @@ show snippets: user.vscode("workbench.action.openSnippets")
 
 # VSCode Snippets
 snip last: user.vscode("jumpToPrevSnippetPlaceholder")
+snip previous: user.vscode("jumpToPrevSnippetPlaceholder")
```
nriley, line 59: You can use `snip (previous | last)`.

nriley (changes requested):
Sorry to nit-pick your literal one-line patch! Thanks.

nriley (approved):

apps/vscode/vscode.talon
jaresty: Sounds good!

PR #1336, opened 2023-12-18
Adds commands to focus on the last application

jaresty:
- Sometimes it is useful to switch back to the last application you were using. This adds commands to do that.


core/windows_and_tabs/window_management_mac.talon
```diff
@@ -0,0 +1,3 @@
+os: mac
+-
+focus last: key(cmd-tab)
```
knausj85, line 3: I'd be inclined to implement this functionality via an action so it's easier to maintain a consistent command set across operating systems. The downside of maintaining multiple .talon in this fashion is that it makes it a little easier for the commands to drift apart. If implemented via an action, we can maintain a window_management file and override the actions as necessary depending on the operating system.

knausj85 (changes requested):

core/windows_and_tabs/window_management_mac.talon
jaresty: I started there but struggled with understanding the existing windows and application management python files. Do you think you could do a PR to demonstrate how that would need to change? I'll take another look when I can. I really appreciate the feedback!

jaresty: I took a stab at this. I don't have a Linux machine handy I took a guess at how to implements the same functionality. I would love a second set of eyes if someone has a chance.

core/windows_and_tabs/windows_and_tabs_linux.py
```diff
@@ -38,3 +38,6 @@ def window_hide_others():
 
     def window_open():
         actions.key("ctrl-n")
+
+    def focus_last():
```
knausj85, line 42: One subtlety with actions is that they need to be declared somewhere first. The app actions are owned by talon and can't be extended. 
 
I would suggest declaring this in app_switcher.py

    def switcher_focus_last():
        """Focus last window/application"

Then for each of the supported operating systems you just need to add the appropriate implementation.  eg the below for [windows_and_tabs_mac.py](https://github.com/talonhub/community/pull/1336/files#diff-91df1befe13e34f8155ad02af92009a9900e4f2d1734ca458d13b2ae536fcaf8)

@ctx.action_class("user")
class UserActions:
    def switcher_focus_last():
        actions.key("cmd-tab")

knausj85 (changes requested):

core/windows_and_tabs/windows_and_tabs_linux.py
jaresty: Okay-how does it look now?

knausj85 (approved):
Looks great, thanks

PR #1337, opened 2023-12-18
Added prefix to disambiguate numbers

jaresty:
- This is one of the common cursorless [recommendations](https://github.com/cursorless-dev/cursorless/wiki/Common-recommendations)

### DIFF
--- a/core/numbers/numbers.talon
+++ b/core/numbers/numbers.talon
@@ -1,3 +1,3 @@
 not tag: user.mouse_grid_showing
 -
-<user.number_string>: "{number_string}"
+numb <user.number_string>: "{number_string}"


pokey:
This is going to break a lot of people's setups when they upgrade. We should first align on whether this is a direction we want to go, and then think if there's a way to do this using our deprecation infrastructure. Cc/ @knausj85

jaresty:
Let me know if there are changes needed on my end to leverage the deprecation infrastructure. Also, I can add you as a contributor if you want to push to my branch to make the changes directly.

pokey:
> I can add you as a contributor if you want to push to my branch to make the changes directly.

no need

image

fidgetingbits:
Agreed this is a tough one. IIRC it's been this way without prefix since I started using talon many years ago, and I suspect most (?) users didn't change it. Could mine the forks to check how often people change it, though that wouldn't entirely be representative of course.

For a change that big I'd almost want to have a more aggressive notification, like a deprecation notification issued on first use without prefix each time you start talon? 

It was one of the first things I changed in my fork I think. That said, given it lived this long without ever being changed I do wonder if it's actually worthwhile to make it the default? To me it comes down to the burden of support on slack mostly? The support burden of leaving it without the prefix as the default may actually be much less than changing it (at least for awhile).

jaresty:
There isn't a way to override this without forking the repository, is there?

fidgetingbits:
Easiest I think is you could edit it in the clone of community and just not commit the change, then when you update the repo git stash first and then git stash pop after. You could create a script or alias to automatically do that when you update the repo. Someone else might have a better idea.

jaresty:
Yeah, I understand that workflow. That's not better than forking IMO. Another approach could be for this prefix to be configured via a setting that isn't committed to the repo, which would unlock folks from having to fork for such a small change.

jaresty:
[Here](https://github.com/talonhub/community/pull/1338) is an alternative approach which allows users to define their own talon configuration file that specifies a tag to enable the numb prefix.

knausj85:
I personally think this is the best route to go. I'll leave it open for others to weigh in.

jaresty:
Some time has passed and nobody's weighed in so far. Does it seem like this might be good to go?

knausj85:
> I've pushed to the #1342 branch with how I think we should approach it. I think we should close this (#1337) and #1338 and move discussion to #1342. I've tested the changes on my fork, but didn't test the actual PR branch, so would appreciate others who weighed in (@knausj85, @pokey, @auscompgeek) to take a look to see if I missed anything.

I think we have sufficient consensus, closing this issue in favor of  #1342

Issue #943, opened 2022-07-31
Dedicated "disable numbers" tag


auscompgeek:
Currently `numbers.talon` depends on the mouse grid, which is rather silly:

https://github.com/knausj85/knausj_talon/blob/bfb282178d78880fad6b69b76f7dd097208f1f65/text/numbers.talon#L1

One of the main reasons why people (soft-)fork knausj_talon is to disable `numbers.talon` altogether, prefixing the number command with `numb` or similar due to Conformer b108 being biased towards numbers.

We should remove the dependency on the mouse grid, and make it easier for people to use knausj_talon without having to make common modifications such as this.

PR #1338, opened 2023-12-20
configurable number prefix

jaresty:
- Attempt to implement optional numb prefix
- The tags need to be logically combined

### DIFF
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -177,6 +177,7 @@ for ten in tens:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
+mod.tag("require_numb_prefix", desc="require number prefix when saying a number")
 ctx.lists["self.number_small"] = number_small_map.keys()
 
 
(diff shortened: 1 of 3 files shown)


jaresty:
This is an alternative approach to requiring all users to [adopt a new prefix](https://github.com/talonhub/community/pull/1337)

auscompgeek:
I think instead of adding complexity to the matcher, we should instead have one tag to disable unprefixed numbers. See #943 for my thoughts

jaresty:
Thanks for the suggestion! I didn't totally understand it, though. Can you show me some pseudocode demonstrating what you mean? (Or another PR that I could look at is ok as well)

jaresty:
Ah, I think I'm getting it. You mean to suggest removing the mouse  grid check altogether?

fidgetingbits:
The approach auscompgeek mentioned seems cleaner to me.

iiuc the idea would be to have a single tag  user.disable_unprefixed_numbers and then you don't need the additional mouse grid testing you have in your new files. 

The mouse grid test can then also be removed from numbers.talon. And then the mouse grid code will need to be modified to set the new user.disable_unprefixed_numbers tag when active in order to turn off unprefixed numbers, so it can then use unprefixed numbers for its own purposes.

jaresty:
Something like this? https://github.com/talonhub/community/pull/1342

knausj85:
I think we have sufficient consensus, closing this issue in favor of  #1342

PR #1342, opened 2023-12-27
configurable number prefix without mouse grid

jaresty:
- Attempt to implement optional numb prefix
- The tags need to be logically combined
- Respond to suggestion that the mouse grid should not be considered

### DIFF
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -177,6 +177,7 @@ for ten in tens:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
+mod.tag("require_numb_prefix", desc="require number prefix when saying a number")
 ctx.lists["self.number_small"] = number_small_map.keys()
 
 
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -177,7 +177,7 @@ for ten in tens:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
-mod.tag("disable_unprefixed_numbers", desc="Require prefix when saying a number")
+mod.tag("prefixed_numbers", desc="Require prefix when saying a number")
 ctx.lists["self.number_small"] = number_small_map.keys()
 
 
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -177,7 +177,7 @@ for ten in tens:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
-mod.tag("require_numb_prefix", desc="require number prefix when saying a number")
+mod.tag("disable_unprefixed_numbers", desc="Require prefix when saying a number")
 ctx.lists["self.number_small"] = number_small_map.keys()
 
 
--- a/core/numbers/numbers_unprefixed.talon
+++ b/core/numbers/numbers_unprefixed.talon
@@ -1,3 +1,5 @@
 not tag: user.disable_unprefixed_numbers
 -
-<user.number_string>: "{number_string}"
+<user.number_string>:
+    insert("{number_string}")
+    user.deprecate_command("2024-01-27", "unprefixed <user.number_string>", "prefixed <user.number_string>")
--- a/core/numbers/numbers_unprefixed.talon
+++ b/core/numbers/numbers_unprefixed.talon
@@ -1,4 +1,4 @@
-not tag: user.disable_unprefixed_numbers
+not tag: user.prefixed_numbers
 -
 <user.number_string>:
     insert("{number_string}")
--- a/core/numbers/numbers_unprefixed.talon
+++ b/core/numbers/numbers_unprefixed.talon
@@ -1,0 +1,3 @@
+not tag: user.disable_unprefixed_numbers
+-
+<user.number_string>: "{number_string}"
--- a/core/numbers/numbers_unprefixed.talon
+++ b/core/numbers/numbers_unprefixed.talon
@@ -2,4 +2,4 @@ not tag: user.prefixed_numbers
 -
 <user.number_string>:
     insert("{number_string}")
-    user.deprecate_command("2024-01-27", "unprefixed <user.number_string>", "prefixed <user.number_string>")
+    user.deprecate_command("2024-01-27", "<number>", "numb <number>")
--- a/settings.talon
+++ b/settings.talon
@@ -76,4 +76,4 @@ settings():
 # By default saying "one" would write "1", however many users find this behavior
 # prone to false positives. If you uncomment this, you will need to say
 # "numb one" to write "1".
-# tag(): user.disable_unprefixed_numbers
+# tag(): user.prefixed_numbers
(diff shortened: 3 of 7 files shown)


fidgetingbits:
Thanks for your patience. I've pushed some changes that I think are more in line with what @auscompgeek originally
suggested in #943, and that will be easier to maintain. I'll detail the changes a bit below, and why I did it this way
instead of how you did it, in case you're interested:

- The name of the 'numb' prefix is no longer in the tag name. It's possible people would change the prefix to something else, so should never include the actual spoken form in the tag name imo. The name "disable_unprefixed_numbers" also fits nicely in that mouse_grid setting the tag becomes more self-explanatory.
- Set the tag in mouse_grid, so that it still works
- Move the unprefixed version of the command in it's own file, rather than the prefixed version. This is because many users add other types of numbers (hexadecimal, etc) to numbers.talon already, which also have prefixes of various forms. So just isolating the non-prefixed version seems cleaner.
- Add documentation to the settings file so that users now how/why to use the tag.

fidgetingbits:
I guess one remaining question is if we should set the tag by default or not.

settings.talon
```diff
@@ -69,5 +69,11 @@ settings():
     # -width windows are resized to stay full-height/width.
     #user.window_snap_screen = "size aware"
 
+    # Uncomment the below to disable support for saying numbers without a prefix.
+    # By default saying "one" would write "1", however many users find this behavior
+    # prone to false positives. If you uncomment this, you will need to say
+    # "numb one" to write "1".
+    #user.disable_unprefixed_numbers = 1
```
auscompgeek, line 76: Shouldn't this example be a `tag()` rather than setting a setting?

jaresty:
my preference would be for the recommended behavior (I.e. not requiring the prefix) be the default.

settings.talon
fidgetingbits: Thanks

pokey (commented):
this looks ok to me. tempted to have it on by default, but then we'd need to do some kind of deprecation process

also, double negatives always confuse me. I wonder if we should have a `allow_unprefixed_numbers` tag that is turned on by default in settings, and the user can comment it out if they don't want unprefixed numbers

fidgetingbits:
> also, double negatives always confuse me. I wonder if we should have a `allow_unprefixed_numbers` tag that is turned on by default in settings, and the user can comment it out if they don't want unprefixed numbers

I considered that but since mouse grid needs to disable non prefixed numbers to use its own nonprefixed number commands, it seemed simpler to have a disable tag. 

Maybe there is a good way for mouse grid to remove the tag when it activates? I don't recall off hand if you can just shut off a a tag like that.

fidgetingbits:
fwiw I'm happy to have prefixed the default as well. 

random thought, but for deprecating stuff I almost wonder if we should introduce a deprecated notification action that pops a notification only the first time you use a deprecated command after starting talon (or every N times maybe), but that also spams the logs on every use. that way it's easier for less technical users to find out (since I suspect many don't read the logs), but still in a way thats not too spammy. it would be useful here and for future depreciation I suspect.

pokey:
> for deprecating stuff I almost wonder if we should introduce a deprecated notification action that pops a notification only the first time you use a deprecated command

I believe that kind of thing is already covered by our [deprecation infrastructure / policy decoumentation](https://github.com/talonhub/community/blob/main/core/deprecations.py)

core/numbers/numbers.py
```diff
@@ -177,6 +177,7 @@ def split_list(value, l: list) -> Iterator:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
+mod.tag("disable_unprefixed_numbers", desc="Require prefix when saying a number")
```
knausj85, line 180: I might suggest just calling this "prefixed_numbers"

knausj85:
I agree the prefix should be the default to improve compatibility with cursorless  / reduce misrecognitions

core/numbers/numbers_unprefixed.talon
```diff
@@ -0,0 +1,3 @@
+not tag: user.disable_unprefixed_numbers
```
knausj85, line 1: I believe we still want
 and not tag: user.mouse_grid_showing
here

pokey:
> I agree the prefix should be the default to improve compatibility with cursorless / reduce misrecognitions

makes sense, tho in that case we should prob use command deprecation infra so it doesn't just break when user updates?

core/numbers/numbers_unprefixed.talon
fidgetingbits: I don't think so because mouse_grid is updated to set the tag, which would disable numbers_unprefixed.talon. As per #943 the idea of doing it the way I did was that the core/numbers code doesn't need to be aware of the existence of mouse grid which I think is preferable. This way mouse grid itself can just toggle the functionality off, so its own functionality won't conflict. 

The downside of this is that it's a double negative tag. To me having core/numbers not having to special case for some plugin is worth the slightly confusing double negative tag, but if others disagree then can change it.

core/numbers/numbers.py
fidgetingbits: See other comment about why it's this way.

fidgetingbits:
> > for deprecating stuff I almost wonder if we should introduce a deprecated notification action that pops a notification only the first time you use a deprecated command
> 
> I believe that kind of thing is already covered by our [deprecation infrastructure / policy decoumentation](https://github.com/talonhub/community/blob/main/core/deprecations.py)

Thanks for the reminder. I just added the relevant updates. Of course wording and such will be subject what tag ends up being settled on.

core/numbers/numbers.py
auscompgeek: Could call it "numbers must be prefixed" if the double negative is too confusing?

fidgetingbits: Ok, I'll update it to be non-double negative tag then. user.prefix_numbers or user.numbers_must_be_prefixed? I think I prefer the latter because it's a bit more self explanatory, but good with whatever others prefer.

core/numbers/numbers_unprefixed.talon
```diff
@@ -0,0 +1,5 @@
+not tag: user.prefixed_numbers
+-
+<user.number_string>:
+    insert("{number_string}")
+    user.deprecate_command("2024-01-27", "unprefixed <user.number_string>", "prefixed <user.number_string>")
```
pokey, line 5:
```suggestion
    user.deprecate_command("2024-01-27", "<number>", "numb <number>")
```

settings.talon
```diff
@@ -71,3 +71,9 @@ settings():
 
 # Uncomment this to enable the curse yes/curse no commands (show hide mouse cursor). See issue #688.
 # tag(): user.mouse_cursor_commands_enable
+
+# Uncomment the below to disable support for saying numbers without a prefix.
+# By default saying "one" would write "1", however many users find this behavior
+# prone to false positives. If you uncomment this, you will need to say
+# "numb one" to write "1".
```
pokey, line 78:
```suggestion
# "numb one" to write "1". Note that this tag will eventually be activated by default
```

pokey (commented):
Looks good! Left a couple final minor comments

jaresty:
Anything left to fix on this? Wanted to make sure you weren't waiting on me 😀

fidgetingbits:
I was mostly waiting for someone else to actually approve after I addressed the last of the suggested changes, but I suppose since pokey said it was fine with the minor changes I can approve it myself.

fidgetingbits (approved):

PR #1343, opened 2023-12-28
Provide a place to configure alternate keys for a given key name.

jaresty:
- This is useful for when users want to share their keys with other people but don't want to require forking the community repository


pokey:
Alternately: do we want to use a talon list file for this rather than a csv? cc/ @knausj85

knausj85:
> Alternately: do we want to use a talon list file for this rather than a csv? cc/ @knausj85

That's my preference, but we need to settle on a migration path I suppose.

jaresty:
I haven't worked with talon list files before. If that's the preferred approach I'm fine with it, but I might need a little help to understand how to do it

jaresty:
Could you please provide a bit of guidance how to start here? I am glad to update the PR-but I don't understand how 😀 . A pointer to related work might be enough to get me started.

jaresty:
Alternatively, if there is more work needed to make talon list files easier to configure, this could be a temporary fix that can be refactored later?

pokey:
> That's my preference, but we need to settle on a migration path I suppose.

Could you elaborate? Not sure I follow

knausj85:
> > That's my preference, but we need to settle on a migration path I suppose.
> 
> @knausj85 Could you elaborate? Not sure I follow
I don't remember what I was thinking at the time. 

I'm inclined to take this as a interim improvement. Since this list is used to construct other lists (special_keys), I think there's some additional details to sort out before we can migrate this to a .talon-list.

pokey:
> Since this list is used to construct other lists (special_keys), I think there's some additional details to sort out before we can migrate this to a .talon-list

Ah right yeah. Tho I think it might be as simple as changing special_keys to a capture

I do wonder if alternate_keys.csv is the right name. It's not particularly clear what that's supposed to do, and that list is kind of an implementation detail. Maybe it should be symbol_keys?

The problem with introducing new csv's casually is you kinda get locked in to that name unless you're willing to implement a migration

I guess we could argue that migrating csvs to talon list will need to happen at some point so that's our opportunity to migrate to a set of files we like better

knausj85:
Converting to capture + using a .talon-list does seem more reasonable. I think I can take a quick swing at this today.

jaresty:
Was just thinking about this one. Did you get a chance to look at that @knausj85 ? If you get a chance to set up a pattern here I'll be excited to see how it works.

knausj85:
> Was just thinking about this one. Did you get a chance to look at that @knausj85 ? If you get a chance to set up a pattern here I'll be excited to see how it works.

I got sidetracked with other things, maybe next weekend

knausj85:
I think we're best served by actually just converting keys to leverage .talon-list. 

https://github.com/talonhub/community/pull/1382 for discussion

jaresty:
Asked a clarifying question in https://github.com/talonhub/community/pull/1382#issuecomment-1932364970 since I don't understand how it addresses the original need here.

jaresty:
Small correction to what I wrote earlier-it seems that the issue is that the talon list files are not available during the outrun code initialization. The current code loads csvs during initialization and so is completely responsible for the lifecycle of that data. When using talon lists, the loading of the lists isn't part of the Python code and since there us a connascence of execution between the loading of the data (csvs) and its usage during initialization, moving to talon lists requires the code to be restructured since the data won't be available during initialization. One path forward could be to delay this processing until it is used (instead if doing it during initialization)

https://connascence.io/execution.html

geobarry:
The ability to customize keys is an important functionality and we know that the procedure for loading csv files will work, I would say do that first and then worry about delayed loading or other attempts to use talon lists later.

jaresty:
I found a way to do this-see this pull request for examples: https://github.com/talonhub/community/pull/1477

nriley:
This has been fixed as we now have a .talon-list implementation.

jaresty:
Makes sense-thanks.

PR #1346, opened 2024-01-02
allow mouse commands to work in zoom mode

mmikeww:
previously they would sporatically work but the zoom overlay window would stay open

### DIFF
--- a/plugin/mouse/mouse.py
+++ b/plugin/mouse/mouse.py
@@ -109,6 +109,11 @@ def gui_wheel(gui: imgui.GUI):
 
 @mod.action_class
 class Actions:
+    def zoom_close():
+        """Closes an in-progress zoom. Talon will move the cursor position but not click."""
+        if eye_zoom_mouse.zoom_mouse.state == eye_zoom_mouse.STATE_OVERLAY:
+            actions.tracking.zoom_cancel()
+
     def mouse_show_cursor():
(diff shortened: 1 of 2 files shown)


mmikeww:
this now allows you to speak "`righty`" WHILE the zoom window is still active

what will now happen is:
1. cancel the zoom
2. in doing so, talon still moves the cursor to the location, it simply doesn't click
3. after the zoomed overlay is closed, talon can now perform the right click, double click, etc

plugin/mouse/mouse.py
```diff
@@ -109,6 +109,11 @@ def gui_wheel(gui: imgui.GUI):
 
 @mod.action_class
 class Actions:
+    def zoom_close():
+        """Closes an in-progress zoom. Talon will move the cursor position but not click."""
+        if eye_zoom_mouse.zoom_mouse.state == eye_zoom_mouse.STATE_OVERLAY:
+            actions.tracking.zoom_cancel()
```
knausj85, line 115: Just curious, do we need this wrapper action? What does zoom_cancel do when zoom mouse isn't enabled / or is enabled and idle?

mmikeww: probably need to ask @lunixbochs , i just put it in there for safety

knausj85 (approved):

PR #1347, opened 2024-01-03
Add ruby op-or-equals command

jaresty:
- without this, saying `op or equals` would be interpreted as `op or`
  followed by `equals`, which results in a syntax error `|| =`

### DIFF
--- a/lang/javascript/javascript.py
+++ b/lang/javascript/javascript.py
@@ -188,6 +188,9 @@ class UserActions:
     def code_operator_equal():
         actions.auto_insert(" == ")
 
+    def code_or_operator_assignment():
+        actions.auto_insert(" ||= ")
+
     def code_operator_not_equal():
         actions.auto_insert(" != ")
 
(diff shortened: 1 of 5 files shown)


auscompgeek:
IMO we should implement this for other languages too. For example JavaScript implements logical and/or assignment as well

it can probably go in https://github.com/talonhub/community/blob/main/lang/tags/operators_assignment.talon with the rest of them

jaresty:
I'm not confident that every language that inherits that file supports it; and, I'm not sure how to do it for every language. My preference would be to keep this small and targeted to what I know and then refactor after it has been implemented everywhere. I only care about Ruby today because using this operator is idiomatic in Ruby. Can we live with this approach for now?

auscompgeek:
> I'm not confident that every language that inherits that file supports it

We don't implement every one of those actions in that file for every language that asserts that tag anyway

jaresty:
Ah, ok. I'm fine with adding it to JavaScript as well then. I'll add that when I can, or please feel free to edit this or submit a new PR with the change if you get to it first.

jaresty:
I updated this based on your feedback. Thanks!

knausj85 (approved):

knausj85 (approved):

knausj85 (approved):

Issue #1306, opened 2023-11-05
Add support for a snippet directory not inside of the community folders


AndrewDant:
I would like to be able to define snippet files somewhere other than the community repo and use them with the new snippet feature, for example to allow defining a set of snippets in a private repository to keep proprietary code in snippets private while still backed up by git. Most likely this would be implemented as a setting which can contain a path for the snippet file watcher to look at in addition to the default.

AndrewDant:
@AndreasArvidsson

AndreasArvidsson:
I've now added a pull request, but I don't really have the time at the moment to fully test it out. Could you give this one a spin? #1312

pokey:
I believe this was fixed in #1312

PR #1348, opened 2024-01-03
coalesce some list updates

lunixbochs:
Snippets and file manager list updates trigger more context refreshes than necessary. Due to an indentation issue, snippets are currently triggering 101 full context re-evaluations every time snippets load/reload (linear to number of snippets).


AndreasArvidsson (dismissed):

pokey (changes requested):
I don't think this is right. `ctx` is local to the outer `for` loop; [defined in its guard](https://github.com/talonhub/community/pull/1348/files#diff-28c99d57b6152f53ca35d94c3471b72f32d5e8bc35a6edc2b4b4ca8388ddc0bcR72)

knausj85:
I'm inclined to merge this after fixing the indentation and open an issue to until someone has the bandwidth to address lunixboch's broader suggestion re: removing snippets setting.

This is at least an improvement in an interim.

pokey:
I don't think this is right. `ctx` is a local variable to the 

> If that's the case, I'm going to strongly recommend you remove the dir setting and use a directory relative to the code, so the contexts can be populated before Talon is ready.

We implemented this feature in response to a user request (https://github.com/talonhub/community/issues/1306), but if untenable, I guess we could roll it back?

> I'll also recommend in the fs.watch only updating the file that changed (if you're not doing that). This code path is currently triggering a lot of work inside Talon.

Makes sense; prob for a follow-up PR though. Filed https://github.com/talonhub/community/issues/1373 to track

pokey (approved):
Looks good to me! I have not tested locally, though

PR #1349, opened 2024-01-03
avoid warnings due to change of apis in talon

saidelike:
```
2024-01-03 09:31:45.551 WARNING C:\knausj_talon\tags\file_manager\file_manager.py:317: DeprecationWarning: SettingsDecl: setting.get() is deprecated. Use settings.get('user.file_manager_auto_show_pickers')
  if gui_folders.showing or setting_auto_show_pickers.get() >= 1:
```

PR #1351, opened 2024-01-07
Use settings.get() in more places

brief:
Builds on #1349 by converting most uses of `setting = mod.setting()` in community to `settings.get()`.

@lunixbochs, would appreciate a quick look to make sure this is correctly using settings.

### DIFF
--- a/apps/tmux/tmux.py
+++ b/apps/tmux/tmux.py
@@ -1,4 +1,4 @@
-from talon import Context, Module, actions
+from talon import Context, Module, actions, settings
 
 mod = Module()
 
@@ -7,7 +7,7 @@ tag: terminal
 and tag: user.tmux
 """
 
-setting_tmux_prefix_key = mod.setting(
+mod.setting(
     "tmux_prefix_key",
     type=str,
     default="ctrl-b",
@@ -19,7 +19,7 @@ setting_tmux_prefix_key = mod.setting(
 class TmuxActions:
     def tmux_prefix():
         """press control and the configured tmux prefix key"""
-        actions.key(f"{setting_tmux_prefix_key.get()}")
+        actions.key(f"{settings.get('user.tmux_prefix_key')}")
 
     def tmux_keybind(key: str):
         """press tmux prefix followed by a key bind"""
--- a/apps/tmux/tmux.py
+++ b/apps/tmux/tmux.py
@@ -19,7 +19,7 @@ mod.setting(
 class TmuxActions:
     def tmux_prefix():
         """press control and the configured tmux prefix key"""
-        actions.key(f"{settings.get('user.tmux_prefix_key')}")
+        actions.key(settings.get("user.tmux_prefix_key"))
 
     def tmux_keybind(key: str):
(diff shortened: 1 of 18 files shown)


apps/tmux/tmux.py
```diff
@@ -19,7 +19,7 @@
 class TmuxActions:
     def tmux_prefix():
         """press control and the configured tmux prefix key"""
-        actions.key(f"{setting_tmux_prefix_key.get()}")
+        actions.key(f"{settings.get('user.tmux_prefix_key')}")
```
nriley, line 22: Think this may have been generated due to .talon → .py action conversion; would just get rid of the format string.

core/help/help_scope.py
```diff
@@ -52,8 +52,11 @@ def print_value(gui: imgui.GUI, path: str, value, ignore: set[str] = {}):
 def format_value(value):
     if isinstance(value, (list, set)):
         value = ", ".join(sorted(value))
-    if isinstance(value, str) and len(value) > setting_max_length.get() + 4:
-        return f"{value[:setting_max_length.get()]} ..."
+    if (
+        isinstance(value, str)
+        and len(value) > settings.get("user.help_scope_max_length") + 4
+    ):
+        return f"{value[:settings.get('user.help_scope_max_length')]} ..."
```
nriley, line 59: Would just retrieve the setting once.

core/snippets/snippets.py
```diff
@@ -37,10 +37,10 @@
 
 
 def get_setting_dir():
-    if not setting_dir.get():
+    if not settings.get("user.snippets_dir"):
         return None
 
-    dir = Path(setting_dir.get())
+    dir = Path(settings.get("user.snippets_dir"))
```
nriley, line 43: Just retrieve once.

nriley (changes requested):
Thanks for doing this!

brief:
Thanks for the review @nriley. Changes made.

The one plugin I haven't updated yet is the mode indicator, mostly because of [this part](https://github.com/talonhub/community/blob/2f7de0d36637f6b24e1be19dd793265785e280db/plugin/mode_indicator/mode_indicator.py#L49-L64).

Changing it throws skipped settings warnings, since I assume the plugin loads before the settings do. Removing it and the reference on line 182 has everything working as it should, but I'm hesitant to do that without first knowing why @AndreasArvidsson included it in the first place.

nriley:
> Changing it throws skipped settings warnings, since I assume the plugin loads before the settings do. Removing it and the reference on line 182 has everything working as it should, but I'm hesitant to do that without first knowing why @AndreasArvidsson included it in the first place.

Yeah, I wonder if that is "approved" usage; it certainly seems like a neat trick if you've got something persistent that depends on a bunch of settings.

AndreasArvidsson:
You just need to add actual strings instead
```
setting_paths = [
    "user.mode_indicator_show",
    "user.mode_indicator_size",
    ...
]
```

brief:
Makes sense. Thanks for the help. That should be everything.

PR #1352, opened 2024-01-09
more emacs commands

rntz:


apps/emacs/emacs.talon
```diff
@@ -204,8 +206,9 @@ search toggle case [fold | sensitive]: user.emacs_meta("c")
 search toggle regex: user.emacs_meta("r")
 
 highlight lines matching [regex]: user.emacs("highlight-lines-matching-regexp")
+highlight phrase: user.emacs("highlight-phrase")
 highlight regex: user.emacs("highlight-regexp")
-unhighlight regex: user.emacs("unhighlight-regexp")
+unhighlight (regex | phrase): user.emacs("unhighlight-regexp")
```
nriley, line 211: Just noticed we have regex here and regex**p** elsewhere. Do you pronounce the "p"? I find it gets eaten a bit if I try to say it out loud. Elsewhere in community we use "expression" (e.g. hunt expression: user.find_toggle_match_by_regex()). I'm OK with this particular change though, makes sense! Perhaps consider `(regex | expression)`.

nriley (changes requested):
Otherwise looks good to me!

apps/emacs/emacs.talon
rntz: I don't think I pronounce the p, but I don't think I've ever noticed a misrecognition either way. Would make sense to standardize on `regex` in the spoken forms, I suppose (emacs itself consistently uses `regexp`).

Re `expression`, I'd tend to prefer to stick close to the naming of the emacs commands generally. But if folks want to consistently add `expression` as an alternative everywhere, that'd be fine. Probably its own PR though.

nriley: I'm OK with just changing everything to `regex`; if other people want `expression` they can file another PR/ticket!

rntz: done

nriley (approved):

PR #1353, opened 2024-01-09
use double quotes for git commit message

rntz:
Double quotes allow you to use contractions (don't, isn't, etc) in the commit message without having to escape them. On the other hand they mean that some special characters need escaping. I find that double quotes are a better default for me. If other folks disagree, feel free to nix this PR.


nriley (approved):
Makes sense to me. Will wait for a bit more agreement before merging!

knausj85:
This looks good to me too. Thanks.

PR #1354, opened 2024-01-11
mouse grid: add method to test if mouse grid is active.

chrisosaurus:
I use this as I have commands which want to ensure a consistent moue grid state before and after they have run.
e.g. If the mouse grid was active before the command runs, they leave it active afterwards, if the mouse grid was closed then it is left closed.


knausj85 (approved):
Looks good, thanks.

PR #1356, opened 2024-01-11
mouse: adding synonyms for starting left drag.

chrisosaurus:
I would regularly say `drag start` which would cause an asterisk to be inserted.


knausj85:
This makes sense from a consistency perspective, thanks. I'll leave this open for others to chime in.

knausj85 (approved):

PR #1357, opened 2024-01-11
terminal command lisa accepts directory name

chrisosaurus:
This makes `lisa` more consistent with `katie`.

### DIFF
--- a/apps/mintty/mintty_win.py
+++ b/apps/mintty/mintty_win.py
@@ -113,8 +113,8 @@ class UserActions:
         """file_manager_open_volume"""
         actions.user.file_manager_open_directory(volume)
 
-    def terminal_list_directories():
-        actions.insert("ls")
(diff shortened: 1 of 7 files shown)


nriley (changes requested):
On the community backlog session we discussed this and felt it was a good idea for consistency, however we would prefer if you made the path argument optional to avoid breaking any existing argumentless uses of the actions.

nriley:
Changes made as I recommended.

nriley (approved):

pyproject.toml
```diff
@@ -12,3 +12,4 @@ pythonpath = [
 
 [tool.pyright]
 reportSelfClsParameterName = false
+reportGeneralTypeIssues = false
```
chdoc: @nriley I assume you added this unrelated change by accident, correct?
```suggestion
```

Issue #1296, opened 2023-10-28
Convert settings in `.talon` files to using literal `bool`s


phillco:
Talon 0.4 shipped on July 24, 2023 which supports boolean literals in Talon files. So you can write:

```talon
user.mode_indicator_show = false
```

...instead of

```talon
user.mode_indicator_show = 0
```

This is easier to parse and also visually distinguishes boolean settings from numerical ones.

~~We might want to hold off for some amount of time (not sure what the SLA is on community supporting older public Talon versions) if people are using <=0.3, so filing this as a tracking ticket.~~

Per @lunixbochs, we don't need to support old public Talon versions, so it's safe to use boolean literals in new functionality. This ticket will track the rewriting of existing literals.

lunixbochs:
I think if you're not going to update Talon you should also not update community or your plugins. Don't worry about supporting older Talon versions in your scripts.

phillco:
Cool, so confirming:

- it's safe to use booleans in new functionality (as well as other new public Talon functionality going forward; we don't have to try to support old public versions)
- we can keep this ticket open for tracking the rewriting of existing numerical setting literals to booleans (probably not a huge rush, this will create a lot of merge conflicts for people who have changed those settings)

PR #1360, opened 2024-01-16
Convert boolean settings to boolean

brief:
Turn boolean settings in `mouse.py` and `file_manager.py` into proper booleans. For now, I've held off on converting the corresponding  `.talon` settings to boolean literals per #1296, but happy to do that too if there are no concerns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,7 @@ https://github.com/talonhub/community/blob/main/settings.talon
 #adjust the scale of the imgui to my liking
 imgui.scale = 1.3
 # enable if you'd like the picker gui to automatically appear when explorer has focus
-user.file_manager_auto_show_pickers = 0
+user.file_manager_auto_show_pickers = false
 #set the max number of command lines per page in help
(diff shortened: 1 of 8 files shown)


knausj85:
This looks good, thanks. I think we should go ahead and convert to boolean literals while we're here

brief:
@knausj85 done.

The only setting I haven't converted is [in c.talon](https://github.com/talonhub/community/blob/72d3e90f938863c926c05e1b696e863cfcc4c9c4/lang/c/c.talon#L28) since I'm not sure it actually does anything: [points to a single reference in c.py](https://github.com/talonhub/community/blob/72d3e90f938863c926c05e1b696e863cfcc4c9c4/lang/c/c.py#L4-L9).

I'll convert if we want to keep it. Can submit a separate pull request to remove it otherwise.

knausj85 (approved):

knausj85:
Yeah, that looks to be a dead setting. I'd say remove it in a separate pull request if you don't mind

PR #1361, opened 2024-01-18
Clarify file location of websites.csv

BoltonBailey:
I have updated the comment here to make it clear that the `websites.csv` file is in the `community/settings` directory. Hopefully this will make it obvious to new users that `websites.csv` cannot be created just anywhere like `.talon` files can.

### DIFF
--- a/core/websites_and_search_engines/websites_and_search_engines.py
+++ b/core/websites_and_search_engines/websites_and_search_engines.py
@@ -13,7 +13,7 @@ mod.list(
 )
 
 # Please do not edit these defaults.  Instead, add / edit your own entries in
-# settings/websites.csv in your user directory.
+# websites.csv in your user/community/settings directory.
 website_defaults = {
     "talon home page": "http://talonvoice.com",
     "talon slack": "http://talonvoice.slack.com/messages/help",


core/websites_and_search_engines/websites_and_search_engines.py
```diff
@@ -13,7 +13,7 @@
 )
 
 # Please do not edit these defaults.  Instead, add / edit your own entries in
-# settings/websites.csv in your user directory.
+# websites.csv in your user/community/settings directory.
```
fidgetingbits, line 16: I think this is actually potentially more confusing in so far as `settings/` is relative to the community repo, whereas `user/community/` is "outside" it, which means the community repo doesn't really know about what that path will be. So for instance, a lot of people end up cloning the community repo with a different name, especially if they fork, and so in those cases `users/community/` will be wrong.

With that in mind I would suggest may be just adding an extra line that gives an example of a possible location for it. So some thing like:

```python
# Please do not edit these defaults. Instead, add / edit your own entries in
# settings/websites.csv in your user directory. For example, if you cloned the 
# talonhub community repo as "community", then the file would be in <talon folder>/user/community/settings/websites.csv.
```

fidgetingbits (changes requested):

nriley:
This is obsolete as we now have `website.talon-list`

PR #1362, opened 2024-01-19
fix #966 by using fixed folders

### LINKED ISSUE #966: If TMPDIR environment variable is set, vscode integration does not accept commands
A random piece of software on my machine (Windows 10) had set this the system environment variable `TMPDIR`, which caused the vscode integration to not accept commands. I believe this is because the function `get_communication_dir_path()` uses `gettempdir()`:
https://github.com/knausj85/knausj_talon/blob/main/apps/vscode/command_client/command_client.py#L211-L224

which [checks](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir) `TMPDIR`, `TEMP`, and `TMP`, in order, to choose a temporary directory.

On the extension side, node's [`os.tmpdir()`](https://nodejs.org/api/os.html#ostmpdir) is used to select a temporary directory:
https://github.com/pokey/command-server/blob/main/src/paths.ts#L4-L12

It's not documented, but I believe this must *not* check the TMPDIR env variable, causing the two ends of the connection to choose different directories and miss each other.

(Not sure how to fix this. You could override the directory lookup, or just switch to using a socket lol.)

### LINKED ISSUE #996: number_small doesn't include double-digit phrases, eg. "one two", only "twelve" is accepted
Dunno why I haven't run into this before, but `number_small` doesn't accept multiple-digit numbers (it didn't even before #993). So you can say `twenty seven` but not `two seven`. Should we fix this?

fidgetingbits:
This is a fix for #966. I've got three different systems where I run into this issue, so want to finally address it for everyone.

There is a corresponding PR for command-server at https://github.com/pokey/command-server/pull/21

I haven't actually tested this on windows, as I don't have it, so I'm not sure if the path I used is sufficient. On this side it is TALON_HOME, and on command-server side it is hardcoded which I think is the same but not positive. Also in the discussion there was a suggestion to make the folder hidden. I left a commented out cmd for setting the folder as hidden, but again not sure if it is right and can't test it, so feedback on if that's the right approach is welcome. Will be good if someone can test/verify the windows side (I added a few windows users as requested reviewers).

Its also worth noting that in #966 discussion, @auscompgeek suggested to use `XDG_RUNTIME_DIR` on Linux, which I had implemented and tested originally. However, when looking into snaps I ran across https://forum.snapcraft.io/t/rethinking-how-we-handle-xdg-runtime-dir/22223 which suggests that they may want to change snaps to no longer have access to the same `XDG_RUNTIME_DIR` folder. If they actually ever implemented that then this would break, so in the end I opted not to use it.

There's also per user temp directories on mac, which you can query using `getconf DARWIN_USER_TEMP_DIR`, but I decided not to keep that code in the end since it keeps the mac/linux case simpler both using `/tmp`, especially in light of not using `XDG_RUNTIME_DIR` in the end. 

ATM I cache the result in a global, and this was mostly because originally I had code that was calling `getconf DARWIN_USER_TEMP_DIR`, but I chose to just leave the global afterwards anyway. 

I've tested the changes using /tmp on both Linux and Mac and it works fine.

- [x] Test on Windows
- [x] Test on Mac
- [x] Test on Linux


lunixbochs:
please don't merge until I do security review

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +197,33 @@ def run_command(
...
 
+# NB: See https://github.com/talonhub/community/issues/966 for why we do OS-specific temp dirs
+def get_platform_specific_communication_dir_path():
     # NB: We don't suffix on Windows, because the temp dir is user-specific
     # anyways
+    suffix = ""
     if hasattr(os, "getuid"):
         suffix = f"-{os.getuid()}"
+    if app.platform == "linux" or app.platform == "mac":
+        return Path("/tmp") / f"{actions.user.command_server_directory()}{suffix}"
```
pokey, line 211: is this going to present a problem for snaps / linux flatpak? cc/ @auscompgeek

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +197,33 @@ def run_command(
...
+    elif app.platform == "windows":
+        # subprocess.run(["attrib","+H","myfile.txt"],check=True)
+        return Path(TALON_HOME / f"{actions.user.command_server_directory()}")
```
pokey, line 214: is `TALON_HOME` guaranteed to never change on Windows? I'd almost be tempted to hardcode it here if we're hardcoding it in https://github.com/pokey/command-server/pull/21 cc/ @lunixbochs

pokey (commented):
Looks fine but I haven't tried it locally. See comments inline

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +197,33 @@ def run_command(
...
-    """
-    suffix = ""
 
+# NB: See https://github.com/talonhub/community/issues/966 for why we do OS-specific temp dirs
+def get_platform_specific_communication_dir_path():
     # NB: We don't suffix on Windows, because the temp dir is user-specific
     # anyways
+    suffix = ""
     if hasattr(os, "getuid"):
         suffix = f"-{os.getuid()}"
```
pokey, line 209: if we're not using this on windows, shall we put it inside the if statement where we actually use it?

apps/vscode/command_client/command_client.py
fidgetingbits: It seems like it will probably be a problem: https://forum.snapcraft.io/t/accessing-tmp-from-snaps/22384/17 . It seems like the suggestion there and elsewhere if you want to have a shared folder is to use bind mounts into the home directory. If that's the case I'm not so sure there's going to be any folder that's reliable to use with snaps, without using some sort of trickery like the bind mount, which will have to be up to the user. That said I'm not super familiar with snaps or flatpaks, so may be missing something.

I would guess that right now vscode as a snap wouldn't actually work with command-server anyway? Since snaps get their own temporary directories, which will differ from what talon gets. So at the very least we're not breaking something that already works. But of course it would be nice to have it also work for snaps if possible.

pokey: wouldn't that argue for using home directory or talon directory in that case?

apps/vscode/command_client/command_client.py
fidgetingbits: I switched it to use hard coded to be consistent. But someone will still need to test it on windows

apps/vscode/command_client/command_client.py
fidgetingbits: Ya you are right. If someone wants to use it in a ram disk an know they're not going to be using a snap, they can create a symbolic link to some other location.

Updated it to use "~/.talon/.comms" for now and dropped the suffix as everything will be per-user now anwyay. I think it makes sense to just put it in talon folder if we're going to go this route. Happy to change the folder name to whatever though.

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +197,34 @@ def run_command(
...
 
+command_server_directory = None
+
+
+# NB: See https://github.com/talonhub/community/issues/966 for why we do OS-specific temp dirs
+def get_platform_specific_communication_dir_path():
+    home_dir = Path(os.path.expanduser("~"))
+    if app.platform == "linux" or app.platform == "mac":
+        return (
+            Path(home_dir) / f".talon/.comms/{actions.user.command_server_directory()}"
```
pokey, line 208: I thought accessing a hidden folder was also a problem for snaps, no? Cc/ @auscompgeek

pokey, line 208: How come we use `.comms` here but not for windows?

fidgetingbits: This is true for most snaps by default iirc, but afaik vscode uses something called classic confinement, which has unrestricted access to files. This makes sense since it's an IDE and users would expected to be able to access anything as their user anyway. So I think we should be okay here (though I haven't tested with the snap yet).

pokey:
Once everything is reviewed and ready to merge, we should do one final test on each platform, given the potential repercussions of this code breaking 😅

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +198,30 @@ def run_command(
...
+
+# NB: See https://github.com/talonhub/community/issues/966 for why we do OS-specific temp dirs
+def get_platform_specific_communication_dir_path():
+    home_dir = Path(os.path.expanduser("~"))
+    directory_name = actions.user.command_server_directory()
+    if app.platform == "linux" or app.platform == "mac":
+        return home_dir / ".talon" / " .comms" / directory_name
+    elif app.platform == "windows":
+        # subprocess.run(["attrib","+H","myfile.txt"],check=True)
+        return home_dir / "AppData" / "Roaming" / "talon" / ".comms" / directory_name
```
nriley, line 212: This also makes little sense on Windows where the dot prefix doesn't "do" anything.

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +198,30 @@ def run_command(
...
+
+
+# NB: See https://github.com/talonhub/community/issues/966 for why we do OS-specific temp dirs
+def get_platform_specific_communication_dir_path():
+    home_dir = Path(os.path.expanduser("~"))
+    directory_name = actions.user.command_server_directory()
+    if app.platform == "linux" or app.platform == "mac":
+        return home_dir / ".talon" / " .comms" / directory_name
+    elif app.platform == "windows":
+        # subprocess.run(["attrib","+H","myfile.txt"],check=True)
```
nriley, line 211: This seems like it may have been some testing code?

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +198,30 @@ def run_command(
     return decoded_contents["returnValue"]
 
 
+command_server_directory = None
+
+
+# NB: See https://github.com/talonhub/community/issues/966 for why we do OS-specific temp dirs
+def get_platform_specific_communication_dir_path():
+    home_dir = Path(os.path.expanduser("~"))
```
nriley, line 206: `Path.home()` is a shorter way to spell this.

nriley (changes requested):

apps/vscode/command_client/command_client.py
fidgetingbits: I left this in for someone with access to Windows in case they want to test using a hidden folder, but didn't uncomment it because I can't test if it works.

apps/vscode/command_client/command_client.py
fidgetingbits: Originally I didn't have it prefixed I think, but it was switched for consistency of paths across all platforms.

apps/vscode/command_client/command_client.py
```diff
@@ -197,20 +198,30 @@ def run_command(
...
+    directory_name = actions.user.command_server_directory()
+    if app.platform == "linux" or app.platform == "mac":
+        return home_dir / ".talon" / " .comms" / directory_name
```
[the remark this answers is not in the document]
fidgetingbits: I'm easy and don't really care either way so happy if others want to change this, but I suppose I'll note I actually _slightly_ prefer it being double hidden. If I enter .talon/ folder to do whatever I do in there, it's definitely got nothing to do with looking at the rpc stuff, which is all just ephemeral data and not meant to be manually messed with; I'd rather just not even see it. I think it's partially why using tmp folders would've been nicer if we actually could...

fidgetingbits:
Aside from the Path.home() bit which will be quick to fix, this is still mostly waiting on someone who uses Windows to test that it actually works as intended. Anyone have time to try?

DonnerWolfBach:
@fidgetingbits I will hopefully have time (about 2-6h) to try it on windows on Saturday and I can find more time after that. 
I assume I would just have to apply this PR to my personal talon community fork (and remove the dirty fixes I Currently Have in Place).
However, how would I go about applying https://github.com/pokey/command-server/pull/21 ?

fidgetingbits:
> @fidgetingbits I will hopefully have time (about 2-6h) to try it on windows on Saturday and I can find more time after that. I assume I would just have to apply this PR to my personal talon community fork (and remove the dirty fixes I Currently Have in Place). However, how would I go about applying [pokey/command-server#21](https://github.com/pokey/command-server/pull/21) ?

Hey @DonnerWolfBach thanks! Reminds me @saidelike recently said they could also test this on Windows.

Easiest is to merge the PR into some temporary branch you want to build from:

```bash
git clone https://github.com/pokey/command-server
cd command-server
git switch -c dev-build
git fetch origin pull/21/head:static-tmp-dir
git merge origin/static-tmp-dir -m "Merge branch 'use-static-tmp-dir' into dev"
```

Then you need to build and install. I do it on nix on both macOS and linux, so merge yet another PR so I can just run nix build. I do it like this:

```bash
git fetch origin pull/22/head:nix-flake
git merge origin/nix-flake -m "Merge branch 'nix-flake' into dev"
nix build
code --force --install-extension result/*.vsix
```

If you have a nix box and can do the above, I think the .vsix file will actually work on windows too. 

But, if you have to build the extension on windows, presumably you just need to
run the same yarn commands as what the flake PR does. But note that I also had to update `package.json` to add vsce to yarn dependencies so it didn't have to be installed separately (not sure off the top of my head where it comes from otherwise). So if you don't want to find where vsce comes from, you may want to do that too, and then you just need to install yarn on windows (however one does that).

Then the manual build will be something like this:

```bash
yarn run compile
yarn vsce package --yarn -o ./command-server.vsix
```

@pokey might be worth adding a small note to the command-server README about how to "officially" build it. I'll file an issue.

apps/vscode/command_client/command_client.py
auscompgeek, line 209: Don't snaps prevent access to `~/.*`?

fidgetingbits: In the default containment profiles yes (I believe you need to request personal confinement to access dotfiles), but from what I read vscode should be okay due to it using classic confinement (which is effectively no confinement). See here from this same PR where I said similar: https://github.com/talonhub/community/pull/1362#discussion_r1460811407

@DonnerWolfBach said he was using VSCodium. I suspect since it's just an open source build of VSCode (?) and an IDE when packaged as a snap it would use the same classic confinement. Maybe if he's using the snap, he could confirm though.

From what I've seen they tend to grant classic confinement to all IDEs under the assumption they will be used to edit any files. For instance neovim snap also has [classic confinement](https://forum.snapcraft.io/t/classic-confinement-request-for-nvim/14405/17).

I think the only other option if we want to assume we need to communicate with a snaps with restricted confinement is /tmp, and the discussion on https://github.com/talonhub/community/issues/966 people couldn't agree, and it seemed some (pokey and rntz at least iirc) were leaning towards using $HOME over /tmp, so I just used $HOME for the sake of getting anything going.

DonnerWolfBach:
> ```shell
> git merge origin/static-tmp-dir -m "Merge branch 'use-static-tmp-dir' into dev"
> ```

The origin in there is a typo isn't it? With it it doesn't work, but without it does, at least on my machine that branch was created locally from your PR

fidgetingbits:
> > ```shell
> > git merge origin/static-tmp-dir -m "Merge branch 'use-static-tmp-dir' into dev"
> > ```
> 
> The origin in there is a typo isn't it? With it it doesn't work, but without it does, at least on my machine that branch was created locally from your PR
> 
> edit: @fidgetingbits (sorry, I am unfamiliar with how github commenting works, not sure whether this is the right way to point out such a small thing)^^'
> 

I modified the commands from a script that uses my local branches rather than the PRs, so likely why. I'll edit the comment in case anyone else tries to use it.

DonnerWolfBach:
> ```shell
> yarn vsce package --yarn -o ./command-server.vsix
> ```

Here I am running into troubles once again:


```
PS C:\Users\Sebastian\Repos\command-server> corepack yarn vsce package --yarn -o ./command-server.vsix
Executing prepublish script 'yarn run vscode:prepublish'...
 ERROR  Command failed: yarn list --prod --json

PS C:\Users\Sebastian\Repos\command-server> corepack yarn list --prod --json

Usage Error: Couldn't find a script named "list".

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...
PS C:\Users\Sebastian\Repos\command-server>
```

Due to lacking knowledge of yarn etc. I don't think I am able to troubleshoot that myself farther than this.

Version info:
```
PS C:\Users\Sebastian\Repos\command-server> corepack yarn vsce --version
2.15.0
PS C:\Users\Sebastian\Repos\command-server> corepack yarn --version
4.2.2
PS C:\Users\Sebastian\Repos\command-server> node --version
v20.13.1
```
(cartago delenda est - I mean I think https://github.com/pokey/command-server/issues/26 really makes sense)

fidgetingbits:
Any chance the instructions saidelike posted on that other issue would help building on windows? He implies 'pnpm compile' is enough. I don't even remember why I used yarn anymore..

saidelike:
> Any chance the instructions saidelike posted on that other issue would help building on windows? He implies 'pnpm compile' is enough. I don't even remember why I used yarn anymore..

Fwiw i used "pnpm" for neovim's command-server because cursorless used "pnpm" and i didn't realise command-server was using "yarn" by default (as indicated in the package.json). Maybe @pokey can indicate how he builds the vscode's command-server? 

Btw, I am happy to test these command-client's changes along with corresponding changes in my neovim command-server extension.

apps/vscode/command_client/command_client.py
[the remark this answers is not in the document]
saidelike: Ya i agree double dot directory is fine. If other agree, we can mark it as resolved

apps/vscode/command_client/command_client.py
saidelike: I don't think it matters much. But yeah the talon user path has no prefixed dot either so there is no consistency at that level anyway. So in this case i think actually using "comms" without dot on Windows is probably better? Happy either way.

apps/vscode/command_client/command_client.py
saidelike: Can anyone with permissions fix this before i test on neovim/Windows?

fidgetingbits: done

apps/vscode/command_client/command_client.py
DonnerWolfBach: I don't use VSCodium in a snap unfortunately. I use the nix package

apps/vscode/command_client/command_client.py
DonnerWolfBach: @nriley  Doesn't the dot prefix still hide a directory on windows in certain views?
Personally I would slightly prefer .comms everywhere for consistency if you compare the talon settings folder  on different systems -  even though of course the name and path of the talon settings folder will be widely different.

DonnerWolfBach:
@fidgetingbits managed to test it on windows 10 with the newest version of both branches. `2024-06-15 10:28:07.578    IO communication_dir_path=WindowsPath('C:/Users/Sebastian/AppData/Roaming/talon/.comms/vscode-command-server')` and indeed cursorless works

@saidelike thanks for the hint, changing yarn to pnpm in the package.json solved the problem

apps/vscode/command_client/command_client.py
auscompgeek: > Doesn't the dot prefix still hide a directory on windows in certain views?

Nope, the hidden bit is separate metadata to the filename on Windows.

knausj85:
I'm going through some of the older PRs. This looks like it's ready to be merged to me? Something I'm missing?

apps/vscode/command_client/command_client.py
knausj85: I would personally vote for the consistency on this bit fwiw, as it is harmless.

fidgetingbits:
> I'm going through some of the older PRs. This looks like it's ready to be merged to me? Something I'm missing?

I think so, with the caveat I think the @pokey command server vscode extension will have to merge the [PR](https://github.com/pokey/command-server/pull/21) and release an updated version first. Otherwise it will break RPC for users.

There was the snap concern about home access, which I've noted I don't think is valid for editors. 

There is also the commented code for setting a hidden file attribute for windows, which I don't know if windows people want and couldn't test, but happy to just have that deleted before merge.

fidgetingbits:
Also @lunixbochs had commented they wanted to do a security review before this gets merged, so will need to take a look.

apps/vscode/command_client/command_client.py
fidgetingbits: Will resolve this since knausj85 agrees, but feel free to open if you want to still debate.

apps/vscode/command_client/command_client.py
fidgetingbits: @knausj85 any thoughts about that part? I'm happy to just delete the commented out bit if windows users don't care to add it

knausj85: I have no concerns with leaving this file unhidden. Thoughts on this @AndreasArvidsson ?

lunixbochs: fyi I'm not sure I want plugins making top-level directories under .talon, hidden or not

AndreasArvidsson: @knausj85 I'm personally not a big fan of hidden files in windows so I'm fine with it being visible. With that being said I'm not sure placing it in the Talon directory is the right choice.

fidgetingbits: Can those of you who have opinions try to decide on a path that is acceptable and I'll update the PRs with whatever it is? I don't care where it is, as long as those who do care agree on something.

AndreasArvidsson: Sounds good. Should we continue that discussion in the issue or here?

Personally I would prefer to keep it in the temp directory and agree with @rntz arguments in the issue that the extension should respect `TMPDIR`.

If that doesn't work for some technical reason we should probably just create a new directory in user home. eg `~/.rpc`.

fidgetingbits: I believe TMPDIR doesn't work as a solution because it can be different path in talon vs vscode. This can definitely be the case on nix, as noted here https://github.com/talonhub/community/issues/966#issuecomment-1236134227 and also unrelated to that plugin which I've run into (which I already forget where as I've been using this PR for ~ 1 year). I originally hardcoded /tmp in the PR, but then eventually we opted for a /home-based path, which unfortunately I now can't find the reason/comment for it seems.

But I think because of all the potential snap/flatpak/nix related issues mentioned it's likely just best to avoid hoping one of those tmpdir/runtime path solutions work everywhere? Pretty sure atm we don't have the capacity to actually test it.

Something like ~/.rpc is fine with me if that's what people like (I just want it fixed at this point :D), but imo it does seem a bit opaque and potentially collision prone vs some name that associates it with talon, which was why I chose to drop the folder in ~/.talon/... ~/.talon-rpc or similar maybe makes more sense?

AndreasArvidsson: If someone on purpose has set a different environment variable in Talon and vscode they kind off asking for trouble. As long as we check the sources in the same order we should get the same result. 

Year that is a real problem because we don't want to break it for existing users.

`~/.talon-rpc` would be fine with me, but I'm not sure the home directory is any more stable than the temp directory since the home directory can also be moved with environment variables.

fidgetingbits: FWIW I agree if it were actually the user just wantonly changing environment variables, it shouldn't something that needs catering to. However, this isn't a case of a user choosing to change TMPDIR, it's being done somewhere behind the scenes, so it's not something they can just choose not to do. 

In addition to the original bug filer, there were numerous (https://github.com/talonhub/community/issues/966#issuecomment-1536142000) cases of it differing in practice, as well as in my systems and for some other Nix users. When it started breaking for me (which prompted finally creating the PR) it's certainly not from me explicitly changing TMPDIR myself. This comment https://github.com/talonhub/community/issues/966#issuecomment-1293379354 also implies this type of thing happens on windows.

In light of that, I don't see how keeping the order checking consistent is enough, unless you mean you have some other mechanism to ensure the paths they are checking in the same order are actually matching? Maybe I'm missing something.

nriley:
On the community backlog session we decided that keeping the current behavior of using the OS native API to retrieve the temporary directory and permitting override via environment variable is likely to cause the fewest issues. If there are any inconsistencies between the behavior of python and node then these should be filed as separate issues and fixed.

PR #1364, opened 2024-01-21
added pop twice to wake functionality

gpunktschmitz:
re-implemented this function in this new PR as I messed up my repo (sorry for the noise)
related PRs: #1363 and #1273


plugin/mouse/mouse.py
```diff
@@ -278,29 +294,33 @@ def show_cursor_helper(show):
 @ctx.action_class("user")
 class UserActions:
     def noise_trigger_pop():
-        if settings.get("user.mouse_enable_pop_stops_scroll") >= 1 and (
-            gaze_job or scroll_job
-        ):
-            # Allow pop to stop scroll
-            stop_scroll()
+        if actions.speech.enabled():
```
pokey, line 297: Could we not use a context for this instead? Ie have the noise_trigger_pop action defined to this in wake mode and the double pop checker in sleep

plugin/mouse/mouse.py
```diff
@@ -53,6 +55,12 @@
     default=0,
     desc="Pop noise clicks left mouse button. 0 = off, 1 = on with eyetracker but not with zoom mouse mode, 2 = on but not with zoom mouse mode",
 )
+mod.setting(
+    "mouse_enable_pop_wake",
+    type=int,
```
pokey, line 60: Let's use a boolean as those are now supported

plugin/mouse/mouse.py
```diff
@@ -239,6 +247,14 @@ def hiss_scroll_down():
         global hiss_scroll_up
         hiss_scroll_up = False
 
+    def pop_twice_to_wake():
+        """Use pop sound to wake from sleep"""
+        global time_last_pop
+        delta = time.time() - time_last_pop
+        if delta >= 0.1 and delta <= 0.3:
```
pokey, line 254: Is the minimum here to prevent false positives? Guessing you just grabbed this from @AndreasArvidsson's impl?

pokey (commented):
Good stuff; left a couple comments

plugin/mouse/mouse.py
AndreasArvidsson: Correct this is from my implementation.

pokey: Cool. What's that 0.1 min for?

AndreasArvidsson: As you said: to prevent false positives

gpunktschmitz: yes - copy from Andreas - so nothing to do here I guess - will mark as resolved

plugin/mouse/mouse.py
gpunktschmitz: changed type to bool

plugin/mouse/mouse.py
gpunktschmitz: I have no clue what to do here - I did not wrap my head around context.

Spiteless: @pokey 

Is that syntax available in non beta talon? This is how it's defined in your repo:

```py
mode: sleep
-
parrot(postalveolar_click): user.wake_all()
```

I'm not sure if that syntax is available for vanilla talon via `noise(pop)`

pokey: I don't believe it is available, but you might ask in the talon slack workspace. Talon scripting channel should do

Spiteless: Yes, it looks like using noise() is unavailable in .talon files, so as written this implementation seems like the best way to go about this currently.

pokey: > Could we not use a context for this instead? Ie have the noise_trigger_pop action defined to this in wake mode and the double pop checker in sleep

I still think this would be better than the big if-else statement

Spiteless: Yeah I was thinking about this as well, what would that format be instead of how it's structured now?

```py
# double_pop_trigger_wake.py

ctx.matches = r"""
mode: sleep
"""

@ctx.action_class("user")
class UserActions:
    def noise_trigger_pop():
        # code etc
```

AndreasArvidsson: I don't see a point in having `pop_twice_to_wake` in a module class. Just have the function itself in the file. I would do something like this?

```py
from talon import Context, settings
import time

ctx = Context()

ctx.matches = r"""
mode: sleep
"""

time_last_pop = 0

def pop_twice_to_wake():
        """Use pop sound to wake from sleep"""
        global time_last_pop
        delta = time.time() - time_last_pop
        if delta >= 0.1 and delta <= 0.3:
            actions.speech.enable()
        time_last_pop = time.time()


@ctx.action_class("user")
class UserActions:
    def noise_trigger_pop():
        if settings.get("user.mouse_enable_pop_wake") == 1:
                pop_twice_to_wake()
```

Spiteless: Nice, this works without having to nest the function within `mouse.py`

Spiteless: Nice, this does feel clearer than having wake functionality nested inside `mouse.py`.

gpunktschmitz: thanks @AndreasArvidsson - I've updated the PR accordingly

core/modes/sleep_mode_pop_twice_to_wake.py
```diff
@@ -0,0 +1,22 @@
...
+
+time_last_pop = 0
+
+
+@ctx.action_class("user")
+class UserActions:
+    def noise_trigger_pop():
+        global time_last_pop
+        if settings.get("user.mouse_enable_pop_wake") == 1:
+            delta = time.time() - time_last_pop
```
alan-barzilay, line 19: wouldnt this raise an error the first time around since `time_last_pop` hasnt been defined with a value yet? there is only its global declaration without any associated value. Not sure if I am missing something...

alan-barzilay: nevermind, I am stupid and hadnt seen line 11

knausj85:
I believe this functionality is included in https://github.com/talonhub/community/pull/1398, which appears ready to be merged. Closing.

PR #1368, opened 2024-01-21
Remove unused setting from c code language

brief:
Removes a [dead setting](https://github.com/talonhub/community/pull/1360#issuecomment-1902757989) from the c code language.

### DIFF
--- a/lang/c/c.py
+++ b/lang/c/c.py
@@ -1,12 +1,6 @@
 from talon import Context, Module, actions, settings
 
 mod = Module()
-mod.setting(
-    "use_stdint_datatypes ",
-    type=int,
-    default=1,
-    desc="Use the stdint datatype naming in commands by default",
-)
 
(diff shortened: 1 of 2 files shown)


knausj85 (approved):

fidgetingbits:
I'm ok with removing this, but I guess I should note that I actually have the functionality related to that setting in my branch, so if there is interest I could send a PR to include the missing bits instead.

https://github.com/fidgetingbits/fidgetingbits-talon/blob/master/lang/c/c.py

nriley:
@fidgetingbits Would seem useful to put the functionality in. Would you rather we merge this in the meantime?

fidgetingbits:
> @fidgetingbits Would seem useful to put the functionality in. Would you rather we merge this in the meantime?

ya that makes sense, won't get to it right away

PR #1371, opened 2024-01-22
Update settings descriptions

brief:
Suggested update to make settings descriptions clearer and more consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,42 +266,7 @@ into each editor.
 
 ## Settings
 
-Several options are configurable via a single settings file out of the box. Any setting can be made context specific as needed (e.g., per-OS, per-app, etc).
-
-https://github.com/talonhub/community/blob/main/settings.talon
-
-```
-# Adjust the scale of the imgui
-imgui.scale = 1.3
-# Automatically show the picker GUI when the file manager has focus
-user.file_manager_auto_show_pickers = false
-# Set the max number of command lines to display per page of help
-user.help_max_command_lines_per_page = 50
-# Set the max number of contexts to display per page of help
-user.help_max_contexts_per_page = 20
-# Set the default scroll amount used when scrolling continuously
-user.mouse_continuous_scroll_amount = 80
-# Stop continuous scroll/gaze scroll with a pop
-user.mouse_enable_pop_stops_scroll = true
-# Choose how pop click should work in 'control mouse' mode
-user.mouse_enable_pop_click = 1
-# Use a hissing noise to scroll continuously
-user.mouse_enable_hiss_scroll = false
-# Hide the continuous scroll/gaze scroll GUI
-user.mouse_hide_mouse_gui = false
-# Hide the cursor when enabling zoom mouse
-user.mouse_wake_hides_cursor = false
-# Set the amount to scroll up/down
-user.mouse_wheel_down_amount = 120
-# Set the amount to scroll left/right
-user.mouse_wheel_horizontal_amount = 40
-# Start mouse grid numbering on the bottom left (vs. top left)
-user.grids_put_one_bottom_left = true
-# Set the number of command history lines to display by default
-user.command_history_display = 10
-# Set the total number of command history lines to keep
-user.command_history_size = 50
-```
+Several options are configurable via a [single settings file](https://github.com/talonhub/community/blob/main/settings.talon) out of the box. Any setting can be made context specific as needed (e.g., per-OS, per-app, etc).
 
 The most commonly adjusted settings are probably
 
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ into each editor.
 
 ## Settings
 
-Several options are configurable via a [single settings file](https://github.com/talonhub/community/blob/main/settings.talon) out of the box. Any setting can be made context specific as needed (e.g., per-OS, per-app, etc).
+Several options are configurable via a [single settings file](settings.talon) out of the box. Any setting can be made context specific as needed (e.g., per-OS, per-app, etc).
 
 The most commonly adjusted settings are probably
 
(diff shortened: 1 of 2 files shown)


knausj85 (approved):

JWaldenback:
I just want to pop in and say that I'm digging your "tidying up Talon" pull requests. It's long overdue and I'm happy someone's taking the time!

README.md
pokey, line 1: I would argue for removing all settings here that already appear in `settings.talon`. And for those that don't, I'd consider adding them to `settings.talon`, possibly commented out

settings.talon
```diff
@@ -1,73 +1,73 @@
 settings():
-    # Adjust the scale of the imgui to my liking
+    # Adjust the scale of the imgui
     imgui.scale = 1.3
 
-    # Enable if you'd like the picker gui to automatically appear when explorer has focus
+    # Automatically show the picker GUI when the file manager has focus
     user.file_manager_auto_show_pickers = false
```
pokey, line 6: If the value is `false`, shouldn't this comment read "Don't automatically..."?

settings.talon
```diff
@@ -1,73 +1,73 @@
...
 
-    # Enable pop click with 'control mouse' mode.
+    # Choose how pop click should work in 'control mouse' mode
     # 0 = off
     # 1 = on with eyetracker but not zoom mouse mode
     # 2 = on but not with zoom mouse mode
     user.mouse_enable_pop_click = 1
 
-    # Enable if you like to use the hissing noise to do mouse scroll
+    # Use a hissing noise to scroll continuously
```
pokey, line 26: As above, prob should read "Disable hissing noise..."

settings.talon
```diff
@@ -1,73 +1,73 @@
...
 
-    # The number of lines of command history to keep in total;
-    # "command history more" to display all of them, "command history less" to restore
+    # Set the total number of command history lines to keep; say "command history more"
+    # to display all of them and "command history less" to display the default
     user.command_history_size = 50
 
-    # Uncomment the below line to add a directory (relative to the Talon user dir) with additional .snippet files
-    # Changing this setting requires a restart of Talon
+    # Uncommment to add a directory (relative to the Talon user dir) with additional
```
pokey, line 51:
```suggestion
    # Uncomment to add a directory (relative to the Talon user dir) with additional
```

pokey (commented):
nice. left a few minor comments

settings.talon
brief: Yes, if as before, we want comments to describe the default value. Since it is much clearer when something is on/off with true/false, I've changed them to describe the functionality: in this case, `true` automatically shows the picker whereas `false` does not. 

I find "Don't automatically..." more confusing: does `false` not show the GUI? Or does it not...not show it.

If you agree with the approach, I can make each comment more explicit: "Automatically show the picker GUI when the file manager has focus. Default is `false`."

README.md
brief: All these settings are included in `settings.talon`.

Shall we strip out the [entire code block](https://github.com/talonhub/community/blob/c6e1969ab516d509f51b9c8acd01bd35d2262d28/README.md?plain=1#L273-L295)? Makes it simpler on first read and eliminates duplication. I'm not sure it means much to someone who's just getting started.

pokey: yes I would, and just link to `settings.talon`

settings.talon
pokey: maybe `# If `true`, automatically ...`?

brief:
Changes made.

Something else I noticed: `user.context_sensitive_dictation` and `user.window_snap_screen` are commented out despite being core capabilities with [default](https://github.com/talonhub/community/blob/c6e1969ab516d509f51b9c8acd01bd35d2262d28/core/text/text_and_dictation.py#L9-L14) [settings](https://github.com/talonhub/community/blob/c6e1969ab516d509f51b9c8acd01bd35d2262d28/core/windows_and_tabs/window_snap.py#L20-L30). Does it make sense to uncomment them and set the correct defaults?

This differs from `user.paste_to_insert_threshold` and `user.snippets_dir`, which both enable functionality if uncommented.

JWaldenback:
To me, it makes perfect sense. Will be easier for new users to understand if it's consistent

README.md
```diff
@@ -266,32 +266,7 @@ into each editor.
...
-#enable pop click with 'control mouse' mode
-user.mouse_enable_pop_click = 1
-#When enabled, the 'Scroll Mouse' GUI will not be shown.
-user.mouse_hide_mouse_gui = false
-#hide cursor when mouse_wake is called to enable zoom mouse
-user.mouse_wake_hides_cursor = false
-#the amount to scroll up/down (equivalent to mouse wheel on Windows by default)
-user.mouse_wheel_down_amount = 120
-```
+Several options are configurable via a [single settings file](https://github.com/talonhub/community/blob/main/settings.talon) out of the box. Any setting can be made context specific as needed (e.g., per-OS, per-app, etc).
```
pokey, line 269: I think you can just use a relative path
```suggestion
Several options are configurable via a [single settings file](settings.talon) out of the box. Any setting can be made context specific as needed (e.g., per-OS, per-app, etc).
```

pokey:
> Something else I noticed: `user.context_sensitive_dictation` and `user.window_snap_screen` are commented out despite being core capabilities with [default](https://github.com/talonhub/community/blob/c6e1969ab516d509f51b9c8acd01bd35d2262d28/core/text/text_and_dictation.py#L9-L14) [settings](https://github.com/talonhub/community/blob/c6e1969ab516d509f51b9c8acd01bd35d2262d28/core/windows_and_tabs/window_snap.py#L20-L30). Does it make sense to uncomment them and set the correct defaults?
> 
> This differs from `user.paste_to_insert_threshold` and `user.snippets_dir`, which both enable functionality if uncommented.

I guess the consistent thing to do would be:

- Leave the line uncommented if it differs from default
- If it is equal to default, change it to not be equal to default, and comment it out

But this PR is a step in the right direction so I'd be happy to accept it as is

fidgetingbits (approved):
lgtm

Issue #546, opened 2021-08-03
Systematize handling of delimiters


rntz:
We have a lot of delimiter-related commands, like `args` and `empty [escape] (string|dubstring)` and `inside (bracket|quotes|percent)` and `(percent|pipe|square) that`, mostly living in `symbols.talon`, generally of the forms

- `[empty] [escaped] <delimiter-name>`
- `inside <delimiter-or-punctuation-name>`
- `<delimiter-or-punctuation-name> that`

But it's not super consistent which ones exist and which don't and what they're all named. It would be good to systematize them using some lists/captures. I have a system locally for this and have been using it for a while and like it, but its conventions don't quite line up with the existing knausj commands (eg. it collides with some knausj formatter names). Not sure what the "right" systematization of the existing commands even ought to be. Thoughts?

nriley:
I don't mind the empty/inside/that behavior, but find the names of the individual delimiters by default in knausj…far from what I would naturally say.  (You can see what I've done in my fork: https://github.com/nriley/knausj_talon/blob/43eae5d5/text/symbols.talon)

We may want to allow customization of delimiters (e.g. "bracket", "brace", "quote") like we do for modifiers.

rntz:
I've come up with a systematization of this that I'm happy with, here it is: https://gist.github.com/rntz/e548ff20a28f9e9bd7c6bad8466670d4

I think we could get it pretty close to the current knausj behavior.

pokey:
That's really nice @rntz!  The only things I'd consider changing:

- "angle" => "diamond", for consistency with other visual pair representations, and to avoid conflict with "angle" for left delimiter itself
- why do you don't you use the capture for the last talon command instead of the list?
- I'd consider calling the list "left_delimiter"

pokey:
We just discussed this issue in a recent Cursorless meet-up, in the context of https://github.com/cursorless-dev/cursorless/pull/2193

Here's where we landed:

## End goal

- Create a list like the one proposed by @rntz in https://gist.github.com/rntz/e548ff20a28f9e9bd7c6bad8466670d4, but use a `.talon-list` file (I don't think those existed when @rntz originally wrote that gist). I think we'd lean towards `user.paired_delimiter` for the name of the list, but that detail's not too important
- Include all the delimiters that @rntz has, though we'd lean towards adding one for a pair of spaces, to replace `"padding"`
[...]

PR #1375, opened 2024-01-29
This is a response to the cursorless PR to add these symbols

jaresty:
- In [this PR](https://github.com/cursorless-dev/cursorless/pull/2193) I tried to add
  cursorless symbols to cursorless but the feedback was that this belongs in community
  instead.
- It does not make sense to refer to cursorless csv files from community, so this opts
  to hard code the values.

### DIFF
--- a/core/edit/cursorless_symbols.talon
+++ b/core/edit/cursorless_symbols.talon
@@ -1,0 +1,50 @@
+empty round: "()"
+empty square: "[]"
+empty curly: "{}"
+empty diamond: "<>"
+empty quad: '""'
+empty twin: "''"
+empty escaped quad: '\\"\\"'
+empty escaped twin: "\\'\\'"
+empty escaped round: "\\(\\)"
+empty escaped curly: "\\{{\\}}"
+tween <user.symbol_key>: user.insert_between("{symbol_key}", "{symbol_key}")
+quad: user.insert_between('"', '"')
+twin: user.insert_between("'", "'")
+ski: user.insert_between("`", "`")
+escaped quad: user.insert_between('\\"', '\\"')
+escaped twin: user.insert_between("\\'", "\\'")
+round: user.insert_between("(", ")")
+escaped round: user.insert_between("\\(", "\\)")
+escaped curly: user.insert_between("\\{{", "\\}}")
+square: user.insert_between("[", "]")
+curly: user.insert_between("{", "}")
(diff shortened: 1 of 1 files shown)


pokey (changes requested):
I don't think it makes sense to hard-code these if we're going to use a talon-list later, and given that they won't match the user's custom settings. I think you mentioned you only intend this as a conversation starter, but I just wanted to make sure someone doesn't merge it without being aware of the bigger conversation

jaresty:
I would prefer to set this up with a talon list but that pattern hasn't been estabilshed yet - I think we're waiting on this PR to set up a pattern there: https://github.com/talonhub/community/pull/1343

jaresty:
FWIW - I don't think this is significantly different from how the current keys work, so likely could be changed to use the new pattern once it has been established.

nriley:
I am worried about the number of short words that this introduces. We already have so many different names for these in community that I think it makes far more sense to wait until we have support for a .talon-list and can reduce the "pollution" created by having multiple names, few of which any given user will likely use.

jaresty:
@nriley  I don't understand what impact talon list support would have. Could you explain a bit more?

nriley:
If a user wanted to *just* use the cursorless (or current community) names for symbols they wouldn't have both in their grammar because they'd just need to customize them in one place. As is we have multiple names for many of the symbols and this PR adds even more.

jaresty:
I see. One distinction here is that this PR introduces a grammar for paired symbols, which doesn't currently exist in the community symbols today. If we are able to get support for configuring symbol names without changing community (see the related PR above, which starts that effort) then that goes a long way to solving my pain point here but still doesn't address the idea of paired symbols. I think pairs symbols are worthwhile to provide a name distinct from individual symbols.

jaresty:
Interesting. That's a good point. I wonder if a better approach would be to make those names configurable...

jaresty:
One issue I see with the current grammar is that it doesn't use the symbol names. I.e. `dub string` doesn't use the symbol name `quote`.

jaresty:
Here is a more targeted pull request that only adds the cursorless names to the existing symbol file: https://github.com/talonhub/community/pull/1381

pokey:
I would argue for closing this PR in favour of the approach proposed in https://github.com/talonhub/community/issues/546#issuecomment-1914536331

PR #1378, opened 2024-02-01
Switch to black pre-commit mirror

pokey:
See https://github.com/psf/black/issues/3405


nriley:
Speed is always useful. Works fine for me.

nriley (approved):

PR #1379, opened 2024-02-01
Update app_name_overrides.windows.csv to use Windows Terminal instead of iTerm2

tankorsmash:
It doesn't appear iTerm is available on Windows, so this should be safe to change.


nriley (approved):
Probably just copied from the Mac version and not changed.

PR #1380, opened 2024-02-03
Add detail to documentation

chrisnicollo:
Currently, the readme file describes several core features of these scripts, but a lot of the functionality is left out. This could make it difficult to get a good sense of what these scripts are capable of on first glance. Also, the current directory structure does not always make it clear where to look to discover and update certain types of commands.

Here I am providing an overview of the subdirectories within `core` and `plugin` for someone to skim through and determine which folders are most likely to have the commands they're looking for. I also added a bit more detail about mouse commands and tags.


nriley:
Thanks, this seems helpful.

What do you think about putting the documentation for the individual subdirectories into those subdirectories and linking to it? The top level README is already getting quite large.

chrisnicollo:
I agree with your suggestion and moved most of my content to subfolders. One thing I'd like to address though is that I don't understand the point of using the command within [screens](https://github.com/talonhub/community/blob/main/core/screens/screens.talon) to show the index of your different computer screens. If I understood the purpose of that, I could give that context within the documentation too.

nriley:
It lets you see how Talon is numbering your screens — which should match your OS, but…, so will help if you use voice commands to move windows between screens (e.g. "snap screen 3"). Thanks!

knausj85 (approved):
This looks great, thanks.

PR #1381, opened 2024-02-03
This adds cursorless names to the existing list of paired delimiters

jaresty:
Adds cursorless aliases to existing symbols list.


jaresty:
This is a response to feedback on this other pull request: https://github.com/talonhub/community/pull/1375,

jaresty:
Were there any thoughts on this approach? It seems to me that not having these available in community might have been unintentional?

pokey:
I would argue for closing this PR in favour of the approach proposed in https://github.com/talonhub/community/issues/546#issuecomment-1914536331

PR #1268, opened 2023-09-30
porting the .csv files to .talon-list files

saidelike:
The goal of this PR is to:
- remove code to generate the .csv files
- include the default .talon-list files
- port code in .python files to use the new .talon-list instead of the .csv

This initial commit is about porting {user.letter}: alphabet.csv -> letter.talon-list

It includes a `.tools/convert_all_csv_files_to_talon-list.py` file in a hidden directory so it is not picked up by talon. Running this script gives something like:

```
C:\Users\User\AppData\Roaming\talon\user\community\.tools> python convert_all_csv_files_to_talon-list.py
Base directory: C:\Users\User\AppData\Roaming\talon\user\community
Skipping unsuppported convertion yet: apps\emacs\emacs_commands.csv
Skipping unsuppported convertion yet: apps\git\git_arguments.csv
Skipping unsuppported convertion yet: apps\git\git_commands.csv
Skipping unsuppported convertion yet: core\app_switcher\app_name_overrides.linux.csv
Skipping unsuppported convertion yet: core\app_switcher\app_name_overrides.mac.csv
Skipping unsuppported convertion yet: core\app_switcher\app_name_overrides.windows.csv
Skipping unsuppported convertion yet: core\homophones\homophones.csv
Skipping unsuppported convertion yet: settings\abbreviations.csv
Skipping unsuppported convertion yet: settings\additional_words.csv
Converting csv file: settings\alphabet.csv -> talon-list file: core\keys\letter.talon-list
Skipping non default csv file settings\bla.csv
Skipping unsuppported convertion yet: settings\file_extensions.csv
Skipping unsuppported convertion yet: settings\search_engines.csv
Skipping unsuppported convertion yet: settings\system_paths.csv
Skipping unsuppported convertion yet: settings\unix_utilities.csv
Skipping unsuppported convertion yet: settings\websites.csv
Skipping unsuppported convertion yet: settings\words_to_replace.csv
Skipping unsuppported convertion yet: tags\emoji\emoji.csv
Skipping unsuppported convertion yet: tags\emoji\emoticon.csv
Skipping unsuppported convertion yet: tags\emoji\kaomoji.csv
```

This script retrieves all the `.csv` files from the `community/` folder and try to convert them to `.talon-list` files. It only support converting those that are default on knausj using a whitelist. Atm, only `alphabet.csv` is implemented and the others are skipped on purpose. Also, a non default `bla.csv` was created in settings in my local folder in order to show it is skipped.

[...]

PR #1382, opened 2024-02-04
convert keys to talon list where possible

knausj85:
1. built upon https://github.com/talonhub/community/pull/1268
2. does not port punctuation_words or symbol_keys, as these are used by create spoken forms
3. add support for keypad keys

see https://github.com/talonhub/community/pull/1382/commits/e25942a5ced9dcdd24a8786831e1c8e8d400f740 for changes actually made on this branch


pokey:
I think the idea would be to relax the requirement of never changing files in your community fork to only change `.talon-list` files. This way the merge conflicts would be rare, and should be straightforward to resolve, due to the simplicity of the talon list format. In the future, Talon package manager would make this even easier

But @knausj85 lmk if that's aligned with your thinking here

jaresty:
I don't currently maintain my own form of community and I don't want to start. I think it makes it harder to contribute changes upstream, which is bad for any open source effort. I don't think this would address the original intention of the csv PR that I submitted, based on this understanding.

jaresty:
One alternative that could work, I think, would be to keep the talon list files in their own repository that is just for configuration. I wouldn't mind forking a repo like that, and it would allow for community to consume the configuration instead of baking it in.

knausj85:
> I think the idea would be to relax the requirement of never changing files in your community fork to only change `.talon-list` files. This way the merge conflicts would be rare, and should be straightforward to resolve, due to the simplicity of the talon list format. In the future, Talon package manager would make this even easier
> 
> 
> 
> But @knausj85 lmk if that's aligned with your thinking here

This aligns with my thinking.


You also don't have to branch, you can create a .talon-list that's more specific with the addition of eg a tag that would take precedence over the default community list. What do you think about this @jaresty?

jaresty:
I think I can make that work, but I still don't think it's ideal. What I would really like to see is a super-simple and intuitive way to provide configuration that does not require understanding the details of how configuration is loaded to make the change you need to make. Most people will do what is easiest to do, and that is to rely on the example that is present and tweak just where needed.

If there is a strong desire to keep this configuration in the community repository, I'd at least like to see it collocated in a `configuration` folder with a README that explains clearly how to override it without changing the repository. I'd be interested to understand, though, if there are concerns around the idea of separating it out to another location? I appreciated the way CSV files were working before - where they set up a set of defaults but gitignored the generated files. I'm currently relying on that in my personal repository to override several talon settings.[1]

[1]: https://github.com/jaresty/mac-dotfiles/tree/master/talon-settings

knausj85:
> I think I can make that work, but I still don't think it's ideal. What I would really like to see is a super-simple and intuitive way to provide configuration that does not require understanding the details of how configuration is loaded to make the change you need to make. Most people will do what is easiest to do, and that is to rely on the example that is present and tweak just where needed.
> 
> 
> 
> If there is a strong desire to keep this configuration in the community repository, I'd at least like to see it collocated in a `configuration` folder with a README that explains clearly how to override it without changing the repository. I'd be interested to understand, though, if there are concerns around the idea of separating it out to another location? I appreciated the way CSV files were working before - where they set up a set of defaults but gitignored the generated files. I'm currently relying on that in my personal repository to override several talon settings.[1]
> 
> 
> 
> [1]: https://github.com/jaresty/mac-dotfiles/tree/master/talon-settings
> 
> 

Yeah, I see where you're coming from.

I think, however, committing talon-lists being a consistency that's sorely lacking at the moment. Specifically, one can essentially delete or disable the community .talon and .talon-lists (once most things are converted) files and customize the entire grammar if needed without touching the associated python.

Re: file location / configuration suggestion

We went thru something similar with .talon files where we decided to colocate them (as best as possible) with their associated python logic. I think this makes sense for .talon-list too for the same reasons (discoverability, ease of navigation  etc).

As an aside, I believe there are long termish plans at the Talon level for a plugin manager and a means of managing the talon/talon-list customizations too.

jaresty:
> Specifically, one can essentially delete or disable the community .talon and .talon-lists (once most things are converted) files and customize the entire grammar if needed without touching the associated python.

Yes! Exactly. Unless you don't want to fork the repository... in which case you can't do this. This is why it would be really nice (IMO) to separate these configuration files out into their own repository since it's intended to be changed by the users.

knausj85:
> > Specifically, one can essentially delete or disable the community .talon and .talon-lists (once most things are converted) files and customize the entire grammar if needed without touching the associated python.
> 
> Yes! Exactly. Unless you don't want to fork the repository... in which case you can't do this. This is why it would be really nice (IMO) to separate these configuration files out into their own repository since it's intended to be changed by the users.

Yeah, I've thought about that. Trouble is, there is a spectrum of users including those that just download the repository as a zip. I hesitate to add new barriers to entry, as the learning curve for talon + git for new folks is so high

jaresty:
I hear you. I definitely am not trying to raise any barriers to nontechnical folks. I think these two interests are not in conflict:
- the desire to make it easy for nontechnical folks to get started
- the desire to make it easy to use this repository without needing to fork it

What about this idea - when community loads for the first time and there isn't a configuration folder at a known location, Python could automatically extract or download this configuration. If it already exists, it wouldn't do anything.

jaresty:
Fwiw. I think that needing to fork the repository to configure adds friction to forcing upstream contributions since it encourages anyone who wants to customize anything at all to fork the repository. I don't want to do so, but I'm tempted myself because I can't change this configuration any other way. Once I fork it, I doubt I'll be motivated enough to contribute back since the friction is pretty high even just considering how many users there are here.

jaresty:
In the talon-ai-tools repository we adopted using a `settings.talon.example` file that can be renamed which is one way to address both. (https://github.com/C-Loftus/talon-ai-tools/blob/main/GPT/gpt-settings.talon.example)

jaresty:
Another option that could work is to make the configuration completely optional... in other words, anywhere that currently requires configuration could fall back to a default value if configuration isn't provided. That would allow new users to skip configuration altogether, while anyone who wanted to could configure alternative values.

jaresty:
One more thought that isn't far off from where this current PR is headed - what if we adopted this change more or less as-is but also created a repository that simply demonstrated how to override the configuration here? That would make it easy for folks who want to work off of community weirdly forking to just fork the configuration repository, which could be kept in sync with this repository by the community?

jaresty:
I realized that this PR is an improvement over the existing functionality: it makes it _possible_ to configure alternate keys without changing the repository where before it was not.

I don't love that it does not provide an easy way to discover how to do so and that there are solutions I can imagine that don't put the burden of figuring out how onto the users who don't want to fork, but this is still an improvement over the existing implementation, so I don't want to stand in its way.

jaresty (approved):

jaresty:
I'd love to see this merged. I'm wondering if there's anything still standing in its way?

jaresty:
What does this mean? I was considering submitting a PR specifically for punctuation and modifiers but it sounds like that may not be easy?

> does not port punctuation_words or symbol_keys, as these are used by create spoken forms

jaresty:
➕  let me know how I can help as well. I'm keen to see this progress.

knausj85:
> What does this mean? I was considering submitting a PR specifically for punctuation and modifiers but it sounds like that may not be easy?
> 
> > does not port punctuation_words or symbol_keys, as these are used by create spoken forms

This file depends upon these and there's no mechanism for getting the active talon-list that's officially sanctioned. CSVs, for better or worse, would be the "correct" approach for these.

https://github.com/talonhub/community/blob/d39c25dc8b2acc394af976515aea6828ead06639/core/create_spoken_forms.py#L11 


> Can I help out with this PR in any way? I wasn't super clear if more needs to be done technically or if this is just a logistical block.
> 
> I think in general having the alphabet and key names in a talon list is really significant benefit for UX and makes it a lot easier to configure and override. I also think that having talon lists is a nice tradeoff that allows both customization in the repository or also outside of it for those who don't want to fork.
> 
> For what it is worth I am also trying to move off a fork. I find that contributing upstream has a fair bit of friction because of merge conflicts each time. I think changes like these are super important and would love to help out to get this in a state to get it merged.

Main holdup is whether we want to try to auto-migrate existing python lists to avoid breaking people.

core/keys/function_key.talon-list
```diff
@@ -0,0 +1,26 @@
+list: user.function_key
+-
+f one key: f1
```
nriley, line 3: Please double check that community by default uses "f one key" instead of "f one" — thanks!

core/keys/keypad_key.talon-list
```diff
@@ -0,0 +1,19 @@
...
+key pad five: keypad_5
+key pad six: keypad_6
+key pad seven: keypad_7
+key pad eight: keypad_8
+key pad nine: keypad_9
+key pad point: keypad_decimal
+key pad plus: keypad_plus
+key pad minus: keypad_minus
+key pad times: keypad_multiply
+key pad stroke: keypad_divide
```
nriley, line 17: Would recommend making these equivalent to whatever the keys are when they are not on the keypad.

core/keys/win/modifier_key.talon-list
```diff
@@ -0,0 +1,10 @@
+list: user.modifier_key
+os: windows
+os: linux
+-
+alt: alt
+control: ctrl
+shift: shift
+super: super
```
nriley, line 8: Might be worth adding a comment to indicate that the super key is the Windows key.

nriley (commented):
Looks good modulo the comments I added and removing all of the CSV files that are being migrated to .talon-list in #1268.

PR #1385, opened 2024-02-10
Update app.exe matching to work with both public and beta Talon

nriley:
- Update mixed-case app.exe: matchers to work with public and beta Talon.
- Update app switcher to work with public and beta Talon.

### DIFF
--- a/apps/adobe/adobe_acrobat_reader_dc.py
+++ b/apps/adobe/adobe_acrobat_reader_dc.py
@@ -2,14 +2,14 @@ from talon import Module
 
 # --- App definition ---
 mod = Module()
-mod.apps.adobe_acrobat_reader_dc = """
+mod.apps.adobe_acrobat_reader_dc = r"""
 os: windows
 and app.name: Adobe Acrobat DC
 os: windows
-and app.exe: /^acrobat.exe$/i
+and app.exe: /^acrobat\.exe$/i
 os: windows
 and app.name: Adobe Acrobat Reader DC
 os: windows
-and app.exe: /^acrord32.exe$/i
+and app.exe: /^acrord32\.exe$/i
 """
 # TODO: mac context and implementation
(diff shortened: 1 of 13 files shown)


nriley:
I've not got Windows at home so this specific set of changes aren't well tested, but wanted to get it out there since these changes will affect Talon beta users. Will test more on Monday if nobody else does over the weekend!

apps/adobe/adobe_acrobat_reader_dc.py
fidgetingbits, line 1: These two are missing the escaped dot

nriley: Thanks. I've got more changes not yet moved over to this branch but will be the weekend until I get back to it.

knausj85:
Should we update the copy context commands to do this regex magic?

nriley:
> Should we update the copy context commands to do this regex magic?

Good call, yes! Will do.

AndreasArvidsson:
Definitely. We should also update create app context:
https://github.com/talonhub/community/blob/c6e1969ab516d509f51b9c8acd01bd35d2262d28/plugin/talon_helpers/create_app_context.py#L79

We should probably even unify the implementation of these two

nriley:
> Definitely. We should also update create app context:
> 
> https://github.com/talonhub/community/blob/c6e1969ab516d509f51b9c8acd01bd35d2262d28/plugin/talon_helpers/create_app_context.py#L79

Done. 

> We should probably even unify the implementation of these two

Agreed, although would prefer to do another time. They generate Windows app matching differently (create_app_context just matches app.exe and talon copy context also matches app.name).

core/app_switcher/app_switcher.py
```diff
@@ -23,6 +25,9 @@
 # a list of the current overrides
 overrides = {}
 
+# apps to exclude from running list
+excludes = set()
```
knausj85, line 29: Probably worth adding a comment that any line without a comma is excluded? This is a bit different from several of the other CSVs, where it's assumed the spoken form and result are the same in this case.

knausj85 (approved):
This look solid to me with one minor comment

core/app_switcher/app_switcher.py
nriley: Agreed, that was a bit too mysterious (and comments in this file were out of date). Updated.

PR #1386, opened 2024-02-18
Improved retrieval of current path from Windows File Explorer

amadea-system:
The old implementation of retrieving the current path from the title of the active File Explorer window only worked as long as the path was under 100 characters. Otherwise the title got truncated and no longer matched a valid path. This adds a new method of getting the path by pulling it directly from the address bar using accessibility elements.

I'm not sure if this works on Windows 11 as I am unable to test that currently. I suspect that with the large amount of changes that have been made to File Explorer it will likely need some adjustments to work properly on Windows 11. However, it should gracefully fall back to the old method if the address bar elements cannot be found.

### DIFF
--- a/apps/windows_explorer/windows_explorer.py
+++ b/apps/windows_explorer/windows_explorer.py
@@ -1,6 +1,7 @@
 import os
+import re
 
-from talon import Context, Module, actions, app, ui
+from talon import Context, Module, actions, app, ui, windows
 
 mod = Module()
 apps = mod.apps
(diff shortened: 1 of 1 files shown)


lunixbochs:
1. don't import from talon.windows.ax, if Element isn't available in talon.ui I'll make it available
2. merging this should be blocked on the 0.5 release, so it should maybe a be a draft PR
3. windows accessibility isn't quite done, much of @nriley's accessibility coding style here is working around some of the incomplete aspects of windows accessibility that will be obsolete when it's finished in 0.5

amadea-system:
Thanks for the feedback!

1. Element doesn't seem to be available in talon.ui yet.
2. Since it definitely sounds like this is very premature, I'll go ahead change this to a draft PR for now.

PR #1390, opened 2024-02-21
Add vscode command to search for views by name

jaresty:


jaresty:
Any thoughts on this one? I find this really handy since there isn't an easy way to search for views built in currently. There are some hard coded for some of the side bars but if you install any extensions you have to map them yourself. This lets you find anything.

apps/vscode/vscode.talon
```diff
@@ -14,6 +14,10 @@ please [<user.text>]:
     user.vscode("workbench.action.showCommands")
     insert(user.text or "")
 
+view [<user.text>]:
```
knausj85, line 17: Would this make more sense as a .talon-list, rather than user.text? Seems like it would be more reliable, with the downside of needing to add an entry for each view.

jaresty: I prefer the test option because I'm using it to discover views I doing know about. It's kind of like "please" for searching views. If I had to delineate all the options, it would be less useful to me. There are several commands already for toggling bars and panels and such but there isn't such a convenient view search option yet (imo)

knausj85: I'm okay with this, but I think we should prefix the command with "go"

jaresty: I'm fine with that.

knausj85 (approved):

PR #1391, opened 2024-02-21
add "server/workspace up/down" commands for consistency

timo:
this mirrors "channel up/down" from the same file, and mirrors the more common "noun verb" order in commands


fidgetingbits:
good idea to keep them consistent. fine by me. Should we deprecate the other two?

timo:
not sure if it's worth the trouble. don't know what the general stance on deprecation is in the community repo. i see the deprecate_notify function but no uses of it.

should we merge this already and create an issue for the deprecation decision of the two commands that are "backwards"?

fidgetingbits:
you want to search for user.deprecate_command use instead.

I believe the stance is to always do it, see https://github.com/talonhub/community/commit/54ce69d2fb6ba1ed44759a14daad9ceef933323c for a recent one.

I'm fine with merging this if you prefer to do a separate PR for deprecation. if you don't want to do it, let me know and I'll throw one together. I agree it's a bit of a hassle, but we don't have any visibility into how many people even use community afaik, and the agreement was that we slowly deprecate stuff, so still seems worth to be on the safe side.

tags/messaging/messaging.talon
```diff
@@ -3,6 +3,8 @@ tag: user.messaging
 # Navigation
 previous (workspace | server): user.messaging_workspace_previous()
```
AndreasArvidsson, line 4: Please deprecate these old commands using `user.deprecate_command`

tags/messaging/messaging.talon
```diff
@@ -3,6 +3,8 @@ tag: user.messaging
...
 next (workspace | server): user.messaging_workspace_next()
+(workspace | server) up: user.messaging_workspace_previous()
```
AndreasArvidsson, line 6: Please also support next/previous in addition to up/down.

AndreasArvidsson (changes requested):

PR #1393, opened 2024-02-25
Add menu choosing helper from pokey_talon

jaresty:


fidgetingbits (approved):
Fine by me.

fwiw I like the idea of rather than this being global it gets tied to a menu tag and then is only enabled for apps that actually assert they have menus, similar to how we do work pages, tabs, etc. I don't think it needs to be done right away though. Not sure what others think.

If it gets associated with a menu tag, then the this would move under the tags folder with the others.

fidgetingbits:
Nice find. I didn't even realize this existed in community already. I'm okay with this as I prefer "choosing" wording as well, as to me drop down actually selecting the entry isn't so intuitive. 

That said I forget what we generally do about adding aliases to existing commands just because some people want them, so someone will have to remind me. Maybe something to write down as a general policy for the repo, if it's not already somewhere.

jaresty:
Personally, I'm not a fan of needing to fork the repo to customize so I'm hoping there's a solution that doesn't require it!

Regarding limiting scope: I was paying with the Homeric app on Mac last night and this command can navigate menus system wide once you have "clicked" with it so I'd like it to work everywhere.

jaresty:
I haven't seen any more feedback on this one. I think no feedback means it's probably not too controversial...?

Are you able to merge @fidgetingbits?

fidgetingbits:
I actually think, for the same reason I already said, that this shouldn't be merged unless we come up with an official policy of dealing with this type of request. That said, I'd rather leave it to others to decide on what the goal of the community repo is, and I don't particularly enjoy debating stuff like this lol.

But fwiw the way I see it, and my understanding of PRs that we generally (there have been exceptions) accept, is that community generally defines one standard way of saying things, and if someone doesn't like it they need to customize, which atm usually means fork. There are too many people customizing stuff (thus so many community forks) to sanely merge in everyone's chosen alias for a given command, especially given the current design.

I know you don't like forking, which is fine, but I'd also like it to be clarified (who I dunno, maybe aegis needs to) if we want one of the goals of the community repo to be preventing people from having to fork (I don't believe it is atm). If we did actually want to go that route, then I'd still want to reject this PR and suggest it instead it should be done in a way that abstracts the spoken command through an action so that anyone can specify whatever they want (similar to how pokey uses <user.teleport> in his repo iirc).

The problem with going with this latter abstraction route is, as far as I know, one of the goals of the talon language is ease of use and readability for non programmers. If every command becomes abstracted in a way to prevent forking, imo the talon files will become unreadable to the common user. Maybe this aegis needs to clarify too.

To me this PR is merging something because you personally find it convenient and don't want to fork. Hypothetically, if I also didn't like to fork, and already had `choosing` used for something else, and we merge this, then who wins? Now I have to fork because you didn't want to? It seems not feasible to avoid this problem, and so we just generally select a single thing that's "good enough" for general users and if people want something else they have to customize. A further example is @nriley uses `pick` for this exact same thing in his pick.talon plugin iirc. He'll presumably never use choosing, so should we now have `(drop down | choosing | pick)`? The more of these aliases we add for personal preference the more it restricts the available words to use for other commands people want. imo its easier to just create a choosing.talon file in your .talon/user (no need to fork for this particular one even ) and not merge this.

All that said, I'm not actually against the idea of a repo (ironically perhaps a fork of community) that has a main goal which is to not have users fork in order to customize, and maybe I'd even use it myself, but my issue here is that I don't think that's actually one of the goals of community in its current form.

Maybe I'm thinking too much and should just merge this, but I'd rather wait for someone that's been around as long or longer than me to comment, because I'm quite sure this isn't the first time we've not merged something like this for the same reason.

@knausj85 @rtnz @pokey @phillco @lunixbochs @AndreasArvidsson @timo

jaresty:
It seems thatmy desire to avoid forking was communicated by my reason for avoiding it hasn't... I definitely understand that there are personal preferences that may differ and that negotiating those differences is challenging. I'm of the mind that where there are significant differences it ought to be configurable or not in this repository at all. As it is, earlier preferences have more weight than anything that comes later because there's no easy way to agree to adopt a change. There are various ways to do many things and there are quite a few that I don't like but have been dealing with because I don't want to maintain a fork. I'm not trying to be difficult here, and I can solve this on my own if I'm out of line here but I think everyone loses out if we choose not to change it just because the existing language got there first even if another word choice is better for more people. (Not saying this choice is, but if it were then it would be a bad outcome).

fidgetingbits:
> It seems that my desire to avoid forking was communicated by my reason for avoiding it hasn't... I definitely understand that there are personal preferences that may differ and that negotiating those differences is challenging. I'm of the mind that where there are significant differences it ought to be configurable or not in this repository at all. As it is, earlier preferences have more weight than anything that comes later because there's no easy way to agree to adopt a change. There are various ways to do many things and there are quite a few that I don't like but have been dealing with because I don't want to maintain a fork. I'm not trying to be difficult here, and I can solve this on my own if I'm out of line here but I think everyone loses out if we choose not to change it just because the existing language got there first even if another word choice is better for more people. (Not saying this choice is, but if it were then it would be a bad outcome). 
> 
> Fwiw - I also like "pick" as much as "choosing" - I'm just not a big fan of the current "drop down (number)" because I find it hard to say accurately. I'm interested though in hearing out other opinions on this because there could be a really good reason why it is the way it is that would persuade me to just keep using it but with a better appreciation of why it's a good design. 

You aren't out of line and I'm not trying to be difficult either fwiw. I also don't use 
drop down, and likely never will, and don't know off hand why it was chosen, aside from I guess it is related to drop down menus. 

I'm not opposed to changing commands like this if they don't make sense, and agree just because things are there first they shouldn't be favored for no reason. If things do change , they should be removed using deprecation vs just adding in more aliases imo. 

But, I don't know how to reasonably select what is actually "better", aside from usually fewer syllables and thematic words are widely considered preferable. I like pick the most of the 3 and have actually been testing using it lately, but I also use tick for ' and I suspect that will occasionally conflict which will inevitably lead me to using something else anyway. That doesn't mean it's not best for everyone else *shrug* Maybe most people already like drop down :D

I recall knausj85 already mentioned why he doesn't want to break out into multiple repos to solve some of these problems, which I guess can be continue to be debated elsewhere. 

fwiw aegis has some vision for a more modular plugin based setup with a super minimal base grammar in the future (no idea on time frame), that might address a lot of your concerns. Not sure, but this may also play into why people aren't rushing to change community's current form. May be worth asking him more about that.

jaresty:
I haven't seen the reasoning for not breaking out into separate repositories. I have heard about the modular plugin system idea. I could be wrong but I think that even a modular plugin system would require a lot of work from community to be usable. For example, I'd expect that many users wouldn't want to add support for every programming language. There's currently a global registry of programming languages that doesn't allow for registration without modifying the base Python code, so it couldn't be used without redesigning it to support languages registering themselves if you wanted to have different language configurations dependent on what plugins were installed. (Maybe not the best example, but this one comes to mind). I think this is resolvable now if there is the desire to do so and the benefit is that anyone could add languages to their own configurations without changing the base repository.

My point is that the problem doesn't go away or change by adding a packaging system, to my understanding. It actually puts more pressure to resolve the problem since it adds additional expectations about how that would work.

jaresty:
I wanted to share one example where `drop down` can cause issues with cursorless terminology. I often use `drop down` when I'm selecting a file from `file hunt`. Sometimes talon thinks I said `chuck down <n>` instead of drop down. This deletes the line n lines below the cursor position while I'm trying to select a file. I can't undo without first closing the file select dialog, and if I pick a file instead vscode will auto save with that change. This is a pretty unpleasant use experience that would be nice to fix for users of vanilla community and cursorless.

knausj85:
The main problem with adding arbitrary aliases is potential conflicts, so I personally prefer to keep aliases to a minimum. This is more problematic for any global commands such as this one.

I don't believe this would play well with this, for example https://github.com/talonhub/community/blob/main/core/homophones/homophones_open.talon#L3

As-is, depending on context, people would need to remember "choose" or "choosing" which seems less than ideal. Plus, choose and choosing likely too close to be reliably distinguished

We could implement an overridable action here to provide a consistent "choose" command

jaresty:
I like the overridable idea!

jaresty:
I'm not sure how to approach it though. Can you help me understand how to implement it?

pokey:
> I'm not sure how to approach it though. Can you help me understand how to implement it?

search for eg `code_comment_block` (whole word only). You'll see how it appears in a `.talon` file, a generic `.py` file, and context-specific `.py` files

jaresty:
I understand how to use indirection to call into a python file. I'm confused about is how you would override it without modifying community. Can you override python files by simply including some additional files that override them?

jaresty:
If some wants to take a stab at this (in case it's easier to demonstrate than explain) I'd be happy to review.

pokey:
I think @knausj85 is proposing that there be one choose action, and if homophones are open, it picks the homophone, otherwise it runs the generic action defined in this PR. So if I'm understanding correctly it's not about users overriding the action, but us overriding the action when homophones are open

pokey:
Also, looks like something strange happened to the commit history here 🤔. The diff is including a bunch of recent mainline changes for some reason

timo:
it prominently features my name there but i'm pretty sure i didn't do anything to this PR

jaresty:
Does it look better now?

timo:
LGTM

timo (approved):

pokey (changes requested):
The diff looks correct, but we still need to figure out what to do about https://github.com/talonhub/community/pull/1393#issuecomment-2041064896. Also, if we're deprecating old commands (which this PR currently does by deleting `dropdown.talon`), we need to use the deprecation infrastructure

jaresty:
I think I'm confused. I thought that by adopting the same wording, we were addressing the comment https://github.com/talonhub/community/pull/1393#issuecomment-2041064896. I also don't understand how to use the deprecation infrastructure. Can you help me out there?

pokey:
The problem with this approach is that you know have a conflict between the "choose" here and the one when homophones is up. Could hack it by adding a context here but better to have an overridable action 

Re deprecation, grep for "deprecate" in the code base; if still not clear how to use it lmk and we can try to improve docs in that file

jaresty:
Thank you for that context. I'm still not sure what an "overrideable action" is. Is there an example I could follow?

pokey:
I'd have another look at https://github.com/talonhub/community/pull/1393#issuecomment-2041451646; that should help figure out how to do an overridable action. If that's still not making sense maybe we could connect briefly before Tuesday cursorless slot like before?

jaresty:
Thanks! Yeah I took another look but I'm still not quite clear. Sure-I'm glad to come at 7:30am pacific before the Tuesday cursorless meetup to chat. Thanks!

jaresty:
I'm starting to get it, I think. Is there any way to define a default and only override in the one case?

jaresty:
How's this approach?

pokey:
That still leaves the user with two places to change the spoken form if they want to change it, but may be a reasonable compromise? @knausj85 wdyt?

Also, any reason to delete the old file and make a new one? It's still doing the same thing; just a new spoken form

And still needs deprecation for old spoken form; there should be instructions if you search for "deprecation" or "deprecate"

jaresty:
Re: deletion-I thought it made sense since the old file name also matches the command name. If I understood the override idea, I think it would still have the same effect (I.e. needing to change in more than one place). I think you could make a talon list and add a spoken form that can be changed in one place there but since there as only two examples so far it seems like too much indirection to me. Happy to restore the original file name.

knausj85:
The talon wiki covers how to create and override actions pretty well: https://talon.wiki/Customization/Talon%20Framework/actions

I still think this is a preferable approach, as there are similar mechanics in many places. 

for example in gaze-ocr
https://github.com/wolfmanstout/talon-gaze-ocr/blob/main/gaze_ocr_disambiguation.talon
with the action approach, we can consolidate to single overridable action and potentially define a more consistent grammar. 

The downside of this approach is we would no longer be able to interact with drop downs while the other stuff is open. It's rare, but I've personally seen it.

jaresty:
I prefer this over an overridable action because I often interact with drop downs in applications that have no talon configuration at all. I could be misunderstanding here, but wouldn't it require every application to define an action for the override approach to work? Or is there a way to set a default behavior?

knausj85:
The thought is the drop down implementation would be the default action implementation, and the command would still be global.

And eg Homophones and gaze ocr could override the action when active. 

This allows us to define one command that does the right thing depending on the active context.

jaresty:
Ok. That makes sense to me. If you have an existing example you could share for setting up a default action please do. I'll look into this later or also please feel free to push to this branch with what you think makes sense.

jaresty:
I'm struggling with how to do this a little bit. This is what I tried:

```
from talon import Context, Module, actions

default_dropdown_select = Context()
mod = Module()


@mod.action_class
class Actions:
    def choose(number_small):
        """Choose the nth item"""
        actions.key("down:{number_small-1} enter")

    def choose_up(number_small):
        """Choose the nth item up"""
        actions.key("up:{number_small} enter")


@default_dropdown_select.action_class("user")
class ChoosingActions:
    def choose(number_small):
        actions.key("down:{number_small-1} enter")

    def choose_up(number_small):
        actions.key("up:{number_small} enter")
```

It's not working and I'm not sure why. This is the error I see in the talon log:

```
2024-05-12 13:39:14.927 ERROR cb error topic="phrase" cb=SpeechSystem.engine_event
   16:      lib/python3.11/threading.py:995 * # cron thread
   15:      lib/python3.11/threading.py:1038* 
   14:      lib/python3.11/threading.py:975 * 
   13:                    talon/cron.py:156 | 
   10: talon/scripting/speech_system.py:369 | 
    9:             talon/engines/w2l.py:742 | 
    7: talon/scripting/speech_system.py:66  | 
    4: talon/scripting/speech_system.py:442 | 
    3:      talon/scripting/executor.py:111 | 
    2:  talon/scripting/talon_script.py:707 | 
    1:  talon/scripting/talon_script.py:610 | 
 in script at /Users/tkma6d4/.talon/user/community/lang/tags/operators_math.talon:15:
 > user.code_operator_greater_than()
NotImplementedError: Action 'user.code_operator_greater_than' exists but the Module method is empty and no Context reimplements it
```

pokey:
You're on the right track!

- I don't think that error message is related. There are no others in your log?
- What happens when you say "choose three"?
- I don't think you need the second class; it should just work with the first class. You should only need the second class when you want to override it, eg when homophones is open, where you'd do what you have there but add a matcher to the context
- Looks like you're missing argument+return value type hints and a docstring on the action definition. Talon is probably rejecting it for that reason, tho usually you'd see an error message

jaresty:
I got this to work now. Thanks for your help! 🙇

auscompgeek:
We should use `deprecate_command` to deprecate the commands.

core/homophones/homophones_open.talon
```diff
@@ -1,10 +1,10 @@
 tag: user.homophones_open
 -
 choose <number_small>:
-    result = user.homophones_select(number_small)
+    result = user.choose(number_small)
     insert(result)
     user.homophones_hide()
```
pokey, line 6: this whole thing needs to be the `choose` action, and then you would remove this whole command entirely. The way you have it, you've created a nondeterministic conflict between the `choose <number_small>` command here and the global one

jaresty: I think I fixed that now. I did discover that the type signatures disagreed so I had to add a return type to the default choose implementation, which I wasn't sure about.

jaresty: Nevermind. It may have been what you originally meant, but at least I figured out a way that seems more simple.

jaresty: If I'm still not getting it, please feel free to push to this branch.

jaresty:
I'm getting a new problem that I don't understand. The homophones work now but now the original 'choose' helper is throwing an error:

```
2024-05-28 21:17:08.035 ERROR cb error topic="phrase" cb=SpeechSystem.engine_event
   16:      lib/python3.11/threading.py:995 * # cron thread
   15:      lib/python3.11/threading.py:1038* 
   14:      lib/python3.11/threading.py:975 * 
   13:                    talon/cron.py:156 | 
   10: talon/scripting/speech_system.py:369 | 
    9:             talon/engines/w2l.py:742 | 
    7: talon/scripting/speech_system.py:66  | 
    4: talon/scripting/speech_system.py:442 | 
    3:      talon/scripting/executor.py:111 | 
    2:  talon/scripting/talon_script.py:707 | 
    1:  talon/scripting/talon_script.py:610 | 
 in script at /Users/tkma6d4/.talon/user/community/plugin/dropdown/dropdown.talon:2:
 > user.choose(number_small)
TypeError: object of type 'NoneType' has no len()
```
I'm kind of lost as to what could be going on here.

jaresty:
I verified that the addition of the homophones user override seems to cause this error. If I remove the override it works.

jaresty:
I checked this a few times and I think it's working correctly now.

pokey:
This is still not quite right. You currently have two rules of the form`choose <number>`. That will cause Talon to randomly pick one. There should only be one rule of this form, have it mapped to a single action, and then override that action depending on context

jaresty:
got it. Removed the redundant rule.

core/homophones/homophones_open.talon
```diff
@@ -1,9 +1,7 @@
 tag: user.homophones_open
 -
-choose <number_small>:
-    result = user.homophones_select(number_small)
-    insert(result)
-    user.homophones_hide()
+tag(): user.choose_selector
+
```
pokey, line 4: what is this tag for?

plugin/dropdown/dropdown.py
```diff
@@ -0,0 +1,17 @@
+from talon import Context, Module, actions
+
+default_dropdown_select = Context()
```
pokey, line 3: this can be removed right?

plugin/dropdown/dropdown.py
```diff
@@ -0,0 +1,17 @@
...
+mod = Module()
+
+mod.tag("choose_selector", desc="Tag for selecting with 'choose'")
```
pokey, line 6: as above, not sure I see what this tag is for

plugin/dropdown/dropdown.talon
```diff
@@ -1,3 +1,8 @@
 # pick item from a dropdown
-drop down <number_small>: key("down:{number_small-1} enter")
-drop down up <number_small>: key("up:{number_small} enter")
+choose <number_small>: user.choose(number_small)
+choose up <number_small>: user.choose_up(number_small)
+
+# DEPRECATED
+drop down <number_small>: user.deprecate_command("2024-05-13", "drop down", "choose")
+drop down up <number_small>:
+    user.deprecate_command("2024-05-13", "drop down up", "choose up")
```
pokey, line 8: we should prob update these dates before we merge

core/homophones/homophones_open.talon
jaresty: I was confused about how the overrides work. Seems unnecessary now.

plugin/dropdown/dropdown.py
jaresty: sure can!

pokey (approved):
Looks good to me! Haven't tested it tho. I'm assuming it works and there's no warnings in the log?

jaresty:
I'm not seeing any additional warnings beyond what I normally do. It works based on my testing it several times.

pokey:
Just took it for a spin locally; works for me! One last thing: the deprecation warning in the log points the user to [`BREAKING_CHANGES.txt`](https://github.com/talonhub/community/blob/main/BREAKING_CHANGES.txt), so we should drop a line there about this deprecation. Otherwise it's good to go!

fidgetingbits:
Curious if we should take the opportunity to change the plugin name from dropdown to something something like "menu_choose" since it's more descriptive?

Also just a thought, homophones are "core" functionality, and now rely on an action defined by this plugin (which sort of implies it's not core? I dunno, seems weird anyway). I wonder if we should move to to core/menu_choose/?

pokey:
> Also just a thought, homophones are "core" functionality, and now rely on an action defined by this plugin (which sort of implies it's not core? I dunno, seems weird anyway). I wonder if we should move to to core/menu_choose/?

this is a great point. I think we should have a hard rule that core cannot depend on anything from plugins

jaresty:
moved it into core.

core/menu_choose/menu_choose.talon
```diff
@@ -0,0 +1,8 @@
+# pick item from a dropdown
+choose <number_small>: user.choose(number_small)
+choose up <number_small>: user.choose_up(number_small)
+
+# DEPRECATED
+drop down <number_small>: user.deprecate_command("2024-05-29", "drop down", "choose")
+drop down up <number_small>:
+    user.deprecate_command("2024-05-29", "drop down up", "choose up")
```
pokey, line 8: should we leave these in the original file?

pokey (commented):
Ok yeah that seems reasonable. Left one more comment and also don't forget adding an entry to breaking changes file as described in https://github.com/talonhub/community/pull/1393#issuecomment-2139353617

core/menu_choose/menu_choose.talon
jaresty: moved it!!

fidgetingbits:
I think all of the issues have been addressed now. Thanks @jaresty for your contribution, effort, and patience.

pokey:
Yes I'll second @fidgetingbits's comment here. I'm quite happy with how this one turned out

PR #1394, opened 2024-02-25
Ad then command from cursorless recommendations

jaresty:


nriley:
Best PR ever?

pokey:
Makes sense to me. We should probably add a comment, though

jaresty:
What should the comment say?

jaresty:
Added!

knausj85:
I might be inclined to call the plugin itself skip or chain, rather than then.

jaresty:
I'm ok with any of those options. Copped the recommendation from the cursorless wiki so I'll let some other folks weigh in before changing.

auscompgeek:
The directory/file name. I think the command name `then` is fine.

plugin/chain/chain.talon
```diff
@@ -0,0 +1,2 @@
+# This makes it easy to chain commands.
```
phillco, line 1:
```suggestion
# This makes it easier to chain commands by letting you hint about command boundaries.
# For example, with Cursorless, the phrase "post line air" is ambiguous as to whether you meant a single command ("post line air", i.e. "move the cursor to the end of the line containing the 'a' hat"), or two separate commands ("post line" to move the cursor to the end of the current line, followed by "air" to insert the letter "a").
# If you know you want the latter, this allows you to say "post line then air" to force that interpretation.
```

phillco (approved):
I suggest expanding the comment a little to make the meaning more obvious, because I don't think it is if you just see the command definition, unless you already know :).

(imo, in community we've historically biased too much towards terseness in these situations.)

cc @pokey if this captures the original intent of the suggestion correctly

phillco:
> I might be inclined to call the plugin itself skip or chain, rather than then.

Sorry I missed this earlier @knausj85, but I actually find either of those names more confusing than the original `then`. This file doesn't enable command chaining or `skip()` -- those are features intrinsic to Talon; it _uses_ them to introduce a new behavior, but naming it skip or chain might confuse people into thinking it's somehow necessary to have this file to have command chaining itself.

I actually thought @jaresty nailed it with his original name, which is pretty perfectly self-descriptive imo, and avoids the possibility for that confusion. Not strongly opinionated since I know this PR has been open for a bit and I'm late to the party, but I'm curious if I missed anything.

(other than that lgtm :)

jaresty:
I'll let @knausj85 respond to the feedback there-I'm ok with either name. Please feel free to push a rename commit if you like, or I will once I get a clear signal.

jaresty:
renamed it back to the original 'then'

fidgetingbits:
I think @phillco reasoning is sound, and since knausj85 mentioned he won't be around around for awhile, I think it's OK not to wait. If it ends up being an issue can change later, but hopefully yhe comment is clear enough now. As such I'm merging, hope everyone is OK with that.

pokey:
yeah this lgtm

PR #1395, opened 2024-02-26
Add vscode terminal helper

jaresty:
It makes sense that the VSCode terminal should behave like a terminal. There's also a corresponding setting that needs to be set in VSCode. cc: @pokey


jaresty:
Added some documentation.

apps/vscode/README.md
```diff
@@ -5,3 +5,11 @@ It is recommended to install the [VSCode talon extension pack](https://marketpla
 ## Cursorless
 
 If you'd like to use Cursorless, follow the instructions in the [cursorless-talon repo](https://github.com/pokey/cursorless-talon).
+
+## Terminal
+
+By default the VSCode terminal is not going to recognize terminal commands from talon. In order for talon to configure terminal commands in VSCode the window has to include a word that it can use to match. This is possible using the VSCode setting to include word terminal in the title. You can do this with this setting in VSCode:
```
pokey, line 11:
```suggestion
In order to use terminal commands when the VSCode terminal is focused, you must add the following line to your [VSCode `settings.json`](https://code.visualstudio.com/docs/getstarted/settings#_settingsjson):
```

apps/vscode/README.md
```diff
@@ -5,3 +5,11 @@ It is recommended to install the [VSCode talon extension pack](https://marketpla
...
+
+```
+"window.title": "${dirty}${activeEditorShort}${separator}${rootName}${separator}${profileName}${separator}${appName}${separator}focus:[${focusedView}]",
+```
```
pokey, line 15:
````suggestion
```

This setting will cause VSCode to include a special string in the window title whenever the terminal is focused. Talon will look for this string in the window title and activate the terminal commands in response.

Note that if you have customizations in your window title that you'd like to keep, the important part is just to ensure that `focus:[${focusedView}]` appears somewhere within your custom window title.
````

File: apps/vscode/vscode_terminal.talon [hunk 3]
```diff
@@ -0,0 +1,7 @@
+app: vscode
+# Treat the VSCode terminal as a terminal. This relies on VSCode to include the word terminal in the window title.
+# Add this setting to your settings:
+#   "window.title": "${dirty}${activeEditorShort}${separator}${rootName}${separator}${profileName}${separator}${appName}${separator}focus:[${focusedView}]",
```
pokey, line 4:
```suggestion
# Looks for special string in window title.
# NOTE: This requires you to add a special setting to your VSCode settings.json
# See [our vscode docs](./README.md#terminal)
```

pokey (commented):
Ok looks good with some minor suggestions. I'm assuming you've tested this locally?

jaresty:
It looks like I've got some debugging to do. For some reason the tag is not quite matching. I'm not sure why yet.

pokey:
fwiw this is the combo that works for me:

- https://github.com/pokey/dotfiles/blob/f7e3c8c632cd898406c8bc56fc4eb859d416e58e/vscode/settings.json#L522
- https://github.com/pokey/pokey_talon/blob/b650d16e1d89c90e60646c8e12da5cb5922bbaf8/apps/vscode/vscode_terminal.talon#L2

jaresty:
It seems that the tag is matching but for some reason the commands are not inputting into the terminal as I would expect. I was expecting them if I tried `lisa all` it would actually do the same thing as it does in terminal. It recognized the command but didn't input anything. I'm not sure why.
image

jaresty:
I updated the configuration with your example. I'm still running into it not typing into my terminal for some reason.

jaresty:
I found it:

```
2024-04-05 17:18:36.199 ERROR cb error topic="phrase" cb=SpeechSystem.engine_event
   16:      lib/python3.11/threading.py:995 * # cron thread
   15:      lib/python3.11/threading.py:1038* 
   14:      lib/python3.11/threading.py:975 * 
   13:                    talon/cron.py:156 | 
   10: talon/scripting/speech_system.py:369 | 
    9:             talon/engines/w2l.py:742 | 
    7: talon/scripting/speech_system.py:66  | 
    4: talon/scripting/speech_system.py:442 | 
    3:      talon/scripting/executor.py:111 | 
    2:  talon/scripting/talon_script.py:707 | 
    1:  talon/scripting/talon_script.py:610 | 
 in script at /Users/tkma6d4/.talon/user/community/tags/terminal/terminal.talon:5:
 > user.terminal_list_directories()
NotImplementedError: Action 'user.terminal_list_directories' exists but the Module method is empty and no Context reimplements it
```

jaresty:
Figured it out.

knausj85:
This looks good, though VSCode does support multiple terminals (power shell, command prompt, git bash, etc). I don't suppose there's a way to differentiate the active one in the title? We could then potentially leverage the existing implementation for each.

jaresty:
I'm not sure... maybe @pokey can help out here. 🙏

pokey:
no I don't believe that's possible. I would just include `tag: terminal` here, and then the user can have a separate file that activates the other terminal tags in a `tag: terminal` context. Do we have docs somewhere on how that works? I believe that problem will appear in other terminal settings, no? Eg whether to activate the `git` commands

knausj85:
Hmmmmm, seems you can create/move terminals as/to regular tabs, which can include the type in the title

image

image

I wonder if this might be a "better" approach, similar to the find window recommendation

pokey:
I'm not sure whether @knausj85's comment is requesting changes or not, but I do still think it's worth thinking about my suggestion in https://github.com/talonhub/community/pull/1395#issuecomment-2041452269. Ie it seems a bit odd that every user should have kubectl commands active?

jaresty:
removed the extra tags.

pokey (approved):
I'm happy, but let's wait for @knausj85 to see if there's some action expected from https://github.com/talonhub/community/pull/1395#issuecomment-2041580824

And if we don't already have it, we should have a guide for how to get terminal working, because if I'm understanding correctly, they need to put a file somewhere that actually activates the tags like `git` etc when you're in `tag: terminal`. Once we have that, we should link to it from the vscode terminal readme section

jaresty:
I tested out [this example](https://github.com/jaresty/mac-dotfiles/blob/02b9e7d83ea4a209d42afa72ebddae3ca545be44/talon-settings/terminal.talon#L3-L6) and it works. I'm not sure there's an example already in community though.

jaresty:
this might be a dumb question but, can you tell me where the community docs are?

jaresty:
I added some more documentation to the VSCode README.

apps/vscode/README.md
```diff
@@ -5,3 +5,26 @@ It is recommended to install the [VSCode talon extension pack](https://marketpla
...
+In order to enable additional terminal commands you will need to set some tags when the terminal tag is active. You can do this by creating a file in your talon settings that looks something like this:
+
+```
+tag: terminal
+-
+tag(): user.generic_unix_shell
+tag(): user.git
+tag(): user.kubectl
+tag(): user.readline
+```
```
pokey, line 30: This is nice. We should eventually move this somewhere more central and link to it from here, because it applies to any terminal emulator app, but for now I'm happy to have it here

fidgetingbits:
I think this probably good as is, as I agree it's not clear what exactly to change atm.

I have some stuff I'll add to community soonish that can leverage the cwd stuff knausj screenshot shows, and can add parsing for that then. 

Could file an issue to follow up to see if there is a way to differentiate the underlying shell (cmd.exe, powershell, etc) when the terminal is moved to a different tab like what knausj85 suggested. I do something similar in neovim terminals, and it's useful, but I don't think it's required for this PR. Also knausj mentioned recently he'll be gone for awhile, so probably not worth waiting on his response here.

PR #1397, opened 2024-03-07
For vscode, define the filename as the first window title component containing a dot.

bjaspan:
The VS Code title format is configurable and does not always contain the filename as the first component. This change preserves existing behavior when the filename is the first component but also supports it in other positions.

### DIFF
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -137,14 +137,10 @@ class WinActions:
         # if title == "":
         #    title = ui.active_window().doc
 
-        if is_mac:
-            result = title.split(" — ")[0]
-        else:
-            result = title.split(" - ")[0]
-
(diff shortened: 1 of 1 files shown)


nriley:
I'm not sure this is really soluble in the way you're trying to do it. Sometimes there are filenames that do not contain a dot at all (a common example in code is LICENSE). Perhaps you can implement this specifically to the VSCode fork that you were using or perhaps there is a way to read the setting from VSCode and interpret it? window.doc will also work on the Mac at least in the desktop version of VSCode — not sure if we are using that yet.

consequentlyopposite:
I agree. I'm going in a different direction; I thought I had closed this already.

bjaspan:
I agree. I'm going in a different direction. I thought I had closed this already.

PR #1399, opened 2024-03-12
Pop twice to repeat

Spiteless:
Add optional tag for pop_twice_to_repeat

Someone suggested this and I thought I'd try it out. It's been working in my local repo for awhile in a slightly different configuration.

## Questions

- Added the repeat file to `/plugin/repeat/`, if it should live somewhere else let me know
- Added enable setting to `settings.talon`, if it should live somewhere else let me know


pokey (commented):
this looks good, tho

- I haven't tested it locally, and
- It doesn't support mapping something else to single pop. Tho that's probably fine? cc/ @AndreasArvidsson

AndreasArvidsson:
Yes assuming that you don't want pop for anything else like clicking this implementation is fine.

Spiteless:
If a user wants to use `pop` noises for other things as well as a repeat what would need to change to make that happen?

I don't use an eye tracker, so pop to wake/pop to repeat have been my only uses for the sound thusfar.

pokey:
I think you'd need to set a timeout on the first pop, and if it fires you run the single-pop action. If the second pop comes before timeout fires, you cancel the timeout

It would be a bit annoying because it would introduce a delay, but I don't think there's any other way to do it. I'd be tempted to leave it out for now

Spiteless:
Closing as duplicate

PR #1401, opened 2024-03-15
stata support

maxbruening:
Adds support for the stata language and application.


knausj85:
I don't know the first thing about stata, so I will leave this open for a bit for other maintainers to take a peek

lang/stata/stata.py
```diff
@@ -0,0 +1,170 @@
...
+
+# functions_common.py
+ctx.lists["user.code_common_function"] = {
+    # base stata
+    "global": "global",
+    "local": "local",
+    "reg": "reg",
+    "regress": "reg",
+    # packages
+    "estadd": "estadd",
```
fidgetingbits, line 24: Just want to confirm you pronounce these literally "estadd" and not "E S T add". If the latter, should change the wording to spell it out.

lang/stata/stata.py
```diff
@@ -0,0 +1,170 @@
...
+ctx = Context()
+
+ctx.matches = r"""
+code.language: stata
+"""
+
+# functions.py
+ctx.lists["user.code_parameter_name"] = {
+    # regressions
+    "v c e cluster": "vce(cluster)",
```
fidgetingbits, line 12: I could be wrong, but I think usually for single letters we usually use capitals, although I'm not sure it actually matters. Can just ignore this suggestion if it works ok as is.

lang/stata/stata.py
```diff
@@ -0,0 +1,170 @@
...
+    "estout": "estout",
+}
+
+
+@ctx.action_class("user")
+class UserActions:
+    # comment_line.py
+    def code_comment_line_prefix():
+        actions.auto_insert("* ")
+        # actions.auto_insert("// ")
```
fidgetingbits, line 42: Can delete this comment?

lang/stata/stata.py
```diff
@@ -0,0 +1,170 @@
...
+    def code_insert_function(text: str, selection: str):
+        text += f" {selection or ''}"
+        actions.user.paste(text)
+
+    # imperative.py
+    def code_block():
+        actions.auto_insert("\n")
+
+    def code_state_if():
+        actions.insert("if  {")
```
fidgetingbits, line 71: Is this intentionally two spaces? It seems like below you use two after `if`'s, `else`'s, `foreach`, but one after `do {` and `for {`.

lang/stata/stata.py
```diff
@@ -0,0 +1,170 @@
...
+        actions.key("enter tab enter left")
+        actions.insert("}")
+        actions.key("up:2 right:2")
+
+    def code_state_else_if():
+        actions.insert("else if  {")
```
fidgetingbits, line 77: two spaces here

lang/stata/stata.py
```diff
@@ -0,0 +1,170 @@
...
+        actions.key("enter tab enter left")
+        actions.insert("}")
+        actions.key("up:2 right:7")
+
+    def code_state_else():
+        actions.insert("else  {")
```
fidgetingbits, line 83: two spaces here

lang/stata/stata.py
```diff
@@ -0,0 +1,170 @@
...
+        actions.key("up tab")
+
+    def code_state_for():
+        actions.insert("for () {")
+        actions.key("enter tab enter left")
+        actions.insert("}")
+        actions.key("up:2 right:4")
+
+    def code_state_for_each():
+        actions.insert("foreach  in {")
```
fidgetingbits, line 101: two spaces here

fidgetingbits (changes requested):
Looks good, couple small suggestions/questions

lang/stata/stata.py
maxbruening: I removed this line. There are multiple ways to comment out a line in stata

lang/stata/stata.py
maxbruening: I reworked the statements. They are now supposed to have two spaces so that after inserting a condition one does not need to add an extra space. With this reasoning the else statement does not need to spaces.

maxbruening:
Thank you for your feedback. I hope I have addressed your comments sufficiently.

lang/stata/stata.py
fidgetingbits: Thanks. I don't think there is a good way atm to supoort both, you'd probably have to use a setting. But probably easier to just have a linter automatically change them. 

fwiw if one of them is for multi-line comments (like C style /* ... */) there is a separate comment action called code_comment_block you could use, but it looks like it's probably not the case here.

apps/stata/stata.talon
```diff
@@ -0,0 +1,7 @@
+# Commands for the Stata Do-File Editor
+os: windows
+app: stata
```
fidgetingbits, line 3: this needs an and before app, otherwise it is treated they are ored and would be active on all windows systems

apps/stata/stata.talon
```diff
@@ -0,0 +1,7 @@
...
+-
+settings():
+    key_wait = 4
```
fidgetingbits, line 6: what's the deal with these delays out of curiosity? is the Stata editor particularly slow on windows? I think it's ok unless someone complains eventually, but since there is another issue in this file I figured I'd ask.

apps/stata/stata_do_file_editor.talon
```diff
@@ -0,0 +1,25 @@
+# Commands for the Stata Do-File Editor
+os: windows
+app: stata
```
fidgetingbits, line 3: app and win.title will need and before them here as well

lang/stata/stata.py
maxbruening: Stata uses the default c-style comment! That should already be activated.

Most stata users use `*`  for comments at the start of a line, `//`  for comments at the end of a line and `/*...*/`  for long multiline comments.  Note that  can also be used at the start of a line whereas `*`  cannot be used at the end of a line.

lang/stata/stata.py
```diff
@@ -0,0 +1,157 @@
...
+        actions.insert("else if  {\n\n}\n")
+        actions.key("up:2 tab up right:4")
+
+    def code_state_else():
+        actions.insert("else {\n\n}\n")
+        actions.key("up:2 tab")
+
+    def code_state_for():
+        actions.insert("forval  {\n\n}\n")
+        actions.key("up:2 tab up right:3")
```
fidgetingbits, line 83: is it possible people would/could configure the stada app to not use 4 space tabs? if so I guess this would break. same goes for the other spots you added tab

fidgetingbits (changes requested):
Thanks, noticed a couple more small things.

apps/stata/stata.talon
maxbruening: Are you sure about this? From https://talon.wiki/Customization/talon-files: "Requirements of different types (e.g. 'app' and 'os') are AND-ed together."

 Using the debug window, the stata commands do not seem to be active when I'm not focusing a stata window.

fidgetingbits: Ya sorry you are right, confused myself last night because I happened to be looking at https://github.com/talonhub/community/blob/main/core/application_matches.py around the same time and it uses `and` everywhere, which I guess is all redundant lol.

apps/stata/stata_do_file_editor.talon
maxbruening: Please see my comment above. I believe Talon "ANDs" these.

fidgetingbits: Yep, marked as resolved. fwiw I just filed https://github.com/talonhub/community/issues/1452

apps/stata/stata.talon
maxbruening: You are right, these do not seem to be strictly necessary for the pull request. I believe I added them because I have some extra commands for stata that may run into trouble otherwise. Since these are not part of the pull request, there  is no need for the file at all.

 I would suggest to keep calling the other file "stata_do_file_editor" because those commands target the "Do-File Editor" of Stata's IDE, not the command window/terminal. The Stata IDE is similar to R Studio or Spyder for python.

apps/stata/stata_do_file_editor.talon
maxbruening: However, I noticed that  for stata 15 the header did not match and improved it accordingly. Newer versions of stata  seem to match.

lang/stata/stata.py
maxbruening: Per default stata inserts 4 spaces per tab. However, this can be configured in the editor. But I guess if someone configured this, talon actually does press tab and this would insert the modified number of spaces, correct?

What would be the alternative? Not adding the tab? `\t`?

fidgetingbits: Ya, that's my worry if someone configures it differently than the default then it will be wrong. I think the only option is to just manually press the keys unfortunately, which will be a bit slower for the user, but at least correct. I believe `\t` will be the same as pressing tab.

fidgetingbits (changes requested):
Thanks. Looking better and better. I also just realized that you are using `app: stata` in a few places but I don't see anything like:

```
mod = Module()
apps = mod.apps
apps.stata = """
...
"""
```

anywhere. I just double checked community and it doesn't seem like there is anything there yet. Do you have that somewhere else in you repo and forgot to add it?

lang/stata/stata.py
maxbruening: Ah, Now I understand.  Will make these changes either tonight or tomorrow.

fidgetingbits:
> I indeed do not have anything like
> 
> ```
>  mod = Module()
> apps = mod.apps
> apps.stata = """
> ...
> """
> ```
> 
> nonetheless, `app.app` Seems to be defined. I guess talon partly automates this? Nevertheless, it may be more robust to manually define this as well?

Huh. No talon doesn't automate it afaik. In the [wiki](https://talon.wiki/Customization/talon-files) under `app` in the table it also notes it must be explicitly declared. It seems like there must be some declaration you have some where, maybe in a separate repo under your talon user folder or something? Otherwise I'm not sure how it would work... 

Anyway, normally there is an explicit app declared. Can just put it in the `app/stata/stata.py` file. Some generic ones are also in `core/application_matches.py` in community, or you can see other stuff in `apps/` for examples.

maxbruening:
So apparently talon does  assign `app.app`  but it is still best practiced to also explicitly declare it. Thus, I added it.

fidgetingbits:
Interesting. I guess I need to test. So if no app is explicitly defined, if you do include a app: it defaults to try to match against app.app? That should get documented on the wiki.

lang/stata/stata.py
maxbruening: I have made the according changes.  The code should now be robust to how many spaces are defined as a tab.

 However, if one works in an editor with auto closing brackets on, one would still run into problems. But I guess this affects all languages in talon.  one could create something like context sensitive dictation I guess. But this goes beyond the scope of this pull request.

fidgetingbits (approved):

fidgetingbits:
Thanks for all the changes and the contribution. Looks good to me now. I'll let it bake for the day in case anyone else pipes in, but otherwise will merge tomorrow morning.

PR #1402, opened 2024-03-18
Meld

maxbruening:
Add basic support for the Meld diff editor on windows.


apps/meld/meld.talon
```diff
@@ -0,0 +1,6 @@
+app: meld
+-
+tag(): user.tabs
+
+change next: key(alt-up)
```
knausj85, line 5: I believe these shortcuts are swapped

apps/meld/meld.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def tab_reopen():
+        print("Meld does not support this action.")
+
+
+@ctx.action_class("user")
+class UserActions:
+    def tab_jump(number):
+        if number < 10:
+            actions.key(f"alt-{number}")
```
knausj85, line 38: This doesn't appear to work, though the shortcut appears to be correct

maxbruening: It works for me on windows 11, But it seems like Meld implements this keyboard shortcut in an unfortunate way:  it seems to work only if the current tab is selected, for example by clicking on the current tab.

maxbruening: See here [https://gitlab.gnome.org/GNOME/meld/-/issues/839#note_2077059]()

knausj85 (approved):

PR #1403, opened 2024-03-18
Support for Foxit Reader for PDFs

maxbruening:
Adds basic support for the foxit reader on windows.


knausj85 (approved):
Looks good, thank you

PR #1404, opened 2024-03-20
Creates the "find" tag and removes that functionality from "find_and_replace"

maxbruening:
The idea is to have a separate "find" tag for very basic search functionality. The reason is that there are many types of applications that are mostly view only (webbrowsers, pdf readers, etc.). These applications would not benefit from the full functionality of the "find_and_replace" tag.

In order not to define some voice commands twice, I made the "find_and_replace" tag automatically activate the new "find" tag as well and removed what would have been duplicate definitions.


maxbruening:
See  [PR 1405](https://github.com/talonhub/community/pull/1405) for an example of how this new "find"  tag may be used for implementing basic search functionality in webbrowsers. In comparison, the full "find_and_replace" tag would create many  voice commands that are impossible to implement in webbrowsers.

tags/find/find.talon
```diff
@@ -0,0 +1,10 @@
+tag: user.find
+-
+hunt this: user.find("")
+hunt this (pace | paste):
+    user.find("")
+    sleep(25ms)
+    edit.paste()
+hunt this <user.text>: user.find(text)
+hunt next: user.find_next()
```
knausj85, line 9: I see these commands exist  (hunt next/previous) in find and replace too. Those duplicates should be removed.

(I can't comment on the correct file 'cause they didn't change)

knausj85 (changes requested):
This makes sense to me, thanks. One minor fix and I think it's good to go.

tags/find/find.talon
maxbruening: The latest commit should correct this.

knausj85 (approved):
Looks good, thanks. I'll leave it open for a bit for others to chime in.

fidgetingbits:
Makes sense and looks good to me. As no one else has chimed in I'll merge this.

PR #1405, opened 2024-03-20
Implements "find" tag for webbrowsers

maxbruening:
The goal of this PR is to add basic search functionality for webbrowsers based on the new "find" tag from the other already merged [PR 1404](https://github.com/talonhub/community/pull/1404).  The code has been tested on windows, but not on mac or Linux.

 I left a commented out version for `def find()` in the PR as this alternative version may be more robust.

I checked and tested that the implemented keyboard actions work in windows for chrome, firefox, opera, edge, and brave.


AndreasArvidsson:
I don't think there are any shortcuts there are general browser. The short cuts are gonna be different depending on platform and application and should be defined in the application context. We should also not use `user.find` when there is `edit.find`. We should definitely just migrate everything to the edit actions.

AndreasArvidsson (changes requested):
I don't think there are any shortcuts that are general enough for all browsers. The shortcuts are gonna be different depending on platform and application and should be defined in the application context. We should also not use `user.find` when there is `edit.find`. We should definitely just migrate everything to the edit actions.

AndreasArvidsson:
Here is a pull request to replace `user.find` with `edit.find`
https://github.com/talonhub/community/pull/1499

maxbruening:
Alright, happy to implement those changes. 
Before I do so: Are we sure that we want to implement the python functions on a per browser basis? The keyboard presses are the same for chrome, firefox, opera, edge, and brave. Given this, wouldn't it be simpler to define those functions once and then if another browser has different shortcuts, the python file for that browser should redefine them?

AndreasArvidsson:
Most of the work is already done in https://github.com/talonhub/community/pull/1499
 I don't think we should put any key bindings under browser it's probably operating system specific in that case. In my pr above I have already added the default operating system shortcuts.

maxbruening:
browser_shortcuts

knausj85 (dismissed):

knausj85:
From a quick test on a number of browsers on both windows and macOS, I believe all we need to do here is enable the tag in browser.talon. Re-requested review for sanity

AndreasArvidsson (approved):

Issue #1406, opened 2024-03-26
Kotlin support


jaresty:
I'm having to do Kotlin lately. I would really appreciate if community could add support for the Kotlin language. There is some existing work in one of the community forks [here](https://github.com/okonomichiyaki/knausj_talon/blob/06ebde9f11e8738f5f82a7c2ff419541f068d769/lang/kotlin/kotlin.py)

auscompgeek:
Was enough support for Kotlin added in #1408 for this to be considered done, or did you have more in mind @jaresty?

jaresty:
Good question. I've got an open PR to add spacing on keywords. I'd like to get that merged and then I can close this issue.

PR #1408, opened 2024-03-27
Add basic kotlin support

jaresty:
- Borrowed from [here](https://github.com/okonomichiyaki/knausj_talon/commit/6542dc38e43b591c2ca2ea1d1360ac198e6eb168)

Addresses https://github.com/talonhub/community/issues/1406


lang/kotlin/kotlin.talon
```diff
@@ -0,0 +1,27 @@
+tag: user.kotlin
```
auscompgeek, line 1: This seems to predate #1256.

```suggestion
code.language: kotlin
```

lang/kotlin/kotlin.py
```diff
@@ -0,0 +1,182 @@
...
+class UserActions:
+    def code_comment_line_prefix():
+        actions.insert("// ")
+
+    def code_operator_indirection():
+        actions.skip()
+
+    def code_operator_address_of():
+        actions.skip()
+
```
auscompgeek, line 19: Actions shouldn't be implemented if they're not implementable. This suppresses the warning about the action being unimplemented when used. We don't `actions.skip()` anywhere else in `lang/`.

```suggestion
```

lang/kotlin/kotlin.py
```diff
@@ -0,0 +1,182 @@
...
+        actions.key("left")
+
+    def code_state_else():
+        actions.insert("else ")
+        actions.key("enter")
+
+    def code_state_switch():
+        actions.insert("switch () ")
+        actions.key("left")
+        actions.edit.left()
```
auscompgeek, line 137: All of these which put the cursor between the `()` should use `actions.user.insert_between()`.

lang/kotlin/kotlin.talon
jaresty: Fixed! Thanks.

lang/kotlin/kotlin.py
jaresty: Removed.

lang/kotlin/kotlin.py
jaresty: Updated these.

auscompgeek:
Possibly silly question: how similar is Kotlin to Java? Would it possibly make sense for our Kotlin support to extend our Java support?

jaresty:
Kotlin is pretty different from Java when you look at it closely. I don't think we'd want to borrow syntax from Java most of the time. For example, Kotlin does not require semicolons and Kotlin is more functional than Java. I think it's better to start with something independent at first.

lang/kotlin/kotlin.py
```diff
@@ -0,0 +1,166 @@
...
+class UserActions:
+    def code_comment_line_prefix():
+        actions.insert("// ")
+
+    def code_operator_lambda():
+        actions.auto_insert(" -> ")
+
+    def code_operator_subscript():
+        actions.insert("[]")
+        actions.key("left")
```
okonomichiyaki, line 19: should this also be using `actions.user.insert_between` ?

okonomichiyaki:
thanks @jaresty for shepherding this into the community repo, and to auscompgeek for comments. I am neither a Kotlin expert nor up to date on the Talon API and patterns. have only used these files for a little Kotlin programming over the past year and was enough for basic stuff but I'm sure can be improved to support more of the language

lang/kotlin/kotlin.talon
```diff
@@ -0,0 +1,27 @@
...
+settings():
+    user.code_private_function_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_protected_function_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_public_function_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_private_variable_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_protected_variable_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_public_variable_formatter = "PRIVATE_CAMEL_CASE"
+
+state var: "var "
+state val: "val "
```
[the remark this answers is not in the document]
jaresty: This sounds like a good idea. I borrowed the modifiers from java but I'm not sure they're correct.

lang/kotlin/kotlin.py
jaresty: Oops I missed that one. Fixed now.

lang/kotlin/kotlin.py
```diff
@@ -0,0 +1,186 @@
...
+    "synchronized": "synchronized",
+    "volatile": "volatile",
+    "transient": "transient",
+    "abstract": "abstract",
+    "interface": "interface",
+    "final": "final",
+}
+
+mod.list("kotlin_modifier", desc="Kotlin Modifiers")
+ctx.lists["self.kotlin_modifier"] = kotlin_modifiers
```
knausj85, line 27: I think we should probably modernize this to use the code_keywords stuff? See JavaScript for an example 

https://github.com/talonhub/community/blob/main/lang/tags/keywords.talon

knausj85 (changes requested):

knausj85 (approved):

PR #1411, opened 2024-04-01
jetbrains: reduce log output, better error output

timo:
until now, finding the vcidea_foo file still resulted in an error that the .vcidea_foo file was not found, which is confusing. also, there's always many lines per command executed, with this patch it's just one when it works.
If there's an error with the vcidea nonce file, or if the port isn't found, there will now be a hint that vscode-idea commands will not work, and when the vcidea file is not found, the actual locations we looked will be put into the log message as well, so you don't have to guess or investigate with the REPL or something

### DIFF
--- a/apps/jetbrains/jetbrains.py
+++ b/apps/jetbrains/jetbrains.py
@@ -76,11 +76,15 @@ def _get_nonce(port):
                     errors_encountered.append(e)
     if nonce is None:
         if any_file_existed:
-            print("could not read voicecode-idea nonce file! jetbrains commands will not work!")
+            print(
+                "could not read voicecode-idea nonce file! jetbrains commands will not work!"
+            )
             for e in errors_encountered:
                 print("  ", e)
         else:
-            print(f"Could not find any voicecode-idea nonce files, .vcidea_{port} or vcidea_{port} in {' or '.join(locations)}! jetbrains commands will not work!")
+            print(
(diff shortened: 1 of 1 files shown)


timo:
```
2024-04-01 20:28:03.909    IO executing jetbrains action EditorCopy
2024-04-01 20:28:03.910    IO Sending action EditorCopy
2024-04-01 20:28:03.911    IO Could not find .vcidea_8654 in tmp or home
2024-04-01 20:28:03.912    IO sending IntelliJ IDEA 8654 blablablabla-blablabla=
```
this is what executing a single command looks like in the log before this patch

timo:
here's real log output from a real user where four commands are sent to the idea plugin
```
2024-04-15 19:05:30.743    IO sending voicecode-idea command action IntroduceField to IntelliJ IDEA 8654 blablablabla-blablabla=
2024-04-15 19:05:42.091    IO sending voicecode-idea command action FindUsages to IntelliJ IDEA 8654 blablablabla-blablabla=2024-04-15 19:05:52.773    IO sending voicecode-idea command action SearchEverywhere to IntelliJ IDEA 8654 blablablabla-blablabla=
2024-04-15 19:06:09.357    IO sending voicecode-idea command action SmartTypeCompletion to IntelliJ IDEA 8654 blablablabla-blablabla=
```

knausj85:
Since https://github.com/talonhub/community/pull/1445 was already approved and substantially similar, we opted to go ahead and merge it. I'm going to close this out.

PR #897, opened 2022-06-23
Place Talon and Python files side-by-side

wenkokke:
Opening this pull request to have a working example for discussion about #669.

My attempt with this restructuring is to *only* focus on renaming files, to ease future merges.

I've not updated any of the imports or file paths, as I mostly intend for this to fuel discussion.

My rationale for organizing is:
- `apps`: concrete commands for an application
- `core`: concrete commands and actions that are always available and any other code can depend on
- `lang`: concrete commands for a programming language
- `tags`: concrete commands with some abstract actions, activated with a tag
- `plugins`: a self-contained set of commands not linked to any application or programming language but whose actions probably should not be depended upon

My aim with grouping each pair of Python and Talon files in their own subdirectory is (1) to make it extremely clear which files belong together, and (2) to encourage some modularisation. As a bonus, these subdirectories can contain a `README.md`, which can really help someone browsing the repository on GitHub.

PR #1416, opened 2024-04-13
Move all talon files into a different directory

jaresty:
The idea behind this is that, since the python files should change less often than the talon files, we might be able to move all of the talon files into their own repository such that most users could share the same python base files and only fork the talon file repository if they wanted to customize their commands.

I think this aligns with Sandy Metz's philosophy to "[depend on things that change less often than you do](https://arc.net/l/quote/rpkmjabx)"


pokey:
I'm a bit hesitant about this direction. I am an advocate of related things being close together. When we have a proper package manager in talon, my understanding is that packages will be able to have their own talon files, so I'm not sure I see how this PR gets us better prepared for when we have a package manager

Fwiw, we actually went in the exact opposite direction two years ago, and I think the reasons for that move still hold today (https://github.com/talonhub/community/pull/897)

jaresty:
I think we've learned that people have reasons for customizing their spoken vocabulary away from the stock commands. This would be the same even if these were split into separate packages. While I also like as a general rule to keep related things together, I could envision a this implementation-vocabulary split still working in a package manager driven system. Imagine this: there is a base package that defines the implementation (Python), and there are a few alternative packages that can be chosen from for the spoken forms (long form, terse, etc). This would allow people who are interested to share their custom vocabularies without having to make the Python code tied to also defining the spoken forms.

nriley:
> I'm a bit hesitant about this direction. I am an advocate of related things being close together. When we have a proper package manager in talon, my understanding is that packages will be able to have their own talon files, so I'm not sure I see how this PR gets us better prepared for when we have a package manager
> 
> Fwiw, we actually went in the exact opposite direction two years ago, and I think the reasons for that move still hold today (#897)

I agree with this and would prefer the strategy proposed in #1418.

jaresty:
That works for me.

PR #1419, opened 2024-04-17
context-sensitive dictation: default to not inserting space

rntz:
Currently `dictation_peek` in context-sensitive dictation inserts a space before selecting text around point. This avoids ever copying with an empty selection, which misbehaves in some applications. However:

1. This makes the whole dance noticeably slower on my machine.
2. In applications where copying an empty selection misbehaves, many of our other commands also fail to work (eg: `select line start` tries to tell whether there is currently a selection to determine whether to cancel it); working around this everywhere would kind of be a nightmare, so I tentatively think it makes more sense to have "fix your editor" be our policy on this problem.
3. In some contexts (eg: the google.com search bar), inserting a space doesn't work consistently because google "helpfully" avoids inserting a space if there already is one.

This PR defaults to *not* inserting anything, but provides a configurable setting, `user.context_sensitive_dictation_dummy_string`, for what dummy string to insert. This makes the dictation dance faster by default but allows users to choose the dummy string behavior if they want, and also to choose a dummy string that will work even in the google address bar.

### DIFF
--- a/core/text/text_and_dictation.py
+++ b/core/text/text_and_dictation.py
@@ -442,13 +442,13 @@ class Actions:
             if before:
                 # Unfortunately, in web Slack, if our selection ends at newline, this
                 # will go right over the newline. Argh.
-                actions.edit.right() # cancel selection
+                actions.edit.right()  # cancel selection
 
         if not right:
-            actions.key(f"backspace:{len(dummy)}") # remove the dummy
+            actions.key(f"backspace:{len(dummy)}")  # remove the dummy
         else:
             for _ in dummy:
-                actions.edit.left()         # go left before dummy
+                actions.edit.left()  # go left before dummy
             # We want to select at least two characters to the right, plus the dummy we
             # inserted, because no_space_before needs two characters in the worst case
             # -- for example, inserting before "' hello" we don't want to add space,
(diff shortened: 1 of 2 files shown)


rntz:
HOLD UP! Don't merge this. I'm worried that its behavior may be problematic if there is already a text selection. Ugh.

maxbruening:
I really like the idea of further improving context sensitive dictation. Generally it might be a good idea to customize the peek behavior on a per application basis?!  in particular, the current version does not work in Microsoft word on windows 11 for me.

Could you just check whether there is a text selection?  first copy,  then if string is empty...  
 
 might increase latency though

rntz:
@maxbruening good idea about checking what the selection is, but yeah, that adds back latency that was otherwise gained. But I don't see any other good way to make this behave correctly if there's already a selection. The standard trick is to do "space backspace" to delete the current selection, but that messes up undo history just like the dummy-space approach, and is more keypresses to boot.

So there's really only two sensible modes here:

1. Insert a dummy string, which can screw up undo history but works quite reliably even when copying an empty selection doesn't work.
2. Don't insert a dummy string, which preserves undo history but will frequently copy with an empty selection.

Neither is particularly faster than the other. I'm no longer sure which should be the default.

rntz:
Ugh, I just ran into trouble with empty selections right here on GitHub. If the cursor is at the end of the text box with no selection and you extend_word_right (alt-shift-right), it puts the cursor into a weird state where inserting more characters does nothing but copying produces an empty string. The current approach of inserting a dummy space avoids this problem because it avoids ever having an empty selection.

I think probably the current behavior should remain the default.

PR #1420, opened 2024-04-19
define "zoom reset" for VSCode

maxbruening:
The default value for "zoom reset"  defined in edit_win.py is "ctrl-0".  This does not work in VS Code. The PR  attempts to fix this.


maxbruening:
The keyboard shortcut in VS Code is "ctrl-numpad_0". An alternative solution may thus be to change the default keyboard shortcut in edit_win.py.  but that  may run the risk of breaking things elsewhere.

knausj85 (approved):

PR #1421, opened 2024-04-21
enable the "user.command_client" tag for "neovim" app

saidelike:
Pushing this file only to allow others to beta test the already packaged cursorless for neovim from https://github.com/hands-free-vim/talon.nvim/tree/beta without having to add the file manually.

Note that the app is "neovim" so it won't trigger any side effect with community which uses the "vim" app only. And the "neovim" app is only defined in https://github.com/hands-free-vim/neovim-talon so it is safe to merge now imho.

NOTE: the apps/vscode/command-client might have to move outside of the "vscode" app since it is now for other apps but it can be moved in a later PR.

### DIFF
--- a/apps/vscode/command_client/neovim.py
+++ b/apps/vscode/command_client/neovim.py
@@ -1,0 +1,18 @@
+from talon import Context, actions
+
+ctx = Context()
+
+ctx.matches = r"""
+app: neovim
+"""
+
+ctx.tags = ["user.command_client"]
+
+
+@ctx.action_class("user")
+class VimActions:
+    def command_server_directory() -> str:
+        return "neovim-command-server"
+
+    def trigger_command_server_command_execution():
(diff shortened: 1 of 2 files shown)


apps/vscode/command_client/neovim.py
```diff
@@ -0,0 +1,18 @@
+from talon import Context, actions
+
+ctx = Context()
+
+ctx.matches = r"""
+app: neovim
```
knausj85, line 6: Is this app already defined somewhere that I missed?

pokey:
@saidelike I think the plan is to close this PR and just fold it into talon.nvim, where the `neovim` app is currently defined. Correct?

apps/vscode/command_client/neovim.py
saidelike: It is only defined in neovim-talon atm.

PR #1422, opened 2024-04-22
micro-optimization for dictation_insert

rntz:
We avoid needing to peek left if both:

1. The string being inserted never needs space before it (eg: it is a period).
2. The string being inserted does not change under auto-capitalization.

This PR extends condition (2) to include the case where `auto_cap = False` has been passed, so we are not doing auto-capitalization at all.


pokey (approved):
Seems reasonable to me, tho I haven't tested it locally and I can't claim to be completely familiar with the formatter code

PR #1423, opened 2024-04-25
Fix upper case executable name

tankorsmash:
A change somewhere along the way made paths consistently lowercase on Windows, so this allows Windows Terminal to be once again detected!


auscompgeek (changes requested):
This needs to be case-insensitive to be compatible with current public Talon.

tankorsmash:
Could you clarify what change you'd like to see? Is there a way I can make this work with the public version without leaving it broken for the beta branch?

If relevant, Windows can run `GiT.exe` `NvIM.exe` etc without issue!

nriley:
Sorry, thought I had committed this change, but apparently not. You can see backwards-compatible fixes in https://github.com/nriley/talon_community/commit/a274e1224b00522f0165f928e2545b266684a656..

nriley:
Several months have gone by and I still failed to merge my changes. I will do that today.

PR #1424, opened 2024-04-28
Add helper to go to the middle of a line

jaresty:
Adds a line command `go mid` which is similar to vim's `gm` command and moves the cursor to the middle of the line.


core/edit/edit.talon
```diff
@@ -185,3 +185,5 @@ redo that: edit.redo()
 # Save
 file save: edit.save()
 file save all: edit.save_all()
+
+go mid: user.go_mid()
```
fidgetingbits, line 189: There is "go top" and "go bottom" which are file commands, so "go mid" could be interpreted as file or line in. The existing line commands are "go line start | head" and "go line start | tail", so I would be tempted to make this "go line mid"? Suppose you could do "go line mid | mid"? However, my only hesitation with using "mid" on its own is that there is https://github.com/talonhub/community/blob/5ebcea7e662f86eb076ac595fc7d58e242fb6822/plugin/mouse/mouse.talon#L23, "mid click"... I've used "click" as an alternative for "touch" before, so this would then conflict with "mid click".

fidgetingbits (changes requested):
Good idea. Mostly ok with it, but a couple small suggestions.

core/edit/edit.talon
jaresty: What do you think about making "go" optional? `[go] line mid`?

fidgetingbits: Ya I think that'd be fine

jaresty: updated - thanks for the feedback!

fidgetingbits (approved):

PR #1425, opened 2024-04-29
Adds additional helpers for refactoring to intellij

jaresty:
this pull request adds commands to go to a line and search for some text and immediately perform an action on that. it makes it very simple to operate on words that you can see without having access to cursorless. This is very helpful in intellij since you don't have cursorless. It also expands on existing refactoring commands by adding specific ones for things like renaming and quick fixing.


fidgetingbits (approved):

phillco (approved):
lgtm, it would be cool if more of these were behind a generic tag across multiple editors but since this pattern matches the existing code there's no need to do that in this change.

Curious about the DFA impact on `<number> <user.text>` since they are both large captures, but we can address that afterwards if it ends up being a problem. Can't assume relative line numbers so we need the larger number capture, and text is useful because IntelliJ will fuzzy match things like came case.

PR #1427, opened 2024-05-01
Allow formatting of abbreviations

jaresty:
- All abbreviations are currently lowercase
- It is useful to be able to add them with formatting
- This makes it possible


knausj85:
Just FYI, the casing of abbreviations is wholly determined by the contents of the csv at the moment. If you define eg. 

TEST,stupid test

it'll output TEST.

My only real concern is a lot of the formatters don't really make sense for abbreviations (smash, camel, snake, dotted etc) at first glance, so this may end up defining a lot of potentially questionable commands that might result in more conflicts?

jaresty:
I'm aware that the casing is determined by the csv. The current default csv includes many abbreviations that are traditionally cased differently, so this seems like it was intended to be used this way? Consider "wip". I have the need to use abbreviations with different casing for variable names in code, so I would like this to work.

fidgetingbits:
I think this is possibly duplicate (and more limited) behaviour of PR #636 which looks like it could be pushed along instead? fwiw I haven't tested either PR to confirm, as I'm just poking around on mobile, so maybe missing something.

jaresty:
🤔 maybe so. Unfortunately it seems that that pull request has conflicts now, so I'm not really sure how to push it forward.

fidgetingbits:
> 🤔 maybe so. Unfortunately it seems that that pull request has conflicts now, so I'm not really sure how to push it forward.

np. I'll take a look to see if I can sort it out.

AndreasArvidsson:
We discussed this in the community meet up. You should probably use the capture `<user.formatters>` instead of the list. Also we feel that we should probably move abbreviation and this new command to `text.talon`.

nriley (approved):

PR #1428, opened 2024-05-03
Fix window maximize not being preserved when moving screen

reecpj:
Now windows maximized on the source screen get properly maximized on the dest screen


knausj85 (approved):

knausj85:
This works well as-is on windows, thanks! MacOS uses a separate property called fullscreen, but it seems we can't programmatically set the window rect when in fullscreen. 

As such, I am going to merge this as-is.

PR #1429, opened 2024-05-06
Add space after Kotlin keywords so that you can chain in keyword commands

jaresty:


lang/kotlin/kotlin.py
```diff
@@ -8,19 +8,19 @@
...
+    "public ": "public",
+    "private ": "private",
+    "protected ": "protected",
+    "static ": "static",
+    "synchronized ": "synchronized",
+    "volatile ": "volatile",
+    "transient ": "transient",
+    "abstract ": "abstract",
+    "interface ": "interface",
+    "final ": "final",
```
auscompgeek, line 23: These are on the wrong side: the keys are the spoken phrase, the values will be what's inserted.

auscompgeek (changes requested):

lang/kotlin/kotlin.py
jaresty: Thanks for pointing that out. Should be fixed now.

auscompgeek (approved):
Looks good. I believe this was the intended way to use this list.

jaresty:
can someone help me with merging this? I don't have write access so I'm unable to merge it on my own.

PR #1430, opened 2024-05-07
Update .pre-commit-config.yaml to fix precommit error

C-Loftus:
Talonfmt requires Python 3.11 to be installed. This line fixes the precommit error and should make it pass. 

I tested this on the ai-tools repo and it passes now
https://github.com/C-Loftus/talon-ai-tools/blob/683638d5aa3a0ad556e747947e9909039ec3385e/.pre-commit-config.yaml#L53

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,3 +50,4 @@ repos:
     hooks:
       - id: talonfmt
         args: ["--in-place"]
+        language_version: python3.11


phillco (approved):
@wenkokke is it possible to make this read from pyproject.toml, the way we do for black?
https://github.com/talonhub/community/blob/main/pyproject.toml#L1-L2

I think we've moved to trying to specify versions in the former, as it reduces the chances for things to go out of sync (e.g., [needing](https://github.com/talonhub/community/pull/1359, https://github.com/talonhub/community/pull/1321)

phillco:
cc @auscompgeek

wenkokke (approved):

wenkokke:
I'm slowly working my way towards patching talonfmt to work with Python 3.12.

C-Loftus:
> This will mean contributors would be required to have Python 3.11 installed. Are we comfortable with that?

Maybe I'm missing something but least in [talon-ai-tools](https://github.com/c-loftus/talon-ai-tools) it just runs on PRs and I don't think anyone ever runs pre-commit locally. Do many people in community run it locally before PRs?

pokey:
> Maybe I'm missing something but least in [talon-ai-tools](https://github.com/c-loftus/talon-ai-tools) it just runs on PRs and I don't think anyone ever runs pre-commit locally. Do many people in community run it locally before PRs?  

I run them locally fairly regularly for cursorless

phillco:
> Do many people in community run it locally before PRs?

I usually prefer formatters that run on save, but I will also run-pre commit locally to make sure I didn't miss anything. It's usually simpler than rebasing the autoformatting commits that pre-commit.ci creates if you ever make followups.

phillco:
> This will mean contributors would be required to have Python 3.11 installed.

Both Talon beta and public have been on 3.11 (3.11.8/3.11.4 currently) for a little bit now. I think that's a reasonable heuristic for a minimum version to assume contributors to have (even if they don't use that environment for pre-commit specifically).

In general I think we're OK with having contributors jump through a few more hoops than users. Although things like pre-commit can be used by users, too -- you can just use it to reformat your local changes even if you don't plan to contribute them back.

wenkokke:
Just released a new version of talonfmt, so we should check if the problem is resolved.

nriley:
We're good! I'm closing this one thanks to @wenkokke.

Issue #1418, opened 2024-04-16
Get community maintainers onto mainline community repo


pokey:
# The problem

Today, none of the community maintainers are using the main branch of community. This fact means that

- we do not experience issues that may happen for new users starting out
- it is harder for users to reproduce commands they see in videos (see https://github.com/cursorless-dev/cursorless/discussions/547)
- it is hard for users to move from the relatively verbose commands in community to commands often used by more experienced voice coders (see #1221)
- it will be difficult for us to move to the future in which Talon has a package manager

# The solution

The idea would be that core maintainers would be able to use the main branch of community, with their own customization repo that sits beside community to enable their own customization. Generally speaking, this will require us to make community more customizable, eg by

- Adding actions that users can override
- Moving more things to `.talon-list` files

These are just general principles; in practice we will need to look at the different places that our core maintainers have changed in their forks, and treat them on a case by case basis. For each change, either:

- The change is a pure improvement to community that just needs to be upstreamed, or
- The change is a personal preference that should be customizable via some mechanism in community, or
- The change is orthogonal to community, and should be able to live in a side repo that the given maintainer clones alongside community without any changes to community

## How do we get there

We propose to create a git subtree of community (similar to what we do with https://github.com/talonhub/community/tree/main/apps/vscode/command_client). This subtree would contain a growing subset of community that we expect all core maintainers to leave untouched, or at least try to keep close to main and upstream changes fairly regularly. The subtree would have a well-defined api surface. Ideally it would be generic enough that @AndreasArvidsson could have this subtree cloned alongside his repo, after having removed the corresponding piece from his repo

Initially, the subtree would have a single directory from core, eg formatters or numbers. We then add more and more to it, until it grows to be most (all?) of the community repo

When Talon has a package manager, the subdirectories from this subtree could become Talon packages, and the maintainers would be able to start using them right off the bat, because they wouldn't need their own custom versions

## Related

- https://github.com/talonhub/community/issues/1384

pokey:
We're planning to do a mob coding session in one of the cursorless meet-ups where we try out this idea for one small part of community, eg numbers or formatters, to see if it is promising

jaresty:
I'm excited to see interest in trying this out! Thank you for putting in the time to experiment. This plan sounds like a really good strategy to me, and I'm excited to hear how it goes.

pokey:
@AndreasArvidsson and I discussed formatters at length. It appears there are several PRs to get formatters to support our use cases:

[...]

PR #1431, opened 2024-05-09
Replacing formatters

AndreasArvidsson:
By request I'm updating the formatters in community with my own implementation. 

* Each formatter can decide how it wants to unformat the text beforehand when doing reformat
* Allow you to reformat strings and still keep delimiters
* Added proper reformat action

All the issues around updating the history in the capture in having actions like `insert_many` I leave to a follow up. We all show probably should have a follow up around the spoken forms and where they are located. Better outlined here: https://github.com/talonhub/community/issues/1418


AndreasArvidsson:
Waiting for talonfmt to be updated for latest python version
https://github.com/talonhub/community/pull/1430

nriley:
We discussed this in the backlog session.

Things look good right now but for a few new pieces we will need to add for the first version of this PR:

* Since we now have differences between formatting and reformatting, it will be helpful to have "format help" allow you to specify a different starting string other than the existing "one two three". Format help should also show the input string in addition to the formatted string.
* We need to pick some spoken forms for the new "reformatter" formatters.

Things that are explicitly out of scope for this version of the PR:

* Allowing extensibility of formatters from outside of formatters.py
* Using a Cursorless-like CSV to allow customization of spoken forms, including the ability to compose formatters with a single spoken form

core/text/formatters.py
```diff
@@ -181,29 +308,20 @@ def every_word(word_func):
     "kebab": "DASH_SEPARATED",
     "packed": "DOUBLE_COLON_SEPARATED",
     "padded": "SPACE_SURROUNDED_STRING",
-    "slasher": "ALL_SLASHES",
-    "conga": "SLASH_SEPARATED",
+    "lasher": "SLASH_SEPARATED",
```
nriley, line 311: These are actually different formatters.

nriley (changes requested):

core/text/formatters.py
```diff
@@ -181,29 +308,20 @@ def every_word(word_func):
...
     "smash": "NO_SPACES",
     "snake": "SNAKE_CASE",
     "string": "SINGLE_QUOTED_STRING",
 }
+
 prose_formatter_names = {
     "say": "NOOP",
     "speak": "NOOP",
-    "sentence": "CAPITALIZE_FIRST_WORD",
+    "sentence": "SENTENCE",
```
nriley, line 320: `CAPITALIZE_FIRST_WORD` is in use in several other places and should be preserved even if it is a misnomer. (Sorry!)

core/text/formatters.py
AndreasArvidsson: done

core/text/formatters.py
AndreasArvidsson: done

AndreasArvidsson:
>     * We need to pick some spoken forms for the new "reformatter" formatters.
Done

>     * Since we now have differences between formatting and reformatting, it will be helpful to have "format help" allow you to specify a different starting string other than the existing "one two three". Format help should also show the input string in addition to the formatted string.
I'm worried it will be confusing if we try to show a new format and a reformat in the same line. I'm wondering if this is something we should have a discussion about in a future meet up and then do in a separate pull request?

nriley:
> > ```
> > * Since we now have differences between formatting and reformatting, it will be helpful to have "format help" allow you to specify a different starting string other than the existing "one two three". Format help should also show the input string in addition to the formatted string.
> > ```
> 
> I'm worried it will be confusing if we try to show a new format and a reformat in the same line. I'm wondering if this is something we should have a discussion about in a future meet up and then do in a separate pull request?

Have you looked at how it looks in the current version? I can't even get `format help` to display with your version:

```
 in script at /Users/nicholas/.talon/user/talon_community/core/help/help.talon:10:
 > user.help_formatters(user.get_formatters_words())
KeyError: 'ALL_UPPERCASE'
```

AndreasArvidsson:
> > > ```
> > > * Since we now have differences between formatting and reformatting, it will be helpful to have "format help" allow you to specify a different starting string other than the existing "one two three". Format help should also show the input string in addition to the formatted string.
> > > ```
> > 
> > 
> > I'm worried it will be confusing if we try to show a new format and a reformat in the same line. I'm wondering if this is something we should have a discussion about in a future meet up and then do in a separate pull request?
> 
> Have you looked at how it looks in the current version? I can't even get `format help` to display with your version:
> 
> ```
>  in script at /Users/nicholas/.talon/user/talon_community/core/help/help.talon:10:
>  > user.help_formatters(user.get_formatters_words())
> KeyError: 'ALL_UPPERCASE'
> ```

I forgot that in community it's all lower case, but all caps. Fixed now.

AndreasArvidsson:
> Thanks. I'm out of time for Talon work this weekend but can take a look later; others can of course feel free to offer suggestions! My concern is that if we have multiple formatters that appear to do the same thing using our existing help system that it'll be even more confusing for new users than it is right now with the existence of prose and code formatters.

I agree, but I'm definitely not sure how to do this in a non confusing way with the current help system.

nriley:
OK. I've got some ideas so can work on prototyping them when I have time (may not be for a couple weeks as I will be working next weekend).

AndreasArvidsson:
We can just disabled the new capitalize formatter and re-enable that when we have a better help menu?

AndreasArvidsson:
Note from last community session. Split formatter help into format and reformat.

AndreasArvidsson:
@nriley I have now added a separate help for reformatting.

AndreasArvidsson:
@nriley How have your testing turned out?

nriley:
Very good - I've been using it for weeks now! Sorry, thought we had already merged this one...

nriley (approved):

PR #1432, opened 2024-05-09
Added edit paragraph actions and commands

AndreasArvidsson:


core/edit/edit_paragraph.py
```diff
@@ -0,0 +1,108 @@
...
+        if is_line_empty():
+            return
+        # Search for start of paragraph
+        actions.edit.extend_paragraph_start()
+        actions.edit.left()
+        # Extend to end of paragraph
+        actions.edit.extend_paragraph_end()
+
+    def extend_paragraph_start():
+        extend_paragraph_start()
```
nriley, line 27: You might want to explain why you don't use the default Talon supplied action.

core/edit/edit.talon
```diff
@@ -161,6 +165,7 @@ paste match: edit.paste_match_style()
 (pace | paste) line: user.paste_line()
```
nriley, line 165: Can we add an optional "to" to make the use of these clearer?

AndreasArvidsson: done

core/edit/edit_paragraph.py
AndreasArvidsson: done

AndreasArvidsson:
@nriley Ready for review again :)

nriley (approved):

PR #1433, opened 2024-05-10
Extract code into separate rpc client

AndreasArvidsson:
Refactor command client so it can be used by multiple applications at once. I need this for my clipboard manager that is active at the same time as vscode.

This pull request basically does two things:
1. Splits the very large `command_client.py` in to several python files in the new `rpc_client` directory. This is done purely mechanically with as minimal changes as possible.
2. Introduces a new Talon action `rpc_client_run_command` that takes the communication directory name as well as the callback to press the keys that starts the request execution. This is what allows us to run multiple applications at once.


apps/vscode/command_client/rpc_client/rpc_client.py
```diff
@@ -0,0 +1,98 @@
...
+        trigger_command_execution: Callable,
+        command_id: str,
+        args: list[Any],
+        wait_for_finish: bool = False,
+        return_command_output: bool = False,
+    ):
+        """Runs a command, using command server if available
+
+        Args:
+            command_id (str): The ID of the command to run.
```
nriley, line 28: Please add documentation for the new arguments.

nriley (changes requested):

apps/vscode/command_client/command_client.py
```diff
@@ -138,142 +52,15 @@ def run_command(
...
-        wait_for_finish=wait_for_finish,
-        return_command_output=return_command_output,
-        uuid=uuid,
+    return actions.user.rpc_client_run_command(
+        actions.user.command_server_directory(),
+        actions.user.trigger_command_server_command_execution,
+        command_id,
+        args,
+        wait_for_finish,
+        return_command_output,
```
phillco, line 61: We're starting to build up a bunch of parameters on this action and I'm worried about the wrong ones getting passed in the future; how do you feel about converting these to kwargs to make them explicit? It's much more verbose but I think it's worth the tradeoff in these situations

```suggestion
    return actions.user.rpc_client_run_command(
        dir_name=actions.user.command_server_directory(),
        trigger_command_execution=actions.user.trigger_command_server_command_execution,
        command_id=command_id,
        args=args,
        wait_for_finish=wait_for_finish,
        return_command_output=return_command_output,
```

AndreasArvidsson:
@knausj85 Will test this.

apps/vscode/command_client/command_client.py
AndreasArvidsson: I don't think you can use named arguments with Talon actions.

apps/vscode/command_client/rpc_client/rpc_client.py
AndreasArvidsson: done

knausj85 (approved):
Tested on Mac and Windows. Appears to work well.

AndreasArvidsson:
@nriley Mind having a look at this so we can merge it?

nriley (approved):

PR #1435, opened 2024-05-12
Cancel any current phrase when Talon is waking

AndreasArvidsson:
When Talon wakes we set a timestamp threshold. On the next command we will compare the phrase timestamp to the threshold and cancel any phrase started before wakeup. This is to prevent commands from being executed if the user wakes Talon using a noise or keypress.


plugin/cancel/cancel.py
```diff
@@ -1,20 +1,70 @@
...
+
+
+@ctx.action_class("speech")
+class SpeechActions:
+    # When Talon wakes we set the timestamp threshold. On the next command we
+    # will compare the phrase timestamp to the threshold and cancel any phrase
+    # started before wakeup. This is to prevent speech said before wake-up to
+    # be interpreted as a command if the user wakes Talon using a noise or
+    # keypress.
+    def enable():
```
knausj85, line 25: We probably need to ensure that this coexists well with 

https://github.com/talonhub/community/blob/main/plugin/listening_timeout/listening_timeout.py

Which also overrides the enable() with the same specificity at the moment

AndreasArvidsson: True. I didn't know about that feature. Do you mind testing it? We could of course hide this new behavior behind a tag or setting if we want to.

knausj85: False alarm, seems to work as-is for me. Nice improvement :)

plugin/cancel/cancel.py
```diff
@@ -1,20 +1,70 @@
...
+        actions.user.cancel_current_phrase()
+        actions.next()
+
+
+@mod.action_class
+class Actions:
+    def cancel_current_phrase():
+        """Cancel/abort current spoken phrase"""
```
knausj85, line 33: Super minor nitpick: do we see a use case for overriding or calling this action elsewhere?  I wonder if ts_threshold should just be set directly in enable()

'Cause I don't think folks would be able to wire up a command to cancel things using this action, as the the timestamp would be set after the command is executed.

nriley: Agreed, for example if you use `speech.toggle` it'll still call `speech.enable`, so there's no need to separate it out.

plugin/cancel/cancel.py
```diff
@@ -1,20 +1,70 @@
...
+
+    # Check if the phrase is before the threshold
+    if ts_threshold != 0:
+        words = phrase["phrase"]
+        start = getattr(words[0], "start", phrase["_ts"])
+        phrase_starts_before_threshold = start < ts_threshold
+        ts_threshold = 0
+        # Start of phrase is before threshold timestamp
+        if phrase_starts_before_threshold:
+            abort_phrase(phrase)
```
nriley, line 53: Consider printing this phrase to the log? It's possible some people would *want* the other behavior; it may be helpful to see what is being discarded so they can disable this if needed. That's what my version does (I didn't realize you were going to submit this PR so rapidly or I wouldn't have spent the time!): https://github.com/nriley/talon_community/blob/480449064a77399948b7491b2b790619c14db262/core/modes/abort.py#L36

plugin/cancel/cancel.py
```diff
@@ -1,20 +1,70 @@
...
+    global ts_threshold
+
+    # Nothing to cancel
+    if "parsed" not in phrase:
+        return
+
+    # Check if the phrase is before the threshold
+    if ts_threshold != 0:
+        words = phrase["phrase"]
+        start = getattr(words[0], "start", phrase["_ts"])
```
nriley, line 48: Note `words[0]` may not exist — just got this in my version:

```
    2: user/talon_community/core/modes/abort.py:32  | start = getattr(words[0], "start", phrase["_ts"])
    1:                                                                ^^^^^^^^
IndexError: list index out of range

```

plugin/cancel/cancel.py
AndreasArvidsson: I actually use that in my own code when I toggle mute in discord. I thought that maybe someone else might have a similar need and having that extra action doesn't really hurt. 
https://github.com/AndreasArvidsson/andreas-talon/blob/53ba2db930f648683092df05f1af80de3e2ac367/games/games.py#L44

plugin/cancel/cancel.py
AndreasArvidsson: Good catch I have that in my own code, but in a earlier part of the call stack so I missed it.

AndreasArvidsson: fixed

plugin/cancel/cancel.py
nriley: OK, I don't feel too strongly about this, sounds like a reasonable argument to keep it separate.

plugin/cancel/cancel.py
AndreasArvidsson: Print added

plugin/cancel/cancel.py
AndreasArvidsson: It's probably an uncommon use case so I could go either way

plugin/cancel/cancel.py
nriley: Thanks. (Python's := operator can be useful in such cases but it's just a stylistic choice…)

AndreasArvidsson: I did some refactoring and I think this turned out quite okay

plugin/cancel/cancel.py
knausj85: It seems reasonable to keep it then. Thanks

nriley (approved):

PR #1436, opened 2024-05-13
Get vscode.py to recognize cursor IDE for macos

colesam:


apps/vscode/vscode.py
```diff
@@ -12,6 +12,8 @@
 and app.bundle: com.microsoft.VSCodeInsiders
 os: mac
 and app.bundle: com.visualstudio.code.oss
+os: mac
+and app.bundle: com.todesktop.230313mzl4w4u92
```
pokey, line 16: this bundle id is surprising, but [appears to be correct and stable](https://forum.cursor.sh/t/cursor-bundle-identifier/779/2)

pokey (approved):
Looks good to me!

colesam:
> I assume this is working on unmodified cursorless-talon, and otherwise unmodifed community?

Yes it is, I reset both of my repositories and tested the change

PR #1438, opened 2024-05-19
Add cursor support for linux

colesam:


pokey (changes requested):
The first file looks good but it appears you've committed a bunch of extra changes

AndreasArvidsson:
To elaborate on pokeys comment. All changes except the two lines in `vscode.py` looks like they don't belong in this pull request.

nriley:
Addressed in #1549.

PR #1442, opened 2024-05-21
file_manager: Check tag before querying current path

auscompgeek:
Also bugfix the clearing of the current page numbers.


phillco (approved):

AndreasArvidsson:
`not in registry.tags:` really isn't a pattern we should ever be using. You're not introducing this here but I just want to say that we should not do this.

nriley (approved):

PR #1444, opened 2024-05-22
allow "help last" as synonym for "help previous"

rntz:
We use "next/last" in various commands, but help uses "next/previous". This matches the button labels (which are "previous" because it's more self-explanatory than "last"), which is good, but is inconsistent with other commands, which is bad. I suggest we support both.


phillco (approved):
Seems reasonable to me.

PR #1445, opened 2024-05-25
jetbrains: Clean up and fix logging and error handling

auscompgeek:
- Remove a bunch of un-useful prints
- Instead of erroring silently when it fails to send the request to the IDE, notify the user
- Remove the unused `extendCommands` variable


apps/jetbrains/jetbrains.py
```diff
@@ -148,7 +135,15 @@ def idea_commands(commands):
 class Actions:
     def idea(commands: str):
         """Send a command to Jetbrains product"""
-        idea_commands(commands)
+        command_list = commands.split(",")
+        try:
+            for cmd in command_list:
+                if cmd:
+                    send_idea_command(cmd.strip())
+                    actions.sleep(0.1)
```
phillco, line 143: Doesn't have to be done in this PR but we should really see if this can be removed. It makes the running of multiple actions like `idea_grab()` unnecessarily slower/jankier.

(I ran with it removed for many months, but I've since diverged from community enough that it'd be good for someone else to test too)

apps/jetbrains/jetbrains.py
```diff
@@ -148,7 +135,15 @@ def idea_commands(commands):
...
+        except Exception as e:
+            app.notify(e)
+            raise e
```
phillco, line 146: I was going to recommend just `raise`, but it looks like in Python 3 this doesn't actually remove any of the context, waited under Python 2? https://stackoverflow.com/a/45986945/303833

phillco (approved):

timo:
https://github.com/talonhub/community/pull/1411 is there and does something similar. which of the two do you prefer @auscompgeek ?

knausj85:
I don't use jetbrains, but it looks like this one is ready to be merged @phillco @auscompgeek ? Trying to close out some older PRs

Maybe the tiebreak between this and https://github.com/talonhub/community/pull/1411 is that this one is already approved :)

apps/jetbrains/jetbrains.py
phillco, line 146:
```suggestion
            raise retching there
```

PR #1446, opened 2024-05-27
Expand split and tab actions, and try to unify naming/conventions

fidgetingbits:
This expands both the tab and split commands associated actions. In trying to unify the grammar/naming, I ended up deprecating almost everything, which may be a bit controversial.

I haven't applied these new action to all of the contexts that are overriding them because I want to first agree on the naming. I also won't add a comment to the breaking changes file until we decide if/what becomes deprecated.

The background here is that we plan to [neovim-talon](https://github.com/hands-free-vim/neovim-talon/tree/main) available as a plugin, and it includes support for significantly more split and tab related functionality than what's currently available in community. Furthermore, I personally use (and I think andreas repo) have more tab-related functoinality. Most of them should be generic and part of community, especially because other apps will benefit. 

I've also chosen to use the command "cross" first split navigation, because I know some people use that in vscode first splits already and @saidelike uses it for vim. I've also started using it since working on this and seems fine.


fidgetingbits:
As noted in #1447, there is `split flip` atm which changes the orientation between vertical and horizontal but I (not sure about others) generally use flip to refer to switching between the most recent "thing" I'm working with be it desktop, window, split, tab, file, folder, etc. So we may want to change that here, to `split axis` or something.

tags/splits/splits.py
```diff
@@ -1,52 +1,128 @@
...
 
     def split_window_down():
-        """Move active tab to lower split"""
+        """Create a split, of the active window, below"""
 
     def split_window_up():
-        """Move active tab to upper split"""
+        """Create a split, of the active window, above"""
 
     def split_window_vertically():
```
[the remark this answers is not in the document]
fidgetingbits: As discussed in a call, there are a couple issues here: 

* Creating a split when on an active window duplicating the current contents (or not) is likely application specific. Eg: vim you can do `:vnew` to create a new empty vertical split, or `ctrl-w v` to duplicate the current split vertically. An approach is to have different actions for each and perhaps have the default implementation of one can call the other.

* Split groups are an app-specific concept. For example in vscode a split has an associated group of tabs (this I guess is a split group), but in vim a single tab will have multiple splits and there isn't a "split group" per se, unless all the splits in the tab were considered the group. So need to be careful how we describe what is done. 
 
App-specific behavior if different than the description of the action declaration should be clarified in the action implementation for that app, where as the action definitions themselves should be as specific as possible, without assuming application behavior.

tags/splits/splits.py
```diff
@@ -1,52 +1,128 @@
...
+        """Goes to last split"""
+
+    def split_focus_number(index: int):
+        """Navigates to the specified split"""
+
+    def split_focus_most_recent():
+        """Focus on the most recently focused split"""
 
+    # Arrangement
+    def split_move_right():
```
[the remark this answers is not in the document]
fidgetingbits: As discussed, I'll rename/reword things to be more clear and also have some more actions. One point of confusion here is that for vscode the split_window_right() implementation (atm) is actually a move, not a creation, so it's dependent on the split layout whether or not a new split is even created. Likely not consistent with other apps implementing splits.

fidgetingbits:
I brought in the tab PR changes here, also significantly expanded the tab actions. For now I've only modified vscode.py to use the new actions just to demo what it would be like, but if/once we agree on naming/conventions all update all of the other apps in community.

I had noted in the tab PR that it's confusing that tabs are part of core/windows_and_tabs/ where as splits are part of tags/splits, so I also opted to move tabs to tags/tabs/ as part of this PR, as I think it's easier to find things and navigate.

I still need to make sure vscode.talon doesn't have any duplication with the new actions I added.

fidgetingbits:
One other thing I suspect people will want to change is atm I make the `split <direction>` default to an empty split, but I guess most people probably currently expect that to clone the active split, so can switch that around depending on people's preference.

tags/tabs/tabs_linux.py
```diff
@@ -0,0 +1,26 @@
+# defines the default app actions for linux
```
saidelike, line 1: the comment is probably not useful?

tags/tabs/tabs_win.py
```diff
@@ -0,0 +1,26 @@
+# defines the default app actions for windows
```
saidelike, line 1: the comment is probably not useful?

tags/splits/splits.py
```diff
@@ -1,52 +1,216 @@
...
+        """Create a new split to the left"""
+
+    def split_create_down():
+        """Create a new empty split to the down"""
+
+    def split_create_up():
+        """Create a new empty split to the up"""
+
+    def split_create_vertically():
+        """Create a new vertical. The left or right orientation is application dependent"""
```
saidelike, line 25: is it an empty split or a clone one? maybe worth adding in the docstring

tags/splits/splits.py
```diff
@@ -1,52 +1,216 @@
...
+
+    def split_create_horizontally():
```
saidelike, line 27: same here

tags/splits/splits.py
```diff
@@ -1,52 +1,216 @@
...
+    def split_clone_up():
+        """Clone the active view into a new split opened the up"""
+
+    def split_clone_vertically():
+        """Clone the active view vertically. The left or right orientation is application dependent"""
+
+    def split_clone_horizontally():
+        """Clone the active view horizontally. The top or bottom orientation is application dependent"""
+
+    def split_create():
```
saidelike, line 49: move up with the others `split_create*`

tags/splits/splits.talon
```diff
@@ -1,17 +1,68 @@
...
+split (new | create): user.split_create()
+
+## Duplicated splits
+split clone right: user.split_clone_right()
+split clone left: user.split_clone_left()
+split clone down: user.split_clone_down()
+split clone up: user.split_clone_up()
+split clone (vertically | vertical): user.split_clone_vertically()
+split clone (horizontally | horizontal): user.split_clone_horizontally()
+split clone (new | create): user.split_clone()
```
saidelike, line 22: remove "create" from here?

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+    def split_close_all():
+        actions.user.vscode("workbench.action.editorLayoutSingle")
+
+    # Navigation
+    def split_focus_previous():
         actions.user.vscode("workbench.action.focusLeftGroup")
 
-    def split_next():
+    def split_focus_next():
         actions.user.vscode_and_wait("workbench.action.focusRightGroup")
```
saidelike, line 239: actions.user.vscode_and_wait("workbench.action.focusNextGroup")

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+    # Destruction
+    def split_close():
+        actions.user.vscode("workbench.action.joinTwoGroups")
+
+    def split_close_all():
+        actions.user.vscode("workbench.action.editorLayoutSingle")
+
+    # Navigation
+    def split_focus_previous():
         actions.user.vscode("workbench.action.focusLeftGroup")
```
saidelike, line 236: workbench.action.focusLastEditorGroup

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+        actions.user.vscode("workbench.action.splitEditorOrthogonal")
+
+    # Destruction
+    def split_close():
+        actions.user.vscode("workbench.action.joinTwoGroups")
+
+    def split_close_all():
+        actions.user.vscode("workbench.action.editorLayoutSingle")
+
+    # Navigation
```
saidelike, line 234:
```
    def split_focus_up():
        actions.user.vscode("workbench.action.focusAboveGroup")

    def split_focus_down():
        actions.user.vscode("workbench.action.focusBelowGroup")

    def split_focus_left():
        actions.user.vscode("workbench.action.focusLeftGroup")

    def split_focus_right():
        actions.user.vscode("workbench.action.focusRightGroup")
```

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
-    def split_window_vertically():
-        actions.user.vscode("workbench.action.splitEditor")
+    # Resizing
+    def split_toggle_orientation():
+        actions.user.vscode("workbench.action.toggleEditorGroupLayout")
 
-    def split_window():
-        actions.user.vscode("workbench.action.splitEditor")
+    def split_toggle_maximize():
+        actions.user.vscode("workbench.action.maximizeEditor")
```
saidelike, line 265: workbench.action.toggleMaximizeEditorGroup

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+
+    def split_layout_reset():
+        actions.user.vscode("workbench.action.evenEditorWidths")
```
saidelike, line 268: workbench.action.editorLayoutSingle

tags/splits/splits.py
```diff
@@ -1,52 +1,216 @@
...
+    def split_toggle_orientation():
+        """Flips the orientation of the active split"""
+
+    def split_toggle_maximize():
+        """Maximizes the active split"""
+
+    def split_layout_reset():
+        """Resets all the split sizes"""
+
+    def split_expand():
```
saidelike, line 131: we could have the base definition just call split_expand_width() and split_expand_height()

tags/splits/splits.py
```diff
@@ -1,52 +1,216 @@
...
+        """Expands the both the width and height of the split"""
+
+    def split_expand_width():
+        """Expands the split width"""
+
+    def split_expand_height():
+        """Expands the split height"""
+
+    def split_shrink():
```
saidelike, line 140: we could have the base definition just call split_shrink_width() and split_shrink_height()

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+
+    def split_expand():
```
saidelike, line 270: this could be removed if we have the base definition just call split_expand_width() and split_expand_height()

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+        actions.user.vscode("workbench.action.increaseViewWidth")
+        actions.user.vscode("workbench.action.increaseViewHeight")
+
+    def split_shrink():
+        actions.user.vscode("workbench.action.decreaseViewWidth")
```
saidelike, line 275: this could be removed if we have the base definition just call split_shrink_width() and split_shrink_height()

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+        actions.user.vscode("workbench.action.pinEditor")
+
+    def tab_unpin():
+        actions.user.vscode("workbench.action.unpinEditor")
+
+    def tab_move_right():
+        actions.user.vscode("workbench.action.moveEditorRight")
+
+    def tab_move_left():
+        actions.user.vscode("workbench.action.moveEditorLeft")
```
saidelike, line 339: workbench.action.moveEditorToLeftGroup

apps/vscode/vscode.py
```diff
@@ -170,53 +170,183 @@ def command_palette():
...
+
+    # Arrangement
+    def tab_pin():
+        actions.user.vscode("workbench.action.pinEditor")
+
+    def tab_unpin():
+        actions.user.vscode("workbench.action.unpinEditor")
+
+    def tab_move_right():
+        actions.user.vscode("workbench.action.moveEditorRight")
```
saidelike, line 336: workbench.action.moveEditorToRightGroup

apps/vscode/vscode.py
fidgetingbits: Should actually be workbench.action.focusPreviousGroup

apps/vscode/vscode.py
fidgetingbits: I don't agree with this one. As the action description says, it's meant to resize all splits to the default, which is what workbench.action.evenEditorWidths does. What you're suggesting would be suitable for split_close_all() (and is what it does).

apps/vscode/vscode.py
fidgetingbits: I don't agree with this. tab_move_right() is meant to move a tab to the right, and in the case of vscode (imo) this should be inside the "group", rather than to a new group. That said, the action is still wrong as it should actually be suffixed with InGroup, so I've fixed them.

Moving a tab right within the group won't move to an adjacent group if that tab is in the right most index. You have to explicitly use moveEditorToRightGroup. I've added another set of actions tab_move_to_split_<direction>(), which will be applicable to vscode, jetbrains, and probably others, but not neovim, etc. 

Since tab_move_up() and tab_move_down() are also actually "group-based" I think, I've switched those to be tab_move_to_split_up() and tab_move_to_split_down() respectively.

I suppose worth noting, if you already have a tab at the right most position and you say tab move right you _might_ expect it to move to the left most index of the group to the right, but this isn't something I think we can easily do from just using available command ids alone. Probably would need a custom extension or something, so outside the scope of this PR.

fidgetingbits:
I did just add one other small split update, but I think for the most part the proposed actions/naming/etc is stable, so is good for others to look at and discuss. I'm leaving it as a draft because of still needing to update all of the other apps using tabs/splits once things are agreed upon.

tags/tabs/tabs.talon
```diff
@@ -0,0 +1,38 @@
...
+
+# Destruction
+tab close: user.tab_close_wrapper()
+tab close all: user.tab_close_all()
+tab (close others | solo): user.tab_close_others()
+
+# Navigation
+[go] tab (last | previous): app.tab_previous()
+[go] tab next: app.tab_next()
+[go] tab <number>: user.tab_focus_index(number)
```
nriley, line 18: The optional `go` causes an issue here if numbers are not prefixed and you are trying to insert a tab character followed by a literal number. The optional go in other cases seems to work well.

tags/tabs/tabs.talon
```diff
@@ -0,0 +1,38 @@
...
+
+# Arrangement
+tab pin: user.tab_pin()
+tab unpin: user.tab_unpin()
+tab move right: user.tab_move_right()
+tab move left: user.tab_move_left()
+tab move to split right: user.tab_move_to_split_right()
+tab move to split left: user.tab_move_to_split_left()
+tab move [to split] up: user.tab_move_to_split_up()
+tab move [to split] down: user.tab_move_to_split_down()
```
nriley, line 34: From the community backlog session — overall we think this is a good way to be moving for tab navigation as some of these tab commands are supported outside editors, for example in browsers. Not necessarily needed as part of this PR but it would be good to get some implementations added.

nriley (changes requested):
From the community backlog session — this is a lot of only peripherally related functionality to be reviewing at one time. It would be helpful to split it at least into:

- Generalization of tab commands and implementation of default tab manipulation behavior by platform
- Implementation of additional split and tab manipulation commands in a way that includes or at least does not preclude implementations for at least a couple editors — we are somewhat worried that this bakes the vscode tab and split model into community in a way that may not be supportable elsewhere

nriley:
@fidgetingbits is is something you might pick up again in future? We discussed the need for some revisions to the split commands on the community backlog session today. Thanks!

fidgetingbits:
Very unlikely for the foreseeable future

PR #1447, opened 2024-05-30
expand tab actions

fidgetingbits:
Similar to #1446 this just expands the available actions for using tabs.

There is one thing I realized which is that here I've used `tab flip` to indicate flipping back between the most recent tabs. I generally use `flip` for going to the most recent thing everywhere in my repo: windows, desktops/workspaces, splits, tabs, files, folder, etc. However, community uses `split flip` to change the horizontal/vertical orientation. Since there is already #1446 open, it might be a good time to change that.


fidgetingbits:
Also I think we should move this to tags/tabs.{py,talon} instead of core/windows_and_tabs to be more consistent, but I won't do it until others way in. Not sure why splits is in tags/ and this has it's own special place.

fidgetingbits:
Since there has been discussion about minimizing forking, and tab is a command I happen to change because I don't find it reliable (I use tabby), I'd also like to propose we make a <user.tab> capture in tabs.py that can be overridden. I'm happy to add it, but I'll wait until others agree.

fidgetingbits:
After a discussion with @saidelike I will make some more changes before I take it back out of draft.

fidgetingbits:
Gonna just merge this into a single PR with the split changes since I want to try to unify the names

PR #1448, opened 2024-05-31
refactor: move command_client folder into core

fidgetingbits:
This is mostly just a nit, but every time I go to find `command_client/` in community I go to the wrong place, because I don't associate it with vscode. Since there's other users of it now, like neovim-talon now, it probably makes sense for this to just become part of `core/`. It could also be suited to `plugin/`, but if we ever decided to move all of the app-specific stuff (only vscode and visual_studio atm) into the associated `apps/<foo>/` folders then this would be `apps/` relying on `plugin/` which we don't want.


pokey (changes requested):
I agree in principle, but we'll need to figure out how it interacts with the git subtree push. I'll need to do a bit of googling / experimentation to make sure it will work. See the README in this directory if you're not sure what I'm referring to

I'm going to mark this as needs changes for now just so we don't merge, and we'll also need to update the README with the new subtree push command once we figure it out

pokey:
> this would be apps/ relying on plugin/ which we don't want.

Hmm not sure bout that one. I def think core shouldn't rely on apps or plugins, but I'm not sure I see an issue with apps relying on plugins

fidgetingbits:
> > this would be apps/ relying on plugin/ which we don't want.
> 
> Hmm not sure bout that one. I def think core shouldn't rely on apps or plugins, but I'm not sure I see an issue with apps relying on plugins

Ya that makes sense actually. In that case I'd probably opt for it to go in plugin? but really I'm good with either.

auscompgeek:
It's been a while since I've used git subtree, but I don't think there's any reason why it wouldn't like a directory move.

AndreasArvidsson:
Replaced by #1678

PR #1449, opened 2024-06-02
expand line command actions

fidgetingbits:
This adds two new actions for lines that allow for something like `f<key>` in vim (and other places) which will just jump to the first occurrence of that character.

The only debatable bit is likely the use of `find` as the command. Open to suggestions though.


jaresty:
I dig it. I'd be interested in your feedback on this one too: https://github.com/talonhub/community/pull/1424 (similar to vim's `gm`)

fidgetingbits:
> I dig it. I'd be interested in your feedback on this one too: #1424 (similar to vim's `gm`). An alternative word that could work maybe would be `char` (`last char`). Or `go char ` and `go last char`.

Ya, I'm a bit iffy on using find here, just happens to be what I had used in vim. I'd be fine with `go char` and `go last char` I think.

jaresty (approved):
🚢

jaresty:
I wanted to share an idea that I just had about this: it would be really cool if we could add a similar command for adding a cursor at the next glyph (meaning: leave a cursor where you are now and then add another one at the next instance of this glyph)

jaresty:
I wish I could merge this for you 😄

fidgetingbits:
> I wanted to share an idea that I just had about this: it would be really cool if we could add a similar command for adding a cursor at the next glyph (meaning: leave a cursor where you are now and then add another one at the next instance of this glyph)

can file an issue or open a separate PR for it. you'd want that to only add a cursor to the next glyph on the same line or next in the file? 

makes me wonder if there should be a file for 'multiple cursor' related commands. I don't ever use them aside from with cursorless, but I suspect lots of others do. would need some common command to indicate you want to add a cursor too. 

"curse" would be good but curse on/off is used for mouse cursor, so a bit confusing. but something like "curse add next char this"?

You could just do what you want in cursorless already with something like (untested) "pre char this and next view char this", but it's a bit wordy.

anyway probably better to discuss in a different issue or PR if you prefer not to use cursorless for that.

jaresty:
👍fwiw you can also use cursorless to navigate to the next glyph. I use multi cursor most often when I don't have access to cursorless because I'm having to work in IntelliJ.

AndreasArvidsson:
We discussed this in the latest community session. 
* We think the action name and description should probably reflect that this is not a normal find/search command and is more like a cursor jump. Something like jump cursor to next instance of character.
* We probably should have an implementation for vim. If we add an implementation of these actions to `vimscript.py`

fidgetingbits:
> We discussed this in the latest community session.
> 
>     * We think the action name and description should probably reflect that this is not a normal find/search command and is more like a cursor jump. Something like jump cursor to next instance of character.
> 

Changed the names and doc strings to reflect the feedback.

>     * We probably should have an implementation for vim. If we add an implementation of these actions to `vimscript.py`

vimscript.py is for the vimscript language, not for the vim app, so not applicable there. community doesn't have any native support for vim atm. Adding it is non-trivial because of mode switching, so it's not as simple as just enabling it if vim is active somewhere.

Most people using vim (afaik) used to use my personal repo, which then got broken out into [talon-vim](https://github.com/fidgetingbits/talon-vim), and then partially got broken into a much simpler implementation [neovim-talon](https://github.com/hands-free-vim/neovim-talon), which is what will be paired with cursorless in neovim. 

The neovim-talon repo will use these actions being added in this PR, so I don't think there _needs_ to be something in community using them right away. I suspect there is probably some emacs equivalent that could be added if you think there really should be something. Or we could add something for https://marketplace.visualstudio.com/items?itemName=ArturoDent.jump-and-select, but I doubt many people are using that in vscode (could be wrong).

fidgetingbits:
I did notice that zsh by default seems to bind a readline shortcut for doing this, but bash doesn't. So will need shell-specific readline overrides, which we don't support atm. I plan to add this to my branch and then port it over eventually after a bunch of testing, but not sure if it needs to be part of this PR. 

I don't mind to wait though if it's preferred to have some real use cases in community. 

I guess maybe of note, I have a vscode plugin scope in my branch so can enable certain commands if a vscode plugin is installed, which could be used to enable jump-and-select extension mentioned above to more cleanly add support for this into vscode as well. I haven't opened a PR to add that to community yet though.

AndreasArvidsson:
Update from the community session. We think that we should probably wait with this until we actually have an implementation in community that utilizes the spoken forms.

fidgetingbits:
No worries. I'll add something that uses it as part of this PR then.

fidgetingbits:
Ok the scope of this is expanding now... but this implements a plugins tag and associated scope so that you can now specify if an app has a specific plugin installed. I also include action implementation for the new line command if the jump-and-select plugin is installed in vscode.

I haven't bothered moving all of the other plugin usage from vscode.talon to use this new `user.plugin_installed` scope, as I think it can be a follow up. But can add it here if preferred.

Also atm the test if the plugin is installed is tested on Linux only. Will test on Mac later, but if someone on Windows can hopefully chime in about where the settings file is located I'll add that too.

knausj85:
Notes from community session:
I really like this plugin idea; I'll be putting it thru some paces over the next week or so to see if we can't get this PR moving again.

apps/vscode/plugins/plugins.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+
+def vscode_update_plugins_list():
+    """Queries the list of vscode plugins"""
+    extensions = pathlib.Path.home() / ".vscode/extensions/extensions.json"
+    if not extensions.exists():
+        return []
+    with extensions.open() as f:
+        return [x["identifier"]["id"] for x in json.loads(f.read())]
```
knausj85, line 23:
```suggestion
        return [x["identifier"]["id"].lower() for x in json.loads(f.read())]
```

Just in case there are case differences, perhaps we should standardize on lower case?

tags/line_commands/line_commands.talon
```diff
@@ -74,3 +74,6 @@ select camel left: user.extend_camel_left()
 select camel right: user.extend_camel_right()
 go camel left: user.camel_left()
 go camel right: user.camel_right()
+
+go char <user.unmodified_key>: user.jump_cursor_to_next_char(unmodified_key)
```
knausj85, line 78: I'm not sure these should live in line commands. Maybe they make sense alongside the plugin and be enabled with the appropriate tag? As it is, this would add some commands that won't work in a lot of applications

tags/plugins/plugins.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def plugins_list_refresh() -> str:
+        """Refresh the list of plugins associated with the active app"""
+
+    def plugins_list_print() -> str:
+        """Print the list of plugins currently associated with the active app"""
+
+
+ui.register("win_focus", scope.update)
+ui.register("win_title", scope.update)
```
knausj85, line 35: Do we need win_title to update the scope?

apps/vscode/plugins/jump-and-select.py
```diff
@@ -0,0 +1,18 @@
...
+ctx = Context()
+ctx.matches = r"""
+    app: vscode
+    user.plugin_installed: arturodent.jump-and-select
+"""
+
+
+@ctx.action_class("user")
+class UserActions:
+    def jump_cursor_to_next_char(char: str):
```
knausj85, line 12: I wonder if these plugin implementations belong in the community repository. Perhaps we should consider moving knees into a separate plugin repository?

knausj85 (changes requested):
This works very well, and I think we should bring it in. Few minor details to sort through.

There were some concerns raised vis-a-vis parsing the extension file (performance?), but this approach makes sense to me.

apps/vscode/plugins/jump-and-select.py
fidgetingbits: Ya, plugins can get out of hand pretty fast, (see https://github.com/fidgetingbits/talon-vim/tree/main/vim/core/plugins) and the vast majority of people won't use most, so I can see the argument to move it somewhere else. I'm good with whatever.

tags/line_commands/line_commands.talon
fidgetingbits: hrm. I do think there is the potential for this to be used elsewhere in community, though not sure of any good example where to add it off the top of my head. But for instance, I already used it in my repo for zsh line editor (would like eventually to PR that), and was going to adapt vim to use it after (though that won't become part of community) and similarly I guess people could add it for emacs (maybe there is no emacs in community). I added a separate vscode example in this PR to community specifically to justify it since we didn't want an otherwise completely dead command, but I do see your point.. 

Having it on the plugin side only doesn't entirely make sense due to the potential to be used inside community, but if you guys decide you don't want it in here and just want to lift the plugin stuff from this PR instead I'm good with that. Can always add it later once there is a second use case.

tags/plugins/plugins.py
fidgetingbits: hrm, I think probably not. I don't entirely remember why I added that. This code was simplified from some vim plugin parsing I had written awhile back though, so it may have been required for vim, but I can't see why it would be for vscode. Will remove.

apps/vscode/plugins/plugins.py
```diff
@@ -0,0 +1,50 @@
...
+
+ctx.matches = r"""
+    app: vscode
+    tag: user.plugins
+"""
+
+
+def vscode_update_plugins_list():
+    """Queries the list of vscode plugins"""
+    extensions = pathlib.Path.home() / ".vscode/extensions/extensions.json"
```
AndreasArvidsson, line 19: This line will not work for all the different clones of vscode. If you look at our [application definitions](https://github.com/talonhub/community/blob/1e3f2e1683c3bacd4cee20d85942db5fe24e8c43/apps/vscode/vscode.py#L8-L51) we support applications like VSCodium and Cursor. Anything we add in the vscode file should ideally work on all of these.

apps/vscode/plugins/jump-and-select.py
knausj85: After pondering, I think this is one of the bridges we cross once it becomes unwieldy. We should at least have an example implementation in community. 🙂

tags/plugins/plugins.py
```diff
@@ -0,0 +1,35 @@
+from talon import Module, actions, ui
+
+mod = Module()
+
+mod.tag("plugins", "Enable commands and contexts for apps that use optional plugins")
+
+
+@mod.scope
+def scope():
+    return {"plugin_installed": actions.user.plugins_list()}
```
knausj85, line 10: We're thinking `actions.user.plugins_list` should have a VScode exclusive implementation for now. For other variants, these could be added by folks.

Long term, we're thinking to explore getting this via the command server.

tags/line_commands/line_commands.py
```diff
@@ -39,3 +39,9 @@ def line_clone(line: int):
         """Clones specified line at current position"""
         actions.edit.jump_line(line)
         actions.edit.line_clone()
+
+    def jump_cursor_to_next_char(char: str):
+        """Jumps the first to the next instance of the specified character on the current line"""
+
+    def jump_cursor_to_prev_char(char: str):
+        """Jumps the first to the previous instance of the specified character on the current line"""
```
phillco, line 47: qq: could you elaborate why limiting search results to the current line is valuable? I thought it was because it would wrap so you could get to stuff closer to the beginning quickly, but it looks like in vim `f` doesn't do that. Probably something we were missing. cc @nriley who tried it

tags/plugins/plugins.py
AndreasArvidsson: I've looked into this and there isn't a simple way from the command server to know this folder path. But we can return a list of all the extensions from the command server.

tags/line_commands/line_commands.py
fidgetingbits: I often fuzzily (as in I don't know the exact index so overshoot) jump to the last character on a line (and don't want to jump off the line) and also don't always want to have to speak multiple commands to go to the end of the line and then search backwards. Also helps if you are working on a single line and definitely don't want to go off that line in the event of a misrecognition/mispeak (a regularity for me).

If others don't think that's actually useful to general users happy to just close.

knausj85:
Alright, I wanna try to get this one moving again.

My take
I like both the plugin functionality and the example go char example. go char's proven useful to me; so as long as no other maintainers feel super strongly, I vote we keep it as-is.

My proposal: 
- scope the plugin list action implementation to Microsoft VSCode proper for now. Other vscode clones can be added similarly as desired by other community members.

- add a comment that the go char commands likely require a plugin to the .talon file

- merge and move on 🙂

PR #1451, opened 2024-06-03
Allow apps to force lang via code.language and add special cased files

fidgetingbits:
This makes some tweaks to language modes:

1. It adds a list of special-cased files that are mixed into the determination of the language, so you can return a language for things even if the file extension itself doesn't imply the language. The languages included aren't actually implemented in community, so I can keep the list empty if preferred, but I figured I'd keep it populated at first to give an example of why it's useful.
2. Adds a command to show the currently forced language
3. #1256 semi-recently went the route of removing language-specific tags. This has the side effect (afaict) that an app that doesn't expose the file extension in its title can no longer force a language (previously it would just `tag: user.<lang>`). Given we removed the tags I assume we don't want them back, so this change goes a different route and adds an action that can be overridden to force a language. It's done in a way that the language being forced by an app will still be overridden by a user manually forcing a language with `force <lang>`. I've included an obsidian skeleton to demonstrate how to use it. Obsidian is a markdown editor that doesn't include file extensions in it's title. I took a look at search.talonvoice.com and see that some people are using tag.markdown (so are their community is not up to date, or they custom implemented tags again) and some others are re-implementing markdown commands. Since I tried to use this app recently, and I don't like forcing a language across my entire system and have to remember to clear it afterwards, I figured I should add a way to do it.

I'm open to other ways to do this if there is a better idea.

### DIFF
--- a/apps/obsidian/obsidian.py
+++ b/apps/obsidian/obsidian.py
@@ -1,13 +1,7 @@
 from talon import Context, Module
 
 mod = Module()
-apps = mod.apps
-apps.obsidian = "app.name: Obsidian"
-
-ctx = Context()
-ctx.matches = r"""
-app: obsidian
-"""
+mod.apps.obsidian = "app.name: Obsidian"
 
 lang_ctx = Context()
 lang_ctx.matches = r"""
--- a/apps/obsidian/obsidian.py
+++ b/apps/obsidian/obsidian.py
@@ -1,0 +1,22 @@
+from talon import Context, Module
+
+mod = Module()
+apps = mod.apps
+apps.obsidian = "app.name: Obsidian"
+
+ctx = Context()
+ctx.matches = r"""
+app: obsidian
+"""
+
+lang_ctx = Context()
+lang_ctx.matches = r"""
+app: obsidian
+not tag: user.code_language_forced
+"""
+
+
+@lang_ctx.action_class("code")
+class CodeActions:
+    def language():
+        return "markdown"
--- a/core/modes/language_modes.py
+++ b/core/modes/language_modes.py
@@ -141,7 +141,7 @@ class Actions:
 
     def code_show_forced_language_mode():
         """Show the active language for this context"""
-        if forced_language
+        if forced_language:
             app.notify(f"Forced language: {forced_language}")
         else:
             app.notify("No language forced")
--- a/core/modes/language_modes.py
+++ b/core/modes/language_modes.py
@@ -1,4 +1,4 @@
-from talon import Context, Module, actions
+from talon import Context, Module, actions, app
 
 # Maps language mode names to the extensions that activate them. Only put things
 # here which have a supported language mode; that's why there are so many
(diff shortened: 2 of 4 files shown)


auscompgeek:
I personally think an app that wants to force a language should override the `code.language` action.

AndreasArvidsson:
> I personally think an app that wants to force a language should override the `code.language` action.

Totally agree. If the problem is that forced language should take precedents over an application specific language then we just need to make the matcher of the forced context more specific. `mode: all` or something.

fidgetingbits:
> I personally think an app that wants to force a language should override the `code.language` action.

Hrm ya that makes sense, is simpler. I'll change it.

Will try the 'mode: all' thing as well, as ya I think we still want forced language to take precedence over overridden code.langauge

AndreasArvidsson:
I will look into if we can make the forced override more specific and always win over application specific

nriley:
To be clear we tried `mode: all` and a couple other things in the community backlog session and couldn't get it to work.

A temporary workaround if you've already got code.language overrides is to do something like this:

https://github.com/nriley/talon_community/commit/b2d584a04805ad65ad411f3ed2fde13f9ca9ae19

but @AndreasArvidsson wasn't too excited about having to do that everywhere and neither am I…

AndreasArvidsson:
It appears that the `app` context has a really high priority. For now I would probably recommend putting `not tag: user.code_language_forced` on the context where you override `code.language`

nriley:
Do you think creating another context for this is better than calling `actions.next`? :-) thought we were trying to avoid having to make changes every time you override `code.language`

nriley:
I'm OK with the `not tag` solution! On the community backlog session apparently people are confused by `actions.next`...

fidgetingbits:
Finally got time to circle back to this. I use the `code.language` override with a context that uses not forced language for obsidian, as suggested. I left in the filename special casing in language_modes.py.

apps/obsidian/obsidian.py
```diff
@@ -0,0 +1,22 @@
+from talon import Context, Module
+
+mod = Module()
+apps = mod.apps
+apps.obsidian = "app.name: Obsidian"
+
+ctx = Context()
```
AndreasArvidsson, line 7: This context appears to be unused

apps/obsidian/obsidian.py
```diff
@@ -0,0 +1,22 @@
+from talon import Context, Module
+
+mod = Module()
+apps = mod.apps
+apps.obsidian = "app.name: Obsidian"
```
AndreasArvidsson, line 5: Please just do `mod.apps.obsidian`

core/modes/language_modes.py
```diff
@@ -57,12 +70,12 @@
     "css": ["c s s"],
     "gdb": ["g d b"],
     "go": ["go", "go lang", "go language"],
-    "r": ["are language"],
+    "r": ["are lang", "are language"],
+    "scm": ["scheme", "s c m", "tree sitter"],
```
AndreasArvidsson, line 74: I don't think this fulfills the criteria `Only put things
# here which have a supported language mode; that's why there are so many`

core/modes/language_modes.py
```diff
@@ -122,3 +139,8 @@ def code_clear_language_mode():
         global forced_language
         forced_language = ""
         ctx.tags = []
+
+    def code_show_forced_language_mode():
+        """Show the active language for this context"""
+        if forced_language
```
AndreasArvidsson, line 145: Should this have in else showing a notification that there is no forced language?

core/modes/language_modes.py
AndreasArvidsson: This entire list is also going away in this pull request
https://github.com/talonhub/community/pull/1528

fidgetingbits: Removed

core/modes/language_modes.py
```diff
@@ -122,3 +138,10 @@ def code_clear_language_mode():
...
```
auscompgeek, line 144:
```suggestion
        if forced_language:
```

knausj85 (approved):
This looks good to me. Leaving open 'till next saturday in case we're holding off for reasons?

AndreasArvidsson (approved):

PR #1453, opened 2024-06-04
feat: Add dark_mode option in settings

Spiteless:
Just learned that dark_mode exists, would like to have an easier settings option for users.

### DIFF
--- a/settings.talon
+++ b/settings.talon
@@ -2,6 +2,9 @@ settings():
     # Adjust the scale of the imgui
     imgui.scale = 1.3
 
+    # If `true`, enable dark mode for talon help menus
+    imgui.dark_mode = false
+
     # If `true`, automatically show the picker GUI when the file manager has focus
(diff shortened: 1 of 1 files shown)


fidgetingbits (approved):

fidgetingbits:
My only minor concern about this is that if people already have `imgui.dark_mode = true` elsewhere in a different settings file, this will potentially clash suddenly and I guess possibly switch them out of dark mode depending on precedence. For now it's kind of an unavoidable problem, so just have to accept it I think. That said, I will wait to see if anyone else says anything.

Spiteless:
We could leave it commented out so that it doesn't override any configs

fidgetingbits:
Oh right, ya that works. Looks like that's the case for a bunch of the settings anyway.

Spiteless:
Changed to `uncomment if` format

PR #1454, opened 2024-06-05
Adds a helper to format and paste

jaresty:
- for example try 'string paste' to paste something as a string


fidgetingbits:
Seems like a good idea. If you keep the order `<user.formatters> paste` I'd almost be tempted to put it into `core/text/text.talon` with the other format-comes-first commands. If you keep it here with the paste commands I'd be inclined to have something like or `paste as <user.formatters>` or similar. That said I'm pretty much good either way, maybe someone else has a stronger preference.

fidgetingbits (approved):

PR #1456, opened 2024-06-07
Add command to focus and snap at the same time

jaresty:
- it's sometimes handy to focus and snap to split two applications side by side

### DIFF
--- a/core/windows_and_tabs/window_management.talon
+++ b/core/windows_and_tabs/window_management.talon
@@ -11,6 +11,12 @@ running list: user.switcher_toggle_running()
 running close: user.switcher_hide_running()
 launch <user.launch_applications>: user.switcher_launch(launch_applications)
 
+focus snap <user.running_applications> <user.running_applications>:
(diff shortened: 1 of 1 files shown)


fidgetingbits:
Seems like a cool idea. I wonder if we should just make it more powerful right away though? Something like:

focus [snap] <user.running_applications> <user.window_snap_position> [and]  <user.running_applications> <user.window_snap_position>

I don't entirely like the idea that atm you'd have to remember which of the windows ends up on left vs right, but that said I guess also having simple to say defaults that don't need explicit direction could still be good.

Also I guess this would behave strangely if the focus results in flipping between desktops for the second app. Not sure the existing apis are good enough to deal with that. Not a big deal imo, but just came to mind.

jaresty:
Here is an alternative that I think might be nicer: https://github.com/talonhub/community/pull/1465

jaresty:
Closing in favor of the alternative.

PR #1457, opened 2024-06-07
tab clone in tag: user.tabs

maxbruening:
I propose to allow users to also say "tab clone". This would make the command consistent with rango. This will become particularly useful once/if community has more applications/contexts that support the "tab duplicate" voice command. Background: I was planning on implementing the command for windows explorer.

To enable compatibility with rango I also created a complementary pull request on that repository: [PR 13](https://github.com/david-tejada/rango-talon/pull/13)


fidgetingbits:
I'm ok with this, even though we normally try not to add multiple ways to say things, since it's trying to sync up with other existing tooling (rango, and we will use clone in neovim-talon as well) 

As noted on slack I have a larger draft PR to rework tabs here https://github.com/talonhub/community/pull/1446

I propose since PR1446 will deprecate lots of tab commands, that we just deprecate "tab duplicate" there in favor of "tab clone", and for now we can merge this PR as is.

I'll wait for at least one other maintainer to agree though.

PR #1460, opened 2024-06-13
Sort help contexts by specificity

AndreasArvidsson:
Instead of just sorting the help context alphabetically we now first sort them by context match specificity. Context with equal specificity are sorted alphabetically.


core/help/help.py
```diff
@@ -438,6 +448,54 @@ def refresh_context_command_map(enabled_only=False):
...
+            return score_b - score_a
+        if short_name_a < short_name_b:
+            return -1
+        if short_name_a > short_name_b:
+            return 1
+        return 0
+
+    return sorted(
+        display_name_to_context_name_map.keys(),
+        key=cmp_to_key(sort_key),
```
auscompgeek, line 495:
```suggestion
        key=get_specificity_score,
```

AndreasArvidsson: Now you're not sorting alphabetically anymore.

auscompgeek: Fair.
```suggestion
        key=lambda name: (get_specificity_score(name), name),
```

AndreasArvidsson: I didn't know you could use a topple like that. Thanks. 
The problem is still that we want specificity to be sorted in descending order. Name should still be in ascending order.

auscompgeek: and this is why I hate `cmp` lol
```suggestion
        key=lambda name: (-get_specificity_score(name), name),
```

AndreasArvidsson: This actually recreates the same behavior with less code. Excellent!

nriley:
We discussed this on the community backlog session today. @AndreasArvidsson and I felt that the current version is essentially a proof of concept and would not be mergeable in its current state. Some potential next steps:

- Figure out some heuristics that create fewer alphabetized sections, at least for stock community matchers, and potentially add separators (lines/blank space) between the alphabetized sections
- Stop using imgui for community's help and instead build something canvas based which would give us a lot more formatting flexibility to be able to explain the richness of the active contexts. See also how we are stuck on #1431 (though we also have another potential short term fix there in splitting it between two different help displays).

If anybody would be interested in working on the first option, feel free to try out/suggest something!

AndreasArvidsson:
What if we group them in three different groups? Application context/with context/global?

AndreasArvidsson:
Something like this

Screenshot 2024-06-23T11-20-06

nriley:
Looks good to me, maybe with some slight wordsmithing "Application-specific:"; "Context-dependent"? We would need to make sure that we don't run over the total # of lines so this doesn't cause help to go off the bottom of the screen.

AndreasArvidsson:
Changing the words is no problem. Updating how max contexts per page is implemented is about bit worse.
Basically this action now needs to be aware of the number of groupings. Not he number of potential groupings, but actual found. 

https://github.com/talonhub/community/blob/3b2e6048fb7832f39cc23ccb0faa83d407a87abc/core/help/help.py#L121-L140

AndreasArvidsson:
Also the setting is number of context per page. Not number of lines per page. I'm very tempted to leave it as it is and the user have to reduce the setting somewhat if the page is to large now. They can always change the setting and go back to the previous behavior if they prefer that?

AndreasArvidsson:
@nriley That pr is now updated with this grouping and your suggested names.

nriley:
Looks great. The only thing that seemed to be an obvious misclassification was text_navigation.talon due to its use of `not app:`:

https://github.com/nriley/talon_community/blob/e11633b3343e25a005dd35991934f6b820f59963/plugin/text_navigation/text_navigation.talon#L1

AndreasArvidsson:
That is harder and requires us to actually parse the context

nriley:
Maybe we can change the matcher for text_navigation.talon then?

AndreasArvidsson:
> Maybe we can change the matcher for text_navigation.talon then?

Probably more filling to use a tag. Can't be only vscode where this should not be active?

nriley:
Yeah, I'm thinking it mostly means "this app has a better way of navigating through text than this", which at the moment means cursorless. Is there something that is getting used consistently by the various non-VSCode cursorless implementations?

AndreasArvidsson:
We do have a Cursorless tag

nriley:
So does that seem reasonable to use — not tag: user.cursorless?

AndreasArvidsson:
Sounds reasonable, but is not application vscode something you added yourself?
https://github.com/talonhub/community/blob/main/plugin/text_navigation/text_navigation.talon

nriley:
Wow, that's embarrassing. Let's go ahead and merge this and I'll clean up my act :-)

nriley (approved):

jaresty:
Is there any trick to seeing this sort? I tried running 'help context' but I didn't see any changes, but I'm not sure I was looking in the right place 🤔

nriley:
This does sort the output of "help context", but is more useful with "help active".

jaresty:
That's what I was missing. Thank you

PR #1464, opened 2024-06-23
Added notepad application context

AndreasArvidsson:


apps/notepad/notepad.talon
```diff
@@ -0,0 +1,5 @@
+app: notepad
+-
+
+tag(): user.tabs
```
phillco, line 4: We live in a frightening future :D

phillco (approved):

apps/notepad/notepad.py
```diff
@@ -0,0 +1,22 @@
...
+ctx.matches = r"""
+app: notepad
+"""
+
+
+@ctx.action_class("win")
+class win_actions:
+    def filename():
+        filename = actions.win.title().split(" - ")[0]
+        if "." in filename:
```
fidgetingbits, line 20: Part of of PR1451 https://github.com/talonhub/community/pull/1451 is to allow extra filename matching where a . isn't included, for example allowing the filename Dockerfile to trigger docker tags. This won't work with that type of file.

fidgetingbits (commented):
One noted issue if a draft PR I have gets accepted eventually.

apps/notepad/notepad.talon
phillco: Actually, is this conditional on Win11 notepad?

AndreasArvidsson: Yes. You think I should remove it for that reason?

apps/notepad/notepad.py
AndreasArvidsson: That pull request doesn't seem to care that all other applications use this implementation already

https://github.com/talonhub/community/blob/b337c2193d30ae67be2a10b3dd4c5487f1c6821c/apps/vscode/vscode.py#L150-L151
https://github.com/talonhub/community/blob/b337c2193d30ae67be2a10b3dd4c5487f1c6821c/apps/notepad++/notepad++_win.py#L70-L72
https://github.com/talonhub/community/blob/b337c2193d30ae67be2a10b3dd4c5487f1c6821c/apps/visualstudio/visual_studio.py#L107-L109
etc..

fidgetingbits: Yep, is why I didn't actually request a change here, just saying.

AndreasArvidsson: Ok since this appears to be the community standard we just leave it for now and have to think about a solution in the future on how to support filenames without a dot in them. Sounds good?

apps/notepad/notepad.talon
fidgetingbits: I would just keep it personally, especially since Windows 10 is EOL in ~1 year anyway? It's usefulness when the feature exists probably outweighs the minor annoyance to have those commands in context when it doesn't? That said I don't use windows and it won't affect me, so take my opinion lightly.

AndreasArvidsson: That was kind of my thinking. Most people will soon be on windows 11. And it's not like we don't have commands available in Talon in some context that doesn't work already.

phillco: Yeah, I think that's fine, I just wanted to discuss it / have our thinking on the record.

If we had an easy `win10`/`win11` tag already available, it might be nice to conditionalize it (although, one extra file 🤢), but agree it's not worth it for this alone. I do think the idea in general is something we would be open to in the future though.

phillco (approved):

PR #1465, opened 2024-06-24
Adds some snap helpers to split windows to the left in various ways

jaresty:


core/windows_and_tabs/window_management.talon
```diff
@@ -17,5 +17,43 @@ snap last [screen]: user.move_window_previous_screen()
 snap screen <number>: user.move_window_to_screen(number)
 snap <user.running_applications> <user.window_snap_position>:
     user.snap_app(running_applications, window_snap_position)
+snap left <user.running_applications> <user.running_applications>:
+    user.switcher_focus(running_applications_1)
+    user.snap_window_to_position("left")
```
phillco, line 22: `user.snap_app` combines these actions, which will reduce the code duplication here.

In theory, it would also allow us to make the rearrangement faster and more elegant -- remember that we can move/resize windows without having to focus the applications they belong to. Although, I imagine you always focus to make sure that they aren't occluded by other applications, and that the other application is second in your application switcher LRU.

core/windows_and_tabs/window_management.talon
```diff
@@ -17,5 +17,43 @@ snap last [screen]: user.move_window_previous_screen()
 snap screen <number>: user.move_window_to_screen(number)
 snap <user.running_applications> <user.window_snap_position>:
     user.snap_app(running_applications, window_snap_position)
+snap left <user.running_applications> <user.running_applications>:
```
phillco, line 20: "snap left" doesn't feel like the right grammar for this, given that the entire screen will be filled by these two applications after the command, right? Given that you also have things like `snap clock`, this second word feels like a noun referring to the arrangement. It seems to me like you would want something like "snap side by side", only shorter?

core/windows_and_tabs/window_management.talon
```diff
@@ -17,5 +17,43 @@ snap last [screen]: user.move_window_previous_screen()
...
+    user.snap_window_to_position("bottom right")
+    user.switcher_focus(running_applications_1)
+    user.snap_window_to_position("left")
+snap counterclock <user.running_applications> <user.running_applications> <user.running_applications>:
+    user.switcher_focus(running_applications_2)
+    user.snap_window_to_position("top left")
+    user.switcher_focus(running_applications_3)
+    user.snap_window_to_position("bottom left")
+    user.switcher_focus(running_applications_1)
+    user.snap_window_to_position("right")
```
phillco, line 57: As you're getting into many commands, this feels like a good candidate for a list that would index into a data structure mapping the name of the arrangement to the list of window positions (a bit like to how `_snap_positions` works today). This would reduce the duplication and make it much easier to add new arrangements.

Then you would just have one command definition:

```
snap {user.window_arrangement} <user.running_applications>+: user.snap_arrangement(window_arrangement, running_applications_list)
```

phillco (changes requested):

core/windows_and_tabs/window_management.talon
jaresty: snap_app doesn't take a string like "right" but a more complex object-I'll need to make a new version of that method to do this, but that's ok I think.

core/windows_and_tabs/window_management.talon
jaresty: I considered "snap split" but you might want to focus on the right or the left split, which led me to think of "snap left" to mean "the main one should be to the left". Open to suggestions though.

core/windows_and_tabs/window_management.talon
jaresty: I like this idea. I'm still pretty uncomfortable with building these dynamic workflows from scratch. Please feel free to change this if it's obvious hope to do so-I'll give it a shot when I have some time.

core/windows_and_tabs/window_management.talon
phillco: You would just need to index into `_snap_positions`. ` <user.window_snap_position>` effectively does this for you if used, but it should be fine to have an action that also takes the string name (seems useful)

core/windows_and_tabs/window_management.talon
phillco: Sure, I'd be happy to do this or walk through it if you want to pair. Either way is fine

core/windows_and_tabs/window_management.talon
phillco: `snap split` sounds best to me.

> but you might want to focus on the right or the left split

Ah, that makes sense now that you point it out. Hm, I'm not sure either.

core/windows_and_tabs/window_management.talon
jaresty: Thank you for the offer. Pairing sounds nice if I can find time. Let me think about it a bit.

phillco: Given that it's a bit more complex than I realized (because we have to indicate which window will be the active one after arrangement, to your point above), it might be best if I prototype it quickly first, just to make sure it's as straightforward as I think it is. But I'm still happy to pair on it afterwards (or anything else, for that matter)

jaresty: Sounds good. Please go ahead-looking forward to seeing what you put together.

core/windows_and_tabs/window_management.talon
jaresty: I could see an argument for making `snap split` always focus the left split as a first pass at this; Maybe we'll learn something that will inform how it could work better.

core/windows_and_tabs/window_management.talon
AndreasArvidsson: The only problem with using despot can form as an identifier is that if you use it in a script somewhere and then changed the spoken form your script is now broken. If you want to have an action that takes a string we should probably separate the spoken form from the identifier.

jaresty: Sorry, I didn't quite follow the suggestion-what do you think we should do, @AndreasArvidsson ?

AndreasArvidsson: Just that the list of available snap positions should have two different strings. One that is a spoken form and one that is an identifier. If I changed a spoken form from `left` to `lefter` the code using `left` is not valid anymore.

jaresty: Makes sense. Is this typically done with a talon list?

AndreasArvidsson: Not typically, but that's definitely one good implementation. Look at four matters for example. 

Here the keys are the identifier. These you can use from your own scripts and they will work even if you changed the spoken form.
https://github.com/talonhub/community/blob/7619d2c369c189667f1ba5f0089b843517e6a507/core/text/formatters.py#L133-L170

Here is the map where the keys is the spoken form that maps to the identifiers. This could have been a Talon list file
https://github.com/talonhub/community/blob/7619d2c369c189667f1ba5f0089b843517e6a507/core/text/formatters.py#L173-L189

jaresty:
I changed this to use the internal apis directly instead. I got rid of snap left in favor of using a custom snap split function. The snapping seems to work pretty well but sometimes the focus is not where I expect it to be. I think that's probably okay though. I think there is still value in having the snap functions available to use at any moment.

core/windows_and_tabs/window_management.talon
jaresty: Thanks that makes sense. I changed this in favor of using the internal apis which I think addresses this concern.

jaresty:
This seems to be working pretty well-is there anything that I'm missing that needs to be looked at?

jaresty:
I learned how to clean this up a bit using talon captures and talon lists. There are only two rules now for splitting with two apps or splitting with three apps. I reused the keys of the the snap window dictionary in python so that I would have some human readable names for the positions that I am snapping to. In addition I switched to `window.focus()` since what I really wanted was for the windows to be focused in the front on the screen. This makes it a lot more usable.

jaresty:
@phillco was I able to address your concerns? Let me know if I can switch things up to make it reasonable for you :)

core/windows_and_tabs/window_snap.py
```diff
@@ -272,14 +276,38 @@ def __init__(self, left, top, right, bottom):
     "fullscreen": RelativeScreenPos(0, 0, 1, 1),
 }
 
+_split_positions = {
+    "split": [
+        _snap_positions["left third"],
+        _snap_positions["center third"],
+        _snap_positions["right third"],
+    ],
```
phillco, line 284:
```suggestion    
    "split": [
    [
        _snap_positions["left"],
        _snap_positions["right"],
    ],
    [
        _snap_positions["left third"],
        _snap_positions["center third"],
        _snap_positions["right third"],
    ]],
```

Let's make the values a list of lists, with different numbers of inputs. That way the data structure can support snapping layouts with 2/3/4/etc windows ergonomically. The action can then look at the number of windows that were passed, and find the layout that matches that length.

As an added bonus, if there isn't one, it can show a useful error explaining this ("layout 'clock' requires 3, 4, or 5 windows, but 2 were given")

The top level will still be string keys for simplicity (the spoken form) so we don't have to duplicate the list for each n-combination.

For the value, instead of a list of lists, we could use a dictionary with numeric keys (indicating the number of applications supported). But that's redundant given that the length of the value itself already has that information; it's less work for the user to just use list of lists. Make the code do the hard work of figuring out which one matches.

core/windows_and_tabs/window_snap.py
```diff
@@ -310,6 +338,30 @@ def snap_app(app_name: str, position: RelativeScreenPos):
...
+        window = _get_app_window(app_one)
+        window_two = _get_app_window(app_two)
+        _snap_window_helper(window_two, _snap_positions["right"])
+        _snap_window_helper(window, _snap_positions["left"])
+        window_two.focus()
+        window.focus()
+
+    def snap_split_three(
+        positions: list[RelativeScreenPos], app_one: str, app_two: str, app_three: str
+    ):
```
phillco, line 352: You can just make this one single action with lots of application parameters, and just make the ones greater than 2 optional:

```suggestion
    def snap_layout(
        positions: list[RelativeScreenPos],
        app_one: str,
        app_two: str,
        app_three: Optional[str],
        app_four: Optional[str],
        app_five: Optional[str]
    ):
```
Might as well support up to 5 in case people want to make more interesting layouts.

I don't think there's an explicit `null` in Talonscript but the shorter-n versions can just leave those parameters off:

```
user.snap_layout("abc", "def", "jkl")
```

`app_four` and `app_five` will thus be `None`.

That way you don't have to duplicate the action for each number of applications.

Also I think it'd be helpful to give this functionality a useful name, to distinguish it from the existing single-window snapping functionality. Since "split" is itself a layout I think it should probably have a broader name like "layout" or "arrangement" (not strongly opinionated on the name itself, just that it has one). Do any of those sound good to you, or perhaps do you have a better one? 😁

core/windows_and_tabs/window_snap.py
phillco: I had a thought that I've talked myself out of: we could simplify this a bit further. Instead of needing to index into `_snap_positions` over and over, you can just use the string keys (`["left
 third", "center third", "right third"]`). The action implementation already has access to `_snap_positions` so it could do a simple array map to get those values. This would make the definitions shorter and less duplicative.

The only downside from my perspective is that in your current version you can use an anonymous `RelativeScreenPos` here without having to put it in `_snap_positions`, i.e. if there was a layout that only made sense as part of arrangement, not individual snapping. Requiring things be in `_snap_positions` means it would have to have a spoken form, which I think is a worse requirement.

You could do something where the values could be either string keys (in which case they are assumed to be `_snap_positions` keys and the action would look them up, or `RelativeScreenPos` objects (in which case the action would use them as is), but .... may not be worth the magic.

So the current way is fine, just posting this comment in case the line of thought is interesting.

core/windows_and_tabs/window_management.talon
```diff
@@ -17,5 +17,9 @@ snap last [screen]: user.move_window_previous_screen()
 snap screen <number>: user.move_window_to_screen(number)
 snap <user.running_applications> <user.window_snap_position>:
     user.snap_app(running_applications, window_snap_position)
+snap <user.window_split_position> <user.running_applications> <user.running_applications>:
+    user.snap_split_two(running_applications_1, running_applications_2)
```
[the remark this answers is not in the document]
phillco: Per discussion, if the version with two windows only really supports splitting side by side, the grammar should look like this. But I think we had a much better idea detailed below, to make the grammar arbitrarily support layouts with different numbers of windows.

core/windows_and_tabs/window_snap.py
```diff
@@ -272,14 +276,38 @@ def __init__(self, left, top, right, bottom):
...
 @mod.capture(rule="{user.window_snap_positions}")
 def window_snap_position(m) -> RelativeScreenPos:
     return _snap_positions[m.window_snap_positions]
 
 
+@mod.capture(rule="{user.window_split_positions}")
+def window_split_position(m) -> list[RelativeScreenPos]:
+    return _split_positions[m.window_split_positions]
+
+
```
phillco, line 307: I don't actually think you need a capture here, the "I want a dictionary so the spoken form is a certain string but the action gets the object that corresponds to that string" mapping can be done with a simple list.

(There are lots of places in community that seem to define unnecessary captures that just wrap a list, for reasons I'm not sure about.)

But I'll follow up to make sure. Feel free to hold off until I've checked 😅

phillco (changes requested):

core/windows_and_tabs/window_snap.py
lunixbochs: lists are a string->string mapping, they can't map to objects

phillco: derp, right.

core/windows_and_tabs/window_snap.py
```diff
@@ -272,14 +276,47 @@ def __init__(self, left, top, right, bottom):
...
+    "split": {
+        2: [_snap_positions["left"], _snap_positions["right"]],
+        3: [
+            _snap_positions["left third"],
+            _snap_positions["center third"],
+            _snap_positions["right third"],
+        ],
+    },
+    "clock": {
+        2: [_snap_positions["left"], _snap_positions["right"]],
```
phillco, line 289:
```suggestion
```

core/windows_and_tabs/window_snap.py
```diff
@@ -272,14 +276,47 @@ def __init__(self, left, top, right, bottom):
...
+        3: [
+            _snap_positions["left"],
+            _snap_positions["top right"],
+            _snap_positions["bottom right"],
+        ],
+    },
+    "counterclock": {
+        2: [_snap_positions["left"], _snap_positions["right"]],
```
phillco, line 297:
```suggestion
```

core/windows_and_tabs/window_management.talon
```diff
@@ -17,5 +17,7 @@ snap last [screen]: user.move_window_previous_screen()
 snap screen <number>: user.move_window_to_screen(number)
 snap <user.running_applications> <user.window_snap_position>:
     user.snap_app(running_applications, window_snap_position)
+snap <user.window_split_position> <user.running_applications> <user.running_applications>+:
```
phillco, line 20:
```suggestion
# <user.running_applications> is here twice to require at least two applications.
snap <user.window_split_position> <user.running_applications> <user.running_applications>+:
```

core/windows_and_tabs/window_snap.py
```diff
@@ -310,6 +345,22 @@ def snap_app(app_name: str, position: RelativeScreenPos):
...
+        positions_by_count: Dict[int, list[RelativeScreenPos]],
+        apps: list[str],
+    ):
+        """Split the screen between multiple applications."""
+        try:
+            positions = positions_by_count[len(apps)]
+        except KeyError:
+            raise NotImplementedError(
+                f"There is no such layout yet defined for {len(apps)} items"
+            )
```
phillco, line 358:
```suggestion
            supported_layouts = ", ".join(positions_by_count.keys())
            message = f"{len(apps)} applications given but chosen layout only supports {supported_layouts}"
            actions.app.notify(message, "Cannot arrange")
            raise NotImplementedError(message)
```

phillco (approved):

phillco (approved):

PR #1466, opened 2024-06-24
Added mute/microphone disabled to mode indicator

AndreasArvidsson:


plugin/mode_indicator/mode_indicator.py
```diff
@@ -180,10 +185,19 @@ def on_update_settings(updated_settings: set[str]):
         update_indicator()
 
 
+def poll_microphone():
+    global current_microphone
```
phillco, line 189:
```suggestion
    # Ideally, we would have a callback instead of needing to poll. https://github.com/talonvoice/talon/issues/624
    global current_microphone
```

plugin/mode_indicator/mode_indicator.talon
```diff
@@ -11,6 +11,8 @@ settings():
     user.mode_indicator_color_alpha = 0.75
     # Grey gradient
     user.mode_indicator_color_gradient = 0.5
+    # Black color for mute (Microphone is disabled/None)
```
phillco, line 14:
```suggestion
    # Black color for when the microphone is muted (set to "None")
```

phillco (approved):
lgtm, my only question is whether the 500ms poll is acceptable to other maintainers. I know historically we've tried to avoid polling where possible, but it's not avoidable in this case. Not really sure if e.g. 1s would be significantly better.

I linked to https://github.com/talonvoice/talon/issues/624, assuming it's the best issue to keep on record for that.

phillco (approved):

PR #1467, opened 2024-06-24
Custom subtitles

AndreasArvidsson:
Custom subtitles and notifications with settings to tweak color, position and timeout.


nriley:
In the backlog session we agreed to change `user.subtitles_all_screens` to an enum including values such as {all screens, screen of focused window/app, screen of current cursor, main screen}.

nriley (changes requested):
(see above)

AndreasArvidsson:
@nriley Updated. I only have one screen on this computer. I'm pretty sure about the code, but if you have multiple screens please give the different options a try.

nriley (approved):

nriley:
@AndreasArvidsson all the options seem to make sense. I made a couple documentation changes — if you're OK with them, I think we're ready to merge.

PR #1470, opened 2024-06-30
Add tag for disabling screenshots (Part of the Better Exam Support Project)

FireChickenProductivity:
Add the tag user.disable_screenshots to make it easier for users to disable screenshot commands in certain circumstances. This was part of the project discussed in a community meeting to add built in functionality to community itself to make it easier to disable certain commands under specific circumstances, such as when taking an exam.

### DIFF
--- a/plugin/screenshot/screenshot.talon
+++ b/plugin/screenshot/screenshot.talon
@@ -1,6 +1,7 @@
 mode: command
-and not tag: user.disable_screenshots
+and not tag: user.screen_shots_disabled
 -
+
 ^grab screen$: user.screenshot()
 ^grab screen <number_small>$: user.screenshot(number_small)
 ^grab window$: user.screenshot_window()
--- a/plugin/screenshot/screenshot.talon
+++ b/plugin/screenshot/screenshot.talon
@@ -1,5 +1,5 @@
 mode: command
-and not tag: user.screen_shots_disabled
+and not tag: user.disable_screenshots
 -
 
 ^grab screen$: user.screenshot()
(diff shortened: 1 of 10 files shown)


tags/feature_disabling/feature_disabling_tags.py
```diff
@@ -0,0 +1,4 @@
+from talon import Module
+
+mod = Module()
+mod.tag("screenshots_disabled", desc="Activating this tag disables screenshot commands")
```
AndreasArvidsson, line 4: Everything in the plugin directory needs to be self contained. ie no dependencies on files outside of that directory. I would just move `mod.tag()` to `screenshot.py`

plugin/screenshot/screenshot.talon
```diff
@@ -1,3 +1,7 @@
+mode: command
```
AndreasArvidsson, line 1: You don't need mode. `not tag: user.screenshot_disabled` is enough

I would probably also go `screenshot` in singular just because that is what we are already using.

AndreasArvidsson (changes requested):

tags/feature_disabling/feature_disabling_tags.py
FireChickenProductivity: Done. Thank you for the suggestion!

plugin/screenshot/screenshot.talon
FireChickenProductivity: I made those changes. I agree with making my name consistent with the existing name.

AndreasArvidsson (approved):

nriley (approved):

PR #1471, opened 2024-07-01
add nix support to app switcher

fidgetingbits:
This tweaks the app switcher code a bit to work when using applications installed via nix on both linux and darwin. 

It also removes the `/usr/bin/` assumption for linux executable path, and uses `which` instead to find the real path instead. Due to pulling out the path that way now, I had to tweak the regex slightly to get it to also match if there's only one argument to the app, as it will no longer be prefixed with a space.

@DonnerWolfBach this should hopefully work for you.


fidgetingbits:
It's probably that the path I added is home-manager specific. I've switched to just reading in all the application folders from XDG_DATA_DIRS, which should hopefully work for you, and is better approach to what I had originally done anyway.

I left the original set of folders because we know they work, and I've not used flatpaks, snaps, and the like, so not sure if they get added XDG_DATA_DIRS (presumably they would..), and don't have a way to test atm.

nriley:
Now we are starting to execute code while evaluating the various platform lists of directories, it may make sense to move them behind the if statements that check for platform.

Similarly it looks like you are executing the code that generates a list of folders based upon `XDG_DATA_DIRS` regardless of platform. Is this because this could be defined on Windows or Linux or could it also be guarded by a platform check?

fidgetingbits:
> Now we are starting to execute code while evaluating the various platform lists of directories, it may make sense to move them behind the if statements that check for platform.
> 
> Similarly it looks like you are executing the code that generates a list of folders based upon `XDG_DATA_DIRS` regardless of platform. Is this because this could be defined on Windows or Linux or could it also be guarded by a platform check?

Agree it makes sense to separate it into per-platform now. I've been test driving some changes for this in my own branch the last couple days on linux/mac, so will lift it into this PR soon.

fidgetingbits:
Cleaned it up per suggestions

DonnerWolfBach:
@fidgetingbits tested it - it works now for me. Thank you very much!
I don't have flatpaks or snaps on my system at the moment though. I can try it though, will take me some time though

knausj85 (approved):
This looks good to me. I tested the change on Mac and windows and everything seems fine.

PR #1472, opened 2024-07-01
Added Warp App

adamambrosius:
Adds a basic configuration to be able to use the warp terminal the same as other terminal apps.


apps/warp/warp.talon
```diff
@@ -0,0 +1,10 @@
+os: mac
+app: warp
```
fidgetingbits, line 2: You'll want to add an explicit match in core/application_matches.py. See [here](https://talon.wiki/Customization/Talon%20Framework/apps#non-explicit-app-header-matching) for why it's working without an explicit match.

fidgetingbits (changes requested):

apps/warp/warp.talon
adamambrosius: Thanks for the pointer. The example had it in it's own python file instead of in the application_matches.py. Let me know if I need to move this there intstead

apps/warp/warp.talon
```diff
@@ -0,0 +1,10 @@
+os: mac
```
nriley, line 1: This is redundant — can remove because it's declared in `mod.apps.warp`.

apps/warp/warp.talon
nriley: No, I think it's fine where it is… we should really get `core/application_matches.py` split out into multiple files rather than vice versa.

nriley (approved):

PR #1476, opened 2024-07-05
Fix typo in plugin/mouse/mouse.py

wenkokke:
The code uses `tracking.control_toggle1` which I assume was caused by some noise.


wenkokke:
I guess it isn't. Is this the legacy naming policy?

PR #1477, opened 2024-07-06
Add an example for how to override key settings

jaresty:


fidgetingbits (changes requested):

fidgetingbits:
Feel free to ignore my suggestion if you think I'm over thinking.

README.md
```diff
@@ -290,6 +290,25 @@ The most commonly adjusted settings are probably
 
 Also, you can add additional vocabulary words, words to replace, search engines and more. Complete the community setup instructions above, then open the `settings` folder to see the provided CSV files and customize them as needed.
 
+## Overriding settings without forking
```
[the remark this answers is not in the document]
jaresty: I'm open to the suggested changes. My Internet is out so can't do it right now but feel free to suggest a change or I'll look at it when I can.

phillco:
```suggestion
## 💡 Tip: Overriding cleanly
```

README.md
```diff
@@ -290,6 +290,25 @@ The most commonly adjusted settings are probably
...
+
+To override a setting without forking the community repository, create talon files with more specific tags than the default. This customizes your setup without affecting the base repository. Any tag will work. For example, setting the language on a talon file or list will override the community settings. This is useful for personal customizations or workflow-specific tweaks.
```
phillco:
```suggestion
You can override Talon lists by creating a new `.talon` file of your own, rather than changing the existing list in the repository.

This reduces how much manual `git merge`-ing you'll have to do in the future, when you go to merge new versions of this repository (colloquially called "upstream") with your local changes. This is because _new_ files you create will almost never conflict with upstream changes, whereas changing an existing files (especially hot spots, like commonly-customized lists) frequently do.

Your override files can even live outside of the `community` repository (anywhere in `~/.talon/user`), if you prefer, further simplifying merging.

To do so, simply create a `.talon` file with a more specific [context header](https://talon.wiki/Customization/talon-files#context-header) than the default. (For example, `lang: en` or `os: mac` main). Talon ensures that the most precise header (your override file) wins.
```

README.md
```diff
@@ -290,6 +290,25 @@ The most commonly adjusted settings are probably
...
+-
+
+# Example
+
+rose: cmd
+troll: control
+shift: shift
+alt: alt
+```
+
```
phillco, line 311:
```suggestion
For example, to override `user.modifier_key`, you could create `modifier_keys_MYNAME.talon`:

```talon
list:  user.modifier_key
language: en
-

# My preferred modifier keys
rose: cmd
troll: control
shift: shift
alt: alt
```

Don't forget that we have a Talon language formatter on GitHub. Also I find it's always helpful to give example filenames, even if they are straightforward to us.

README.md
phillco: I think we should probably focus on describing the benefits in simple terms (which I think for most people is "how do I make changes in the best possible way?" / "how do I reduce the pain of merging with community?" which is absolutely something people run into).

I think it's great that you're raising the need to reduce forking, but I don't think that philosophy is something the average new user would readily be familiar with (keep in mind a decent number of users aren't familiar with git and start by downloading the repository as a .zip), and leading with "without forking" sort of assumes that you already know why that's valuable. But making changes that won't be a pain to merge in the future is something almost all of our users have to deal with, so I would personally lead with that.

I think if we get more content about making changes cleanly/sustainably, we could probably break that out to a dedicated page/article, and from there get into depth about the philosophy of forking community, for people who are interested.

(As as aside, I wish we had a better term to capture the classical OSS meaning of the word "fork". GitHub has blurred the term significantly, since they encourage everyone to make a fork in order to open a pull request, and it's a painless one click operation that isn't an obviously bad thing. I think this makes it more difficult to talk about why "forking" is a bad thing, because some people will just think you mean creating your own copy (necessarily, sometimes!) of the repository on the web. "hard fork" is the closest term I can think of that comes closest to the original meeting, but I'd be interested in if there are any others. This will probably help your broader discussion goals by avoiding this ambiguity.)

phillco (changes requested):

README.md
phillco: I think what you're referring to as "tags" is best described as the context header, which can include tags, but also other matchers, like `os`, that are not tags.

phillco:
Looks like I reviewed an older version, but if you need me to rebase my suggestions I can

jaresty:
@phillco  would you mind rebasing your suggestions? I'm inclined to accept them all but it's a bit awkward since they are behind the latest commit.

README.md
jaresty: I'm fine with changing the wording to emphasize the benefits you called out here. The target audience I had in mind is not the average talon user but rather someone coming to talon who might be open / interested in contributing to community. Those folks would likely already be familiar with the benefits of avoiding forking but wouldn't know how to manage the configuration necessary to do so without struggling (as I did) because of the complexity involved in configuring community without editing the repository.

README.md
AndreasArvidsson: Agree tags is just a subset of the available matchers.

Also in the suggestion it should probably be a new `.talon-list` file

Also `~/.talon` doesn't exist on windows machines. Something like `TALON_HOME` is probably a better way of referring to it.

nriley:
On the community backlog session we discussed using a purpose-built tag for this purpose so we do not have to abuse things like "os". As an analog to CSS we discussed creating a tag called user.important that would always be active as declared in community and then encouraging people to use this.

Thoughts?

jaresty:
I love that idea, @nriley

README.md
phillco: > Also ~/.talon doesn't exist on windows machines. Something like TALON_HOME is probably a better way of referring to it.

Good point. We can just say "Talon home directory", per https://talonvoice.com/docs/#getting-started.

phillco:
That's a nice idea, as `lang: en` technically isn't correct either (but is a lot less problematic than `os`)

README.md
jaresty: I updated to match this text.

README.md
```diff
@@ -290,6 +290,27 @@ The most commonly adjusted settings are probably
 
 Also, you can add additional vocabulary words, words to replace, search engines and more. Complete the community setup instructions above, then open the `settings` folder to see the provided CSV files and customize them as needed.
 
+## 💡 Tip: Overriding cleanly
+
+You can override Talon lists by creating a new `.talon` file of your own, rather than changing the existing list in the repository.
```
AndreasArvidsson, line 295: This needs to be `.talon-list` not `.talon`

README.md
```diff
@@ -290,6 +290,27 @@ The most commonly adjusted settings are probably
...
+This reduces how much manual `git merge`-ing you'll have to do in the future, when you go to merge new versions of this repository (colloquially called "upstream") with your local changes. This is because _new_ files you create will almost never conflict with upstream changes, whereas changing an existing files (especially hot spots, like commonly-customized lists) frequently do.
+Your override files can even live outside of the `community` repository (anywhere in the Talon user directory), if you prefer, further simplifying merging.
+To do so, simply create a `.talon` file with a more specific [context header](https://talon.wiki/Customization/talon-files#context-header) than the default. (For example, `lang: en` or `os: mac` main). Talon ensures that the most precise header (your override file) wins.
```
AndreasArvidsson, line 298: Same here

README.md
```diff
@@ -290,6 +290,27 @@ The most commonly adjusted settings are probably
 
 Also, you can add additional vocabulary words, words to replace, search engines and more. Complete the community setup instructions above, then open the `settings` folder to see the provided CSV files and customize them as needed.
 
+## 💡 Tip: Overriding cleanly
+
+You can override Talon lists by creating a new `.talon` file of your own, rather than changing the existing list in the repository.
+This reduces how much manual `git merge`-ing you'll have to do in the future, when you go to merge new versions of this repository (colloquially called "upstream") with your local changes. This is because _new_ files you create will almost never conflict with upstream changes, whereas changing an existing files (especially hot spots, like commonly-customized lists) frequently do.
```
AndreasArvidsson, line 296: Existing `file`

README.md
AndreasArvidsson, line 298: And here

AndreasArvidsson, line 298: We prefer the wording `specific` instead of precise

AndreasArvidsson (changes requested):

AndreasArvidsson (approved):

phillco (approved):

PR #1482, opened 2024-07-06
Adds some configuration recommendations to contributing read me

jaresty:


CONTRIBUTING.md
```diff
@@ -11,3 +11,9 @@ Each of the principles is numbered for easy referencing. The body is formatted a
 
 - P03 - Use the `app.bundle` matcher for apps on OSX. This is the least ambiguous way of referring to a particular program.
 - P04 - Use both `app.name` and `app.exe` matchers for apps on Windows. That is the context should OR together one matcher of each type. Apparently the [MUICache](https://www.magnetforensics.com/blog/forensic-analysis-of-muicache-files-in-windows/) can break, perhaps making one of these matchers stop working.
+
+# Configuration
+
+Sometimes it is useful to allow for users to specify the configuration settings that they would like. We often use the `settings.talon` file to do this.
+
+Since some users prefer to set settings using a file outside of this repository, it is easier for those users if we do not hard code the settings values into the `settings.talon` file. Instead, it is helpful to provide an example setting that is commented out in that file and set a default value in Python code in case the setting is not present in the `settings.talon` file.
```
knausj85, line 19: I slightly disagree with this. I think settings.talon should contain the sensible defaults in one place. It's also easy for users to override elsewhere by introducing their own more specific talon file via a tag, hostname, or similar. Maybe we should add a note about that for this?

jaresty: I hear you. If it's a sensible default, it could make sense for the default not be necessary to be configured at all-essentially allowing the code to define the default if it is not passed in as configuration. That would make the code more robust in the case that the configuration was not present. As a new user when I encountered this file for the first time I found it extremely confusing because it seemed like there was no way to override these settings without modifying the repository. I understand how to do that now but it took me months to figure out. In the talon AI tools repository we followed this pattern instead:

1. Include example settings commented out in the settings talon file
2. If the configuration is not passed default it in python code

By using this approach we were able to have sensible defaults but also make it obvious how to override things if you wanted to set your own settings. Does that make sense? I don't want to die on this hill, so if this is not really something you want to negotiate about then I understand. My feedback as a new user is that it was difficult to learn how to use for me.

AndreasArvidsson:
We talked this over on the community meetup and we don't think that we should enforce all default settings to be commented by default. For now you just need to make your own settings file more specific if you don't want to edit the default one.

PR #1483, opened 2024-07-07
Added gamepad support

AndreasArvidsson:
* Added default gamepad bindings. Hidden behind a tag.
* Added gamepad tester ui

I'm not sure about that bindings being disabled by default and hidden behind a tag. That makes it so that whenever someone wants to try out a gamepad they need to figure out how to enable it first. If you don't have a gamepad connected these bindings will do nothing. Maybe we should have it be active by default and have a tag to disable it in case someone wants to use their own custom gamepad bindings? This code is flexible enough that you can override every single key in a Talon action so you shouldn't really need to do anything else, but you know.
What are peoples take on this? Enabled or disabled by default?


jaresty:
My vote would be to enable it by default since it is more discoverable that way.

jaresty:
I'm imagining it would be quite nice if you could toggle talon listening by holding down a gamepad button. I also wonder if everyone would want the same behavior. Is there an easy way to override behaviors for one or two buttons?

AndreasArvidsson:
That's one of the things I use a gamepad button for

Just overwrite the action. Every button has a down/pressed and an up/release action

jaresty:
That's great. Can you overwrite the action in a separate git repo? (I.e. without forking?)

jaresty:
Fwiw, if two of us want to configure a button to toggle talon listening, that might be enough to suggest it could be a reasonable default.

lunixbochs:
Enabling some gamepad bindings by default may be incredibly surprising for some users, as it will trigger if they play a game with a gamepad.

I'm confused why you need the actions here outside the gamepad tester - gamepad overrides should work fine from .talon without the added layer of indirection.

AndreasArvidsson:
> That's great. Can you overwrite the action in a separate git repo? (I.e. without forking?)

Absolutely. The purpose of the Talon context system is that the most specific context wins no matter where in the user directory that file is.

AndreasArvidsson:
> Enabling some gamepad bindings by default may be incredibly surprising for some users, as it will trigger if they play a game with a gamepad. I'm confused why you need the actions here outside the gamepad tester - gamepad overrides should work fine from .talon without the added layer of indirection.


> Enabling some gamepad bindings by default may be incredibly surprising for some users, as it will trigger if they play a game with a gamepad. I'm confused why you need the actions here outside the gamepad tester - gamepad overrides should work fine from .talon without the added layer of indirection.

Oh it does that work now? I'm pretty sure gamepad was one of those things where you got duplication if you specified multiple. That is why I'm doing the context on the python side instead. I just tried it and appears to be working. Then I can simplify this pull request greatly. Thanks!

knausj85 (approved):

AndreasArvidsson:
So the question is do we merge this? 
* Today if you specify multiple identical gamepad bindings it does not behave the same way as voice commands. All of them run. 
* With the above in mind we have two different options:
    1. Only ship the gamepad tester. User can look at that and then implement their own bindings. The drawback here is that things like controlling the mouse cursor and scrolling needs python code. Also if the user wants the buttons on their gamepad to do different things in different context they running to the above problem.
    2. Do what I done in this pr. Add a Talon action for each button that the user then can override by a more specific context. This is my preferred solution and what I use personally.

@knausj85 @nriley @pokey @phillco

knausj85:
I'd vote for #2.

Is the running of multiple bindings a Talon bug?

AndreasArvidsson:
Since everything else in Talon is operating on specificity I would assume so?

jaresty:
#2 sounds amazing to me 🙏 (just my unsolicited 2 cents 😅)

AndreasArvidsson:
@phillco Could you just test this since I know you have a gamepad?

AndreasArvidsson:
Sounds like people are overall happy with this. We can always revisit this with future updates if it's not perfect after more people have a chance to test it out.

PR #1485, opened 2024-07-07
Fixing display of scroll mode for wheel gaze

bradmajewski:


bradmajewski:
Forced pushed due to needing to fix the author of the commit

plugin/mouse/mouse.py
```diff
@@ -206,10 +206,12 @@ def mouse_scroll_stop():
 
     def mouse_gaze_scroll():
         """Starts gaze scroll"""
+        # this calls stop_scroll, which resets continuous_scroll_mode
+        start_cursor_scrolling()
+
         global continuous_scroll_mode
```
AndreasArvidsson, line 212: Feedback from the community meetup. The global should be the first line in the function otherwise this looks good.

AndreasArvidsson (changes requested):
Feedback from the community meetup. The global should be the first line in the function otherwise this looks good.

AndreasArvidsson (approved):

PR #1486, opened 2024-07-08
Add are you sure functionality

FireChickenProductivity:
Implement are you sure functionality where you can use an action to register a function to be performed of the user confirms that it should. This offers a standard way to handle potentially destructive commands by making misrecognitions far less likely to activate them.


AndreasArvidsson:
Feed back from the community session. Looks good, but we would want buttons instead of text in the UI.

FireChickenProductivity:
I replaced the spoken forms text with buttons as requested.

AndreasArvidsson (approved):

PR #1487, opened 2024-07-08
Add helpers for appending to the clipboard

jaresty:
- Sometimes you want to append to the clipboard instead of replacing it

### DIFF
--- a/core/edit/edit.py
+++ b/core/edit/edit.py
@@ -93,6 +93,11 @@ class Actions:
         for _ in range(n):
             actions.edit.word_right()
 
+    def cut_append():
+        """Cut and append the selected text to the clipboard"""
+        clip.set_text(clip.get() + "\n" + actions.edit.selected_text())
+        actions.key("delete")
+
     def cut_word():
         """Cut word under cursor"""
         actions.edit.select_word()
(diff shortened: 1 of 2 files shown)


AndreasArvidsson:
Notes from the community meetup. 
We think this is a bit confusing and should probably be part of a proper clipboard manager instead of just appending your existing clipboard entry. 
This needs a bit more polish before we could push it to community, but its close:
https://github.com/AndreasArvidsson/clippy

jaresty:
FWIW - the use case for this is distinct from clipboard history.  Especially when working with LLMs, it is useful to collect context in a single location - I would not want it to be separate clipboard entries, for that reason.

AndreasArvidsson:
Couldn't you just store it in a variable in Talon instead. If you just need a storage and not actual clipboard use?

jaresty:
I use the clipboard as a way to collect context.  In the talon-ai-tools repository, you can run prompts on the clipboard contents directly.  It is possible to solve in other ways, but personally I will probably implement this as a helper for myself if the community maintainers opt not to take it.

jaresty:
FWIW - I do love the idea of having a builtin clipboard manager, though! 🙇

jaresty:
I'm closing this as it seems like it's not going anywhere-we've also found a different approach that works fine for talon-ai-tools.

PR #1488, opened 2024-07-08
Add a helper to copy a phrase

jaresty:
- Sometimes it is useful to set the clipboard text directly.


AndreasArvidsson:
See previous answer
https://github.com/talonhub/community/pull/1487#issuecomment-2228736024

PR #1490, opened 2024-07-09
Unify go back and go forward implementations

AndreasArvidsson:
This implements standardized go back and go forward voice commands as well as matching actions to be used in applications system wide


nriley:
Makes sense — thanks!

Issue #1495, opened 2024-07-19
Copy block command conflicts with cursorless


jaresty:
Some new commands were introduced to community recently that conflict with the cursorless copy block command. This conflict is unfortunate because there is no simple way to fix this without editing this repository.

Related to this issue as well : https://github.com/cursorless-dev/cursorless/issues/428

fidgetingbits:
Also relevant I think: https://github.com/talonhub/community/issues/1412

Probably implementing the actions proposed by #1412 will fix a lot of these problems. I think it will also improve the fallback functionality in cursorless

AndreasArvidsson:
https://github.com/talonhub/community/pull/1498
One this had merged you can just with a tall long list file change enable or disable spoken forms without changing community.

AndreasArvidsson:
Fixed with #https://github.com/talonhub/community/pull/1498

jaresty:
Thank you!

jaresty:
is this fixed on the cursorless side too? I still seem to be running into a conflict

AndreasArvidsson:
There is no planned change in cursorless.
You need to use a different spoken forms in community and cursorless. Or make sure not both are enabled at thr damer time.

jaresty:
Oh-I thought there was a plan for cursorless to override the same command name. If there isn't a plan to change cursorless, doesn't that mean that the default configuration is still a conflict? (I.e. copy is the same?)

AndreasArvidsson:
That is a potential future use, but not something needed to solve the conflicts now.
I changed the default in community to paragraph so the conflict is gone.

jaresty:
ah! I see.  yes, that is indeed fixed.  Thank you!

PR #1496, opened 2024-07-19
Move cursorless conflicts out

jaresty:
- This stopgap solution will allow for people to avoid conflicts with cursorless commands while we wait for a better solution

This addresses: #1495


AndreasArvidsson:
Here is a solution where you can override with just a Talon list
https://github.com/talonhub/community/pull/1498

I'm also not sure if we are supposed to use tags that are not defined in community. For now Talon doesn't throw an error to my knowledge, but that might just be accidental and in the future this might create errors in the log. 

@lunixbochs What do you say about using the Cursorless tag in community? Will Talon be happy with `not tag: user.some_tag` if that tag is not defined?

jaresty:
Amazing - glad to see the talon list version. I think that's preferable to this anyway, so I'm happy to close in favor of that solution.

PR #1497, opened 2024-07-19
File manager commands improvements

alexander-clarke:
Previously the file manager commands were quite messy and some commands were not consistent so I cleaned it up.

Added an optional `[select | cell]` for all select commands, making them all consistent

Also enforced ordering of non numbered item, and then numbered item

Does have a slight breaking change in that
`open <number>`
has now moved too
`open num <number>`
And now instead we have the command
`open {user.file_manager_files}`

But this does mean we have consistency across open, select, and follow for both files and folders, which I think is worth the small breaking change.

Thoughts?


tags/file_manager/file_manager.talon
```diff
@@ -8,25 +8,25 @@ manager close: user.file_manager_hide_pickers()
 manager refresh: user.file_manager_update_lists()
 go <user.system_path>: user.file_manager_open_directory(system_path)
 (go parent | daddy): user.file_manager_open_parent()
+^follow {user.file_manager_directories}$:
```
jaresty, line 11: Does this conflict with cursorless's "follow" command?

alexander-clarke: I don't believe so as I don't think anything implements cursorless and also the file manager tag
and if they did directories names are going to be different to most cursorless targets

But either way this isn't a new command, just a moved command in the same file

jaresty: Oh I see that now. I misread the diff. Thanks for the clarification.

alexander-clarke: yeah the diff is a bit confusing as the old file is a bit confusing so had to move around a lot of the lines

knausj85 (approved):
This is a sensible change, thanks! I'll leave the pull request open for others to consider for a bit.

PR #1498, opened 2024-07-20
Implemented compound edit command

### LINKED ISSUE #1412: Change text editing grammar to be more generic
# The problem

We have a large set of commands in community that consist of a (partial) cross product of "actions" with "modifiers" (to use Cursorless terminology): {select, copy, cut, delete, etc} x {word, line, file, etc}. For example, "select line", "select word", "select file", "copy word", etc. Because the commands in this cross product are just defined one by one, we can end up with inconsistency, as well as difficulty changing things.

In addition, it's more difficult for users to customize the spoken forms for these actions. Eg they may want to change `"select"` to `"take"`. Today, that would require forking and changing many lines in their `.talon` files.

Another problem is that the user doesn't have consistency between their global editing grammar and Cursorless, which leads to cognitive dissonance for users that use Cursorless. Beyond Cursorless, other grammars may want to leverage similar actions. For example @rokubop is working on a grammar for sophisticated actions based on a grid for use in gaming / drawing. This grammar would have actions such as `"copy"`, `"select"`, etc

# The solution

We propose to create the following

- a list for actions such as `"copy"`", `"select"`, etc. This list would be defined using a `.talon-list`
- a list for "modifiers" such as `"word"`", `"line"`, `"file"`, etc, also as a `.talon-list`
- a rule `<user.action> <user.modifier>` that would first run the modifier, then the action. For example, the user could say `"copy line"`, and that would first call `actions.edit.select_line()`, then call `actions.edit.copy()`
- In Cursorless, we could leverage these lists, and use the `user.cursorless` tag to override the Talon action that implements the above rule to instead run the command using Cursorless

To make things easier, we actually have a lot of the code to make this work already written in Cursorless. We use this code to implement a set of fallback actions that make it so that if you issue Cursorless commands when your focus is in a text box in VSCode, rather than in the main editor area, it will fallback to using Talon actions. For example, just as described above, if you say `"copy line"`, it will first call `actions.edit.select_line()`, then call `actions.edit.copy()`. See [here](https://github.com/cursorless-dev/cursorless/blob/main/cursorless-talon/src/fallback.py) for the implementation. 

The way we've implemented it is that the Cursorless talon grammar sends the usual json payload to Cursorless, and if Cursorless detects that your focus is in a text box, it will do nothing, but instead return a json payload describing what fallback actions Talon should perform.

We could adopt a somewhat similar (but simpler) approach universally. Instead of sending a json payload to Cursorless and waiting for the response, we would define a spoken grammar, as described above, for the subset of actions / "modifiers" that are supported everywhere, and these captures would directly run the desired Talon actions. Note that we could then remove some of the definitions from Cursorless's fallback and just use the version built into community

Credit for this idea, as well as for the fallback code, goes primarily to @AndreasArvidsson

### LINKED ISSUE #1493: There no longer needs to be a distinction between edit.find and user.find
Now we have a separate user.find tag we can collapse these definitions.

AndreasArvidsson:
Compound of action(select, clear, copy, cut, paste, etc.) and modifier(word, line, etc.) commands for editing text. eg: `"select line"`, `"clear all"`

Fixes #1412
Fixes #1493

Note that there are support for some actions in here that I have not added a spoken form for. "insert" for example.
https://github.com/talonhub/community/blob/df1dc66815c875bcc942649c65cf242da0231a51/core/edit/edit_command_actions.py#L75-L77

This one is a provision for Cursorless or something similar to use and an example of how you can extend this implementation with more complex actions compared to what you can do with just an identifier in a Talon list. The other action `applyFormatter` is definitely something we can wire up in a future update.

Also note that there are a lot of edit actions left out of this implementation. In some cases because they don't really fit the action/modifier paradigm like `"go left"`. Calling the modifier `character`  as `left` isn't really correct. And in other cases we can definitely do a follow up and fill out this implementation some more.


core/edit/edit_command_actions.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __str__(self):
+        return self.type
+
+
+EditAction = Union[
+    EditSimpleAction,
+    EditInsertAction,
+    EditFormatAction,
+]
```
auscompgeek, line 37:
```suggestion
@dataclass
class EditAction:
    type: str

    def __str__(self):
        return self.type


@dataclass
class EditInsertAction(EditAction):
    type = "insert"
    text: str


@dataclass
class EditFormatAction(EditAction):
    type = "applyFormatter"
    formatters: str
```

AndreasArvidsson: Why? `EditAction` is was get pasts into the action and needs to be of every kind.

core/edit/edit_command_actions.py
```diff
@@ -0,0 +1,83 @@
...
+        return simple_action_callbacks[action_type]
+
+    match action_type:
+        case "insert":
+            assert isinstance(action, EditInsertAction)
+            return lambda: actions.insert(action.text)
+
+        case "applyFormatter":
+            assert isinstance(action, EditFormatAction)
+            return lambda: actions.user.formatters_reformat_selection(action.formatters)
```
auscompgeek, line 81: I'd move this into a method of `EditAction`, which the subclasses override

core/edit/edit_command_actions.py
auscompgeek: If they subclass `EditAction` then `EditAction` is effectively already a union of every kind.

AndreasArvidsson: True. I missed the inheritance part. I basically only do Talon in python and there isn't much call for inheritance there :D
You definitely right that that would work, but I'm not sure if it's strictly better.

core/edit/edit_command.py
```diff
@@ -0,0 +1,48 @@
+from talon import Module, actions
+
+from .edit_command_actions import EditAction, get_action_callback
+from .edit_command_modifiers import EditModifier, get_modifier_callback
+
+# In some cases there already is a "compound" talon action for a given action and modifier
+compound_actions = {
```
pokey, line 7: switch these to tuple keys, eg `("select", "word")`

pokey: you could also
```py
select = SimpleAction("select")

word = SimpleModifier("word")

{
    (select, word): ...
}
```

core/edit/edit_command_actions.py
AndreasArvidsson: We discuss this in the Cursorless meetup today. . Cursorless should be able to context override `edit_command` so putting the implementation on the classes themself is probably not the best way. 
The dataclasses are basically DTOs/descriptors. It's the action that can have different context specific implementations. Does that make sense?

knausj85 (approved):
This looks good to me.

PR #1499, opened 2024-07-21
Replace `user.find` with `edit.find`

### LINKED ISSUE #1493: There no longer needs to be a distinction between edit.find and user.find

AndreasArvidsson:
We should definitely not implement user namespace actions we're Talon already have defined them on internal namespaces.

Fixes #1493


apps/eclipse/eclipse_win.py
```diff
@@ -42,9 +42,16 @@ def toggle_comment():
 
 @ctx.action_class("edit")
 class EditActions:
-    def find(text: str):
+    def find(text: str = None):
         actions.key("ctrl-f")
-        actions.insert(text)
+        if text:
+            actions.insert(text)
```
auscompgeek, line 48: I believe this matches the default Windows implementation, right? Should we just remove the matching ones?

AndreasArvidsson: Yes we should probably have an operating system defaults not every single application needs to specify this. I'm on it.

AndreasArvidsson: Done

tags/find/find.py
```diff
@@ -1,16 +1,27 @@
...
 
+ctx = Context()
 
-@mod.action_class
-class Actions:
-    def find(text: str):
-        """Finds text in current editor"""
+ctx_mac = Context()
+ctx_mac.matches = r"""
+os: windows
```
auscompgeek, line 10: ?

tags/find/find.py
```diff
@@ -1,16 +1,27 @@
...
+class EditActions:
+    def find(text: str = None):
+        actions.key("ctrl-f")
+        if text:
+            actions.insert(text)
+
+
+@ctx_mac.action_class("edit")
+class MacEditActions:
+    def find(text: str = None):
```
auscompgeek, line 24: Assuming this is meant to target macOS, the only correct macOS generic implementation exists in `edit_mac.py`.

tags/find/find.py
AndreasArvidsson: wops copy paste error :D

tags/find/find.py
AndreasArvidsson: Okay I think we move these implementations to edit instead

AndreasArvidsson:
@auscompgeek We now have a platform specific implementation in their corresponding edit files. There only exists application specific overrides where they differ.

nriley (approved):

apps/vscode/vscode.py
```diff
@@ -169,6 +167,11 @@ class MacUserActions:
     def command_palette():
         actions.key("cmd-shift-p")
 
+    def find(text: str = None):
```
nriley, line 170: This is defining an override for `user.find`, not `edit.find`.

nriley (commented):
Sorry, missed this!

apps/vscode/vscode.py
AndreasArvidsson: Good catch!
I made a follow up here
https://github.com/talonhub/community/pull/1511

PR #1501, opened 2024-07-21
Allow `-` in snippet body

### LINKED ISSUE #1310: Add support for - in snippet bodies
I was trying to add a snippet for generating snippets, such as:

```
language: snippet
---

name: snippetDeclaration
phrase: snippet
insertionScope: statement
-
name: $0
phrase:
insertionScope:
-
---
```

This runs into a parser error from `snippet_parser.py:parse_document()`: 
```
2023-11-10 10:34:06.115    IO ERROR | snippet.snippet:3 | Found multiple '-' in snippet document '
```

I tried to do some naive escaping (`\-`) but that didn't work. I don't have time at the moment to fix it properly or check if there's some other way to escape, so filing.

AndreasArvidsson:
Fixes #1310


core/snippets/snippets_parser.py
```diff
@@ -192,17 +192,13 @@ def parse_document(
     file: str, line: int, optional_body: bool, text: str
 ) -> Union[SnippetDocument, None]:
     parts = re.split(r"^-\r?\n?$", text, flags=re.MULTILINE)
```
auscompgeek, line 194: Can we more simply use `text.split("-\n", 1)` here? `\r\n` should be translated to `\n` by `open` universal newlines mode.

AndreasArvidsson: I also want to split if there is no trailing newline. That would be a snippet with an empty body. An unlikely case, but I still want to support it.

AndreasArvidsson:
Feedback from community session. Don't do split instead just find the separator.

AndreasArvidsson:
@nriley Now we use search instead of split

core/snippets/snippets_parser.py
```diff
@@ -191,18 +191,21 @@ def parse_file_content(file: str, text: str) -> list[SnippetDocument]:
 def parse_document(
     file: str, line: int, optional_body: bool, text: str
 ) -> Union[SnippetDocument, None]:
-    parts = re.split(r"^-\r?\n?$", text, flags=re.MULTILINE)
+    match = re.search(r"^-\r?\n?$", text, flags=re.MULTILINE)
```
nriley, line 194: Think you can remove the `\r?\n?` from here as they will never match. Here's an example:

```python
>>> x = 'foo\n-\nbar\n-\nbaz'
>>> re.search(r'^-$', x, flags=re.MULTILINE)
<re.Match object; span=(4, 5), match='-'>
>>> re.search(r'^-\n$', x, flags=re.MULTILINE)
```

nriley: Actually, even better, can just use `re.split(r"^-$", text, maxsplit=1, flags=re.MULTILINE)` — this will give you `context_text` and (optionally) `body_text` in a list.

nriley (changes requested):

core/snippets/snippets_parser.py
AndreasArvidsson: Done

nriley (approved):

PR #1506, opened 2024-07-23
Enable additional language extensions

AndreasArvidsson:
Noticed that html snippets were not working. For some reason a lot of common extensions were not supported. I'm tempted to enable more of the disabled ones, but for now I'm enabled stuff that I know people are using.


nriley:
Seems reasonable for the snippets. Would it make sense to add any comments or indication that there aren't full [programming] language modes for these?

AndreasArvidsson:
> Seems reasonable for the snippets. Would it make sense to add any comments or indication that there aren't full [programming] language modes for these?

Is that not true for some of the other languages in that list as well?

AndreasArvidsson:
Only enable html and put a comment that this is not a full language mode, but just snippet support today

AndreasArvidsson:
@nriley Deactivated the other language id and added a comment about html

core/modes/language_modes.py
```diff
@@ -9,18 +9,17 @@
     "batch": "bat",
     "c": "c h",
     # 'cmake': 'cmake',
-    # 'cplusplus': 'cpp hpp',
+    # "cpp": "cpp hpp",
```
nriley, line 12: Not sure why you are renaming `cplusplus` to `cpp`?

AndreasArvidsson: That is the default abbreviation.

nriley: I'm concerned it might break existing people with modes (e.g. #777).

AndreasArvidsson: Yes since it wasn't even active I kind of feel this is quite harmless. I will leave that decision to you. If you prefer I can revert

AndreasArvidsson: @knausj85 any opinion?

knausj85: This is what we match on with code.language, yes? cplusplus feels more clear in that case.

AndreasArvidsson: Correct
Okay I have reverted to the previous identifier.

nriley (approved):

PR #1508, opened 2024-07-26
Added navigation steps command

AndreasArvidsson:
Implemented new more powerful navigation grammar
* go left
* go left left down
* go 5 left 2 down
*  go word left
* go 2 words right
* go 5 left down down 4 words left

This is fully backwards compatible with the existing `go left`, `go word right` and `go left left down down` grammars. Only here combining them in a single (more powerful) command

### DIFF
--- a/core/edit/edit.talon
+++ b/core/edit/edit.talon
@@ -13,13 +13,10 @@ next one: edit.find_next()
 scroll up: edit.page_up()
 scroll down: edit.page_down()
 
-go word left: edit.word_left()
-go word right: edit.word_right()
-
-go left: edit.left()
-go right: edit.right()
-go up: edit.up()
-go down: edit.down()
+# go left, go left left down, go 5 left 2 down
+# go word left, go 2 words right
+go <user.navigation_step>+:
+    user.perform_navigation_steps(navigation_step_list)
 
 go line start | head: edit.line_start()
 go line end | tail: edit.line_end()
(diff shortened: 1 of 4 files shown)


pokey:
I feel you are starting down the path to rebuilding https://github.com/talonhub/community/tree/main/plugin/text_navigation

AndreasArvidsson:
> I feel you are starting down the path to rebuilding https://github.com/talonhub/community/tree/main/plugin/text_navigation

The purpose of this is that it's fully backwards compatible width existing commands. We don't want users to have to learn multiple different ways of navigating when one grammar can do it all. I don't think that many people use to navigate command because it's to complicated.

pokey:
Not saying is is necessarily a bad thing, just want to make sure we think through the relationship here and whether we want to borrow anything from there

nriley (approved):

nriley:
I'm OK with this. I have some alternative commands for "go word left" and "go word right" (I use compass directions for these) — not sure it is worth accommodating these as, e.g. "go 2 east down"?

Also this does allow "go 2 words down" even though this is nonsensical; again, not sure if this is worth fixing.

AndreasArvidsson:
> I'm OK with this. I have some alternative commands for "go word left" and "go word right" (I use compass directions for these) — not sure it is worth accommodating these as, e.g. "go 2 east down"? Seems like this would require some significant rearchitecting when I'm happy doing what I do now, e.g. "go east second".
> 
> Also this does allow "go 2 words down" even though this is nonsensical; again, not sure if this is worth fixing.

You just need to update the spoken form of your arrow keys and that's it.

nriley:
Yeah, I approved it anyway, just noting a few potential caveats!

PR #1509, opened 2024-07-26
Add functions to toggle mouse drag and gaze scrolling on and off

bradmajewski:
Adds additional mouse actions to:

- Toggle drag on and off.
- Toggle between gaze scrolling on and  all scrolling off.

I created these so I can use a single foot pedal to control mouse dragging or gaze scrolling.


plugin/mouse/mouse.py
```diff
@@ -143,6 +143,13 @@ def mouse_drag_end():
         for button in buttons_held_down:
             ctrl.mouse_click(button=button, up=True)
 
+    def toggle_mouse_drag(button: int):
+        """If the button is held down, release all held buttons, else start dragging"""
+        if button in list(ctrl.mouse_buttons_down()):
+            self.mouse_drag_end()
```
AndreasArvidsson, line 149: I wonder if we only should release the button specified in the `button` argument?
Something like `ctrl.mouse_click(button=button, up=True)`

plugin/mouse/mouse.py
```diff
@@ -143,6 +143,13 @@ def mouse_drag_end():
         for button in buttons_held_down:
             ctrl.mouse_click(button=button, up=True)
 
+    def toggle_mouse_drag(button: int):
```
AndreasArvidsson, line 146: I would name this `mouse_drag_toggle` to be consistent with the other mouse actions

plugin/mouse/mouse.py
```diff
@@ -221,6 +228,13 @@ def mouse_gaze_scroll():
             actions.tracking.control_toggle(True)
             control_mouse_forced = True
 
+    def toggle_mouse_gaze_scroll():
```
AndreasArvidsson, line 231: I would name this `mouse_gaze_scroll_toggle` to be consistent with the other mouse actions

AndreasArvidsson (changes requested):

plugin/mouse/mouse.py
```diff
@@ -221,6 +228,13 @@ def mouse_gaze_scroll():
             actions.tracking.control_toggle(True)
             control_mouse_forced = True
 
+    def mouse_gaze_scroll_toggle():
+        """If not scrolling, start gaze scroll, else stop scrolling."""
+        if continuous_scroll_mode == "":
+            self.mouse_gaze_scroll()
```
AndreasArvidsson, line 234: You should use `actions.user.` instead of `self.`

plugin/mouse/mouse.py
```diff
@@ -143,6 +143,13 @@ def mouse_drag_end():
         for button in buttons_held_down:
             ctrl.mouse_click(button=button, up=True)
 
+    def mouse_drag_toggle(button: int):
+        """If the button is held down, release the button, else start dragging"""
+        if button in list(ctrl.mouse_buttons_down()):
+            ctrl.mouse_click(button=button, up=True)
+        else:
+            self.mouse_drag(button=button)
```
AndreasArvidsson, line 151: You should use `actions.user.` instead of `self.`

AndreasArvidsson (approved):

PR #1510, opened 2024-07-26
Adds "go to tab" to Apple Terminal application.

bradmajewski:
This adds the "go to tab" command to the apple terminal application.

Relies on "Use #-1 through #-9 to switch tabs" being checked in application settings. (This is the default setting.)

It is possible to programmatically detect this using a shell command. I didn't do that because I don't know a lot of python and it adds complexity.

The setting is on
```bash
> defaults read com.apple.Terminal "Command1Through9SwitchesTabs" 
The domain/default pair of (com.apple.Terminal, Command1Through9SwitchesTabs) does not exist
> echo $!
847
```
The setting is off
```bash
> defaults read com.apple.Terminal "Command1Through9SwitchesTabs" 
0
> echo $!
0
```

### DIFF
--- a/apps/apple_terminal/apple_terminal.py
+++ b/apps/apple_terminal/apple_terminal.py
@@ -82,6 +82,7 @@ class UserActions:
         if number <= 9:
             actions.key(f"cmd-{number}")
 
+
 @ctx.action_class("app")
 class app_actions:
     # other tab functions should already be implemented in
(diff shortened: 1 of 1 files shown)


nriley:
This feature is not specific to tabs and will also switch between windows by number (personally I don't use multiple tabs in my Terminal windows). So not sure this is a great idea to implement with this action.

We could consider implementing this feature for "standard" tabs across multiple apps by using accessibility. Here's a quick example of how to activate the second tab in the active Terminal window.

```
ui.apps(bundle="com.apple.Terminal")[0].active_window.element.children.find_one(AXRole='AXTabGroup', max_depth=0).children.find(AXRole="AXRadioButton")[1].perform("AXPress")
```

PR #1515, opened 2024-07-30
Don't use not mode

### LINKED ISSUE #1437: New user defined modes automatically contain mode switching commands in contradiction to the talon wiki
When a user defines a new mode, this mode automatically contains mode switching commands like "command mode" or "sleep all".  This is due to the headers of the files [modes_not_dragon.talon](https://github.com/talonhub/community/blob/main/core/modes/modes_not_dragon.talon) and [modes.talon](https://github.com/talonhub/community/blob/main/core/modes/modes.talon).  

According to the [talon wiki](https://talon.wiki/Customization/Talon%20Framework/modes) this behaviour runs against the primary purpose of a new user defined mode: "So why might you add a custom mode? The main reason is because you want to disable all normal voice commands so only the ones in your mode are active. An example might be where you were using a full screen computer game, and wanted to eliminate potential conflicts with commands outside the game context."

So I guess either the wiki or community should change. 

Personally, I think community should change. I stumbled across this because I wanted to have a new empty mode so that I could use different voice software in combination with Talon (i.e. voice from the other software but Talon noises).

AndreasArvidsson:
Users should be allowed to make their own custom modes and not had any existing commands interfere.

This is what a Talon wiki describes.
https://talon.wiki/customization/talon%20framework/modes/

Fixes #1437

We could of course go even more explicit and not even have these combination files. The sleep command could be in a command mode file and in the dictation mode file etc.

Note that the dragon files are still enabled in all nodes. I don't know this engine well enough to make changes and I don't want to break something.


core/modes/command_and_dictation_mode.talon
```diff
@@ -1,4 +1,5 @@
-not mode: sleep
+mode: command
+mode: dictation
```
nriley, line 2: Can you please update the breaking changes file to note that if users have their own custom modes defined, they will no longer be able to use the standard words for toggling back to command or dictation mode unless they add the additional modes to the context header of this file (or add the corresponding command(s) to their custom mode directly).

nriley (changes requested):

nriley (approved):

PR #1521, opened 2024-07-31
Catch and handle errors when closing draft editor

alexander-clarke:
Sometimes submitting drafts can fail as it cannot find the window to focus on to submit the draft. As it is currently that draft can then be lost.

These changes catches any errors when trying to focus on the previous window, and informs the user how to fix this.

### DIFF
--- a/plugin/draft_editor/draft_editor.py
+++ b/plugin/draft_editor/draft_editor.py
@@ -1,4 +1,4 @@
-from talon import Context, Module, actions, settings, ui
+from talon import Context, Module, actions, settings, ui, app
 
 mod = Module()
 mod.tag("draft_editor_active", "Indicates whether the draft editor has been activated")
@@ -82,11 +82,24 @@ class Actions:
 
     def draft_editor_submit():
         """Submit/save draft editor"""
-        close_editor(submit_draft=True)
+        global last_draft
+        last_draft = close_editor()
+        try:
+            actions.user.switcher_focus_window(original_window)
+        except Exception:
+            app.notify("Failed to focus on window to submit draft, manually focus on intended destination and use draft submit again")
+        else:
+            actions.sleep("300ms")
+            actions.user.paste(last_draft)
+
 
     def draft_editor_discard():
         """Discard draft editor"""
-        close_editor(submit_draft=False)
+        close_editor()
+        try:
+            actions.user.switcher_focus_window(original_window)
+        except Exception:
+            app.notify("Failed to focus on previous window, leaving editor open")
 
     def draft_editor_paste_last():
         """Paste last submitted draft"""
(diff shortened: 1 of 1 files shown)


plugin/draft_editor/draft_editor.py
```diff
@@ -104,15 +118,11 @@ def get_editor_app() -> ui.App:
...
 
-def close_editor(submit_draft: bool):
+def close_editor() -> str:
     global last_draft
     remove_tag("user.draft_editor_active")
     actions.edit.select_all()
     selected_text = actions.edit.selected_text()
     actions.edit.delete()
     actions.app.tab_close()
-    actions.user.switcher_focus_window(original_window)
```
nriley, line 114: Recommendation from @AndreasArvidsson in the community backlog session is that we keep `close_editor` with its current behavior and simply save `last_draft` prior to attempting to focus the original window. It is fine if you want to display a notification that focusing the window failed — that is, move the logic that you have included in `draft_editor_submit/draft_editor_discard` back into `close_editor`.

nriley (dismissed):

plugin/draft_editor/draft_editor.py
alexander-clarke: Yep that makes sense, moved the logic into `close_editor` now

plugin/draft_editor/draft_editor.py
```diff
@@ -104,15 +104,26 @@ def get_editor_app() -> ui.App:
...
+def close_editor(submit_draft: bool) -> None:
     remove_tag("user.draft_editor_active")
     actions.edit.select_all()
     selected_text = actions.edit.selected_text()
     actions.edit.delete()
     actions.app.tab_close()
-    actions.user.switcher_focus_window(original_window)
-    actions.sleep("300ms")
     if submit_draft:
+        global last_draft
```
AndreasArvidsson, line 114: The community standard is to always have globals as the first line in the function

AndreasArvidsson, line 114: In community we prefer if all globals are defined on the first line of the function

plugin/draft_editor/draft_editor.py
```diff
@@ -104,15 +104,26 @@ def get_editor_app() -> ui.App:
     raise RuntimeError("Draft editor is not running")
 
 
-def close_editor(submit_draft: bool):
-    global last_draft
+def close_editor(submit_draft: bool) -> None:
     remove_tag("user.draft_editor_active")
     actions.edit.select_all()
     selected_text = actions.edit.selected_text()
```
AndreasArvidsson, line 110: I would just replace this line with
```
if submit_draft:
        last_draft = actions.edit.selected_text()
```
Then we store the selected text as early as possible in case of anything else later throwing an exception

AndreasArvidsson (changes requested):

plugin/draft_editor/draft_editor.py
alexander-clarke: Fixed, is this something that should be added to the pre-commit as a check?

plugin/draft_editor/draft_editor.py
alexander-clarke: makes sense, fixed

plugin/draft_editor/draft_editor.py
AndreasArvidsson: Sounds reasonable

AndreasArvidsson (approved):

Issue #1373, opened 2024-01-23
Make snippet watcher more fine-grained


pokey:
The snippet watcher code is currently triggering a lot of work inside Talon. We'd like to only update the file that changed in the fs.watch

Alternately, we could remove support for custom user snippets dir, but that would mean dropping support for https://github.com/talonhub/community/issues/1306

_Originally posted by @lunixbochs in https://github.com/talonhub/community/issues/1348#issuecomment-1875631471_

AndreasArvidsson:
The problem is that snippets can override each other. We have a priority system where typescript reacts wins over typescript and so on. The scenario where if they have removed a definition and we now should fall back to a previous one requires us to keep the content of all files and then figure out what lists have actually updated. Fully doable, but not as simple as just rereading a single file.

PR #1528, opened 2024-08-03
Migrate language modes to CSV

FireChickenProductivity:


core/user_settings.py
```diff
@@ -75,3 +76,138 @@ def append_to_csv(filename: str, rows: dict[str, str]):
...
+def _compute_values_packed_into_column(values):
+    output = io.StringIO()
+    writer = csv.writer(output, delimiter=";")
+    writer.writerow(values)
+    result = output.getvalue().strip()
+    return result
+
+
+def _obtain_rows_from_csv(path):
+    with open(str(path), "r") as f:
```
nriley, line 137: Please use `open(str(path), newline='')` to benefit from the standard newline handling.

core/user_settings.py
```diff
@@ -75,3 +76,138 @@ def append_to_csv(filename: str, rows: dict[str, str]):
...
+                name, values_text = row
+                new_spoken_forms_text = ""
+            else:
+                if len(row) > 3:
+                    print(
+                        f'"{filename}": More than three values in row: {row}.'
+                        + " Ignoring the extras."
+                    )
+                name, values_text, new_spoken_forms_text = row[:3]
+            name = name.strip()
```
nriley, line 166: Please strip all of the columns in the CSV, not just the first one.

nriley (changes requested):

core/user_settings.py
FireChickenProductivity: I made the requested change.

core/user_settings.py
FireChickenProductivity: Instead of stripping the other columns directly, I strip every value within the other columns.

core/modes/language_modes.py
```diff
@@ -74,20 +20,162 @@
...
+        [
+            "typescriptreact",
+            ("tsx",),
+        ],
+        # ['vba', ('vba',),],
+        [
+            "vimscript",
+            ("vim", "vimrc"),
+        ],
+        # htm doesn't actually have a language moded, but we do have snippets.
```
nriley, line 141: Please add "l" and remove "d" after the word "mode"

core/modes/language_modes.py
```diff
@@ -74,20 +20,162 @@
...
+SETTINGS_FILENAME = "language_modes.csv"
+settings_filepath = compute_csv_path(SETTINGS_FILENAME)
+
+LANGUAGE_HEADERS = ["language", "extensions", "spoken_forms"]
+
+
+def make_sure_settings_file_exists():
+    # Maps language mode names to the extensions that activate them and language spoken forms. Only put things
+    # here which have a supported language mode; that's why there are so many
+    # commented out entries. TODO: make this a csv file?
```
nriley, line 38: Please update these comments to reflect the current state of things.

core/modes/language_modes.py
```diff
@@ -74,20 +20,162 @@
...
+    )
+
+
+make_sure_settings_file_exists()
+
+
+@resource.watch(settings_filepath)
+def load_language_modes(path: str):
+    make_sure_settings_file_exists()
+    global language_extensions
```
nriley, line 158: Please move all of your global statements to the top of the function.

core/modes/language_modes.py
```diff
@@ -1,65 +1,11 @@
...
-    "tex": ["tech", "lay tech", "latex"],
-}
+from talon import Context, Module, actions, resource
+
+from ..user_settings import (
+    compute_csv_path,
+    compute_spoken_form_to_key_dictionary,
+    create_three_columns_csv_from_default_if_nonexistent,
+    get_key_value_pairs_and_spoken_forms_from_three_column_csv,
+)
```
nriley, line 8: Please consider adding a CHANGELOG documenting this change such that users who have customized the language extensions will know what to do.

nriley (changes requested):

core/modes/language_modes.py
FireChickenProductivity: I moved them to the top of the function as requested.

core/modes/language_modes.py
FireChickenProductivity: I added a CHANGELOG file as requested.

core/modes/language_modes.py
FireChickenProductivity: I fixed the comment.

core/modes/language_modes.py
FireChickenProductivity: I tried to update the comments to reflect the current state of things. Let me know if you think I missed anything.

FireChickenProductivity:
This is ready for review again.

FireChickenProductivity:
Snippets apparently uses language_ids from language_modes.py, so it might be worth making sure that the snippets system still works. 

The following code for dealing with snippets will not get reran on updating the CSV.

```
# Create a context for each defined language
for lang in language_ids:
    ctx = Context()
    ctx.matches = f"code.language: {lang}"
    context_map[lang] = ctx
```

I could address this by creating a callback registration mechanism inside language_modes.py.

nriley:
> Snippets apparently uses language_ids from language_modes.py, so it might be worth making sure that the snippets system still works.
> 
> The following code for dealing with snippets will not get reran on updating the CSV.
> 
> ```
> # Create a context for each defined language
> for lang in language_ids:
>     ctx = Context()
>     ctx.matches = f"code.language: {lang}"
>     context_map[lang] = ctx
> ```
> 
> I could address this by creating a callback registration mechanism inside language_modes.py.

We could declare an action that returns the list of languages?

FireChickenProductivity:
> We could declare an action that returns the list of languages?
Yes, but that would not cause the snippets code to update in response to the csv updating.

nriley:
> > We could declare an action that returns the list of languages?
> 
> Yes, but that would not cause the snippets code to update in response to the csv updating.

Gotcha. Not sure if we have an example of this, but an action that triggers when the list is updated seems reasonable.

FireChickenProductivity:
> > > We could declare an action that returns the list of languages?
> > 
> > 
> > Yes, but that would not cause the snippets code to update in response to the csv updating.
> 
> Gotcha. Not sure if we have an example of this (I couldn't find one on a cursory search of community), but an action that triggers when the list is updated seems reasonable.

I plan on writing code to allow registering callbacks to handle language mode updates.

FireChickenProductivity:
I theoretically have the code working, but I do not use the snippets system, so I will need to learn how it works before I can test it properly.

AndreasArvidsson:
Notes form their community session. We're not sure that introducing a event system especially for this use case is desirable.
I will have a look and see if I can get rid of the dependency on language ids from the snippets. Let's talk more about this on the next meetup. 
If necessary another developer can finish this.

FireChickenProductivity:
I am going to be busy for a while starting next week, so I am fine with another developer taking this PR.

AndreasArvidsson:
Okay I looked into the snippet dependency on language ids and the result is not that good. Basically the only reason why I'm using the language ids from the language modes file and not the actual snippets themselves is because of performance. Dynamically creating contexts is problematic for Talon. Even to the extent that it have been suggested we should remove the custom snippet dir setting.

The current implementation creates all the contexts once before Talon is started/ready
https://github.com/talonhub/community/blob/9af1757205be0a7458116892380e61c5153ef76c/core/snippets/snippets.py#L32-L36

Issue talking about the problem with updating snippet contexts
https://github.com/talonhub/community/issues/1373

The implementation in this pull request removes the performance gain in creating the contexts before Talon starts. In that case we could just get the language ids from the snippet files themself. I can't really say how much of a problem it is to create the context dynamically, but I definitely feel that if we are going to create them dynamically we should not do it based upon the language mode file, but the snippet files themself.

core/snippets/snippets.py
```diff
@@ -23,17 +23,26 @@
...
+def create_context_map():
+    global context_map
+    context_map = {
+        # `_` represents the global context, ie snippets available regardless of language
+        "_": Context(),
+    }
+
+    # Create a context for each defined language
+    for lang in language_ids:
+        ctx = Context()
```
AndreasArvidsson, line 38: If we decide to keep this implementation we definitely should check if the context already exists before creating a new one

AndreasArvidsson:
My preferred solution would be to get the language identifiers from the snippet files themselves directly when the file is read so the context would be created before Talon is ready. The only scenario where we would dynamically create a context would be if you add a new snippet language which hopefully is not that common. The problem with this approach is that we have a setting for custom snippet dir. This setting is not available until Talon is ready. The implementation in this pull request might be the the best solution available at the moment.

core/modes/language_modes.py
```diff
@@ -74,20 +22,165 @@
...
+    ]
+    create_three_columns_csv_from_default_if_nonexistent(
+        SETTINGS_FILENAME, LANGUAGE_HEADERS, default_csv_contents
+    )
+
+
+make_sure_settings_file_exists()
+
+
+@resource.watch(settings_filepath)
```
AndreasArvidsson, line 159: How does this handle if the file doesn't exist?

FireChickenProductivity: Last time I tested the code, it created the file if it did not exist.

knausj85:
> Okay I looked into the snippet dependency on language ids and the result is not that good. Basically the only reason why I'm using the language ids from the language modes file and not the actual snippets themselves is because of performance. Dynamically creating contexts is problematic for Talon. Even to the extent that it have been suggested we should remove the custom snippet dir setting.
> 
> The current implementation creates all the contexts once before Talon is started/ready
> 
> https://github.com/talonhub/community/blob/9af1757205be0a7458116892380e61c5153ef76c/core/snippets/snippets.py#L32-L36
> 
> Issue talking about the problem with updating snippet contexts #1373
> 
> The implementation in this pull request removes the performance gain in creating the contexts before Talon starts. In that case we could just get the language ids from the snippet files themself. I can't really say how much of a problem it is to create the context dynamically, but I definitely feel that if we are going to create them dynamically we should not do it based upon the language mode file, but the snippet files themself.

To me, it sounds like we're might be better off keeping this in python. Maybe we should instead clean up the extension definition stuff e.g. using a data class to define the list of extensions/associated spoken forms to make it more easily understandable/modifiable, and live with the need to branch this particular file for performance reasons. Plus, the alternative solutions to this PR require a talon restart if we modify the file, which is downer.

AndreasArvidsson:
In the community session today we talked further on this topic. We think we should probably have a new python file `code_languages.py` where we have the language definitions. This python file can then be used by the language modes and the snippets. 

We want to have these language definitions typed and something like this could probably work: 
```py
string_or_list = str | list[str]
lang = tuple[str, string_or_list] | tuple[str, string_or_list, string_or_list]

languages: list[lang] = [
    ("python", ["python"]),
    ("java", "java"),
    ("javascript", ["javascript"], ["something something"]),
]
```

nriley:
Migrated to an issue (#1585).

PR #1529, opened 2024-08-06
Handle blank entries in homophones.csv

geobarry:
added a line to function update_homophones in homophones.py to remove words that are empty strings from the list of homophones created from each line of the CSV file. The issue is that CSV file format is not standard, and programs like MS Excel try to force each line to have the same number of "cells" (i.e. commas). This results in almost every line having on extra comma which is read as a "word" that is an empty string, and then these empty string "homophones" are matched with other words' empty string homophones resulting in every word being a homophone of every other word in the file.

### DIFF
--- a/core/homophones/homophones.py
+++ b/core/homophones/homophones.py
@@ -119,7 +119,6 @@ def raise_homophones(word_to_find_homophones_for, forced=False, selection=False)
         return
 
     # Move current word to end of list to reduce searcher's cognitive load
-
     valid_homophones_reordered = list(
         filter(
             lambda word_from_list: word_from_list.lower()
--- a/core/homophones/homophones.py
+++ b/core/homophones/homophones.py
@@ -134,7 +134,6 @@ def raise_homophones(word_to_find_homophones_for, forced=False, selection=False)
         and quick_replace
         and not force_raise
     ):
-
         if word_to_find_homophones_for == active_word_list[0].lower():
             new = active_word_list[1]
         else:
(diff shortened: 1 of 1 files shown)


core/homophones/homophones.py
```diff
@@ -118,6 +119,7 @@ def raise_homophones(word_to_find_homophones_for, forced=False, selection=False)
         return
 
     # Move current word to end of list to reduce searcher's cognitive load
+
     valid_homophones_reordered = list(
```
phillco, line 123:
```suggestion
    valid_homophones_reordered = list(
```

core/homophones/homophones.py
```diff
@@ -133,14 +135,14 @@ def raise_homophones(word_to_find_homophones_for, forced=False, selection=False)
         and quick_replace
         and not force_raise
     ):
+
         if word_to_find_homophones_for == active_word_list[0].lower():
```
phillco, line 139:
```suggestion
        if word_to_find_homophones_for == active_word_list[0].lower():
```

nriley (approved):

core/homophones/homophones.py
```diff
@@ -102,6 +102,7 @@ def raise_homophones(word_to_find_homophones_for, forced=False, selection=False)
 
     # We support plurals, but very naively. If we can't find your word but your word ends in an s, presume its plural
     # and attempt to find the singular, then present the presumed plurals back. This could be improved!
+
     if word_to_find_homophones_for in all_homophones:
         valid_homophones = all_homophones[word_to_find_homophones_for]
```
phillco, line 107:
```suggestion
    if word_to_find_homophones_for in all_homophones:
        valid_homophones = all_homophones[word_to_find_homophones_for]
```

phillco:
(Just removing the extra blank lines)

PR #1530, opened 2024-08-06
Implemented action `edit.delete_all()`

AndreasArvidsson:


core/edit/edit.py
```diff
@@ -75,6 +75,10 @@ def select_word():
         actions.edit.word_left()
         actions.edit.extend_word_right()
 
+    def delete_all():
```
fidgetingbits, line 78: I think I'd prefer to not have something that relies on select_all() in a ctx-based implementation if it can be avoided, as then it will conflict with an app that needs a non-selection-based delete_all() implementation. 

If needed the @ctx.action_class("edit") version could just call user.delete_all(), which could have a default defined in the @mod.action_class that uses select_all(). This way user.delete_all() can be overridden if needed.

fwiw in the the old version https://github.com/talonhub/community/pull/1498/files#diff-2169d0699cd298adec8d401b6646edf196e2142c268a308ea592ee71454d3f83 it was also in @mod.action_class.

fidgetingbits (changes requested):

core/edit/edit.py
AndreasArvidsson: The most specific context wins. That means if you have a context on a specific application or operating system that would win over this.

I very strongly believe we should not create user namespace copies of existing Talon actions

fidgetingbits: Ah yes. I do recall that discussion somewhere recently now that you mention it. As long as it can still be replaced by something else I'm fine with it.

fidgetingbits (approved):

PR #1531, opened 2024-08-06
Avoid hardcoded xdg-open path and resolve incoming paths

fidgetingbits:
This fixes using xdg-open while on nixos and also resolves paths passed in, so if someone uses ~/ or similar in their path it'll work.


core/edit_text_file.py
```diff
@@ -67,7 +71,11 @@ def edit_text_file(path):
         # we use xdg-open for this even though it might not open a text
         # editor. we could use $EDITOR, but that might be something that
         # requires a terminal (eg nano, vi).
-        open_with_subprocess(path, ["/usr/bin/xdg-open", path])
+        open_cmd = shutil.which("xdg-open")
```
auscompgeek, line 74: I had raised that this would be a problem on NixOS back in the original PR too: https://github.com/talonhub/community/pull/847#discussion_r878703305

Time to start this discussion again I guess. cc @rntz

fidgetingbits: If we are concerned about malicious binaries injected into PATH folders enough to avoid calling which, then I can change it to test a set of hardcoded paths to try find it and include /run/current-system/sw/bin/ for nixos, though I'm not sure that will always work for all possible setups. Could cache the path after the first go to not have to search through a list each time too. 

There is a which call in PR #1471, to fix a similar issue on nixos, that will need to change depending on what gets decided.

nriley: It's now been a couple months. Nobody on the community backlog session this week was a Linux or NixOS user so we really didn't have much constructive to add, however we weren't against whatever you might agree on. Is there some consensus here?

fidgetingbits: Thanks for the bump. I'll just switch to hardcoded paths under the assumption we don't want to rely on PATH, since there hasnt been any discussion. I'm pretty sure that there are other places where this will need to be changed in talon community, but I'll try to find them when I update this.

auscompgeek: I honestly don't think it's a major issue. It's pretty common to set environment variables like EDITOR and BROWSER to values that depend on PATH. Caching the lookup would be a good idea though.

Honestly if an attacker has write access to your PATH on a desktop, you probably have bigger problems.

fidgetingbits: Ok, I'll leave it as is then, in which case it can probably be merged unless there are any other concerns.

lunixbochs: do you need `which`? depending on how you launch the subprocess, you can just pass `xdg-open`

fidgetingbits:
Switched to catching xdg-open not being found and notifying, rather than calling which every time per aegis suggestion.

knausj85 (approved):

PR #1532, opened 2024-08-08
override peek implementation for libreoffice

chdoc:
This only matches for `libreoffice` on Linux. I don't have access to Windows or Mac OS, so I don't know what `libreoffice` calls itself on those operating systems.

### DIFF
--- a/apps/libreoffice/libreoffice.py
+++ b/apps/libreoffice/libreoffice.py
@@ -1,4 +1,5 @@
 from typing import Callable
+
 from talon import Context, Module, actions, settings
 
 mod = Module()
@@ -13,9 +14,10 @@ ctx.matches = r"""
 app: libreoffice
 """
 
+
 # Unlike most other applications, LibreOffice does not move the cursor to the
 # start/end of the selection when pressing left/right. Instead, it only moves
-# the cursor one step. 
+# the cursor one step.
 @ctx.action_class("user")
 class UserActions:
     def dictation_peek(left, right):
@@ -26,12 +28,12 @@ class UserActions:
             actions.edit.extend_word_left()
             actions.edit.extend_word_left()
             before = actions.edit.selected_text()
-            repeat_action(actions.edit.right,len(before))
+            repeat_action(actions.edit.right, len(before))
         if right:
             actions.edit.extend_word_right()
             actions.edit.extend_word_right()
             after = actions.edit.selected_text()
-            repeat_action(actions.edit.left,len(after))
+            repeat_action(actions.edit.left, len(after))
         return (before, after)
 
 
--- a/apps/libreoffice/libreoffice.py
+++ b/apps/libreoffice/libreoffice.py
@@ -1,0 +1,40 @@
+from typing import Callable
+from talon import Context, Module, actions, settings
+
+mod = Module()
+
+mod.apps.libreoffice = """
+os: linux
+and app.exe: soffice.bin
+"""
+
+ctx = Context()
+ctx.matches = r"""
+app: libreoffice
+"""
+
+# Unlike most other applications, LibreOffice does not move the cursor to the
+# start/end of the selection when pressing left/right. Instead, it only moves
+# the cursor one step. 
+@ctx.action_class("user")
+class UserActions:
+    def dictation_peek(left, right):
+        # clobber selection if it exists
+        actions.key("space backspace")
+        before, after = None, None
+        if left:
+            actions.edit.extend_word_left()
+            actions.edit.extend_word_left()
+            before = actions.edit.selected_text()
+            repeat_action(actions.edit.right,len(before))
+        if right:
+            actions.edit.extend_word_right()
+            actions.edit.extend_word_right()
+            after = actions.edit.selected_text()
+            repeat_action(actions.edit.left,len(after))
+        return (before, after)
+
+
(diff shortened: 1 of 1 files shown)


apps/libreoffice/libreoffice.py
```diff
@@ -0,0 +1,42 @@
+from typing import Callable
+
+from talon import Context, Module, actions, settings
+
+mod = Module()
+
+mod.apps.libreoffice = """
+os: linux
+and app.exe: soffice.bin
+"""
```
AndreasArvidsson, line 10: You can add the windows matcher
```
os: windows
and app.exe: soffice.bin
```

apps/libreoffice/libreoffice.py
```diff
@@ -0,0 +1,42 @@
...
+            repeat_action(actions.edit.right, len(before))
+        if right:
+            actions.edit.extend_word_right()
+            actions.edit.extend_word_right()
+            after = actions.edit.selected_text()
+            repeat_action(actions.edit.left, len(after))
+        return (before, after)
+
+
+def repeat_action(action: Callable, count: int):
```
AndreasArvidsson, line 40: I would remove this function and just inline it. It's actually fewer lines.

apps/libreoffice/libreoffice.py
AndreasArvidsson, line 1: This doesn't to work for me when I try it
Given this text and cursor position `aaa |bbb` if I dictate `Test` I end up with `aaa bTestbb`
Just running `dictation_peek(True, True)` leaves me with `aaa b|bb`.

AndreasArvidsson (changes requested):

apps/libreoffice/libreoffice.py
chdoc: I'm assuming you tried this on Windows. When you extend a selection by several words to the right and then press the "left" key, where does your cursor end up?

AndreasArvidsson: Yes I did
It ends up one character to the left of the right edge

chdoc: Now this is puzzling. Are you absolutely certain that your talon actually picked up "my" implementation? Because the behavior you are describing is basically what I have been seeing with the generic implementation. Maybe you recall our conversation about my problems in getting talon to pick up on this.

AndreasArvidsson: It was running because I did get the print out I added to your code to make sure, but now when I try the same code again it appears to be working as I have no idea what I did wrong. It appears to be working now at least :)

chdoc: While this does have me slightly worried, I'm glad I'm not the only one encountering this kind of weirdness.

apps/libreoffice/libreoffice.py
chdoc: I don't think inlining this increases clarity. In fact, I copied the function from `core/edit/edit_navigation_steps.py`. So I would prefer moving this to the right place instead, whatever the right place is.

chdoc:
Apologies for being silent on this for so long. I realized that, while this works really well for a normal text, it still fails miserably when editing text in tables: then running out of words `actions.edit.extend_word_left` transitions to selecting entire table cells, and then all bets are of. Not sure what to do about this.

AndreasArvidsson:
Feedback from the community backlog session. If we can't actually get this to work reliably in both text and tables we are kind of leaning towards actually just disabling context sensitive dictation by default in libre office with a comment about the failure mode for tables. Then we can actually merge this change and if people want to use it it's up to them.

apps/libreoffice/libreoffice.py
phillco: I agree with Andreas, I found this harder to parse. For a two-line function I agree we should just inline it - otherwise it leads the reader to try to understand why the function was necessary (performance? etc).

While `edit_navigation_steps.py` does have many functions like this, it has so many that the pattern becomes more understandable.

phillco (changes requested):

AndreasArvidsson:
Looks like clip.mime().html has <td>s etc for a table selection. We could potentially use this to determine if we are in a table and in that case don't do this cursor dance.

chdoc:
> Looks like clip.mime().html has s etc for a table selection. We could potentially use this to determine if we are in a table and in that case don't do this cursor dance.

I don't understand how this is supposed to work. After all, by the time we can use `clip` to detect whether we have selected a table, we're already in the middle of the "cursor dance". 

That being said, I disagree that "just disabling smart insertion by default" is a viable option. I find it absolutely indispensable when editing normal text, even though it is utterly useless (or a major hindrance) in certain contexts. Maybe this shouldn't be a fixed setting for a given application, instead there should be a command to quickly toggle that. How would one go about implementing this?

AndreasArvidsson:
All Talon settings are already context based. This means you can enable or disable context sensitive dictation in any application or other matchable context. What we suggested was that it by default should be specifically disabled in libre office with a comment describing why. People can still enable it if they want to.

knausj85:
Converting this to draft until someone has time to pick it up again

nriley:
Closing with @chdoc 's blessing. He is no longer using LibreOffice — if anyone else has a bright idea of how to fix this feel free to reopen it.

PR #1534, opened 2024-08-10
Anchor the VSCode please command

jaresty:
- I was frequently getting this triggered in the middle of other commands
- I don't think it makes sense to chain this command

### DIFF
--- a/apps/arc/arc_mac.talon
+++ b/apps/arc/arc_mac.talon
@@ -4,7 +4,7 @@ os: mac
 tag(): browser
 tag(): user.tabs
 
-please [<user.text>]:
+^please [<user.text>]$:
     key("cmd-l")
     sleep(200ms)
     insert(user.text or "")
--- a/apps/eclipse/eclipse.talon
+++ b/apps/eclipse/eclipse.talon
@@ -8,7 +8,7 @@ tag(): user.splits
 tag(): user.tabs
 # splits.py support end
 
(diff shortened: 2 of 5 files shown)


nriley (changes requested):
From the community backlog session — this now creates a difference between the VSCode and the other please implementations in community. I would recommend you consider addressing #1536 or modifying the other implementations at the same time as this one.

jaresty:
I'm opting to simply anchor the current please commands
- since jetbrains is already different, I'm not sure which way to go if I were to try and consolidate with the single overrideable helper.
- if it is preferable to come up with a single overrideable helper, I'm happy to do so but would like some guidance what to do with jetbrains

auscompgeek:
I don't see how the JetBrains equivalent behaves any differently to the VSCode one. They both open a command palette and optionally insert some text.

jaresty:
It behaves the same, but the naming is different. I heard that the ask was to make a single talon definition that had overridden implementations in each application. If I do that for jetbrains, users who rely on the "action" usage will not have that anymore. Is that what I should do?

jaresty:
Ok, I think I'll put together a version of this that consolidates on a single overrideable command so we can discuss/decide which works better.

jaresty:
Here is a version of this that does that: https://github.com/talonhub/community/pull/1537

nriley:
Replaced by #1537

Issue #1536, opened 2024-08-10
Add a please action for use in command palettes and the like


nriley:
This voice command is defined in many places in community already, and with changes like #1534 these implementations can drift from one another.

jaresty:
I'm suspicious that there may be some valid reasons to diverge (if the please behavior is to automatically trigger, chaining might be possible). But just to understand clearly, what is the ask here?

nriley:
Either change all the please implementations to anchor, or implement please once with an action overridden in each app's context.

jaresty:
I can do that. I was trying to target my original fix to just where I was feeling pain since I'm not sure how it will impact others (since I'm not using please elsewhere). I'll update the PR to accommodate this additional requirement.

PR #1537, opened 2024-08-11
Created a centralized please command

jaresty:
To avoid divergence in the grammar for please commands, it is useful to define a centralized tag and allow it to be overridden in command implementations.

Addresses https://github.com/talonhub/community/issues/1536


jaresty:
I was able to test this ini VSCode and jet brains on the mac but I was unable to test the other implementations to make sure they're working. I could use a hand with testing on other platforms to make sure these things still work, though the implementations appear correct to me right now.

apps/jetbrains/jetbrains.py
```diff
@@ -298,7 +298,7 @@ def select_range(line_start: int, line_end: int):
         # if it's a single line, select the entire thing including the ending new-line5
         if line_start == line_end:
             actions.user.idea(f"goto {line_start} 0")
-            actions.user.idea("action EditorSelectLine"),
+            (actions.user.idea("action EditorSelectLine"),)
```
nriley, line 301: Looks like this was a mistake that got auto formatted to make it more obvious — just remove the trailing comma.

nriley (commented):
Makes sense to use a tag for this to avoid "promising what we can't deliver". With the exception of the tiny nit I mentioned, the only other thing I can think that might is whether we'd like to continue sending an empty string for "please" by itself or define two voice commands and an action with an optional argument. I don't feel strongly one way or the other.

apps/jetbrains/jetbrains.talon
```diff
@@ -14,11 +15,7 @@ smart: user.idea("action SmartTypeCompletion")
 (done | finish): user.idea("action EditorCompleteStatement")
 # Copying
 grab <number>: user.idea_grab(number)
-# Actions
-(action | please): user.idea("action GotoAction")
-(action | please) <user.text>:
```
AndreasArvidsson, line 19: We probably want deprecation warning for the command `action [<user.text>]`

apps/arc/arc.py
```diff
@@ -20,6 +20,11 @@ def tab_close_wrapper():
         actions.sleep("180ms")
         actions.app.tab_close()
 
+    def please_command(command: str):
```
AndreasArvidsson, line 23: We probably want to call this `command_search` instead. Also the argument should probably be optional.

AndreasArvidsson (changes requested):

apps/jetbrains/jetbrains.talon
jaresty: I'm away this week but will make this change when I return.

apps/arc/arc.py
jaresty: I tried to do this but I'm not sure if I did it right. Please check my work-thank you!

AndreasArvidsson: I guess there is a discussion if we want to default to an empty string or None.
https://github.com/talonhub/community/blob/940fe694bdb1b5ea752fc2e62154a255f055734e/plugin/talon_draft_window/draft_talon_helpers.py#L224

Also I personally feel that if the argument is optional we should do an if statement an only do sleep and insert if we actually have an argument.
https://github.com/AndreasArvidsson/andreas-talon/blob/114911424dc568d833b865f64f8607ac83275a6d/apps/vscode/vscode_git.py#L63-L67

AndreasArvidsson: No, not of its optional. You just need to do it away I linked above, but you need to call it from a separate command. `(arg)` doesn't work if `arg` is optional.

jaresty: I went through and added the conditions.  Do you think I should add the separate command? I wasn't sure about that.

AndreasArvidsson (approved):

apps/vivaldi/vivaldi.py
```diff
@@ -53,6 +53,12 @@ def tab_close_wrapper():
     def tab_jump(number: int):
         actions.key(f"ctrl-{number}")
 
+    def command_search(command: str = ""):
+        actions.user.vivaldi_toggle_quick_commands()
```
nriley, line 57: I would recommend you remove this separate action as it is not used in any other places.

nriley (changes requested):

apps/vivaldi/vivaldi.py
jaresty: Are you suggesting that I remove the vivaldi_toggle_quick_commands action? I notice that it is being used [here too](https://github.com/talonhub/community/blob/e781ba8c81042e027f12ced8e8d87b685a1031a5/apps/vivaldi/vivaldi.py#L29). I don't quite understand the suggestion, though I appreciate the feedback.

auscompgeek: Since that action is only used in this file, I think it makes sense to replace that usage with a call to this action.

tags/please_command/please_command.talon
auscompgeek, line 1: These files should be named after the tag, not the name of the command.

apps/vivaldi/vivaldi.py
jaresty: Can you give me exact code? I don't have this application so I don't want to make a mistake since I can't test it myself-thank you!

jaresty: Oh, I get it. You mean the new "please" action. That makes sense.

jaresty: I think I addressed this - @nriley  does that look good now? Let me know if there's more to do here. Thanks!

nriley: Looks good. Note that you can click the "chasing arrows" icon next to a reviewer if you want to request a re-review.

nriley (approved):

PR #1540, opened 2024-08-13
Enable readline and python tags in talon repl

fidgetingbits:
Similar to #1451 this uses a context using not forced language to override `code.language` and enable python lang features inside the talon repl. It also makes the title matching more restricting, as this currently matches on anything with repl in the name, and there are many repls that aren't talon or python...

I also enable readline so that common navigation commands work inside.

### DIFF
--- a/apps/talon/talon_repl/talon_repl.py
+++ b/apps/talon/talon_repl/talon_repl.py
@@ -1,0 +1,19 @@
+from talon import Context, Module
+
+mod = Module()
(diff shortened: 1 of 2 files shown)


phillco (approved):

phillco:
I noticed a few readline bugs in the Talon REPL last I looked, but that's more of a bug report for Talon, it clearly intends to implement them

PR #1541, opened 2024-08-13
Add mission control to focus for app switcher

thetomcraig-aya:
This adds functionality for application switching on MacOS.  The actual equivalent on MacOS is the Application Switcher (Command + Tab), but there is no built in way for this, to be persistent on screen with a single hot key.  The best alternative is Mission Control, which will show all of the open windows.


thetomcraig-aya:
Made it with wrong github account

PR #1542, opened 2024-08-13
Add mission control to focus for app switcher

thetomcraig:
This adds functionality for application switching on MacOS.  The actual equivalent on MacOS is the Application Switcher (Command + Tab), but there is no built in way for this, to be persistent on screen with a single hot key.  The best alternative is Mission Control, which will show all of the open windows.


nriley:
The way to do this without depending on the activation keystroke not being reset is `user.dock_send_notification("com.apple.expose.awake")`.

I do have this defined in my fork but have not suggested it for merging as there is no way to navigate Mission Control from the keyboard or via accessibility. I guess it might be useful to people who use an eye tracker or another pointing device.

phillco:
+1, +1 to also using the dock notification API if it works instead of the key shortcut, since it's possible people would redefine it

nriley (changes requested):
See above.

thetomcraig:
Also, @nriley now I understand what you're saying about not being able to navigate Mission Control with the keyboard. But I do find using with an eye tracker pretty nice 🙂

thetomcraig:
Closing this per discussion above

phillco:
@thetomcraig -- confused by your last comment, I think @nriley both wanted is to merge, we just wanted to change it to use  `user.dock_send_notification("com.apple.expose.awake")`. :)

I think it's fine if the interface it shows isn't voice navigable, it's still useful for mixed users, and better than what we have today (which is nothing)

thetomcraig:
> @thetomcraig -- confused by your last comment, I think @nriley both wanted is to merge, we just wanted to change it to use `user.dock_send_notification("com.apple.expose.awake")`. :)
> 
> I think it's fine if the interface it shows isn't voice navigable, it's still useful for mixed users, and better than what we have today (which is nothing)

@phillco @nriley sorry I misunderstood what was meant.  I thought you were saying you were going to merge the `dock_send_notification` change.

thetomcraig:
> @thetomcraig yup, that's what I meant, happy to merge if you change it out

Just put that in!  I put in your `dock` files into the `apps` folder as well.

apps/dock/dock.talon
```diff
@@ -0,0 +1,5 @@
+os: mac
+-
+^desktop$: user.dock_send_notification("com.apple.showdesktop.awake")
+^window$: user.dock_send_notification("com.apple.expose.front.awake")
```
nriley, line 4: Apparently this stopped working since I implemented it. Does it work for you? I tested both on macOS 14 and 15 and on neither OS did it do anything.

Don't think this is a barrier to merging though!

nriley (approved):

apps/dock/dock.talon
thetomcraig: Ah you're right, I just poked around and it looks like it's now `com.apple.expose.awake` (without `front`).  I put that in a PR here, mind taking a look?

https://github.com/talonhub/community/pull/1608

PR #1544, opened 2024-08-14
Added phrase enders list

AndreasArvidsson:
Instead of just `"over"`, adds a list of symbols that can be used to end/terminate a phrase. Also supports values for these phrases so you can end your phrase with a symbol that will be inserted. eg `"say hello there slap"` => `hello there\n`

This implementation is quite simple with the drawback that there is no synchronization between the spoken forms in this list and existing voice commands for adding these symbols. 

Slack reference
https://talonvoice.slack.com/archives/C9MHQ4AGP/p1723643676364569

### DIFF
--- a/core/text/phrase_ender.talon-list
+++ b/core/text/phrase_ender.talon-list
@@ -1,8 +1,4 @@
 list: user.phrase_ender
 -
 
-over:          ""
-space:         " "
-question:      "?"
-bang:          "!"
-slap:          "\n"
+over: ""
--- a/core/text/phrase_ender.talon-list
+++ b/core/text/phrase_ender.talon-list
@@ -1,0 +1,9 @@
+list: user.phrase_ender
+-
+
+over:          ""
+void:          " "
+step:          " "
+question:      "?"
+bang:          "!"
+slap:          "\n"
--- a/core/text/phrase_ender.talon-list
+++ b/core/text/phrase_ender.talon-list
@@ -2,8 +2,7 @@ list: user.phrase_ender
 -
 
 over:          ""
-void:          " "
-step:          " "
+space:         " "
 question:      "?"
 bang:          "!"
 slap:          "\n"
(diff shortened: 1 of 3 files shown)


core/text/text.talon
```diff
@@ -2,13 +2,15 @@
 phrase <user.text>$:
     user.add_phrase_to_history(text)
     insert(text)
-phrase <user.text> over:
+phrase <user.text> {user.phrase_ender}:
     user.add_phrase_to_history(text)
-    insert(text)
+    insert("{text}{phrase_ender}")
 {user.prose_formatter} <user.prose>$: user.insert_formatted(prose, prose_formatter)
 {user.prose_formatter} <user.prose> over: user.insert_formatted(prose, prose_formatter)
```
bradmajewski, line 9: Should this line be changed as well?

```
{user.prose_formatter} <user.prose> {user.phrase_ender}: user.insert_formatted("{prose}{phrase_ender}", prose_formatter)
```

core/text/phrase_ender.talon-list
```diff
@@ -0,0 +1,8 @@
+list: user.phrase_ender
+-
+
+over:          ""
+space:         " "
+question:      "?"
+bang:          "!"
+slap:          "\n"
```
phillco, line 8: This is a bit different from the standard `slap` (`edit.line_insert_down()`) behavior if you're in the middle of a line, right?

The newline will be inserted directly after the phrase, breaking the line, whereas slap normally inserts anewline after the end of the current line. If you're already at the end I think it behaves the same.

AndreasArvidsson: True, but this is just about ending with a particular symbol. Adding support for edit actions in the middle of a phrase is possible, but definitely more complicated the current implementation. We could call it something else if you think the overloaded phrase is confusing. This is just what I'm personal running with.

nriley (approved):

PR #1545, opened 2024-08-15
Added pop to stop mouse drag setting

AndreasArvidsson:


settings.talon
```diff
@@ -27,6 +27,9 @@ settings():
     # If `true`, stop continuous scroll/gaze scroll with a pop
     user.mouse_enable_pop_stops_scroll = true
 
+    # If `true`, stop mouse drag with a pop
+    user.mouse_enable_pop_stops_drag = true
```
AndreasArvidsson, line 31: I made the decision to enable this as default here sins the above `mouse_enable_pop_stops_scroll` already does that

nriley (approved):

PR #1546, opened 2024-08-15
Code cleanup for held mouse buttons

AndreasArvidsson:


plugin/mouse/mouse.py
```diff
@@ -132,15 +132,14 @@ def mouse_wake():
     def mouse_drag(button: int):
         """Press and hold/release a specific mouse button for dragging"""
         # Clear any existing drags
-        self.mouse_drag_end()
+        actions.user.mouse_drag_end()
```
AndreasArvidsson, line 135: I was surprised this even worked. We should definitely call Talon actions through the action interface.

plugin/mouse/mouse.py
```diff
@@ -132,15 +132,14 @@ def mouse_wake():
...
 
         # Start drag
         ctrl.mouse_click(button=button, down=True)
 
     def mouse_drag_end():
         """Releases any held mouse buttons"""
-        buttons_held_down = list(ctrl.mouse_buttons_down())
-        for button in buttons_held_down:
+        for button in ctrl.mouse_buttons_down():
```
AndreasArvidsson, line 142: `ctrl.mouse_buttons_down()` returns `set[int]`. No need to wrap in a list.

plugin/mouse/mouse.py
```diff
@@ -151,11 +150,7 @@ def mouse_sleep():
 
         show_cursor_helper(True)
         stop_scroll()
-
-        # todo: fixme temporary fix for drag command
-        button_down = len(list(ctrl.mouse_buttons_down())) > 0
```
AndreasArvidsson, line 156: This is basically a recreation of the logic in `mouse_drag_end`, but worse.

plugin/mouse/mouse.py
```diff
@@ -132,15 +132,14 @@ def mouse_wake():
     def mouse_drag(button: int):
         """Press and hold/release a specific mouse button for dragging"""
         # Clear any existing drags
-        self.mouse_drag_end()
```
AndreasArvidsson, line 135: I was surprised this even worked. We should definitely call Talon actions through the action interface.

nriley (approved):

PR #1547, opened 2024-08-17
Skip non speakable abbreviation values

AndreasArvidsson:
Today a user gets warnings in the Talon log if the abbreviation value is not speakable

### DIFF
--- a/core/abbreviate/abbreviate.py
+++ b/core/abbreviate/abbreviate.py
@@ -454,7 +454,7 @@ abbreviations_list = get_list_from_csv(
     default=abbreviations,
 )
 
-# Matches letters and space. Talon accepts nothing else as a spoken form.
+# Matches letters and spaces, as currently, Talon doesn't accept other characters in spoken forms.
 PATTERN = re.compile(r"^[a-zA-Z ]+$")
 abbreviation_values = {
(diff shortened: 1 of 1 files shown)


AndreasArvidsson:
@nriley This should help with your log warnings

core/abbreviate/abbreviate.py
```diff
@@ -452,9 +454,15 @@
     default=abbreviations,
 )
 
+# Matches letters and space. Talon accepts nothing else as a spoken form.
```
phillco, line 457:
```suggestion
# Matches letters and spaces, as currently, Talon doesn't accept other characters in spoken forms.
```

I believe there was some discussion of newer model supporting numeric literals

phillco (approved):

nriley:
Note that this will only work for English language abbreviations. I tried creating a pattern that (within the limited Unicode awareness of Python's re module) would approximately match a letter in any language but this still barfs (e.g.) when using a German abbreviation on English Conformer. Short of adding a pattern for each supported language I think this is fine…

AndreasArvidsson:
I thought about that, but came to the conclusion that people are probably not using a list of english abbreviations for non english dictation.

PR #1548, opened 2024-08-19
Select final web area to avoid devtools

brollin:
On mac os, I was having an issue with `browser.host` which can be reproduced with the following steps.
1. put the following in a talon file:
```
browser.host: www.google.com
-
testing: app.notify("testing")
```
2. in firefox, open google.com
3. say "testing", verify you get the notification
4. open dev tools, focus another app, and then focused firefox again
5. say "testing", you will NOT get the notification 

It seems that in this scenario browser.host context matcher is no longer respected. I hunted down the issue to this line.

My change is probably pretty brittle, but at least works on mac (the only operating system where this code applies) chrome and firefox.

Some slack discussion took place here: https://talonvoice.slack.com/archives/C9MHQ4AGP/p1724030602939929

### DIFF
--- a/tags/browser/browser_mac.py
+++ b/tags/browser/browser_mac.py
@@ -27,7 +27,7 @@ class BrowserActions:
         except IndexError:
             return ""
         try:
-            web_area = window.element.children.find_one(AXRole="AXWebArea")
+            web_area = window.element.children.find(AXRole="AXWebArea")[-1]
             address = web_area.AXURL
         except (ui.UIErr, AttributeError):
             try:
--- a/tags/browser/browser_mac.py
+++ b/tags/browser/browser_mac.py
@@ -1,5 +1,4 @@
 from talon import Context, actions, app, mac, ui
-from talon.mac import applescript
 
 ctx = Context()
 ctx.matches = r"""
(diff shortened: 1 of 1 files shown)


tags/browser/browser_mac.py
```diff
@@ -27,7 +27,7 @@ def address():
         except IndexError:
             return ""
         try:
-            web_area = window.element.children.find_one(AXRole="AXWebArea")
+            web_area = window.element.children.find(AXRole="AXWebArea")[-1]
```
phillco, line 30: Is there a attribute of the correct webview that we can try to target instead (and add to the query), rather than trying to guess at the indices in the list? Seems like this would be more brittle over time

nriley: Yes, we can filter the URL. I cleaned this up a bit so it filters dev tools web areas by URL on both Edge and Firefox; may need to do some more testing with further Chromium-based browsers as I don't have any others installed. I also replaced (slower) AppleScript with (faster) appscript for a backup strategy. Firefox has no useful scripting support so this would only work in Chromium-based browsers (and Safari, which has its own implementation).

https://github.com/nriley/talon_community/commit/4860223edead4f00a02d4470a08c3f0b733bc0e2

nriley (changes requested):
See comment.

brollin:
Thanks for the help here. My talon/python environment is still somewhat broken, and without documentation of those APIs I just settled for the brittle solution. This is looking much better. On Mac, it works on both Firefox and Chrome for me. Hopefully it's all right that I used your code in my commit @nriley?

nriley:
To be clear, yes it's fine if you use it!

nriley (approved):

PR #1549, opened 2024-08-23
Update vscode.py to include Cursor for Linux as an option

JasoonS:
I've been using cursorless with Cursor the last few weeks with this.


phillco (approved):

nriley:
Cursor's macOS bundle ID (`com.todesktop.230313mzl4w4u92`) is already added. If someone can get the Windows .exe then we can get platform parity.

PR #1551, opened 2024-08-25
Support for GitHub and Jujitsu CLIs

qubitz:
Found myself using these quite often and decided to finally get around to contributing them. I'm sure there's a few standard practices that I'm missing. I primarily replicated the setup from git cli resulting in duplicate code with the csv handling. Not sure how to handle that.

I'm very ill practiced with talon. All suggestions welcome!

PS: I'd understand if jujitsu was too niche of an application to include by default with community 👌.


apps/jujitsu/jujitsu_commands.csv
```diff
@@ -0,0 +1,79 @@
...
+show
+sparse
+sparse edit
+sparse list
+sparse reset
+sparse set
+split
+squash
+status
+Subcommand, Spoken form (optional)
```
auscompgeek, line 67: That's an interesting subcommand

auscompgeek (commented):
I think we want to move away from CSVs for these sorts of lists, and use .talon-list instead

apps/jujitsu/jujitsu_commands.csv
qubitz: 😅 auto sort.

qubitz:
> I think we want to move away from CSVs for these sorts of lists, and use .talon-list instead

I agree, it is a much better system, it's just that I'm a little hesitant as there's not as many [examples](https://github.com/search?q=repo%3Atalonhub%2Fcommunity%20.talon-list&type=code). I'll make some attempts this weekend or next.

phillco:
> Jujitsu

Is it not [Jujutsu ](https://github.com/martinvonz/jj/blob/main/README.md)? :)

PR #1556, opened 2024-09-02
override edit.selected_text (preserve/restore selection after copy)

chdoc:
For quite a while, have been annoyed by homophone selection (specifically `phones word`) not working in `emacs`. I finally tracked this down to the following issue: the call to `selected_text` inside of  ` user.homophones_show_selection` actually clears the selection. As a consequence, "choose" appends the selected homophone to the original word, instead of replacing it. 

Clearing the selection is the expected behavior when copying in `emacs`, but this changes `selected_text` to restore the selection after copying.

Note: This still requires enabling `delete-selection-mode`, but I would deem that acceptable when editing natural language text.

Context:
```
phones word:
    edit.select_word()
    user.homophones_show_selection()

choose <user.formatters> <number_small>:
    result = user.homophones_select(number_small)
    insert(user.formatted_text(result, formatters))
    user.homophones_hide()
```

### DIFF
--- a/apps/emacs/emacs.py
+++ b/apps/emacs/emacs.py
@@ -1,7 +1,7 @@
 import logging
 from typing import Optional
 
-from talon import Context, Module, actions, settings
+from talon import Context, Module, actions, settings, clip
 
 mod = Module()
 mod.setting(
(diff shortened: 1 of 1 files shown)


chdoc:
I just realized that this by itself will not solve the issue. In fact, the change that I proposed causes issues with dictation mode. I will have to investigate this further.

nriley:
Closing with the blessing of @chdoc .

PR #1559, opened 2024-09-07
Fix regressions in system path support from #1268.

nriley:
Use `hostname:` rather than `host:` to (correctly) match a hostname.

Generate a default `system_paths-<hostname>.talon-list` akin to the
prior CSV.

Also makes some minor tweaks to the generated spoken forms:

- only use "profile" for Windows; "home" for other platforms
- add "talon recordings" (because I want to use it in the wiki)


knausj85 (approved):
This looks good, tested it.

Surprised the host thing didn't cause errors in the talon log. Good catch.

PR #1565, opened 2024-09-25
Add tab clone command and improve file organization

alexander-clarke:
Implements the tab clone command in vscode. As you can't have two tabs of the same file open in the same tab group it splits the window vertically which clones the tab left and right.

Also cleans up the functions, grouping all of the splits,py, and tabs.py functions together

### DIFF
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -223,6 +223,14 @@ class UserActions:
     def split_window():
         actions.user.vscode("workbench.action.splitEditor")
 
+    def split_number(index: int):
+        """Navigates to a the specified split"""
+        if index < 9:
+            if is_mac:
+                actions.key(f"cmd-{index}")
+            else:
+                actions.key(f"ctrl-{index}")
+
     # splits.py support end
 
     # multiple_cursor.py support begin
@@ -254,6 +262,9 @@ class UserActions:
     def multi_cursor_skip_occurrence():
         actions.user.vscode("editor.action.moveSelectionToNextFindMatch")
 
+    # multiple_cursor.py support end
+
+    # tabs.py support begin
     def tab_jump(number: int):
         if number < 10:
             if is_mac:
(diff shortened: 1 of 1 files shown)


apps/vscode/vscode.py
```diff
@@ -223,6 +223,14 @@ def split_window_vertically():
     def split_window():
         actions.user.vscode("workbench.action.splitEditor")
 
+    def split_number(index: int):
+        """Navigates to a the specified split"""
```
nriley, line 227: Please remove docstrings from implementations of actions.

apps/vscode/vscode.py
```diff
@@ -223,6 +223,14 @@ def split_window_vertically():
...
+        if index < 9:
+            if is_mac:
+                actions.key(f"cmd-{index}")
+            else:
+                actions.key(f"ctrl-{index}")
```
phillco, line 232: From the meetup: is there a VS Code action that we can use to do this?

It looks like @AndreasArvidsson  added `andreas.openEditorAtIndex` in his extension which makes me suspect it is not in the base VS Code

phillco: Oh looks like there is `workbench.action.openEditorAtIndex1` throuh `workbench.action.openEditorAtIndex9` :p. Shall we use that instead?

nriley (changes requested):
Would recommend you remove dependencies on specific keyboard bindings where you can. Thanks!

apps/vscode/vscode.py
AndreasArvidsson: The main purpose of my implementation is that it supports tabs higher than 9 as well as negative numbers (counting from the back).

nriley: Does it make sense for community to try to use your action if it's present?

AndreasArvidsson: I don't know. Ideally it would be a different grammar that supported larger numbers and also negative numbers. For community is probably better to strive for simplicity and more widespread support.

alexander-clarke: `workbench.action.openEditorAtIndex1` does the tab number rather than the split number, used `workbench.action.focusFirstEditorGroup` etc.

apps/vscode/vscode.py
alexander-clarke: Done

alexander-clarke:
Moved the tabs split to a vscode action now. I wasn't sure if the ordinals are available elsewhere to just use but I defined them in the function for now

knausj85 (approved):

PR #1568, opened 2024-10-07
Add talon open debug

peagha:
Command for opening debug window:
image

### DIFF
--- a/plugin/talon_helpers/talon_helpers.talon
+++ b/plugin/talon_helpers/talon_helpers.talon
@@ -1,4 +1,5 @@
 talon check updates: menu.check_for_updates()
+# the debug window is only available in the talon beta
 talon open debug: menu.open_debug_window()
 talon open log: menu.open_log()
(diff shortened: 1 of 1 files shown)


nriley:
This command is beta-only. At one point we had discussed having a separate branch for the beta version, or perhaps a tag. What do others think?

nriley:
We discussed this on the backlog session. We feel like because this is a low-impact feature, that it is fine to merge but should not be considered a precedent for further use of beta-only features in community.  Could you please add a comment that indicates that this functionality is beta-only? Thanks!

plugin/talon_helpers/talon_helpers.talon
```diff
@@ -1,4 +1,5 @@
 talon check updates: menu.check_for_updates()
+talon open debug: menu.open_debug_window()
```
nriley, line 2: Please add a comment as discussed.

knausj85 (approved):

Issue #1512, opened 2024-07-28
Elixir support


jaresty:
It would be nice if community had language configuration for the elixir language.

nriley:
I think this is done?

PR #1571, opened 2024-10-11
First pass at adding a language mode for elixir

jaresty:
I am not an elixir expert but I think this covers the basics. I wasn't able to find an elixir example in community forks so I put this together on my own. 

When this gets merged I will be on the lookout for feedback to improve it incrementally in subsequent changes.

### DIFF
--- a/lang/elixir/elixir.talon
+++ b/lang/elixir/elixir.talon
@@ -32,6 +32,8 @@ state rescue: "rescue"
 state after: "after"
 state end: "end"
 
+op pipe: " |> "
+
 # Elixir-specific keywords and symbols
 true: "true"
 false: "false"
--- a/lang/elixir/elixir.talon
+++ b/lang/elixir/elixir.talon
@@ -37,8 +37,7 @@ true: "true"
 false: "false"
 nil: "nil"
 
-[state] raise {user.elixir_exception}:
-    user.insert_between("raise ", "")
+[state] raise {user.elixir_exception}: user.insert_between("raise ", "")
 
 [state] rescue {user.elixir_exception}: "rescue {elixir_exception}"
 
--- a/lang/elixir/elixir.talon
+++ b/lang/elixir/elixir.talon
@@ -35,13 +35,7 @@ state end: "end"
 op pipe: " |> "
 
 # Elixir-specific keywords and symbols
-true: "true"
-false: "false"
-nil: "nil"
 
 [state] raise {user.elixir_exception}: user.insert_between("raise ", "")
 
 [state] rescue {user.elixir_exception}: "rescue {elixir_exception}"
-
-comment line: user.code_comment_line_prefix()
-comment block: user.code_comment_block_prefix()
--- a/lang/elixir/elixir.talon
+++ b/lang/elixir/elixir.talon
@@ -35,7 +35,6 @@ state end: "end"
 op pipe: " |> "
 
 # Elixir-specific keywords and symbols
-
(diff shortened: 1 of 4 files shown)


jaresty:
Addresses #1512

jaresty:
Related: https://github.com/cursorless-dev/cursorless/pull/1844

lang/elixir/elixir.talon
```diff
@@ -0,0 +1,47 @@
...
+true: "true"
+false: "false"
+nil: "nil"
+
+[state] raise {user.elixir_exception}: user.insert_between("raise ", "")
+
+[state] rescue {user.elixir_exception}: "rescue {elixir_exception}"
+
+comment line: user.code_comment_line_prefix()
+comment block: user.code_comment_block_prefix()
```
phillco, line 47: `user.code_comment_line`
`user.code_comment_block`

Unclear why the block commit uses "block comment" instead of comment block as the grammar 😭

lang/elixir/elixir.talon
```diff
@@ -0,0 +1,47 @@
...
+state def: "def "
+state defp: "defp "
+state if: "if "
+state else: "else"
+state case: "case "
+state cond: "cond do"
+state try: "try do"
+state rescue: "rescue"
+state after: "after"
+state end: "end"
```
phillco, line 33: I could swear we had a system for making these a list, but inspecting Python javascript in java, they all seem to put these up in the Talon file raw 🤷‍♂️. It might be nice to do that but it can be a followup

lang/elixir/elixir.talon
```diff
@@ -0,0 +1,47 @@
...
+
+op pipe: " |> "
+
+# Elixir-specific keywords and symbols
+true: "true"
+false: "false"
```
phillco, line 39: `tag: user.code_data_bool`, you'll need to use a `state` prefix though

but it might be worth a followup issue or PR if booleans are important enough to have unprefixed

lang/elixir/elixir.talon
```diff
@@ -0,0 +1,47 @@
...
+nil: "nil"
```
phillco, line 40: `tag: user.code_data_null`

phillco (changes requested):
The majority looks good to me, just need to reduce some of the bespoke spoken forms that can actually be implemented via tags

lang/elixir/elixir.talon
jaresty: It turns out that elixir does not have multiline comments, which was a surprise to me but I don't think I need to support it now.

lang/elixir/elixir.talon
```diff
@@ -0,0 +1,41 @@
...
+state cond: "cond do"
+state try: "try do"
+state rescue: "rescue"
+state after: "after"
+state end: "end"
+
+op pipe: " |> "
+
+# Elixir-specific keywords and symbols
+
```
phillco, line 38:
```suggestion
# Elixir-specific keywords and symbols
```

lang/elixir/elixir.talon
```diff
@@ -0,0 +1,41 @@
...
+[state] raise {user.elixir_exception}: user.insert_between("raise ", "")
```
phillco, line 39:
```suggestion
# Elixir-specific keywords and symbols
[state] raise {user.elixir_exception}: user.insert_between("raise ", "")
```

phillco (approved):

PR #1574, opened 2024-10-12
Add more per-platform keys in talon-lists.

nriley:
All platforms:
- keypad_enter
- insert (inadvertently removed in #1554)
- Allow dictating keypad keys directly without "press"

Mac:
- "function"/"globe" for globe/fn modifier (increasingly used by the OS)
- "return" for return (as printed on Apple keyboards)
- "enter" for [keypad] enter

The only one I'm a bit worried about is the last one. It is correct per Apple's definitions but it's possible some users are used to saying "enter" (though it's not the correct name for the key) and will get unexpected behavior; in most Mac apps I've tried, Return and Enter behave identically.

### DIFF
--- a/core/keys/keypad_key.talon-list
+++ b/core/keys/keypad_key.talon-list
@@ -1,5 +1,6 @@
 list: user.keypad_key
 -
+
 key pad zero: keypad_0
 key pad one: keypad_1
 key pad two: keypad_2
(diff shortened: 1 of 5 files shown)


knausj85 (approved):
I'm good with the change. I'll leave it open for others to poke.

phillco (approved):

PR #1576, opened 2024-10-16
Remove unprefixed numbers command

AndreasArvidsson:
This is a source of misrecognitions for new users. We have now had the deprecation message for nine months so I think it's fine to remove this.

### DIFF
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -177,7 +177,7 @@ for ten in tens:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
-mod.tag("prefixed_numbers", desc="Require prefix when saying a number")
+mod.tag("unprefixed_numbers", desc="Doesn't require prefix when saying a number")
 ctx.lists["self.number_small"] = number_small_map.keys()
 
 
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -177,7 +177,7 @@ for ten in tens:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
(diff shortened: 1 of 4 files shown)


jaresty:
Just to double check my understanding: after this change it will mean that everyone will need to say 'numb' in order to input a number?

AndreasArvidsson:
> Just to double check my understanding: after this change it will mean that everyone will need to say 'numb' in order to input a number?

Yes unless they change the command.

jaresty:
Thanks for the clarification. This is how I'm using it, so seems fine to me.

phillco (approved):

nriley:
We have frequently had questions in Slack asking how to allow unprefixed numbers and stop displaying the notification. We decided on the community backlog session to implement a user.unprefixed_numbers tag instead and the default will then become prefixed numbers without any warning.

nriley (changes requested):
See comment.

AndreasArvidsson:
Fixed

core/numbers/numbers.py
```diff
@@ -177,7 +177,7 @@ def split_list(value, l: list) -> Iterator:
 number_small_map = {n: i for i, n in enumerate(number_small_list)}
 
 mod.list("number_small", desc="List of small numbers")
-mod.tag("prefixed_numbers", desc="Require prefix when saying a number")
+mod.tag("unprefixed_numbers", desc="Dont require prefix when saying a number")
```
knausj85, line 180: Minor nitpick: should numbers.talon be disabled when this tag is active?

knausj85 (approved):

core/numbers/numbers.py
AndreasArvidsson: I'm not sure. I don't know if people want both active or only one of them?

nriley (approved):
Think this is good for a first attempt, addressing most of the issues we have heard about in slack, and we can always change things if there are further complaints.

PR #1578, opened 2024-10-20
Laying Out Windows Command

jaresty:
This pull request adds a configurable layout command. The layout command allows you to lay out multiple windows around the screen in prearranged configurations. With a single command, you can arrange multiple windows, and if you repeat the same command, it will rotate them. Here are some example arrangements:

Halves: Split the screen into two halves. The first window goes to the left half and the second goes to the right. Thirds: Split the screen into thirds, arranging from left to right. Clock: Arrange one window on the left half, and split the right from top to bottom.

When arranging windows if you specify nothing it will arrange in order of windows from top to bottom-in other words, the most recent three windows that you have interacted with will be snapped into the arrangement. If you want more control you can specify windows by saying an application name or using an ordinal such as 'second' to refer to the second window from the top of the window manager (the second most recently used window). If you want to skip a particular position when arranging windows, you can use the word 'gap' to skip a position. You can also use the word 'all' to refer to the rest of the windows available filling up all available slots. Here are some examples:

layout clock: Arrange the most recent three windows in a clockwise layout
layout halves chrome slack: Arrange chrome and slack in a split screen.
layout halves gap slack: Arrange slack on the right (skipping the first placement, which would have been on the left)
layout clock second all: Move these second from the top window to the first position, rearranging all other windows accordingly.
If you repeat any of these commands without interacting with the window using the mouse, it will rotate the arrangement.

TODO community meetup feedback
===

- [x] Create a fallback in `snap_window_to_position` in case people were hardcoding the spoken forms - we don't want them to break. We can just check if it's not in the map and see if the HammerCase version is, and show an alert but make the code work for now, can remove after six months
- [x] Create Talon list for the spoken forms of layouts
- [x] possibly convert snap identifiers to SHOUT_CASE to match formatters
- [x] lay → layout
- [x] better pull request description


core/windows_and_tabs/window_snap.py
```diff
@@ -345,22 +362,22 @@ def snap_app(app_name: str, position: RelativeScreenPos):
...
-            window.focus()
+        windows = [_get_app_window(app) for app in apps]
+        _snap_layout(positions, windows)
+
+    def snap_window_layout(
+        positions: list[RelativeScreenPos],
+        num_windows: int,
+    ):
+        """Snap next layout using the top three application windows"""
+        top_n_windows = ui.windows()[:num_windows]
```
jaresty, line 378: Some of these windows aren't snappable-that can result in an error. I don't know how to filter such windows here though.

core/windows_and_tabs/window_snap.py
```diff
@@ -302,14 +328,170 @@ def __init__(self, left, top, right, bottom):
...
+        snapped_windows.append(snapped_windows.pop(0))
+    for window in snapped_windows:
+        window.focus()
+        actions.sleep("180ms")
+    actions.sleep("500ms")
+    last_focus_was_done_by_snap_layout = True
+    currently_snapping_layout = False
+
+
+def _top_n_windows() -> list[Any]:
```
nriley, line 378: Consider calling this something like `_windows_under_active_window`?

core/windows_and_tabs/window_snap.py
```diff
@@ -302,14 +328,170 @@ def __init__(self, left, top, right, bottom):
...
+def window_split_position(m) -> list[RelativeScreenPos]:
+    if hasattr(m, "number_small"):
+        return _split_positions[m.window_split_positions][m.number_small]
+    else:
+        min_key = min(_split_positions[m.window_split_positions].keys())
+        return _split_positions[m.window_split_positions][min_key]
+
+
+@mod.capture(rule="all")
+def all_windows(m) -> list[Any]:
```
nriley, line 403: This also seems a strange name for the capture as it does not return all windows.

core/windows_and_tabs/window_snap.py
```diff
@@ -302,14 +328,170 @@ def __init__(self, left, top, right, bottom):
 }
 
 
+def _snap_next(windows: list[Any], target_layout: RelativeScreenPos) -> Optional[int]:
+    print(windows)
```
nriley, line 332: Would recommend you remove your debugging code if you feel confident of its quality.

core/windows_and_tabs/window_snap.py
```diff
@@ -212,6 +223,21 @@ def __init__(self, left, top, right, bottom):
         self.bottom = bottom
         self.right = right
 
+    def __str__(self):
+        # Return a string representation of the RelativeScreenPos instance
+        return f"RelativeScreenPos(left={self.left}, top={self.top}, right={self.right}, bottom={self.bottom})"
+
+
+class WindowLayout:
```
nriley, line 231: Consider making this into a `dataclass`.

core/windows_and_tabs/window_snap.py
```diff
@@ -302,14 +328,170 @@ def __init__(self, left, top, right, bottom):
...
+        if window is None:
+            return index
+        try:
+            _snap_window_helper(
+                window,
+                target_layout,
+            )
+            return index
+        except Exception as e:
+            print(f"Failed to snap window: {e}")
```
nriley, line 343: Would recommend you use notifications rather than just a log message to identify errors to the user.

core/windows_and_tabs/window_management.talon
```diff
@@ -17,8 +17,7 @@ snap last [screen]: user.move_window_previous_screen()
 snap screen <number>: user.move_window_to_screen(number)
 snap <user.running_applications> <user.window_snap_position>:
     user.snap_app(running_applications, window_snap_position)
-# <user.running_applications> is here twice to require at least two applications.
-snap <user.window_split_position> <user.running_applications> <user.running_applications>+:
-    user.snap_layout(window_split_position, running_applications_list)
+snap <user.window_layout>: user.snap_layout(window_layout)
```
nriley, line 20: This capture is ambiguous with the one in line 14. Would recommend you remove the other voice command or make it such that either the number or target windows are required in your `user.window_layout` capture.

nriley (commented):
From the community backlog session: we had a few specific comments on your implementation but are not sure the complexity of this implementation it's something we want to have in community.

core/windows_and_tabs/window_management.talon
jaresty: To avoid the ambiguity I'm going to try changing the name to see what you think. Invoking the layouts with no arguments is useful however so I don't want to require an argument.

core/windows_and_tabs/window_snap.py
jaresty: As many applications have hidden windows this is not really an error case-not in any significant sense. The zoom application for example always has a hidden window in front that cannot be snapped which would cause this all the time.

core/windows_and_tabs/window_snap.py
jaresty: I renamed it to reflect the fact that the windows are under the active window but we don't know whether they are resizable or not-there is not really a good way to detect that windows can be resized other than to try and resize them so some of these windows will be skipped.

core/windows_and_tabs/window_snap.py
jaresty: If the consensus is that this is a better thing to notify for than I'm happy to do it, though I expect that most people will be notified almost every time they run the command-sometimes even multiple times. I'm happy to handle this however is recommended.

phillco:
I think it's worth the complexity to get this in, personally. Window management with layouts is a good fit for voice, given the powerful nature of the grammar (and vs what could be done by the keyboard), so I think it's worth the complexity. It also improves functionality that's already been added to community. I also foresee building other workflows on this. Lastly, I think the layout logic could be moved to its own file, if needed so the commands can just be ignored if unwanted.

Therefore I intend to work on reviewing with a goal of merging this with @jaresty -- if other maintainers disagree, just reach out to me and we can chat; I couldn't attend the session but got the sense that the opposition wasn't too strongly opinionated (and I think the above context might help). I'll double check with maintainers before merging though.

core/windows_and_tabs/window_snap.py
phillco: Yeah, if this is an expected case for invisible windows that will happen frequently, I would change the print statement to be less scary (if we have it at all). I would also put the application name and window title in the message to see what it was.

core/windows_and_tabs/window_snap.py
```diff
@@ -299,14 +325,169 @@ def __init__(self, left, top, right, bottom):
...
+    return None
+
+
+def _snap_layout(window_layout: WindowLayout):
+    """Split the screen between multiple windows."""
+    global last_focus_was_done_by_snap_layout
+    global currently_snapping_layout
+    currently_snapping_layout = True
+    import copy
```
phillco, line 348: Imports inside of methods are usually discouraged

core/windows_and_tabs/window_snap.py
```diff
@@ -299,14 +325,169 @@ def __init__(self, left, top, right, bottom):
...
+        except Exception as e:
+            print(f"Failed to snap window: {e}")
+    return None
+
+
+def _snap_layout(window_layout: WindowLayout):
+    """Split the screen between multiple windows."""
+    global last_focus_was_done_by_snap_layout
+    global currently_snapping_layout
+    currently_snapping_layout = True
```
phillco, line 347: This sort of thing should be set in a try/finally, in case there's an exception

core/windows_and_tabs/window_snap.py
```diff
@@ -299,14 +325,169 @@ def __init__(self, left, top, right, bottom):
...
+            remaining_windows = remaining_windows[snapped_window_index + 1 :]
+
+    if window_layout.should_rotate and len(snapped_windows) > 0:
+        snapped_windows.append(snapped_windows.pop(0))
+    for window in snapped_windows:
+        window.focus()
+        actions.sleep("180ms")
+    actions.sleep("500ms")
+    last_focus_was_done_by_snap_layout = True
+    currently_snapping_layout = False
```
jaresty, line 371: Change  to a timestamp instead of using sleep

core/windows_and_tabs/window_snap.py
phillco, line 348: Same comment about imports inside of functions

core/windows_and_tabs/window_snap.py
```diff
@@ -299,14 +325,169 @@ def __init__(self, left, top, right, bottom):
...
+
+
+def _snap_layout(window_layout: WindowLayout):
+    """Split the screen between multiple windows."""
+    global last_focus_was_done_by_snap_layout
+    global currently_snapping_layout
+    currently_snapping_layout = True
+    import copy
+
+    target_layout = copy.deepcopy(window_layout.layout)
```
phillco, line 350: Is deep copying really needed here?

core/windows_and_tabs/window_snap.py
```diff
@@ -299,14 +325,169 @@ def __init__(self, left, top, right, bottom):
...
+        snapped_window_index = _snap_next(remaining_windows, target_layout.pop(0))
+        if snapped_window_index is not None:
+            snapped_windows.insert(0, remaining_windows[snapped_window_index])
+            remaining_windows = remaining_windows[snapped_window_index + 1 :]
+
+    if window_layout.should_rotate and len(snapped_windows) > 0:
+        snapped_windows.append(snapped_windows.pop(0))
+    for window in snapped_windows:
+        window.focus()
+        actions.sleep("180ms")
```
phillco, line 368: Sleeping for a fixed amount of time is going to be flaky and should be avoided wherever possible. If we're trying to see when the focus change has completed, you can make a helper action that will wait in a loop with a specified timeout.

core/windows_and_tabs/window_snap.py
```diff
@@ -299,14 +325,169 @@ def __init__(self, left, top, right, bottom):
...
+    actions.sleep("500ms")
```
phillco, line 369: As we discussed, if this is to catch late focus events in the callback incorrectly making the code think the user manually focused something, let's have it store the timestamp that the last layout attempt completed instead. This will allow you to make the grace period longer as well without making the action itself take longer, which can be problematic.

phillco (changes requested):
Let's move this into its own file for isolation purposes. You can refactor the existing actions to support snapping an arbitrary window (which I think would already be a good change) which I think will remove your dependency on this file's private members.

``` 
def snap_window(position: RelativeScreenPos, window: Optional[Window]) -> None:
        """Move the active window to a specific position on its current screen, given a `RelativeScreenPos` object."""
        if window is None:
            window = ui.active_window()
        _snap_window_helper(window, position)

    def snap_window_to_position(position_name: str, window: Optional[Window]) -> None:
        """Move the active window to a specifically named position on its current screen, using a key from `_snap_positions`."""
        actions.user.snap_window(_snap_positions[position_name], window=window)```

File: core/windows_and_tabs/window_snap.py (hunk 11 above)
Comment (jaresty) [reply core/windows_and_tabs/window_snap.py:350]: Since we're modifying the underlying data structure I think some level of copying is necessary. Maybe we could get away with the shallow copy? I'm not really a python expert here but I want to make sure that we don't modify the structure that the running process depends on-and I am modifying the underlying array to keep track of where I am in the snapping process.

File: core/windows_and_tabs/window_snap.py (hunk 6 above)
Comment (jaresty) [reply core/windows_and_tabs/window_snap.py:339]: Sure I can do that

File: core/windows_and_tabs/window_snap.py (hunk 12 above)
Comment (jaresty) [reply core/windows_and_tabs/window_snap.py:368]: I think I can remove the window focus sleeping-it was something I was trying but I think it's unnecessary if I instead use a top level timestamp to manage tracking which thing caused focus.

**Comment by jaresty** [comment]
I made quite a few changes based on your feedback so it would be really helpful if you could give it another look-thanks!

File: core/windows_and_tabs/window_layout.py [hunk 14]
```diff
@@ -0,0 +1,248 @@
...
+    """Split the screen between multiple windows."""
+    global last_focus_was_done_by_snap_layout
+    global lay_out_start_time
+
+    lay_out_start_time = datetime.now()
+
+    target_layout = copy.deepcopy(window_layout.layout)
+    remaining_windows = window_layout.windows
+    snapped_windows = []
+    snapped_window_index = 0
```
phillco, line 94:
```suggestion
    snapped_window_idx = 0
```

phillco (changes requested):

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,248 @@
...
+        try:
+            actions.user.snap_target_window_to_position(
+                window,
+                target_layout,
+            )
+            return index
+        except Exception as e:
+            print(
+                f"Failed to snap window for application (this is normal) {window.app.name} {window.title}: {e}"
+            )
```
phillco, line 80: Unsure about the `type(e).__name__`

phillco (commented):
.

jaresty:
I made most of the changes that we discussed today. I'm holding off on the 'lay again' helper that we discussed though because there are some things I'm not sure about how to handle in terms of edge cases so I want to focus on this functionality first. Specifically I wasn't sure what to do if you did not have a previous layout  since I am currently using none as a way to signal that the layout had been manually adjusted, which might complicate which behavior is correct when someone then tries to 'lay again'-I'm open to suggestions here.

jaresty:
@nriley I see that you marked this as a draft - I'm not planning to change any more unless I get feedback to do so. Is it ok to mark it ready for review now?

nriley:
> @nriley I see that you marked this as a draft - I'm not planning to change any more unless I get feedback to do so. Is it ok to mark it ready for review now?

Sure! We just wanted to separate it out during our review session as it was still being worked on.

jaresty:
Thanks! I heard from @phillco that I do still have some feedback coming so I'll wait until we meet to discuss.

core/windows_and_tabs/window_snap.py
```diff
@@ -216,72 +217,115 @@ def __str__(self):
...
+    "left third": "LeftThird",
+    "right third": "RightThird",
+    "left two thirds": "LeftTwoThirds",
+    "right two thirds": "RightTwoThirds",
+    "center small": "CenterSmall",
+    "left small": "LeftSmall",
+    "right small": "RightSmall",
+    "left large": "LeftLarge",
+    "right large": "RightLarge",
+    "top left": "TopLeft",
```
phillco, line 292: We should go ahead and make this a Talon list

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+    windows: list[Window]
+    should_rotate: bool
+    finish_time: float
+
+
+layout_in_progress: Optional[WindowLayout] = None
+last_layout: Optional[WindowLayout] = None
+
+
+SPLIT_POSITIONS = {
```
jaresty, line 48: Move this to the top of the file after the imports

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+        3: [
+            "Right",
+            "TopLeft",
+            "BottomLeft",
+        ],
+    },
+}
+
+
+def snap_next(windows: list[Window], target_layout: str) -> Optional[int]:
```
jaresty, line 74: Add a dock string explaining what this function does and what it does especially when it returns none at the end

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+    for idx, window in enumerate(windows):
+        if window is None:
+            return idx
+        try:
+            actions.user.snap_window_to_position(
+                target_layout,
+                window,
+            )
+
+            return idx
```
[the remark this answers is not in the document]
jaresty: Explain that none means a gap

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+@mod.capture(rule="all")
+def all_candidate_windows(m) -> list[Window]:
+    return filter_nonviable_windows(ui.windows())
+
+
+@mod.capture(rule="gap")
+def skip_window(m) -> list[Window]:
+    return [None]
```
jaresty, line 144: May be change this into a python singleton object

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+    windows = []
+    if hasattr(m, "layout_item_list"):
+        windows += [window for sublist in m.layout_item_list for window in sublist]
+
+    if hasattr(m, "all_candidate_windows"):
+        windows += [w for w in m.all_candidate_windows if w not in windows]
+    return windows
+
+
+def pick_split_arrangement(
```
jaresty, line 206: Add types to the input arguments as well

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+        not window_was_specified
+        and last_layout is not None
+        and perf_counter() - last_layout.finish_time > 1,
+        0,
+    )
+
+
+ctx = Context()
+
+ctx.lists["user.window_split_positions"] = SPLIT_POSITIONS.keys()
```
jaresty, line 252: Move this to the top right after defining it on the module

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+
+            return idx
+        except (UIErr, AttributeError) as e:
+            print(
+                f'Failed to snap {window.app.name}\'s "{window.title}" window ({type(e).__name__} {e});  this is normal; continuing to the next'
+            )
+    return None
+
+
+def snap_layout(window_layout: WindowLayout):
```
jaresty, line 92: rename window layout to just layout

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+    """Split the screen between multiple windows."""
+    global layout_in_progress
+    global last_layout
+
+    layout_in_progress = window_layout
```
jaresty, line 97: wrap this in a try/finally

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+    target_layout = copy.deepcopy(window_layout.layout)
```
jaresty, line 98: Change layout property to snap positions

jaresty: Rename variable to snap positions

jaresty: Can we do the copy in the capture instead?

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,265 @@
...
+            print(
+                f'Failed to snap {window.app.name}\'s "{window.title}" window ({type(e).__name__} {e});  this is normal; continuing to the next'
+            )
+    return None
+
+
+def snap_layout(window_layout: WindowLayout):
+    """Split the screen between multiple windows."""
+    global layout_in_progress
+    global last_layout
```
jaresty, line 95: These can be moved onto the same line a,be

core/windows_and_tabs/window_management.talon
```diff
@@ -17,8 +17,7 @@ snap last [screen]: user.move_window_previous_screen()
 snap screen <number>: user.move_window_to_screen(number)
 snap <user.running_applications> <user.window_snap_position>:
     user.snap_app(running_applications, window_snap_position)
-# <user.running_applications> is here twice to require at least two applications.
-snap <user.window_split_position> <user.running_applications> <user.running_applications>+:
-    user.snap_layout(window_split_position, running_applications_list)
+layout <user.window_layout>: user.snap_layout(window_layout)
```
knausj85, line 20: I really like this functionality. This is an example of a complicated grammar that might benefit from a little documentation. It's a little hard to quickly piece together what can be done just by looking at the python or talon files. 

Maybe a small readme makes sense for this functionality or a few comments in this file?

jaresty: Sounds good. I added to read me describing how to use it.

knausj85:
Trying to put this thru some paces this week.

I can't seem to get this command or any similar ones to fire on this branch at the moment:

'layout halves chrome slack'

Bug perhaps? Or am I missing something?

jaresty:
I just tried it and it worked for me, however I don't use chrome so I used `layout halves arc zoom`. Do you see anything in the logs? You also might want to try pulling with a rebase just to be sure you haven't diverged since I rebase the branch on main.

jaresty:
If it's useful I'd be happy to take a look at it with you at some point.

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,282 @@
...
+def skip_window(m) -> list[Window]:
+    return [Gap()]
+
+
+@mod.capture(rule="<user.running_applications>")
+def application_windows(m) -> list[Window]:
+    return [
+        window
+        for app in m.running_applications_list
+        for window in actions.self.get_running_app(app).windows()
```
knausj85, line 178: Is there a reason filter_nonviable_windows isn't called here?

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,282 @@
...
+def filter_nonviable_windows(windows: List[Window]) -> list[Window]:
+    active_window = ui.active_window()
+
+    # Many invisible non-resizable windows are identifiable because they exist above the current window
+    # in the z-index
+    all_windows = ui.windows()
+    active_window_idx = all_windows.index(active_window)  # type: ignore
+    return list(
+        filter(
+            lambda w: (isinstance(w, Gap) or all_windows.index(w) >= active_window_idx),
```
knausj85, line 149: I believe we should leverage this function https://github.com/talonhub/community/blob/6dd27a1d759dcbefd6ac13a62d72896ca2b84944/core/windows_and_tabs/windows_and_tabs.py#L31 for improved detection of invalid windows

So far, I have been able to address most of the issues I've encountered on windows with a lightly modified version of this function. 
```
def is_window_valid(window: ui.Window) -> bool:
    """Returns true if this window is valid for focusing"""
    return (
        not window.hidden
        # On Windows, there are many fake windows with empty titles -- this excludes them.
        and len(window.title) > 0
        and (window.title not in ("Chrome Legacy Window"))
        # This excludes many tiny windows that are not actual windows, and is a rough heuristic.
        and window.rect.width > window.screen.dpi
        and window.rect.height > window.screen.dpi
    )
```

core/windows_and_tabs/window_layout.py
jaresty: I wasn't having good a result with the existing filter nonviable windows method; I'm happy to try again after making the changes that you proposed.

core/windows_and_tabs/window_layout.py
jaresty: I can't figure out how to call it-can you point me to an example?

jaresty: This is what I was trying to do, though I tried a few variations that were somewhat similar without luck:

```
lambda w: (isinstance(w, Gap) or actions.app.is_window_valid(w)),
```

core/windows_and_tabs/window_layout.py
jaresty: It seems to be working well now after using the improved filter function.

core/windows_and_tabs/window_layout.py
```diff
@@ -0,0 +1,286 @@
...
+        filter(
+            lambda w: (isinstance(w, Gap) or is_window_valid(w)),
+            windows,
+        )
+    )
+
+
+mod = Module()
+mod.list(
+    "window_split_positions",
```
knausj85, line 159: Nit-picky, but I'd suggest renaming this and the associated talon file to "window_layout_list" or similar. 

This is mostly so it's more obvious it's associated with window_layout.py, and to differentiate from the split stuff that exists elsewhere when searching for one or the other. It's also a bit more consistent with the arguments in the actions.

knausj85: Actually, maybe not. :)

knausj85:
I've poking at a windows-specific issue where the windows don't always resize/position correctly that I'd like to resolve before merging. This might be an issue with talon or the API used to resize windows, tbd.

nriley:
From the community backlog session — it seems like we are approaching this PR being ready to merge. We discussed that @knausj85 will test a little bit more on the Mac and we will consider merging this behind a tag where it is disabled by default.

knausj85:
I've put this through a few paces on macOS and it look solid to me. I am comfortable merging this behind an opt in tag that is disabled by default. `experimental_window_layout` or something similarly appropriate

jaresty:
That seems fine to me. Could I have a hand implementing that tag? I gave it a try this morning and couldn't seem to get it working, and I'm not sure what I'm doing wrong.

knausj85:
I went ahead and implemented the tag. Please give it a try on your end when you have time. 

We should probably also update the readme to account for the tag

settings.talon
```diff
@@ -119,3 +119,7 @@ settings():
 # By default you need to say "numb one" to write "1". If you uncomment this,
 # you can say "one" to write "1".
 # tag(): user.unprefixed_numbers
+
+# Uncomment a blow to enable the experimental window layout commands
```
```suggestion
# Uncomment to enable the experimental window layout commands
```

jaresty:
I updated the readme and am trying out the tag and haven't run into any issue so far. Thank you!

PR #1581, opened 2024-10-24
Preserve fullscreen status when moving screens

knausj85:


nriley (commented):
On the community backlog session we were little confused about what this PR is attempting to accomplish, and on which platform.

On Windows, fullscreen and maximize appear to be the same thing, so I don't think this accomplishes anything.

On macOS, maximized is not a thing and this code exits before it does anything. I initially thought that the PR might have turned off fullscreen, moved it to another display and then turned it back on but this does not seem to be the case.

I don't have a Linux desktop available for testing.

knausj85:
Fullscreen (f11) and maximized are certainly two different things on windows, that sounds like a talon bug me. 

I didn't catch that as I was originally trying to get this to work on Mac (but couldn't), and it seemed to just work on windows

PR #1583, opened 2024-10-30
Fix maximize split VSCode action

david-tejada:
It looks like the vscode command has changed.


nriley (approved):
Thanks for submitting this. I made the same change in my own version but forgot to upstream it!

david-tejada:
No problem. I switched to having my customizations in a different repository and having community mostly unchanged. Now is much easier for me to create PRs.

Issue #1245, opened 2023-08-06
Consistent customisable way to use decimals


auscompgeek:
In the CSS support there's a couple of commands that recognise decimals:

https://github.com/talonhub/community/blob/d9c6373661eb937fc6fc1a29df4bfdce54b6914a/lang/css/css.talon#L27-L29

This only recognises how decimals are typically taught to be read in English-speaking classrooms (e.g. `three point one four` for `3.14`). Some people might like to be able to dictate the fractional part of decimals like whole numbers though, e.g. `twelve point thirty four` for `12.34`. Others might prefer to say `dot` instead of `point`.

We should add a `decimal_string` capture to allow this to be reused in other commands and customised globally.

auscompgeek:
Something else just came to mind: not all locales use `.` for decimals, instead using `,`. For programming though we'll still want a way to dictate decimals using `.` – not sure what the best way to do this is.

auscompgeek:
Oh, interesting, just took a look at how we do it in dictation mode and we have a similar construction to what we have in CSS:

https://github.com/talonhub/community/blob/d9c6373661eb937fc6fc1a29df4bfdce54b6914a/core/text/text_and_dictation.py#L49-L51

AndreasArvidsson:
I would just extend the normal number command with this ability. We can either do two separate commands in the Talon file or make a capture that enables both?
eg
https://github.com/AndreasArvidsson/andreas-talon/blob/bd3e75779295d9f139d8bc614ecb54e50899d704/core/numbers/numbers.py#L207-L213

PR #1590, opened 2024-11-09
Added support for decimal numbers

AndreasArvidsson:
`"numb fifty five point two"` -> `55.2`


AndreasArvidsson:
@nriley I ready for review :)

nriley (approved):

auscompgeek:
Does this address #1245?

nriley:
> Does this address #1245?

Mostly, I think. I wonder if we want to replace our dictation mode implementation with this too?

AndreasArvidsson:
> > Does this address #1245?
> 
> Mostly, I think. I wonder if we want to replace our dictation mode implementation with this too?

In my own repository I am using the same prose capture for dictation. That means I automatically get parity between the prose formatters and dictation mode.
I definitely think we should use the same captures whenever possiblepLooks like in community we do have some work to clean this up.

PR #1592, opened 2024-11-11
Make numbers more consistent

### LINKED ISSUE #1245: Consistent customisable way to use decimals

AndreasArvidsson:
Makes numbers more consistent between prose formatters and the global number command. 

Increase consistency and additional delimiters for different locales

Fixes #1245

### DIFF
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -234,7 +234,7 @@ def number_prose_with_colon(m) -> str:
 
 
 @mod.capture(
-    rule="<user.number_string> | <user.number_signed_string> | <user.number_prose_with_dot> | <user.number_prose_with_comma> | <user.number_prose_with_colon>"
+    rule="<user.number_signed_string> | <user.number_prose_with_dot> | <user.number_prose_with_comma> | <user.number_prose_with_colon>"
 )
 def number_prose_unprefixed(m) -> str:
     return m[0]
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -284,7 +284,7 @@ def number_small(m) -> int:
     return number_small_map[m.number_small]
 
 
-@mod.capture(rule="[negative|minus] <number_small>")
+@mod.capture(rule="[negative | minus] <number_small>")
 def number_signed_small(m) -> int:
     """Parses an integer between -99 and 99."""
     number = m[-1]
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -249,7 +249,7 @@ def number(m) -> int:
     return int(m.number_string)
 
 
-@mod.capture(rule="[negative | minus] <user_number_string>")
+@mod.capture(rule="[negative | minus] <user.number_string>")
 def number_signed_string(m) -> str:
     """Parses a (possibly negative) number phrase, returning that number as a string."""
     number = m.number_string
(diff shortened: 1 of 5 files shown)


core/numbers/numbers.py
```diff
@@ -243,30 +243,58 @@ def number_string(m) -> str:
...
+    return f"-{number}" if (m[0] in ["negative", "minus"]) else number
+
+
+@ctx.capture("number_signed", rule="<user.number_signed_string>")
+def number_signed(m) -> int:
+    """Parses a (possibly negative) number phrase, returning that number as a integer."""
+    return int(m.number_signed_string)
+
+
+@mod.capture(rule="<user.number_string> ((dot | point) <number_string>)+")
```
knausj85, line 265:
```suggestion
@mod.capture(rule="<user.number_string> ((dot | point) <user.number_string>)+")
```

core/numbers/numbers.py
```diff
@@ -243,30 +243,58 @@ def number_string(m) -> str:
...
+def number_prose_with_dot(m) -> str:
+    return ".".join(m.number_string_list)
+
+
+@mod.capture(rule="<user.number_string> (comma <number_string>)+")
```
knausj85, line 270:
```suggestion
@mod.capture(rule="<user.number_string> (comma <user.number_string>)+")
```

core/numbers/numbers.py
```diff
@@ -243,30 +243,58 @@ def number_string(m) -> str:
...
 def number(m) -> int:
     """Parses a number phrase, returning it as an integer."""
     return int(m.number_string)
 
 
-@ctx.capture("number_signed", rule=f"[negative|minus] <number>")
-def number_signed(m):
-    number = m[-1]
-    return -number if (m[0] in ["negative", "minus"]) else number
+@mod.capture(rule="[negative | minus] <user_number_string>")
```
knausj85, line 252:
```suggestion
@mod.capture(rule="[negative | minus] <user.number_string>")
```

knausj85 (changes requested):
Few typos I think, otherwise I think this change makes sense

core/numbers/numbers.py
```diff
@@ -243,30 +243,58 @@ def number_string(m) -> str:
...
+def number_signed_string(m) -> str:
+    """Parses a (possibly negative) number phrase, returning that number as a string."""
+    number = m.user_number_string
```
knausj85, line 255:
```suggestion
    number = m.number_string
```

knausj85:
There's one other inconsistency that might be worth addressing while in here, or as a separate issue: 
can't dictate negative numbers with decimals.

AndreasArvidsson:
Fixes applied.

Negative decimal numbers for something I thought about but since we have a specific signed numbers for integers do we need something similar for decimals or should we just make one capture supporting both? I think we should leave that as a separate issue.

knausj85 (approved):
looks good, thanks. I agree with handling the negative numbers in a separate issue.

AndreasArvidsson:
Issue filed
https://github.com/talonhub/community/issues/1615

PR #1594, opened 2024-11-12
Refactor mouse scrolling code

AndreasArvidsson:
mouse.py is to large so I have extracted all the code related to mouse growling into a separate file. I also took the liberty to clean up the code an simplify as much as I could.


knausj85 (approved):
This looks good to me.

PR #1595, opened 2024-11-12
Support double digit number small

### LINKED ISSUE #951: `<number_small>` doesn't support utterances of the form eg "two five"
Would be good to be consistent with `<number>` here.  Causes https://github.com/cursorless-dev/cursorless/issues/897

AndreasArvidsson:
Fixes #951

### DIFF
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -214,7 +214,7 @@ for d1 in range(1, 10):
     for d2 in range(10):
         number_small_map[f"{digit_list[d1]} {digit_list[d2]}"] = f"{d1}{d2}"
 
-mod.list("number_small", desc="List of small numbers")
+mod.list("number_small", "List of small (0-99) numbers")
 mod.tag("unprefixed_numbers", desc="Dont require prefix when saying a number")
 ctx.lists["user.number_small"] = number_small_map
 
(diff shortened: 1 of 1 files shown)


core/numbers/numbers.py
```diff
@@ -208,11 +208,15 @@ def split_list(value, l: list) -> Iterator:
 for ten in tens:
     number_small_list.append(ten)
     number_small_list.extend(f"{ten} {digit}" for digit in digit_list[1:])
-number_small_map = {n: i for i, n in enumerate(number_small_list)}
+number_small_map = {n: str(i) for i, n in enumerate(number_small_list)}
+# Add support for double single digits. eg. "five one" -> 51
+for d1 in range(1, 10):
+    for d2 in range(10):
+        number_small_map[f"{digit_list[d1]} {digit_list[d2]}"] = f"{d1}{d2}"
```
knausj85, line 215: I wonder if we should just consolidate this logic into get_spoken_form_under_one_hundred?

AndreasArvidsson: That would mean updating get_spoken_form_under_one_hundred to support that. That just runs a single pass from start to end, but to get the double digit feature you actually need nested loops. Is it worth making get_spoken_form_under_one_hundred that much more complicated?

knausj85: I personally really like this, as it creates a single point of failure. Thanks!

knausj85 (approved):

PR #1599, opened 2024-11-14
Extract code languages into separate file

### LINKED ISSUE #1585: Migrate language definitions to `code_languages.py`
From https://github.com/talonhub/community/pull/1528#issuecomment-2336158817:

> In the community session today we talked further on this topic. We think we should probably have a new python file `code_languages.py` where we have the language definitions. This python file can then be used by the language modes and the snippets. 
> 
> We want to have these language definitions typed and something like this could probably work: 
> ```py
> string_or_list = str | list[str]
> lang = tuple[str, string_or_list] | tuple[str, string_or_list, string_or_list]
> 
> languages: list[lang] = [
>     ("python", ["python"]),
>     ("java", "java"),
>     ("javascript", ["javascript"], ["something something"]),
> ]
> ```
>

AndreasArvidsson:
Fixes #1585

### DIFF
--- a/core/modes/language_modes.py
+++ b/core/modes/language_modes.py
@@ -1,79 +1,5 @@
+from .code_languages import code_languages, code_special_file_map
 from talon import Context, Module, actions, app
-
-# Maps language mode names to the extensions that activate them. Only put things
-# here which have a supported language mode; that's why there are so many
-# commented out entries. TODO: make this a csv file?
-language_extensions = {
-    # 'assembly': 'asm s',
-    # 'bash': 'bashbook sh',
-    "batch": "bat",
-    "c": "c h",
-    # 'cmake': 'cmake',
-    # "cplusplus": "cpp hpp",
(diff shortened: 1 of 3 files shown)


core/modes/language_modes.py
```diff
@@ -127,7 +51,7 @@ class Actions:
     def code_set_language_mode(language: str):
         """Sets the active language mode, and disables extension matching"""
         global forced_language
-        assert language in language_extensions
+        assert language in code_languages
```
knausj85, line 54: I don't believe this works, as code_languages is not a dictionary or list of strings

knausj85 (changes requested):
This looks good to me other than my single comment

core/modes/language_modes.py
AndreasArvidsson: You're completely right. I forgot to update that one.
Fixed now

knausj85 (approved):

PR #1600, opened 2024-11-14
Make draft editor more reliable

AndreasArvidsson:
Make submission of draft editor more reliable

Don't close the draft if we didn't manage to get the document text


knausj85 (approved):
looks good to me

PR #1603, opened 2024-11-16
Additional mouse clean up

AndreasArvidsson:


plugin/mouse/mouse.py
```diff
@@ -84,13 +90,14 @@ def mouse_sleep():
 
     def copy_mouse_position():
         """Copy the current mouse position coordinates"""
-        position = ctrl.mouse_pos()
-        clip.set_text(repr(position))
+        x = actions.mouse_x()
+        y = actions.mouse_y()
```
phillco, line 94:
```suggestion
        x, y = actions.mouse_x(), actions.mouse_y()
```

or
```
x, y = ctrl.mouse_pos()
```

plugin/mouse/mouse.py
```diff
@@ -34,6 +34,12 @@
 )
 
 
+@ctx.action_class("main")
```
phillco, line 37: Remind me why we use main again instead of user?

phillco (approved):
Two comments otherwise looks good to me

plugin/mouse/mouse.py
phillco: Could add a mouse_pos() action - I think getting the tuple is useful

AndreasArvidsson: We could. I saw that we're actually calling ctrl.mouse_pos in multiple files. I would could make a separate pr changing all of those?

plugin/mouse/mouse.py
AndreasArvidsson: `main` is for talent actions without a prefix. `mouse_click`, `key`, etc

PR #1604, opened 2024-11-16
Added setting for scroll acceleration

AndreasArvidsson:


plugin/mouse/mouse_scroll.py
```diff
@@ -184,7 +195,13 @@ def mouse_scroll_continuous(new_scroll_dir: Literal[-1, 1]):
...
 def scroll_continuous_helper():
     scroll_amount = settings.get("user.mouse_continuous_scroll_amount")
-    y = scroll_amount * scroll_dir / 10
+    acceleration_setting = settings.get("user.mouse_continuous_scroll_acceleration")
+    acceleration_speed = (
+        1 + min((time.perf_counter() - scroll_ts) / 0.5, acceleration_setting - 1)
+        if acceleration_setting > 1
+        else 1
+    )
+    y = scroll_amount * acceleration_speed * scroll_dir / 10
```
phillco, line 204: I know the existing code does this, but do we know why we have a setting for `mouse_continuous_scroll_amount` if we're just going to divide it by 10 arbitrarily 😭. Maybe the setting used to be different and we needed to change its meaning?

settings.talon
```diff
@@ -24,6 +24,9 @@ settings():
     # Set the scroll amount for continuous scroll/gaze scroll
     user.mouse_continuous_scroll_amount = 80
 
+    # Set the maximum acceleration factor when scrolling continuously. 1=constant speed/no acceleration.
+    user.mouse_continuous_scroll_acceleration = 1
```
phillco, line 28: Could we document the type of acceleration (ie linear)?

plugin/mouse/mouse_scroll.py
```diff
@@ -7,6 +8,7 @@
 scroll_job = None
 gaze_job = None
 scroll_dir: Literal[-1, 1] = 1
+scroll_ts: float = 0
```
phillco, line 11: Thought about suggesting `scroll_start_ts` which is a tiny bit clearer but uglier. Not strongly opinionated. Feel free to go either way.

plugin/mouse/mouse_scroll.py
AndreasArvidsson: I have no idea why and was very tempted to change it, but that would of course break every one who has changed the default setting.

settings.talon
AndreasArvidsson: done

plugin/mouse/mouse_scroll.py
AndreasArvidsson: done

plugin/mouse/mouse_scroll.py
knausj85: One of many great mysteries lost to time. Might be worth writing up an issue if either if you feel strongly that we should revisit

knausj85 (approved):
It works as advertised.

plugin/mouse/mouse_scroll.py
AndreasArvidsson: Agreed
https://github.com/talonhub/community/issues/1605

PR #1608, opened 2024-11-17
Fix MacOS "window" Command

thetomcraig:
This fixes the mac "window" command, which is now `com.apple.expose.awake`


nriley:
Unfortunately this doesn't work; it does the same thing as "focus".  I spent a bunch of time trying to figure out if there was a replacement CoreDock notification but can't find one.

However if the frontmost app has a dock icon there is a workaround using accessibility.

PR #1609, opened 2024-11-17
Move formatters to talon lists

knausj85:
Move formatters to talon lists for ease of customization


core/text/reformatters.talon-list
```diff
@@ -0,0 +1,5 @@
+list: user.reformatters
```
nriley, line 1: I realize you didn't name this but I just noticed it. While we're not consistent (…) most lists are named with a singular rather than plural noun. I don't think we need to worry about backwards compatibility here since reformatters are so new and I couldn't find any references on search.talonvoice.com aside from copies of community.

nriley (changes requested):

knausj85:
Oops I broke a help function. I'm going to convert this to draft until I can take a peek

core/text/formatters.py
```diff
@@ -421,6 +375,18 @@ def formatter_immune(m) -> ImmuneString:
     return ImmuneString(str(value))
 
 
+def get_all_phrase_and_prose_formatters() -> Tuple[Dict[str, str], Dict[str, str]]:
```
phillco, line 378: Per the discussion in the meetup, does this work?

```suggestion
def get_all_phrase_and_prose_formatters() -> tuple[dict[str, str], dict[str, str]]:
```

If so, let's use that everywhere because it doesn't require the type import, so it's easier to write. I can try see if we can make a new rule for it

test/stubs/talon/__init__.py
```diff
@@ -176,6 +176,12 @@ class Settings:
     """
 
 
+class Registry:
+    """
+    Implements something like talon's registry
```
phillco, line 181:
```suggestion
    Implements something like Talon's registry
```

core/text/formatters.py
knausj85: Seems to work.

nriley (changes requested):

core/text/formatters.py
```diff
@@ -421,6 +375,18 @@ def formatter_immune(m) -> ImmuneString:
...
+    """Returns dictionary of all phrase formatters and a dictionary of all prose formatters"""
+    all_phrase_formatters = {}
+    prose_formatters = {}
+
+    all_phrase_formatters.update(registry.lists["user.code_formatter"][0])
+    all_phrase_formatters.update(registry.lists["user.prose_formatter"][0])
+    all_phrase_formatters.update(registry.lists["user.reformatter"][0])
+    prose_formatters.update(registry.lists["user.prose_formatter"][0])
```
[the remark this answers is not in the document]
knausj85: I'd be inclined to separate the list action into a separate PR and sweep thru everything.

I changed the index to -1 earlier today, so not quite clear if it's still somehow appearing on your side (https://github.com/talonhub/community/commit/4899a1968a5b714a9d5dd82a38b3ada5e6373f62)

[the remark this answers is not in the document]
nriley: Looks like it happened because I had an in-progress review from this morning. I'll take a look tomorrow…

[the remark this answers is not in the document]
knausj85: Since Andreas added the action, I've updated the code to use it

core/text/formatters.py
```diff
@@ -421,6 +375,25 @@ def formatter_immune(m) -> ImmuneString:
     return ImmuneString(str(value))
 
 
+def get_all_phrase_and_prose_formatters() -> tuple[dict[str, str], dict[str, str]]:
+    """Returns dictionary of all phrase formatters and a dictionary of all prose formatters"""
```
nriley, line 379: This is the only place you call this a "phrase formatter". I get it (not a word formatter) but it's a bit confusing. Elsewhere (e.g. in the lists above) we just call this a "formatter".

Think we should have some documentation of all the different formatter types and how/where they can be used since it's become even more complicated.

knausj85: I adjusted the names and description of the function. I'd be inclined to separate further documentation into a separate issue.

core/text/formatters.py
```diff
@@ -362,10 +314,12 @@ def shrink_to_string_inside(text: str) -> tuple[str, str, str]:
     return text, "", ""
 
 
-@mod.capture(rule="{self.formatters}+")
+@mod.capture(
+    rule="({user.code_formatter} | {user.prose_formatter} | {user.reformatter})+ "
```
AndreasArvidsson, line 318:
```suggestion
    rule="({user.code_formatter} | {user.prose_formatter} | {user.reformatter})+"
```

AndreasArvidsson (approved):

nriley (approved):

PR #1610, opened 2024-11-17
Implement app Exposé with accessibility since the CoreDock notification broke

nriley:
Obsoletes #1608.

### DIFF
--- a/apps/dock/dock.py
+++ b/apps/dock/dock.py
@@ -19,8 +19,9 @@ class Actions:
 
 @ctx.action_class("user")
 class UserActions:
-    def dock_app_expose():
-        from pathlib import Path
+    def dock_app_expose(app: Optional[App]):
+        if app is None:
+            app = ui.active_app()
 
         app_name = Path(ui.active_app().path).stem
         dock_items = ui.apps(bundle="com.apple.dock")[0].children.find(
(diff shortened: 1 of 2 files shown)


apps/dock/dock.py
```diff
@@ -13,9 +13,33 @@ class Actions:
...
         """Send a CoreDock notification to the macOS Dock using SPI"""
 
+    def dock_app_expose():
+        """Activate app Exposé using the macOS Dock item for the frontmost app"""
+
 
 @ctx.action_class("user")
 class UserActions:
+    def dock_app_expose():
+        from pathlib import Path
```
phillco, line 23: Imports should probably be at the file module level unless it's needed to break a circular import

phillco, line 23: Why not something like this?

```suggestion
    def dock_app_expose(app: Optional[App]):
        if app is None:
            app = ui.active_app()
```

Than a user could choose to add:

```<user.running_applications> window expose user.dock_app_expose(running_applications) ```

I've generally taken this pattern in my own scripts rather than hardcoding `ui.active_app()` inside of it; it makes the action more flexible (even if we don't ship this particular grammar in community, although I think it's useful)

Comment (nriley) [reply apps/dock/dock.py]: `user.running_applications` returns a `str` rather than a `ui.App`. But this works:

```
^window <user.running_applications>$:
    app = user.get_running_app(running_applications)
    user.dock_app_expose(app)
```

Seems OK?

phillco (approved):

apps/dock/dock.py
phillco: Oops, correct.

PR #1611, opened 2024-11-23
Remove deprecated insert cursor action

AndreasArvidsson:


nriley (approved):
Over Phil's objection

PR #1612, opened 2024-11-23
Delimiter pairs

### LINKED ISSUE #546: Systematize handling of delimiters

AndreasArvidsson:
1. Adds a Talon list defining delimiter pairs
2. Adds a voice command to insert delimiter pairs 
3. Adds a voice command to wrap selection with delimiter pairs 
4. Extends compound edit command to wrap with delimiter pairs

These are the spoken forms I personal use (except for padding). Do we want to use something else?

What do we do with all the existing symbol commands?
https://github.com/talonhub/community/blob/bb240ba1cf958bd3e2c7791e5a57d09d205adc81/plugin/symbols/symbols.talon#L10-L45

Fixes #546


core/edit/delimiter_pair.py
```diff
@@ -0,0 +1,32 @@
+from talon import Module, actions
+
+mod = Module()
+
+mod.list("delimiter_pair", "List of matching pair delimiters")
+
+
+@mod.capture(rule="{user.delimiter_pair}")
+def delimiter_pair(m) -> list[str]:
```
phillco, line 9: When do we use `list` vs `List`?

core/edit/delimiter_pair.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@mod.action_class
+class Actions:
+    def delimiter_pair_insert(pair: list[str]):
+        """Insert a delimiter pair <pair> leaving the cursor in the middle"""
+        left, right = pair
+        actions.insert(f"{left}{right}")
+        for _ in right:
+            actions.edit.left()
```
phillco, line 24: Use `insert_between`?

core/edit/delimiter_pair.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def delimiter_pair_wrap_selection(pair: list[str]):
+        """Wrap selection with delimiter pair <pair>"""
+        left, right = pair
+        selected = actions.edit.selected_text()
+        actions.insert(f"{left}{selected}{right}")
+        for _ in right:
+            actions.edit.left()
```
phillco, line 32: ditto

core/edit/delimiter_pair.py
```diff
@@ -0,0 +1,32 @@
...
+mod.list("delimiter_pair", "List of matching pair delimiters")
+
+
+@mod.capture(rule="{user.delimiter_pair}")
+def delimiter_pair(m) -> list[str]:
+    pair = m.delimiter_pair.split()
+    assert len(pair) == 2
+    open = pair[0] if pair[0] != "space" else " "
+    close = pair[1] if pair[1] != "space" else " "
+    return [open, close]
```
phillco, line 14:
```suggestion
    # "space" requires a special written form because Talon lists are whitespace insensitive
    open = pair[0] if pair[0] != "space" else " "
    close = pair[1] if pair[1] != "space" else " "
    return [open, close]
```

(Feel free to rewrite, I don't fully understand the issue yet)

nriley (changes requested):
Please move the additional ways to insert delimiters into a separate file, remove those that have spoken forms equivalent to the once being introduced and deprecate all the rest.

AndreasArvidsson:
Changes performed. Also migrated deprecated symbol commands to a separate file and called the deprecated command action.

core/edit/insert_between.py
```diff
@@ -5,8 +5,8 @@
 
 @mod.action_class
 class module_actions:
-    def insert_between(before: str, after: str):
+    def insert_between(before: str, after: str, middle=""):
```
knausj85, line 8:
```suggestion
    def insert_between(before: str, after: str, middle: str=""):
```

I'm pretty sure import will fail without the type specified

core/edit/edit.talon
```diff
@@ -125,12 +125,18 @@ new line above: edit.line_insert_up()
...
 (pad | padding) <user.symbol_key>+:
     insert(" ")
     user.insert_many(symbol_key_list)
     insert(" ")
 
+# Insert delimiter pairs
+<user.delimiter_pair>: user.delimiter_pair_insert(delimiter_pair)
+
+# Wrap selection with delimiter pairs
+<user.delimiter_pair> that: user.delimiter_pair_wrap_selection(delimiter_pair)
```
knausj85, line 138: There is one subtlety with the new wrap command versus the deprecated commands: 
the cursor winds up within the delimiters when wrapping; previously, the cursor would wind up outside. I'm not entirely sure this change is undesirable. If the intent is to replace the deprecated commands, should we preserve the previous functionality?

knausj85: I suppose there are three scenarios we could cover with the grammar.

1. cursor between the delimiters. Done.
2. Wrapping selection with decimeters. Done, but need to settle the cursor position.
3. Patch in the delimiters, cursor after. I believe this was the original purpose of the "empty <delimiter>" commands, but they no longer behave as such. Should we bother with this?

core/edit/edit.talon
```diff
@@ -125,12 +125,18 @@ new line above: edit.line_insert_up()
...
 # Insert padding with optional symbols
-(pad | padding): user.insert_between(" ", " ")
+(padding): user.insert_between(" ", " ")
 (pad | padding) <user.symbol_key>+:
     insert(" ")
     user.insert_many(symbol_key_list)
     insert(" ")
 
+# Insert delimiter pairs
+<user.delimiter_pair>: user.delimiter_pair_insert(delimiter_pair)
```
knausj85, line 135: It seems slightly weird to define these in edit vs symbols.talon. I personally don't really care, but adds another place to find symbol-y things.

core/edit/edit.talon
AndreasArvidsson: Personally I prefer the cursor ending up inside but we can change it if people prefer. Should we get a few additional voices on this?

core/edit/edit.talon
AndreasArvidsson: moved

core/edit/edit.talon
knausj85: Yeah, let's see what @nriley and @phillco think :)

core/edit/edit_command_actions.py
```diff
@@ -1,36 +1,68 @@
...
+from typing import Callable, Union
 
 from talon import Module, actions
 
 
+# "simple" actions are actions that don't require any arguments. Only a type.
+# select, copy, delete, etc.
 @dataclass
-class EditAction:
+class EditSimpleAction:
```
phillco, line 10:
```suggestion
class EditSimpleAction:
    """An action without additional parameters or fields"""
```

nriley (approved):

phillco (approved):

nriley (approved):

PR #1619, opened 2024-11-24
Added tag for draft editor application running

### LINKED ISSUE #751: Switch draft editor running check to tag
https://github.com/knausj85/knausj_talon/blob/d57d65ae65a4b2d542ccfe44bc369569d46479c5/draft_editor/draft_editor.talon#L1

It's probable that this line isn't even doing anything because bools aren't supposed to be evaluated there

AndreasArvidsson:
Fixes #751

### DIFF
--- a/plugin/draft_editor/draft_editor.py
+++ b/plugin/draft_editor/draft_editor.py
@@ -3,6 +3,10 @@ from talon import Context, Module, actions, app, settings, ui
 mod = Module()
 mod.tag("draft_editor_active", "Indicates whether the draft editor has been activated")
 mod.tag(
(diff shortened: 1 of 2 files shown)


knausj85:
I'm not sure I understand why we need to make this change. The scope seems preferable so people don't accidentally assert the tag, and it works as is?

knausj85:
If anything, I would actually be inclined to move the focused tag over to a scope

AndreasArvidsson:
I guess the concern is that tags are more easily understandable and reliable and that the scope is unnecessarily complicated for a binary value.

nriley (approved):

PR #1621, opened 2024-11-24
move file manager to dynamic lists

knausj85:
WIP

How valuable are the imgui thingies?


alexander-clarke:
I find them quite useful particularly when there are multiple files with similar or difficult to pronounce names. Showing which file is which index is very useful in that context.

knausj85:
Note for my future self: dynamic lists are currently a beta only feature, so pausing this effort.

PR #1622, opened 2024-11-25
Update powershell to add option not to update the title

alexander-clarke:
Currently when using powershell the title is forced to be updated after every directory change automatically
This is updated by inserting a command into powershell which updates the title, which clutters your terminal quite quickly.

The better solution is that the user updates their powershell prompt function so that this is done automatically, and then turn off the forced title refresh.

This pull request adds the talon setting to turn off the automatic forced title refresh, and adds a readme on how to do this.

Additionally I've changed open file to actually automatically open the file. If the user just wants to insert the string of the file they can already use select file command. 

I'm not sure if it's best to add the powershell specific readme file or add the documentation in the project readme, or elsewhere, so that can move.

### DIFF
--- a/apps/powershell/README.md
+++ b/apps/powershell/README.md
@@ -11,3 +11,5 @@ function prompt {
   "$pwd" + '> '
 }
 ```
+
+Then you can set the setting `user.powershell_always_refresh_title` to false.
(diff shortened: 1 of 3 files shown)


knausj85:
looks good to me, thanks

knausj85 (approved):

PR #1627, opened 2024-11-30
Add address bar tag

knausj85:
Implement common grammar for applications that support address bars

I think it makes sense for this to be separate from the navigation tag, as not all applications support both.


tags/file_manager/file_manager.py
```diff
@@ -69,6 +73,11 @@
 ctx.lists["self.file_manager_files"] = []
 
 
+@ctx_file_manager.capture("user.address", rule="<user.system_diretory>")
```
AndreasArvidsson, line 76: Could we go directly to the system directory list instead of the capture? This appears to be done with the web browser context.

knausj85: Fixed. Mismerge from my local stuff where user.system_diretory actually exists

AndreasArvidsson (approved):

apps/finder/finder.py
```diff
@@ -75,3 +75,12 @@ def file_manager_select_file(path: str):
         """selects the file"""
         actions.key("home")
         actions.insert(path)
+
+    def address_bar_focus():
+        actions.key("cmd-shift-g")
+
+    def address_bar_copy_address():
+        actions.user.key("alt-cmd-c")
```
nriley, line 83: Don't think we want `user.` here.

nriley (changes requested):
Finder (and some other file browsers) don't really have an "address bar"; perhaps consider `user.address` instead?

knausj85:
> Finder (and some other file browsers) don't really have an "address bar"; perhaps consider `user.address` instead?

user.address works for me. Done.

nriley (approved):

nriley:
@knausj85 Only noticed this after I started merging with my own fork...

What do we plan to do about `user.address_focus` versus `browser.focus_address`?

knausj85:
> @knausj85 Only noticed this after I started merging with my own fork...
> 
> What do we plan to do about `user.address_focus` versus `browser.focus_address`?

I really don't think there's anything to be done. The browser implementation of `user.address_focus` users the existing implementation of browser.focus_address. 

This was why I slightly preferred user.address_bar.

nriley:
Gotcha, sorry I missed that!

PR #1628, opened 2024-12-01
Add cursorless support for jetbrains

asoee:


phillco (commented):
This requires your new extension, right? Can we detect if they have that or not? I guess it will error out either way but it might be nice if we can show a more precise error. (I know a lot of people were confused by the error the existing code shows that you don't have the old plugin because it's very esoteric)

jaresty:
Would it be possible to merge this change behind a tag? In other words only enable it if the tag is enabled? It would make it easier for folks to try out the change...just a thought.

jaresty:
For what it's worth, you can enable this without modifying the community repo: https://github.com/jaresty/mac-dotfiles/commit/dff0250d9d2148a6383b762620359f2f8e36d147

asoee:
@phillco: this PR is backwards compatible with the old plugin, except that it ALSO enables cursorless commands.
in honesty, there is 3 changes in this PR:
- Fix a bug in the error handling for send_idea_command, where the err handling itself caused a new exception, that was hiding the original error.
- Add the tag to enable cursorless for jetbrains.
- Add the extra stuff to the command server, to make it work with jetbrains

the tag part could of course be changed to do like for vscode here:
https://github.com/cursorless-dev/cursorless-talon/blob/main/src/apps/cursorless_vscode.py

asoee:
This PR now just enables the command server for jetbrains, following the way it is done for vscode.
enabling the cursorless tags is moved to cursorless-talon PR here: 
https://github.com/cursorless-dev/cursorless/pull/2695/files#diff-430a53a203d37ee06335d1e1b4ac6965ad818f615dbe2b453e5cb1250d4602ff

nriley (approved):

PR #1629, opened 2024-12-02
Allow Passing Window to Snap User Actions and Introduce Configurable Snap Name List

jaresty:
## Background:
We want to be able to call these snap positions from another module (#1578), but the existing action takes a spoken form. We'd rather not build a dependency on those, because they can be brittle if the user changes them; it's become a best practice recently (formatters, etc) to use static identifiers instead, and use a Talon list to provide the spoken form.

## Change:

- This change changes the keys of the snap positions dictionary to have static identifiers, instead of the previous spoken forms.
- A talon list is introduced to map the spoken forms to the new identifiers.
- To reduce user friction, as people are probably calling the existing action with a spoken form, we'll attempt to convert it to the new identifier if we can, to allow the action to succeed, although we will still show a deprecation warning so they can update their code.


core/windows_and_tabs/window_snap.py
```diff
@@ -316,13 +319,32 @@ def window_split_position(m) -> Dict[int, list[RelativeScreenPos]]:
 
 @mod.action_class
 class Actions:
-    def snap_window(position: RelativeScreenPos) -> None:
+    def snap_window(
+        position: RelativeScreenPos, window: Optional[Window] = None
+    ) -> None:
         """Move the active window to a specific position on its current screen, given a `RelativeScreenPos` object."""
```
phillco, line 325: We should update this docstring

core/windows_and_tabs/window_snap.py
```diff
@@ -316,13 +319,32 @@ def window_split_position(m) -> Dict[int, list[RelativeScreenPos]]:
...
+    def snap_window_to_position(
+        position_name: str, window: Optional[Window] = None
+    ) -> None:
         """Move the active window to a specifically named position on its current screen, using a key from `_snap_positions`."""
-        actions.user.snap_window(_snap_positions[position_name])
+        position: Optional[RelativeScreenPos] = None
+        if position_name in _snap_positions:
+            position = _snap_positions[position_name]
+        else:
+            screaming_snake_position = actions.user.formatted_text(
```
phillco, line 338:
```suggestion
            # Previously this function took a spoken form, but we now have constant identifiers in `_snap_positions`.
            # If the user passed a previous spoken form instead, see if we can convert it to the new identifier.
            new_key = actions.user.formatted_text(
```

core/windows_and_tabs/window_snap.py
```diff
@@ -316,13 +319,32 @@ def window_split_position(m) -> Dict[int, list[RelativeScreenPos]]:
...
+                position_name, "ALL_CAPS,SNAKE_CASE"
+            )
+            actions.user.deprecate_command(
+                "2024-12-02",
+                f"snap_window_to_position('{position_name}')",
+                f"snap_window_to_position('{screaming_snake_position}')",
+            )
+            position = _snap_positions[screaming_snake_position]
```
phillco, line 346: This should probably be a three-clause if:

- If a valid identifier was passed (works)
- if a previous spoken form was passed that matches a valid identifier when converted (convert, show deprecation warning → works)
- Something else was passed (just return KeyError or something similar, don't show the deprecation warning)

Presently this is mixing 2 and 3 because you're assuming the converted key is always present here without checking first; I would separate 3 out. Otherwise if I pass a nonsense identifier, it will still try to offer the deprecation code conversion, assuming I passed a spoken form, which isn't quite right.

jaresty: Makes sense. I updated it.

core/windows_and_tabs/window_snap.py
```diff
@@ -316,13 +319,35 @@ def window_split_position(m) -> Dict[int, list[RelativeScreenPos]]:
...
+        """Move a window (defaults to the active window) to a specifically named position on its current screen, using a key from `_snap_positions`."""
+        position: Optional[RelativeScreenPos] = None
+        if position_name in _snap_positions:
+            position = _snap_positions[position_name]
+        else:
+            # Previously this function took a spoken form, but we now have constant identifiers in `_snap_positions`.
+            # If the user passed a previous spoken form instead, see if we can convert it to the new identifier.
+            new_key = actions.user.formatted_text(position_name, "ALL_CAPS,SNAKE_CASE")
+            if new_key in _snap_positions:
+                actions.user.deprecate_command(
```
AndreasArvidsson, line 342: `deprecate_command` is only really used in a voice command. This is more `deprecate_action`. For some reason that action doesn't take a replacement argument, but I would probably add an optional replacement argument to `deprecate_action` an use that instead.

AndreasArvidsson (changes requested):

AndreasArvidsson (approved):

core/windows_and_tabs/window_snap.py
```diff
@@ -316,17 +319,39 @@ def window_split_position(m) -> Dict[int, list[RelativeScreenPos]]:
...
+            if new_key in _snap_positions:
+                actions.user.deprecate_action(
+                    "2024-12-02",
+                    f"snap_window_to_position('{position_name}')",
+                    f"snap_window_to_position('{new_key}')",
+                )
+                position = _snap_positions[new_key]
+                actions.user.snap_window(position, window)
+            else:
+                raise KeyError()
```
nriley, line 350: In other cases `KeyError` includes a parameter which is the key that was not found in the dictionary. Can you do this here?

core/windows_and_tabs/window_snap.py
```diff
@@ -316,17 +319,39 @@ def window_split_position(m) -> Dict[int, list[RelativeScreenPos]]:
...
 
     def move_window_next_screen() -> None:
         """Move the active window to a specific screen."""
-        _move_to_screen(ui.active_window(), offset=1)
+        actions.user.snap_window_to_position("top right", ui.active_window())
```
nriley, line 354: Not sure why you are using "top right" here. This is only supposed to move to another screen not move the window relative to the screen? Even if this is intentional, we are wondering why you are using the words "top right" instead of an identifier as you had implemented in this PR?

nriley (changes requested):

core/windows_and_tabs/window_snap.py
jaresty: Whoops - that was test code I didn't mean to commit. Will remove-was testing the fall back

phillco:
Can you make the title reflect everything the change does?

jaresty:
Adjusted the title to mention the new talon list. Let me know if you think I should add more.

phillco (approved):

phillco:
Ah weird @nriley's old requested changes was keeping this one in the "reviewed needed" state even though a new review had been requested.

PR #1632, opened 2024-12-03
More robust cancel that checks `_ts`

AndreasArvidsson:
Mimic and sometimes Dragon don't have this key.

### DIFF
--- a/plugin/cancel/cancel.py
+++ b/plugin/cancel/cancel.py
@@ -45,7 +45,7 @@ def pre_phrase(phrase: Phrase):
 
     # Check if the phrase is before the threshold
     if ts_threshold != 0:
-        start = getattr(words[0], "start", phrase["_ts"])
+        start = getattr(words[0], "start", None) or getattr(phrase, "_ts", ts_threshold)
         phrase_starts_before_threshold = start < ts_threshold
         ts_threshold = 0
         # Start of phrase is before threshold timestamp
--- a/plugin/cancel/cancel.py
+++ b/plugin/cancel/cancel.py
@@ -45,6 +45,7 @@ def pre_phrase(phrase: Phrase):
 
     # Check if the phrase is before the threshold
     if ts_threshold != 0:
+        # NB: mimic() and Dragon don't have this key.
         start = getattr(words[0], "start", None) or getattr(phrase, "_ts", ts_threshold)
         phrase_starts_before_threshold = start < ts_threshold
         ts_threshold = 0


plugin/cancel/cancel.py
```diff
@@ -45,7 +45,7 @@ def pre_phrase(phrase: Phrase):
 
     # Check if the phrase is before the threshold
     if ts_threshold != 0:
-        start = getattr(words[0], "start", phrase["_ts"])
+        start = getattr(words[0], "start", None) or getattr(phrase, "_ts", ts_threshold)
```
phillco, line 48:
```suggestion
        # NB: mimic() and Dragon don't have this key.
        start = getattr(words[0], "start", None) or getattr(phrase, "_ts", ts_threshold)
```

phillco (approved):

plugin/cancel/cancel.py
```diff
@@ -45,7 +45,8 @@ def pre_phrase(phrase: Phrase):
 
     # Check if the phrase is before the threshold
     if ts_threshold != 0:
-        start = getattr(words[0], "start", phrase["_ts"])
+        # NB: mimic() and Dragon don't have this key.
+        start = getattr(words[0], "start", None) or getattr(phrase, "_ts", ts_threshold)
         phrase_starts_before_threshold = start < ts_threshold
```
phillco, line 50: Can they both be None? If so will None < <ts_threshold, ie some int> throw an error?

AndreasArvidsson: They can. That's why we have `ts_threshold` as the last default. That means that `delta` is zero and nothing happens.

plugin/cancel/cancel.py
```diff
@@ -45,7 +45,8 @@ def pre_phrase(phrase: Phrase):
 
     # Check if the phrase is before the threshold
     if ts_threshold != 0:
-        start = getattr(words[0], "start", phrase["_ts"])
+        # NB: mimic() and Dragon don't have this key.
+        start = getattr(words[0], "start", None) or getattr(phrase, "_ts", ts_threshold)
```
auscompgeek, line 49: This behaves very differently to previous. This is now looking for `phrase._ts` instead of `phrase["_ts"]`.

Does a Phrase have both key access and attribute access?

auscompgeek (changes requested):

plugin/cancel/cancel.py
phillco: Ah! Right. I'm not used to `getattr`

plugin/cancel/cancel.py
AndreasArvidsson: You're totally right. This should have been a `.get` on phrase. Good catch.
https://github.com/talonhub/community/pull/1633

PR #1636, opened 2024-12-11
Fix select/copy/cut way left/right

alexander-clarke:
Currently edit modifier "way left" and "way right" uses actions.user.select_line_start/end

This action based upon it's comment description should be the same as actions.edit.extend_line_start/end but they have different behaviour.

Behaviour when using the current user action: before doing an extend line action it does a edit.left action, leading to a different selection than intended.

Prior to the move to the compound edit command `select way left` was using the actions.edit.extend_line_ action

I'm also not sure if the older action is actually necessary when we have the edit versions?


alexander-clarke:
Relates to and fixes #1630

nriley (approved):
We agreed on the community backlog session that this is more consistent with the other selection commands.

Issue #1561, opened 2024-09-08
keys.py - figure out customization of symbol_key and punctuation lists


knausj85:
After merging https://github.com/talonhub/community/pull/1555, we may want to sort out how to ease the customization of symbol_key and  punctuation list

- Currently, symbol_key_words is imported by create_spoken_forms.py. We may want to re-evaluate whether we really utilize this functionality so we can move to .talon-list. Otherwise, gotta explore other options
- `punctuation_words` (used to populate the punctuation list) is a subset of `symbol_key_words` (used to populate the symbol_key list) that's intended to be available in both command and dictation. `symbol_key_words` is only used in command mode.  It's unclear how we can cleanly port this to .talon-list files without duplication. Some options that come to mind...
     - Without the create_spoken_form dependency, we could simply update the existing symbol_key capture to utilize both lists, and port to .talon-list
     - a CSV with a custom parser where we can somehow specify the intended scope. 
     - simply split this out into its own python file for easier customization
     - insert other options here :)

AndreasArvidsson:
I would probably just remove that feature from create spoken forms and instead just strip symbols. Once that is done we can just update captures to use both lists. No duplication needed.

knausj85:
I'm inclined to agree with the approach of simply removing the symbols from create spoken forms. I'll mark this as to discuss for next time (hopefully)

PR #1638, opened 2024-12-14
Remove symbols from create spoken forms & migrate to *.talon-list

knausj85:
https://github.com/talonhub/community/issues/1561


knausj85:
This is ready for folks to test in order to decide whether we want to proceed with the change.

core/create_spoken_forms.py
```diff
@@ -15,14 +14,12 @@
 DEFAULT_MINIMUM_TERM_LENGTH = 2
 EXPLODE_MAX_LEN = 3
 FANCY_REGULAR_EXPRESSION = r"[A-Z]?[a-z]+|[A-Z]+(?![a-z])|[0-9]+"
```
knausj85, line 16: strip apostrophes, ensure everything else is a space.

knausj85:
After our discussion, I've settled on a compromise of keeping support for a handful of symbols. Unfortunately, this does mean folks have to define them in two places. However, I feel it is a step in the right direction. 

This is ready for review.

core/keys/symbol_key.talon-list
```diff
@@ -0,0 +1,56 @@
...
+quote: "'"
+question: ?
+apostrophe: "'"
+L square: [
+left square: [
+brack: [
+bracket: [
+left bracket: [
+square: [
+R square: ]
```
nriley, line 15: Would recommend you capitalize R everywhere (my preference) or nowhere; right now it's inconsistent.

core/keys/symbol_key.talon-list
```diff
@@ -0,0 +1,56 @@
...
+tilde: ~
+bang: !
+down score: _
+underscore: _
+paren: (
+brace: {
+left brace: {
+curly bracket: {
+left curly bracket: {
+r brace: }
```
nriley, line 36: We have "R" here but not "L".

nriley (changes requested):

core/keys/symbol_key.talon-list
knausj85: This variant was apparently never in keys.py. Added.

core/keys/symbol_key.talon-list
knausj85: Fair enough, this was auto converted and apparently inconsistent in the original. Fixed

nriley (commented):
Looks good to me, but I haven't run with it yet!

core/create_spoken_forms.py
```diff
@@ -6,12 +6,18 @@
 
 from talon import Module, actions
 
-from .keys.keys import symbol_key_words
 from .numbers.numbers import digits_map, scales, teens, tens
 from .user_settings import track_csv_list
 
 mod = Module()
 
+# Keep support for a handful of symbols.
```
AndreasArvidsson, line 14: Elaborate on why we're keeping these three

knausj85: Done

AndreasArvidsson (approved):

knausj85:
I'm having second thoughts about this one due to the response on https://github.com/talonvoice/talon/issues/680

Maybe we're better off doing something similar to the code languages stuff and defining these in a separate python file.

AndreasArvidsson:
> I'm having second thoughts about this one due to the response on [talonvoice/talon#680](https://github.com/talonvoice/talon/issues/680)
> 
> Maybe we're better off doing something similar to the code languages stuff and defining these in a separate python file.

I guess that's our best option at the moment.

nriley (commented):
Based on feedback from aegis and discussion at the community backlog session, we have decided that this is not the approach we want to take going forward.

Issue #1572, opened 2024-10-12
Reimplement or remove libraries_gui?


nriley:
This functionality has been broken for some time (e.g., it tries to modify in place the tag frozenset) and it uses the Talon registry in a way that is not recommended.

On the community backlog session we were not sure if this functionality actually gets any use or would be useful in practice.

Also see #1251.

nriley:
On the community backlog session we discussed this again briefly and decided to remove this as it does not seem anybody is complaining about its lack of functionality.

PR #1639, opened 2024-12-15
Remove nonfunctional [code_]libraries_gui support

nriley:
See #1572.


knausj85 (approved):

knausj85:
I think libraries_gui.py needs to be purged as well? Otherwise looks good

knausj85 (approved):

PR #1640, opened 2024-12-15
Restore no-op behavior for word formatter

nriley:


auscompgeek (approved):
@phillco I think [raises a good point](https://talonvoice.slack.com/archives/CJ2EE3LEB/p1733272273099959?thread_ts=1733179823.125689&cid=CJ2EE3LEB) in favour of this:

> having a no-op version is useful for the case where you have acronyms in your additional words (like "iCloud") which the formatters will otherwise mangle.

PR #1641, opened 2024-12-15
migrate more edit commands

knausj85:


core/edit/edit_command_modifiers.talon-list
```diff
@@ -3,12 +3,18 @@ list: user.edit_modifier
...
 way left: lineStart
 line end: lineEnd
 way right: lineEnd
+up: lineUp
+down: lineDown
 file start: fileStart
 way up: fileStart
 file end: fileEnd
 way down: fileEnd
+left: left
```
knausj85, line 19: Everything works as expected, but adding left and right adds some more-or-less worthless commands to the grammar since it extends by a character. 

"cut left" "copy left" are pretty dubious.

I wonder if left / right / word left / up / down should be re-worked to be consistent with our arrows grammar e.g. copy 2 left, copy 2 up etc instead.

knausj85: Ensure this idea works with classic repeaters

knausj85: > Ensure this idea works with classic repeaters

@nriley FYI eg cut word left 3rd vs cut 3 word left will render different results for likely obvious reasons.

Since cut is destructive, only the final word would be on the clipboard with the classic repeaters, whereas cut 3 word left can be implemented to do what I would expect: three words on the clipboard. 

Copy/Selection actions work as expected with either.

core/edit/edit_command.py
```diff
@@ -1,41 +1,177 @@
...
 from .edit_command_actions import EditAction, run_action_callback
 from .edit_command_modifiers import EditModifier, run_modifier_callback
 
-# In some cases there already is a "compound" talon action for a given action and modifier
+mod = Module()
+mod.setting(
+    "edit_command_word_selection_delay",
+    type=int,
+    default=75,
+    desc="Sleep required between word selections",
```
nriley, line 11: Can you explain a bit more as to why this is necessary?

nriley (changes requested):

core/edit/edit_command.py
knausj85: Added comment above the setting.

nriley:
Working well so far. In merging it into my own repo the largest issue I ran into was that there's no consistency between navigation (`go …`) and editing (`select …`). Seems that we should consider reusing `edit_command_modifiers_repeatable` in `navigation_step` since you would typically want to use the same words in both cases. Probably best in another PR; can file as an issue if you want.

nriley (approved):

knausj85:
I agree with consolidating those grammars; let's do a separate PR/issue for it if you don't mind.

nriley:
Filed #1728. Thanks!

PR #1645, opened 2024-12-19
Add comma-separated list formatter

BlueDrink9:


core/text/formatters.py
```diff
@@ -239,6 +239,7 @@ def de_camel(text: str) -> str:
     CodeFormatter("SNAKE_CASE", "_", lower, lower),
     CodeFormatter("DASH_SEPARATED", "-", lower, lower),
     CodeFormatter("DOT_SEPARATED", ".", lower, lower),
+    CodeFormatter("COMMA_SEPARATED", ", ", lower, lower),
```
AndreasArvidsson, line 242: We already have a formatter with this name. See line 229 above. If there is something with the behavior of the old formatter you want to change please write a test for that.

core/text/code_formatter.talon-list
```diff
@@ -4,6 +4,8 @@ all cap: ALL_CAPS
 all down: ALL_LOWERCASE
 camel: PRIVATE_CAMEL_CASE
 dotted: DOT_SEPARATED
+listed: COMMA_SEPARATED
```
AndreasArvidsson, line 7: We already have a formatter "list" in the reformat Talon list. If you want to be able to insert we should just move that to code formatters instead.

core/text/code_formatter.talon-list
```diff
@@ -4,6 +4,8 @@ all cap: ALL_CAPS
...
+args: COMMA_SEPARATED
```
AndreasArvidsson, line 8: "args" is used in community to insert parentheses. We probably shouldn't overload this with another definition.

AndreasArvidsson (changes requested):

core/text/formatters.py
BlueDrink9: Oops, sorry for missing that

core/text/code_formatter.talon-list
BlueDrink9: I can't find where that reformatter list gets used in the docs, nor in any cheatsheets - How does it get used? Why would we want to move 'list' as a verb rather than adding a new transitive verb 'listed'?

AndreasArvidsson: Everything that uses the formatters capture. eg
https://github.com/talonhub/community/blob/25cf36af08984a74acf4b4127a521621de5bfafa/core/text/text.talon#L16

Because we already have a spoken form for this and we don't want two different spoken forms for the same thing. Either leave it as a reformatter or move it to code formatters.

BlueDrink9: Thanks, done that now

BlueDrink9:
Rebased and addressed review comments

AndreasArvidsson (approved):

PR #1649, opened 2024-12-21
Add continuous scrolling speed adjustment

### LINKED ISSUE #1648: implement scrolling speed adjustment commands
implement ability to change continuous scrolling speed while scrolling.

FireChickenProductivity:
Add the ability to adjust continuous scrolling speed while scrolling by dictating a number_small. The speed becomes the speed setting multiplied by the dictated number divided by ten. The acceleration gets reset every time the speed gets changed. The scrolling speed reverts to the default after the current scroll is finished. 
closes #1648


FireChickenProductivity:
I added the ability to set the scrolling speed when starting continuous scrolling.

plugin/mouse/mouse_scroll.py
```diff
@@ -11,10 +11,11 @@
 scroll_start_ts: float = 0
 hiss_scroll_up = False
 control_mouse_forced = False
+continuous_scrolling_speed_factor = 1
 
 mod = Module()
 ctx = Context()
-
+continuous_scrolling_context = Context()
```
AndreasArvidsson, line 18: Remove and just use the above context instead

plugin/mouse/mouse_scroll.py
```diff
@@ -84,13 +90,13 @@ def mouse_scroll_right(amount: float = 1):
         x = amount * settings.get("user.mouse_wheel_horizontal_amount")
         actions.mouse_scroll(0, x)
 
-    def mouse_scroll_up_continuous():
+    def mouse_scroll_up_continuous(speed_factor: float = 10):
```
AndreasArvidsson, line 93: change type to integer
`10` could be a constant at the top of the file

plugin/mouse/mouse.talon
```diff
@@ -86,10 +86,10 @@ wheel tiny [down]: user.mouse_scroll_down(0.2)
 wheel tiny [down] here:
     user.mouse_move_center_active_window()
     user.mouse_scroll_down(0.2)
-wheel downer: user.mouse_scroll_down_continuous()
-wheel downer here:
+wheel downer [<number_small>]: user.mouse_scroll_down_continuous(number_small or 10)
```
AndreasArvidsson, line 89: pass none instead of 10 and make it optional in the actions

AndreasArvidsson (changes requested):

plugin/mouse/mouse.talon
FireChickenProductivity: As none can only be passed in beta, do you think it would be better to just have a separate version of the command for the argument or to do something like use "" or -1 to denote none? I personally think a separate version would be easier to read, but I do not have a strong opinion either way.

AndreasArvidsson: Unfortunately making separate commands is the best way for now I think.

FireChickenProductivity: I made the change, but there is a bug I should fix before merge.

FireChickenProductivity:
I found a bug that should be resolved before merge.

FireChickenProductivity:
Actually, the "issue" seems to happen without my pull request. Is it intended for some reason for using the continuous scrolling action twice in the same direction to stop the scrolling completely?

AndreasArvidsson:
Yes that is on purpose. For people using a foot switch, noise or other to toggle the scrolling on and off.

AndreasArvidsson:
@knausj85 Will test this. If someone else has time before that please go ahead.

plugin/mouse/mouse_scroll.py
```diff
@@ -181,13 +206,17 @@ def mouse_scroll_continuous(new_scroll_dir: Literal[-1, 1]):
         continuous_scroll_mode = "scroll down continuous"
```
knausj85, line 206: As written, I believe this will always display "scroll down continuous"

plugin/mouse/continuous_scrolling.talon
```diff
@@ -0,0 +1,3 @@
+tag: user.continuous_scrolling
```
knausj85, line 1: I think we need to ensure this coexists well with the tag user.numbers_unprefixed. Since the talon files have the same specificity, it's ambiguous which one talon will choose. Probably just add `not tag: user.continuous_scrolling` to that file

lunixbochs: specificity doesn't work on commands right now besides exact matches. it's pretty complicated to factor in .talon specificity in the command parser.

plugin/mouse/mouse_scroll.py
```diff
@@ -1,20 +1,22 @@
 import time
-from typing import Literal
+from typing import Literal, Optional
 
 from talon import Context, Module, actions, app, cron, ctrl, imgui, settings, ui
 
+DEFAULT_CONTINUOUS_SCROLLING_SPEED_FACTOR: int = 10
```
knausj85, line 6: I'm inclined to make this a setting. Any reason not to do so?

plugin/mouse/continuous_scrolling.talon
knausj85: In this case, the commands are not an exact match. unprefixed numbers use a different capture, so I don't think an action really nets us anything in this context either.

knausj85:
> Actually, the "issue" seems to happen without my pull request. Is it intended for some reason for using the continuous scrolling action twice in the same direction to stop the scrolling completely?

I initially thought this was a bug too; these commands used to be compatible with repeaters and such. I guess I missed that memo!

If needed, we can explore how to make these more friendly for hotkeys / pedals in a separate issue

I tested this a bit and it's a great improvement. I've a few minor suggestions before we merge.

knausj85:
> Actually, the "issue" seems to happen without my pull request. Is it intended for some reason for using the continuous scrolling action twice in the same direction to stop the scrolling completely?

There is a separate few bugs with this logic though that we might as well fix while we're here:
- the imgui should probably close & tag be cleared when this happens. 

I took a quick swing at fixing that.

plugin/mouse/mouse_scroll.py
FireChickenProductivity: I can make that change.

FireChickenProductivity:
I identified an issue where if the scrolling speed becomes less than one but non-zero the result is no scrolling. Should I address that by rounding up? (This can happen when trying to set the scrolling speed to a small number)

FireChickenProductivity:
It looks like float values for the scrolling speed get truncated, so changing the scrolling speed from some values to certain other values through voice commands has no effect. I am not sure if we should try to address that to make increasing speed value always increase the actual scrolling speed, but the issue is worth noting.

plugin/mouse/mouse_scroll.py
FireChickenProductivity: I fixed that.

plugin/mouse/mouse_scroll.py
FireChickenProductivity: I made it a setting.

plugin/mouse/continuous_scrolling.talon
```diff
@@ -0,0 +1,3 @@
...
+-
+<number_small>: user.mouse_scroll_set_speed(number_small)
```
nriley, line 3: On the community backlog session we recommended that you add a graphical display of the presence of this feature.

plugin/mouse/mouse_scroll.py
```diff
@@ -115,10 +127,12 @@ def mouse_gaze_scroll_toggle():
 
     def mouse_scroll_stop() -> bool:
         """Stops scrolling"""
-        global scroll_job, gaze_job, continuous_scroll_mode, control_mouse_forced
+        global scroll_job, gaze_job, continuous_scroll_mode, control_mouse_forced, continuous_scrolling_speed_factor, ctx
```
nriley, line 130: You don't need `ctx` here unless you're writing directly to it.

plugin/mouse/mouse_scroll.py
```diff
@@ -138,6 +152,24 @@ def mouse_scroll_stop() -> bool:
...
+        if speed is None:
+            continuous_scrolling_speed_factor = 1.0
+        else:
+            continuous_scrolling_speed_factor = speed / settings.get(
+                "user.mouse_continuous_scroll_speed_quotient"
+            )
+
+    def mouse_is_continuous_scrolling():
+        """Returns whether continuous scroll is in progress"""
+        if continuous_scroll_mode:
```
nriley, line 169: Consider `return continuous_scroll_mode != ""`

plugin/mouse/mouse_scroll.py
```diff
@@ -162,38 +194,54 @@ def noise_trigger_hiss(active: bool):
                 actions.user.mouse_scroll_stop()
 
 
-def mouse_scroll_continuous(new_scroll_dir: Literal[-1, 1]):
-    global scroll_job, scroll_dir, scroll_start_ts, continuous_scroll_mode
+def mouse_scroll_continuous(
+    new_scroll_dir: Literal[-1, 1],
+    speed_factor: Optional[int] = None,
+):
+    global scroll_job, scroll_dir, scroll_start_ts, ctx
```
nriley, line 201: Please remove `ctx` as above

nriley (changes requested):

knausj85 (approved):
This looks good to me, thanks.

nriley (approved):

PR #1650, opened 2024-12-22
Refactor code operators

### LINKED ISSUE #1597: Discussion: Improve code operators
## The problem
Implementing operators in a programming language today is very verbose. This is due to the fact that each operator is its own action. This creates a lot of boiler plate and there is no easy overview of the operators in a particular language.

eg 97 lines in java : https://github.com/talonhub/community/blob/c5f32adef6db9a722e949423f2203cb7d474594f/lang/java/java.py#L93-L190

## My implementation - Talon list with typed class
My solution to this is to use a Talon list. The benefit with this is that operators that is not supported by the language is not speakable and its implementation is quite compact. I even added a class so I could get typed operators in my list.

eg 23 lines in java: https://github.com/AndreasArvidsson/andreas-talon/blob/ab634074d15197eca86f7bff5c0fb549799faa46/languages/java/java.py#L12-L35

### Pros
* Terse / compact
* Only supported operators in the grammar
* The keys are typed so you get code completion and can't make a typo
* Can be implemented one language at the time

### Cons
*  Spoken form is defined in python instead of a Talon file
    In my implementation I just use the class fields as the spoken form. For community we could either default to that and have a override map or we could define all the spoken forms in a map from the start.
* A Talon list file doesn't lend itself well to optional grammar. eg this operator can be said in six different ways:
    https://github.com/talonhub/community/blob/8eb2f32eb8b3312b50e3b402efab958ade401161/lang/tags/operators_bitwise.talon#L11

## The middle ground solution
One way to do it would be to keep the community Talon files but instead of each operator calling a separate action we just have a single `code_operator(id: str)` that is used by all the operators. We would then per language have something very similar to the operator's class in my example that just defines each operator's value but don't assign them to a Talon list. Instead the class is used by an action.
eg: `(op | logical | bitwise) (ex | exclusive) or: user.code_operator("bitwise_exclusive_or")`

### Pros
* No spoken forms are defined in python
* We keep the flexibility of the Talon files
* Almost as compact per language as my implementation

### Cons
* We still need to use the operator tags and the multiple Talon files
* Operators that are not supported by the language might be available in the grammar
* Needs to be implemented in all languages at the same time

## Summary

I'm very happy with my implementation for my own needs, but for community I think the middle ground solution is the best. That would basically involve: 
1. Create a dict/class for each programming language with its operators 
2. Implement the new operator's action
3. Replace all the old calls to the operator's actions in the Talon files with the new action

Any opinions? @phillco @knausj85 @nriley

AndreasArvidsson:
Fixes #1597

I think this is where we landed in our latest discussion. Please have a look and give feedback. I would also want a community user to test drive this thoroughly.


FireChickenProductivity:
I get exceptions when I use this (using the latest beta on Mac).  I simply cannot use any of the operator commands. I tried merging the latest community changes into my local copy of the branch to see if that helped, but I still got the following exceptions:

2025-01-02 14:53:45.624 ERROR cb error topic="phrase" cb=SpeechSystem.engine_event
   16:      lib/python3.11/threading.py:1002* # cron thread
   15:      lib/python3.11/threading.py:1045* 
   14:      lib/python3.11/threading.py:982 * 
   13:                    talon/cron.py:159 | 
   10: talon/scripting/speech_system.py:396 | 
    9:             talon/engines/w2l.py:743 | 
    7: talon/scripting/speech_system.py:69  | 
    4: talon/scripting/speech_system.py:518 | 
    3:      talon/scripting/executor.py:112 | 
    2:  talon/scripting/talon_script.py:823 | 
    1:  talon/scripting/talon_script.py:725 | 
 in script at /Users/name/.talon/user/knausj_talon/lang/tags/operators.talon:5:
 > user.code_operator(code_operators_math)
NotImplementedError: Action 'user.code_get_operators' exists but the Module method is empty and no Context reimplements it


When coding in java, I instead get the following

2025-01-02 14:58:37.865 ERROR     7:          talon/scripting/talon_script.py:721| 
    6:          talon/scripting/talon_script.py:314| 
    5:               talon/scripting/actions.py:88 | 
    4: user/knausj_talon/lang/tags/operators.py:88 | operators: Operators = actions.user.co..
    3:                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    2:               talon/scripting/actions.py:69 | 
    1:                 talon/scripting/types.py:155| 
TypeError: Function 'user.code_get_operators' incompatible with prototype 'user.code_get_operators'
  Signature: () -> user.knausj_talon.lang.tags.operators.Operators
  Expected:  () -> Optional[user.knausj_talon.lang.tags.operators.Operators]
  - Return type mismatch. Found 'user.knausj_talon.lang.tags.operators.Operators', expected 'Optional[user.knausj_talon.lang.tags.operators.Operators]' (or no annotation)

[The below error was raised while handling the above exception(s)]
2025-01-02 14:58:37.868 ERROR cb error topic="phrase" cb=SpeechSystem.engine_event
   16:      lib/python3.11/threading.py:1002* # cron thread
   15:      lib/python3.11/threading.py:1045* 
   14:      lib/python3.11/threading.py:982 * 
   13:                    talon/cron.py:159 | 
   10: talon/scripting/speech_system.py:396 | 
    9:             talon/engines/w2l.py:743 | 
    7: talon/scripting/speech_system.py:69  | 
    4: talon/scripting/speech_system.py:518 | 
    3:      talon/scripting/executor.py:112 | 
    2:  talon/scripting/talon_script.py:823 | 
    1:  talon/scripting/talon_script.py:725 | 
 in script at /Users/name/.talon/user/knausj_talon/lang/tags/operators.talon:2:
 > user.code_operator(code_operators_assignment)
TypeError: Function 'user.code_get_operators' incompatible with prototype 'user.code_get_operators'
  Signature: () -> user.knausj_talon.lang.tags.operators.Operators
  Expected:  () -> Optional[user.knausj_talon.lang.tags.operators.Operators]
  - Return type mismatch. Found 'user.knausj_talon.lang.tags.operators.Operators', expected 'Optional[user.knausj_talon.lang.tags.operators.Operators]' (or no annotation)

AndreasArvidsson:
Talon is a little bit finicky about context return types. Please try now.

FireChickenProductivity:
"op plus" worked for java but not python.

2025-01-02 16:11:32.931 ERROR cb error topic="phrase" cb=SpeechSystem.engine_event
   16:      lib/python3.11/threading.py:1002* # cron thread
   15:      lib/python3.11/threading.py:1045* 
   14:      lib/python3.11/threading.py:982 * 
   13:                    talon/cron.py:159 | 
   10: talon/scripting/speech_system.py:396 | 
    9:             talon/engines/w2l.py:743 | 
    7: talon/scripting/speech_system.py:69  | 
    4: talon/scripting/speech_system.py:518 | 
    3:      talon/scripting/executor.py:112 | 
    2:  talon/scripting/talon_script.py:823 | 
    1:  talon/scripting/talon_script.py:725 | 
 in script at /Users/name/.talon/user/knausj_talon/lang/tags/operators.talon:5:
 > user.code_operator(code_operators_math)
NotImplementedError: Action 'user.code_get_operators' exists but the Module method is empty and no Context reimplements it


 I got it to work locally by replacing the default definition with the following and then checking for 7 instead of none. It seems that having an action that only returns none causes problems.

def code_get_operators() -> Optional[Operators]:
        """Get code operators dictionary"""
        return 7

 I could also get away with the following:

def code_get_operators() -> Optional[Operators]:
        """Get code operators dictionary"""
        if 'chicken':
            return None

AndreasArvidsson:
Yeah I forgot that Talon interprets `return None` as an action not being implemented. I think I have a fix for this now using a try catch instead.

lang/tags/operators.py
```diff
@@ -0,0 +1,207 @@
...
+    POINTER_INDIRECTION: NotRequired[Operator]
+    POINTER_ADDRESS_OF: NotRequired[Operator]
+    POINTER_STRUCTURE_DEREFERENCE: NotRequired[Operator]
+
+
+@mod.action_class
+class Actions:
+    def code_operator(id: str):
+        """Insert a code operator"""
+        operators: Operators = actions.user.code_get_operators()
```
knausj85, line 88: The fallback mechanism might work better without this line ;)

AndreasArvidsson: Very true :D
Fixed now

lang/tags/operators.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def code_get_operators() -> Operators:
+        """Get code operators dictionary"""
+
+
+# Fallback is to rely on the legacy actions
+def operators_fallback(id: str) -> None:
+    match id:
+        # code_operators_array
+        case "subscript":
```
knausj85, line 112:
```suggestion
        case "SUBSCRIPT":
```

knausj85:
This is working well on my end. I'll put it through a few more paces this week, but I think we're probably good to go. Very clean!

I think we just need to decide what to do about all the variants that have been removed. Maybe we create a _deprecated.talon file for each. Or we just live with it

AndreasArvidsson:
Thank you for testing!

I guess we can have a discussion about that on Saturday? That gives you almost a week to thoroughly test it as well.

nriley:
On the community backlog session we were worried about breaking existing code and decided to add deprecated implementations of the existing actions that will display a message and be implemented in terms of their replacements.

AndreasArvidsson:
Okay the old commands are re added and deprecated. In some cases there are weird overlaps with a lot of optionals, but I did my best.

nriley:
Thanks. I hadn't registered that this would result in a major change to the available operator voice commands as well as the actions, and had assumed you'd implement deprecation at the level of the actions rather than the voice commands. Use of the removed actions in your own code will result in an error, which I guess is OK! I would recommend adding something to `BREAKING_CHANGES.txt` though (wonder if this would be better as a `BREAKING_CHANGES.md`...)

How do others feel about the canonicalization of voice commands such that each operator voice command starts with `op` and there is only one way of saying each operator? These changes also make it clearer where to customize the commands, but could also increase misrecognitions, e.g. between `op and` and `op in`.

Note that I'm not saying to keep all the old ways! Some obviously need replacement, e.g. `(op | logical | bitwise) (ex | exclusive) or equals:` makes no sense.

I'll try running with this for a week, but if others who voice code more than I do could provide feedback, it'd be helpful before we ask all the users of community to start retraining themselves!

AndreasArvidsson:
Please do and we can have a discussion about it next week :)

nriley:
Looking at this more carefully I see you only migrated the Java language mode and *kept* the actions because they're still used by the duplicate language modes. So we should have some way (ticket?) of keeping track of which language modes still use the individual actions, then we can actually do what I originally thought this PR did in deprecating/removing the actions.

lang/tags/operators.py
```diff
@@ -0,0 +1,204 @@
...
+
+    # code_operators_pointer
+    POINTER_INDIRECTION: NotRequired[Operator]
+    POINTER_ADDRESS_OF: NotRequired[Operator]
+    POINTER_STRUCTURE_DEREFERENCE: NotRequired[Operator]
+
+
+@mod.action_class
+class Actions:
+    def code_operator(id: str):
```
nriley, line 86: `id` is a Python builtin function so would recommend against using it as a parameter. (Surprised nothing warned about this!)

nriley (changes requested):

lang/tags/operators.py
AndreasArvidsson: Oh that ship has thoroughly sailed already :D
https://search.talonvoice.com/search/?q=file%3A%5C.py%24%20%5C(id%3A%20&fold_case=auto&regex=true&context=true

AndreasArvidsson:
> Looking at this more carefully I see you only migrated the Java language mode and _kept_ the actions because they're still used by the duplicate language modes. So we should have some way (ticket?) of keeping track of which language modes still use the individual actions, then we can actually do what I originally thought this PR did in deprecating/removing the actions.

An issue with checkboxes work maybe?

lang/tags/operators.py
```diff
@@ -0,0 +1,204 @@
...
+mod.list(
+    "code_operators_math", desc="List of code operators for mathematical operations"
+)
+mod.list("code_operators_pointer", desc="List of code operators for pointers")
+
+
+Operator = str | Callable[[], None]
+
+
+class Operators(TypedDict):
```
auscompgeek, line 29: Instead of annotating literally every key as `NotRequired`, can we make it the default here instead?
```suggestion
class Operators(TypedDict, total=False):
```

AndreasArvidsson: Thank you!

nriley (approved):

Issue #1637, opened 2024-12-14
Remove uses of `from talon_plugins import eye_zoom_mouse`


nriley:
From aegis in Slack:

looks like community is directly importing and using `eye_zoom_mouse`, this is about to break, remember talon_plugins is not in the public api set

nriley:
On the community backlog session we discussed that we would need an alternative to `eye_zoom_mouse.zoom_mouse.state` to be able to replace our existing usage.

PR #1657, opened 2024-12-26
Remove dependency on `talon_plugins`

AndreasArvidsson:
`talon_plugins` has been removed in the latest beta. As far as I know there is no replacement to check if we are zoomed in or not.

Fixes https://github.com/talonhub/community/issues/1637

### DIFF
--- a/core/deprecations.py
+++ b/core/deprecations.py
@@ -164,7 +164,8 @@ class Actions:
     def deprecate_action(time_deprecated: str, name: str, replacement: str = ""):
         """
         Notify the user that the given action is deprecated and should
-        not be used into the future.
+        not be used into the future; the action `replacement` should be used
+        instead.
         """
 
         id = f"action.{name}.{time_deprecated}"
--- a/core/deprecations.py
+++ b/core/deprecations.py
@@ -161,7 +161,7 @@ class Actions:
         )
         warnings.warn(msg, DeprecationWarning, stacklevel=3)
 
-    def deprecate_action(time_deprecated: str, name: str):
+    def deprecate_action(time_deprecated: str, name: str, replacement: str = ""):
         """
         Notify the user that the given action is deprecated and should
         not be used into the future.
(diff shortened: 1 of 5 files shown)


plugin/mouse/mouse.py
```diff
@@ -38,8 +37,7 @@
 class Actions:
     def zoom_close():
         """Closes an in-progress zoom. Talon will move the cursor position but not click."""
-        if eye_zoom_mouse.zoom_mouse.state == eye_zoom_mouse.STATE_OVERLAY:
-            actions.tracking.zoom_cancel()
+        actions.user.deprecate_action("2024-12-26", "user.zoom_close")
```
auscompgeek, line 40: Can we point users to call `tracking.zoom_cancel` directly instead?

AndreasArvidsson: That is what I'm doing in the Talon file. `deprecate_action` have no argument for replacement. I was kinda leaning towards maybe adding that.

phillco (approved):

plugin/mouse/mouse.py
AndreasArvidsson: Optional replacement argument added

core/deprecations.py
```diff
@@ -170,9 +170,11 @@ def deprecate_action(time_deprecated: str, name: str):
         id = f"action.{name}.{time_deprecated}"
 
         deprecate_notify(id, f"The `{name}` action is deprecated. See log for more.")
+        replacement_msg = f' Instead, use: "{replacement}".' if replacement else ""
 
         msg = (
             f"The `{name}` action is deprecated since {time_deprecated}."
+            f"{replacement_msg}"
             f' See {os.path.join(REPO_DIR, "BREAKING_CHANGES.txt")}'
```
auscompgeek, line 178: I guess we want an entry in BREAKING_CHANGES.txt?

auscompgeek (approved):
Just need docs, otherwise LGTM

core/deprecations.py
AndreasArvidsson: True. Done

phillco (approved):

PR #1671, opened 2025-01-08
Improve exception message when communication directory is not found

AndreasArvidsson:


apps/vscode/command_client/rpc_client/rpc_client.py
```diff
@@ -44,7 +44,7 @@ def rpc_client_run_command(
         if not communication_dir_path.exists():
             if args or return_command_output:
                 raise Exception(
-                    "Must use command-server extension for advanced commands"
+                    "Communication directory not found. Must use command-server extension for advanced commands"
```
phillco, line 47: I'd ask if we can link to it, but realized this is the generic code path. It would be nice if each implementation (VSCode, JetBrains, etc) could provide a "help" action that explains how to fix the issue (could just linked to the extension page or even open it)

phillco (approved):

PR #1677, opened 2025-01-11
vscode: copy, paste, redo and undo through command palate

BlueDrink9:
Makes these actions more reliable than the default of sending keypresses, especially if the user has rebound those keys or is in a different editor mode (eg with vim extension)


BlueDrink9:
Also added copy and paste while I'm there, for the same reason.

nriley:
On the community backlog session we discussed that this will make the experience worse for people who do not have the command server installed but at this point this should be a rare occurrence and the command server should essentially be considered a dependency if you are using visual studio code and Talon.

nriley (approved):

BlueDrink9:
Good to know. Would rt be possible to make the whole class definition dependent on whether it can reach the server? Then the default edit shortcuts would be used as a fallback

I also found a few other commands that would be useful to send via server, eg find. I'll open a PR for those, since the decision is to go with the command versions generally

lunixbochs:
won't this break undo commands when the edit view isn't focused?

nriley:
@lunixbochs good point, we should probably revert this.

@BlueDrink9 we did discuss that, @AndreasArvidsson should be able to clarify.

BlueDrink9:
Confused why undo would break outside the editor window - is it because ctrl+z does different commands outside the edit window?

BlueDrink9:
I presume the command server can tell us if the editor window is focused - maybe this just needs a bit more smarts

AndreasArvidsson:
I just tried `editor.action.clipboardPasteAction` and that worked fine in the search widget, settings panel and the terminal. There might be a scenario were these actions fail, but as far as I can tell vscode has done a good job of implementing them across the application.

nriley:
If those cases work, the only example I can think of where it would break would be system open and save dialogs; these are modal. Sometimes you will paste a file name into those, for example.

AndreasArvidsson:
You're right. In the save dialog `editor.action.clipboardPasteAction` doesn't work and we get a time out error from the command client. We probably need to revert this.

BlueDrink9:
Looking at coming back to this. I know from @pokey's advocacy work that https://github.com/microsoft/vscode/issues/97790 is now sufficient to find out whether the editor is focused, but I can't figure out how to go about parsing it such that we can get that information. Would anyone mind guiding me to the right place to look for that?

AndreasArvidsson:
https://github.com/talonhub/community/blob/a49ff1a76fcd0fe37610338d36e1fd5286dbedc4/apps/vscode/README.md#terminal

https://github.com/talonhub/community/blob/a49ff1a76fcd0fe37610338d36e1fd5286dbedc4/apps/vscode/vscode_terminal.talon

BlueDrink9:
New PR here for review btw: [vscode: copy, paste, redo and undo through command palate (2nd attempt) by BlueDrink9 · Pull Request #1777 · talonhub/community](https://github.com/talonhub/community/pull/1777)

PR #1678, opened 2025-01-12
Move command client into core folder

AndreasArvidsson:
The reason the command client resides in the vscode folder is because it was designed for vscode originally, but now it's actually used for more applications so I'm moving it to the core folder instead.

I will take care of the subtree. I have successfully run the split command on my computer.


knausj85 (approved):

knausj85:
I'll leave this to you to merge when you're ready.

PR #1687, opened 2025-01-17
More jetbrains macos matchers

asoee:


asoee:
Mistake.. too much in one PR

PR #1690, opened 2025-01-17
Resurrect go language support

asoee:
Trying to keep it simple for a start


lang/go/go.talon
```diff
@@ -0,0 +1,37 @@
...
+settings():
+    user.code_private_function_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_protected_function_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_public_function_formatter = "PUBLIC_CAMEL_CASE"
+    user.code_private_variable_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_protected_variable_formatter = "PRIVATE_CAMEL_CASE"
+    user.code_public_variable_formatter = "PRIVATE_CAMEL_CASE"
+
+(variadic | spread): "..."
+assign: " := "
```
nriley, line 28: This is confusing as " = " is inserted via "op assign". According to the go language specification this is a "short variable declaration"; perhaps "declare" would be better?

lang/go/go.py
```diff
@@ -0,0 +1,243 @@
...
+
+ctx.lists["user.code_common_function"] = {
+    # golang buildin functions
+    "append": "append",
+    "length": "len",
+    "make": "make",
+    # formatting
+    "format print": "fmt.Printf",
+    "format sprint": "fmt.Sprintf",
+    "format print L": "fmt.Println",
```
nriley, line 54: Are you missing something here? Seems like it would make more sense to have this as "print line". Rust uses "print line".

nriley (changes requested):

lang/go/go.talon
asoee: wasn't aware that assign was already used.
declare works for me too

lang/go/go.py
asoee: print line it is

nriley (approved):

PR #1697, opened 2025-01-22
Make trigger shortcut more obscure

BlueDrink9:
The current shortcut [clashes with some extensions](https://github.com/cursorless-dev/cursorless/issues/880#issuecomment-2606151500)

This seems to work on my linux configuration with codium, whereas the f17 mapping didn't


BlueDrink9:
Matches with https://github.com/cursorless-dev/command-server/pull/41

AndreasArvidsson:
This only changes the key for Linux, but the change for the server side changes it for all three platforms.

PR #1700, opened 2025-01-22
feat: add user setting for hiss debounce time

BlueDrink9:
The default is far too short for me and kept triggering if I uttered "say something" or "sentence Something", or any other sibilant sentence.


core/noise.py
```diff
@@ -31,7 +38,10 @@ def noise_trigger_hiss_debounce(active: bool):
     """Since the hiss noise triggers while you're talking we need to debounce it"""
     global hiss_cron
     if active:
-        hiss_cron = cron.after("100ms", lambda: actions.user.noise_trigger_hiss(active))
+        hiss_cron = cron.after(
+            str(settings.get("user.hiss_scroll_debounce_time")) + "ms",
```
knausj85, line 42: I'd suggest using an F-string maybe, but otherwise looks good to me
```suggestion
            f"{settings.get("user.hiss_scroll_debounce_time")}ms",
```

phillco: hm I thought we had flynt running automatically in CI to do that

knausj85 (approved):

PR #1702, opened 2025-01-23
Added support for VSCode palette

halidaee:
Added talon commands for opening VSCode's general command search.

### DIFF
--- a/apps/vscode/vscode.talon
+++ b/apps/vscode/vscode.talon
@@ -16,6 +16,13 @@ go view [<user.text>]:
     user.vscode("workbench.action.openView")
     insert(user.text or "")
 
+# Palette Command Search
+palette: user.vscode("workbench.action.showCommands")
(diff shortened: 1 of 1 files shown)


auscompgeek (changes requested):
There's already a command for this:

https://github.com/talonhub/community/blob/c9a4c0987eddfca98eeba73834452d25dbbf0c23/apps/vscode/vscode.talon#L9

https://github.com/talonhub/community/blob/c9a4c0987eddfca98eeba73834452d25dbbf0c23/tags/command_search/command_search.talon#L3

https://github.com/talonhub/community/blob/c9a4c0987eddfca98eeba73834452d25dbbf0c23/apps/vscode/vscode.py#L284-L287

halidaee:
ah my bad! I expected it to show up under VSCode in the help active panel, but since it's under a tag it is under command search context instead.

PR #1705, opened 2025-01-25
Support description in snippets

AndreasArvidsson:
Support a description field in snippet files

### DIFF
--- a/core/snippets/snippets_parser.py
+++ b/core/snippets/snippets_parser.py
@@ -11,6 +11,7 @@ class SnippetDocument:
     line_body: int
     variables: list[SnippetVariable] = []
     name: str | None = None
+    description: str | None = None
     phrases: list[str] | None = None
     insertionScopes: list[str] | None = None
     languages: list[str] | None = None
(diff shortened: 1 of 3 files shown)


core/snippets/snippets_parser.py
```diff
@@ -48,10 +49,12 @@ def create_snippets(documents: list[SnippetDocument]) -> list[Snippet]:
 
 
 def create_snippet(
-    document: SnippetDocument, default_context: SnippetDocument
+    document: SnippetDocument,
+    default_context: SnippetDocument,
 ) -> Snippet | None:
     snippet = Snippet(
-        name=document.name or default_context.name,
+        name=document.name or default_context.name or "",
```
nriley, line 56: Am I missing something here? According to the documentation the name is required but you have an option for no name?

AndreasArvidsson: It's just that the type doesn't support None. We are validating that there is a name a few lines below
https://github.com/talonhub/community/blob/c9a55b56f00851b826df4a05ca88cefd4a2898b9/core/snippets/snippets_parser.py#L74-L76

nriley (approved):

PR #1706, opened 2025-01-26
Added contacts feature, documented in core/contacts/README.md.

wolfmanstout:
I have been holding onto this one in my fork for a while, yet it's one of the most useful extensions to community I've written. Long ago I used to put contacts into vocabulary, but they end up conflicting with other words. By adding explicit suffixes and a predictable grammar, it is so much easier to work with them. As a bonus, you never have to remember email addresses again.


core/contacts/README.md
wolfmanstout, line 1: If preferred, I could add this section directly to the top-level readme, or link to it from there.

core/text/text_and_dictation.py
```diff
@@ -120,14 +120,28 @@ def word(m) -> str:
         )
 
 
-@mod.capture(rule="({user.vocabulary} | <phrase>)+")
+@mod.capture(rule="({user.vocabulary} | <user.prose_contact> | <phrase>)+")
```
wolfmanstout, line 123: We could also add this to <user.word> above -- just bear in mind that it'll take multiple words to speak it, and some result in multiple words.

core/text/text_and_dictation.py
```diff
@@ -136,14 +150,27 @@ def prose(m) -> str:
...
+        ")+"
+    )
 )
 def raw_prose(m) -> str:
     """Mixed words and punctuation, auto-spaced & capitalized, without quote straightening and commands (for use in dictation mode)."""
     return apply_formatting(m)
 
 
-# For dragon, omit support for abbreviations
+# For dragon, omit support for abbreviations and contacts
```
wolfmanstout, line 173: Maybe it works with Dragon -- I just don't know because I don't use that anymore.

core/contacts/contacts.py
```diff
@@ -0,0 +1,293 @@
...
+@track_file("contacts.json", default="[]")
+def on_contacts_json(f):
+    global json_contacts
+    try:
+        contacts = json.load(f)
+    except Exception:
+        logging.exception("Error parsing contacts.json")
+        return
+
+    json_contacts = []
```
knausj85, line 118: (I'm probably missing something from the parsing logic above that makes from_json necessary).  

but since the contacts are data class, could just do
```
    json_contacts = [Contact(**contact) for contact in contacts]
```

knausj85: Oh wait, nevermind there's a dictionary in there :)

knausj85 (approved):
This looks great and is working very well for me, thanks! I'll leave it open for now for other maintainers to chime in.

core/edit_text_file/edit_text_file_list.talon-list
```diff
@@ -13,3 +13,5 @@ unix utilities: tags/terminal/unix_utility.talon-list
 abbreviations: settings/abbreviations.csv
 file extensions: settings/file_extensions.csv
 words to replace: settings/words_to_replace.csv
+contacts: settings/contacts.json
```
AndreasArvidsson, line 16: This should probably be contacts json to make it really clear which one you are using

AndreasArvidsson:
Feedback from the community backlog session. We think this one looks good, but due to the somewhat sensitive nature of storing email addresses and user names. We think that we probably want a different folder than settings to store these in. We can then git ignore that new folder separately with a comment in git ignore about the sensitive nature. The reason for this is that a lot of people remove the settings directory from git ignore.
We recommended that we call this new folder `private`.

core/edit_text_file/edit_text_file_list.talon-list
wolfmanstout: Good call; done.

wolfmanstout:
> Feedback from the community backlog session. We think this one looks good, but due to the somewhat sensitive nature of storing email addresses and user names. We think that we probably want a different folder than settings to store these in. We can then git ignore that new folder separately with a comment in git ignore about the sensitive nature. The reason for this is that a lot of people remove the settings directory from git ignore. We recommended that we call this new folder `private`.

Sounds good, done.

.gitignore
```diff
@@ -4,6 +4,7 @@ __pycache__
 .idea/
 # Locally generated
 /settings
+/private
```
knausj85, line 7:
```suggestion
# we highly recommended against removing the private folder from the gitignore file
# the directory is used to store sensitive information such as contacts
/private
```

core/contacts/README.md
```diff
@@ -0,0 +1,62 @@
+# Contacts
+
+This directory provides a versatile `<user.prose_contact>` capture that can be
+used to insert names and email addresses using a suffix. The contact list may be
+provided through `contacts.json` in this directory, `contacts.csv` in the
+settings directory, or both.
```
knausj85, line 6:
```suggestion
provided in the private directory via the `contacts.json` file, the `contacts.csv` file,  or both.
```
I believe as written these both need to live in the private directory?

wolfmanstout:
Thanks Jeff and Andreas! Appreciate the final adjustments!

brief:
Thanks so much for adding this functionality, @wolfmanstout!

Two questions:
- According to the README, saying `Jonathan Doh email` should get the email address, but that doesn't work for me. I can only get it with `phrase Jonathan Doh email`. Is that right?
- Should the [customize commands](https://github.com/wolfmanstout/wolfmanstout_talon/blob/8452160413d0e3634cb8d6768e1b1e26b41bdddf/core/edit_text_file/edit_text_file_list.talon-list#L16) point to `private/contacts`?

wolfmanstout:
> Thanks so much for adding this functionality, @wolfmanstout!
> 
> Two questions:
> 
>     * According to the README, saying `Jonathan Doh email` should get the email address, but that doesn't work for me. I can only get it with `phrase Jonathan Doh email`. Is that right?
> 
>     * Should the [customize commands](https://github.com/wolfmanstout/wolfmanstout_talon/blob/8452160413d0e3634cb8d6768e1b1e26b41bdddf/core/edit_text_file/edit_text_file_list.talon-list#L16) point to `private/contacts`?

Thanks! I just sent #1746 to address these. The first of these is expected behavior. This is embedded in a capture within `<user.text>` and `<user.prose>`, not directly as commands. In that sense it is equivalent to vocabulary. We could definitely consider adding these directly as commands, since the risk of conflicts is much lower than vocabulary. It would certainly be trivial to implement; just add this to `text.talon`:
```
<user.prose_contact>: "{prose_contact}"
```

What is a little unusual about adding these as commands is that they use a suffix syntax instead of prefix. This is intentional for a couple reasons:
- It is easier to remember mid-name that you should say "name" afterwards than to plan ahead.
- It leads to a consistent and fairly intuitive possessive form, e.g.: "Jonathan full names" -> "Jonathan Doh's". This would be difficult to do if the syntax were reversed, because there's no way (AFAIK) to add "name <name>s" to the grammar. I guess the names could be preprocessed to add an apostrophe S, but that's pretty hacky and may be hard to say/recognize for some names (e.g. "James's").

PR #1715, opened 2025-01-31
Remove snippet language inheritance

### LINKED ISSUE #1712: Remove snippet language inheritance
Today in snippets we inherit snippet from one language to another. For example typescript inherits all javascript snippets.

This behavior is defined here: https://github.com/talonhub/community/blob/b57dac88d69a06c22aea76dc6f47bab895a32f56/core/snippets/snippets.py#L126-L147

This was made in an effort to make the snippet files as small as possible. In hindsight this was a bad decision. First of all is not at all obvious to users which languages are inherited and there is no simple way for them to extend this list. I recommend that we remove this and update the snippet files to just explicitly include all languages that are needed.

@phillco @pokey @nriley @knausj85

AndreasArvidsson:
While I was at it I added missing languages and a few snippets

Fixes #1712

### DIFF
--- a/core/snippets/snippets.py
+++ b/core/snippets/snippets.py
@@ -89,7 +89,8 @@ def update_snippets():
         insertions_phrase_map = {}
         wrapper_map = {}
 
-        for lang_super in get_super_languages(lang):
+        # Assign global snippets to all languages
+        for lang_super in ["_", lang]:
             snippets, insertions, insertions_phrase, wrappers = create_lists(
                 lang,
                 lang_super,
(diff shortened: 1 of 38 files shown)


core/snippets/snippets.py
```diff
@@ -123,30 +124,6 @@ def get_snippets() -> list[Snippet]:
     return result
 
 
-def get_super_languages(language: str) -> list[str]:
```
AndreasArvidsson, line 126: This is the principal logic that was removed.

core/snippets/snippets.py
```diff
@@ -200,7 +177,7 @@ def create_lists(
 
 
 def on_ready():
-    fs.watch(str(SNIPPETS_DIR), lambda _1, _2: update_snippets())
+    fs.watch(SNIPPETS_DIR, lambda _1, _2: update_snippets())
```
AndreasArvidsson, line 180: Type change in Talon

core/snippets/snippets/html.snippet
```diff
@@ -1,23 +1,10 @@
-language: html
+language: html | javascriptreact | typescriptreact
 ---
 
-name: element
```
AndreasArvidsson, line 4: This are moved to xml.snippet

core/snippets/snippets/javascript.snippet
```diff
@@ -39,18 +39,9 @@ const $1 = ($2) => {
 }
 ---
 
-name: reactUseState
```
AndreasArvidsson, line 42: These are moved to javascriptreact.snippet

core/snippets/snippets/talon.snippet
```diff
@@ -7,5 +7,5 @@ insertionScope: command
 
 $0.insertionFormatter: NOOP
 -
-$0: user.vscode("$CLIPBOARD")
+$0: user.run_rpc_command("$CLIPBOARD")
```
AndreasArvidsson, line 10: No need for this command to be vscode only.

core/snippets/snippets.py
```diff
@@ -200,7 +177,7 @@ def create_lists(
...
 
     if get_setting_dir():
         fs.watch(str(get_setting_dir()), lambda _1, _2: update_snippets())
```
nriley, line 183: Might as well remove `str()` here

nriley (changes requested):
OK if you add a message to breaking changes

PR #1716, opened 2025-01-31
Update gitignore match rules

chrisnicollo:
As it is currently written, the `.gitignore` file does not properly ignore `.bak`s. It also seems to not properly ignore `.DS_Store`s. Here I add the wildcard character to these patterns to make sure they are correctly matched against.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,5 @@ __pycache__
 # Locally generated
 /settings
 .vscode/settings.json
-.DS_Store
-.bak
+*.DS_Store
+*.bak
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,5 @@ __pycache__
 # Locally generated
 /settings
 .vscode/settings.json
-*.DS_Store
+.DS_Store
 *.bak


nriley:
From the community backlog session — we agree it is past time to add these type of ignores.

Adding ignores for temporary files created by common editors from https://github.com/github/gitignore/blob/main/Global/Emacs.gitignore and https://github.com/github/gitignore/blob/main/Global/Vim.gitignore would make sense.

.gitignore
```diff
@@ -5,5 +5,5 @@ __pycache__
 # Locally generated
 /settings
 .vscode/settings.json
-.DS_Store
-.bak
+*.DS_Store
```
nriley, line 8: This ignore works for me. Can you give an example of where (platform, filename) where this helps?

nriley (changes requested):

.gitignore
chrisnicollo: I'm a Windows user and recognized that as file type used on Mac systems. For some reason, I forgot that that was a file name and not a file extension, so I thought an asterisk was needed beforehand to make it match. I'll go ahead and change that back. As for adding ignores for temporary files created by vim and emacs, I'm not super familiar with the temporary files those programs create. The only files I'm recognizing within those example `.gitignore`s are swap files, so I don't have the knowledge to know which rules would be useful to add.

nriley: OK, go ahead and fix that and I'll put the vim/emacs stuff in as a separate PR. Thanks!

nriley (approved):

PR #1717, opened 2025-02-01
Add to homophones

chrisnicollo:
This adds new homophones to the list in `homophones.csv`. It also sorts the list such that each line has its entries ordered alphabetically, and the lines themselves are ordered alphabetically. The ordering is done case-insensitively.

### DIFF
--- a/core/homophones/homophones.csv
+++ b/core/homophones/homophones.csv
@@ -267,7 +267,6 @@ gamble,gambol
 gel,jell
 gene,Jean
 jibe,gibe
-gui,gooey
 guild,gild
 gilder,guilder
 guilt,gilt
--- a/core/homophones/homophones.csv
+++ b/core/homophones/homophones.csv
@@ -40,6 +40,7 @@ apprize,apprise
 arc,ark
 ascent,assent
 assistance,assistants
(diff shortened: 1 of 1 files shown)


nriley:
The current intra-line sorting order is intentional; see #318. Can you update your PR such that resorting within lines is not intermixed with new homophones? Thanks!

core/homophones/homophones.csv
```diff
@@ -252,10 +262,12 @@ friar,fryer
 fur,fir
 gaffe,gaff
 gate,gait
+gator,gaiter
 gamble,gambol
 gel,jell
 gene,Jean
 jibe,gibe
+gui,gooey
```
nriley, line 270: 'gui' is not a word so would remove this; consider adding to the default abbreviation list instead.

nriley (changes requested):
Looks good with one exception; I've got some of these in my own `homophones.csv`.

nriley (approved):

chrisnicollo:
As a followup, is there a reason for the current inter-line sorting? Because if not, I still think that can be improved by fixing the alphabetical sorting.

PR #1718, opened 2025-02-01
Update snippets to use multiple language variants

### LINKED ISSUE #1373: Make snippet watcher more fine-grained

### LINKED ISSUE #1707: Extend snippet functionality
Once https://github.com/cursorless-dev/cursorless/pull/2747 is merged we have a way of migrating Cursorless snippet to the community snippet format. Having one single easy to use format for all Talon users is a nice improvement, but the Cursorless snippet format did have two features that the community format is missing. My intent is to extend the community snippet format to support these.

The two Cursorless snippet features missing from community snippets are:
1. Use snippets in a non active editor with a different programming language. 
    eg you are in a markdown file and say `"snip air after blue bat"`. `blue bat` is in another split and in python. The Cursorless snippet implementation would pick the language for the editor you are inserting the snippet into regardless of what the active/focused editor's language is.
2. Different snippet bodies for the same snippet name/phrase and language depending on the parse tree.
   eg in javascript a function looks like `function foo() { }`, but a method in a class looks like `foo() { }`. With the Cursorless snippet format you can just say `"snip funk"` and Cursorless would figure out which one of the two snippet to use. To do this Cursorless snippets have two extra fields `scopeTypes` and `excludeDescendantScopeTypes` to make its match with.

# Proposed solution

For both these features we need to send multiple snippets from Talon to Cursorless. Today community does its snippet insertion using the action `def insert_snippet(body: str)`. What we could do is to add an action like `def insert_snippet_variants(snippets: list[Snippet])`. The default module implementation would just be to pick the first snippet corresponding to the active language and then call `insert_snippet` . Cursorless on the other hand could context override `insert_snippet_variants` to send all snippet variants to the Cursorless extension using the command client. After that Cursorless can pick the best snippet. Thanks @phillco for this idea.

We still need to figure out how to either implement or replace `scopeTypes` and `excludeDescendantScopeTypes` in the community format.
The reason why we need both of these fields is `scopeTypes` would specify `class` so Cursorless knows to use the method snippet body when you're inside a class and `excludeDescendantScopeTypes` would specify `function` so that Cursorless knows to NOT use a method snippet when inside a function. 

Example snippets

```
name: functionDeclaration
phrase: funk
---

language: javascript
-
function $1 () {
    $0
}
---

language: javascript
scopeTypes: class
excludeDescendantScopeTypes: function
-
$1 () {
    $0
}
```

## Problems

1. Ideally we would find a way to express this functions/method behavior using only one field.
2. How can we handle Cursorless context overriding an action for users with an old community version? What is the best way to check or catch the errors so users don't get a warning in the log about a missing module definition?

AndreasArvidsson:
* Added actions to get multiple snippets for the same name. Can be used by extensions like Cursorless to make better decisions about which snippet to use. 
* Simplified the snippet code and added stronger types
* Only update Talon lists if they have actually changed. 

Fixes #1707
Fixes #1373


core/snippets/snippets.py
```diff
@@ -52,155 +63,161 @@ def get_setting_dir():
...
-        ctx.lists.update(
-            {
-                "user.snippet": insertion_map,
-                "user.snippet_with_phrase": insertions_phrase_map,
-                "user.snippet_wrapper": wrapper_map,
-            }
-        )
+def get_preferred_snippet(snippets: list[Snippet]) -> Snippet:
+    lang: Union[str, set[str]] = actions.code.language()
+    languages = list([lang]) if isinstance(lang, str) else lang
```
nriley, line 108:
```suggestion
    languages = [lang] if isinstance(lang, str) else lang
```

core/snippets/snippets.py
```diff
@@ -52,155 +63,161 @@ def get_setting_dir():
...
+
     for snippet in snippets:
-        if snippet.languages is not None:
-            for lang in snippet.languages:
-                result[lang].append(snippet)
-        else:
-            result["_"].append(snippet)
-    return result
+        # Map snippet names to actual snippets
+        if snippet.name not in name_to_snippets:
```
nriley, line 150: You can use `name_to_snippets.setdefault` here.

core/snippets/snippets.py
```diff
@@ -52,155 +63,161 @@ def get_setting_dir():
...
+            name_to_snippets[snippet.name] = []
+        name_to_snippets[snippet.name].append(snippet)
 
+        # Map languages to phrase / name dicts
+        for language in snippet.languages or [GLOBAL_ID]:
+            if language not in language_to_lists:
+                language_to_lists[language] = SnippetLists()
```
nriley, line 157: Can also use `setdefault` here

core/snippets/snippets.py
```diff
@@ -52,155 +63,161 @@ def get_setting_dir():
...
+    snippets_map = name_to_snippets
+    update_contexts(language_to_lists)
+
+
+def update_contexts(language_to_lists: dict[str, SnippetLists]):
+    global_lists = language_to_lists[GLOBAL_ID] or SnippetLists()
+
+    for lang, lists in language_to_lists.items():
+        if lang not in languages_state_map:
+            print(f"Found snippets for unknown language: {lang}")
```
nriley, line 180: `app.notify` here?

core/snippets/snippets.py
```diff
@@ -52,155 +63,161 @@ def get_setting_dir():
...
-                if var.wrapper_phrases is not None:
-                    for phrase in var.wrapper_phrases:
-                        wrappers[phrase] = f"{id_lang}.{var.name}"
 
-    return snippets_map, insertions, insertions_phrase, wrappers
+def get_snippets_from_files() -> list[Snippet]:
+    files = glob.glob(f"{SNIPPETS_DIR}/**/*.snippet", recursive=True)
+
+    if get_setting_dir():
+        files.extend(glob.glob(f"{get_setting_dir()}/**/*.snippet", recursive=True))
```
nriley, line 209: `pathlib.Path` objects have a `.glob` which is recursive by default.

core/snippets/snippets.py
```diff
@@ -52,155 +63,161 @@ def get_setting_dir():
...
+
+    return result
 
 
 def on_ready():
-    fs.watch(str(SNIPPETS_DIR), lambda _1, _2: update_snippets())
+    fs.watch(SNIPPETS_DIR, lambda _1, _2: update_snippets())
 
     if get_setting_dir():
         fs.watch(str(get_setting_dir()), lambda _1, _2: update_snippets())
```
nriley, line 223: Can remove the `str()` again here

core/snippets/snippets.py
```diff
@@ -52,155 +63,161 @@ def get_setting_dir():
...
+
+    for file in files:
+        result.extend(create_snippets_from_file(file))
+
+    return result
 
 
 def on_ready():
-    fs.watch(str(SNIPPETS_DIR), lambda _1, _2: update_snippets())
+    fs.watch(SNIPPETS_DIR, lambda _1, _2: update_snippets())
```
nriley, line 220: Give the `_`-prefixed arguments to the `lambda` meaningful names.

nriley (changes requested):

core/snippets/snippets.py
```diff
@@ -52,128 +63,155 @@ def get_setting_dir():
...
     for snippet in snippets:
-        if snippet.languages is not None:
-            for lang in snippet.languages:
-                result[lang].append(snippet)
-        else:
-            result["_"].append(snippet)
-    return result
+        # Map snippet names to actual snippets
+        name_to_snippets.setdefault(snippet.name, [])
+        name_to_snippets[snippet.name].append(snippet)
```
nriley, line 151: Sorry, forgot to mention that the return value of `setdefault` is also useful!

```suggestion
        name_to_snippets.setdefault(snippet.name, []).append(snippet)
```

core/snippets/snippets.py
```diff
@@ -52,128 +63,155 @@ def get_setting_dir():
...
 
+        # Map languages to phrase / name dicts
+        for language in snippet.languages or [GLOBAL_ID]:
+            language_to_lists.setdefault(language, SnippetLists())
+            lists = language_to_lists[language]
```
nriley, line 156:
```suggestion
            lists = language_to_lists.setdefault(language, SnippetLists())
```

nriley (changes requested):

AndreasArvidsson:
@nriley Requested changes have been made

knausj85 (approved):
It looks good to me

nriley (approved):

nriley:
This generates errors/notifications on stock community now. Should we should consider defining these languages?

```
2025-02-02 14:08:14.752    IO Found snippets for unknown language: cpp
2025-02-02 14:08:14.756    IO Found snippets for unknown language: shellscript
2025-02-02 14:08:14.756    IO Found snippets for unknown language: xml
2025-02-02 14:08:14.757    IO Found snippets for unknown language: json
2025-02-02 14:08:14.757    IO Found snippets for unknown language: csv
2025-02-02 14:08:14.757    IO Found snippets for unknown language: scm
```

PR #1727, opened 2025-02-02
Ignore Emacs and Vim temporary files (follow-up from #1716).

nriley:


knausj85 (approved):

.gitignore
```diff
@@ -7,3 +7,13 @@ __pycache__
...
+# Emacs (from https://github.com/github/gitignore/blob/main/Global/Emacs.gitignore)
+*~
+\#*\#
+.\#*
+# Vim (from https://github.com/github/gitignore/blob/main/Global/Vim.gitignore)
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
+[._]s[a-rt-v][a-z]
+[._]ss[a-gi-z]
+[._]sw[a-p]
```
auscompgeek, line 19: We could probably get away with just the following?
```suggestion
[._]*.swp
```
I think it'll be relatively rare for someone to open the same file in multiple instances simultaneously

PR #1729, opened 2025-02-02
Fix missing language definitions for snippets

AndreasArvidsson:
#1718 added more snippets. Due to an oversight we are now missing a few language definitions that the snippets are relying on which gives the user errors.

### DIFF
--- a/core/snippets/snippets/commentLine.snippet
+++ b/core/snippets/snippets/commentLine.snippet
@@ -18,8 +18,3 @@ language: xml | html
 -
 <!-- $0 -->
 ---
-
-language: scm
--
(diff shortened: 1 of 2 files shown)


core/snippets/snippets/commentLine.snippet
```diff
@@ -18,8 +18,3 @@ language: xml | html
 -
 <!-- $0 -->
 ---
-
-language: scm
```
AndreasArvidsson, line 22: I don't think this one is necessary. scm is quite unusual and we only have a single sniper for it.

nriley (approved):

PR #1735, opened 2025-02-04
Created list of math comparison operators with `"is"` prefix

AndreasArvidsson:
With our latest updates to the operators we deprecated a bunch of spoken forms. Specifically deprecating `"is equal"` and replaced it with `"op equal"` feels like a regression. It also conflicts with `"op equals"` for assignment.

I have here split the math operators into a separate list for comparisons. These comparison operators use `"is"`
instead of `"op"`


lang/tags/operators_math.talon
```diff
@@ -19,29 +19,17 @@ op (exponent | to the power [of]):
     user.code_operator("MATH_EXPONENT")
 
 # comparison operators
-is equal:
-    user.deprecate_command("2025-01-19", "is equal", "op equal")
-    user.code_operator("MATH_EQUAL")
+is equal: user.code_operator("MATH_EQUAL")
```
knausj85, line 22: This creates a duplicate command versus [pace](https://github.com/talonhub/community/blob/d49b52e170487233ca9df5c28c278200c6956dcc/lang/tags/operators.talon#L7), no?

knausj85: I'd be inclined to remove the duplicates, otherwise this looks good to me.  I'm fine with the variants that aren't straight duplicates eg `is greater [than] or equal`,  `is (less | below) [than]` etc

AndreasArvidsson: Done

knausj85 (approved):

PR #1744, opened 2025-02-09
Add "import from" snippet for Python

afallou:
## Why?
We didn't have any such snippet for Python, and it's something that's used often.

## What?
- Move existing Javascript `import from` snippet to separate file and add similar Python construct.
- Add `import` statement.


core/snippets/snippets/import.snippet
```diff
@@ -0,0 +1,14 @@
+name: importFromStatement
+phrase: import from
+
+---
+
+language: javascript | typescript | javascriptreact | typescriptreact
+-
+import $0 from "$1";
```
afallou, line 8: Note that the original snippet from @AndreasArvidsson has `import $0 from "$0";`, I'm not sure what the intent was (my only guess is that `import foo from 'foo'` is somewhat common with npm libs).
This one is more generally applicable, and consistent with Python (where `from foo import foo` is quite rare).

`$1` is set to the name of the import location in both cases, as that allows using suggestions when you got to `$0`.

core/snippets/snippets/import.snippet
```diff
@@ -0,0 +1,14 @@
+name: importStatement
+phrase: import
+
+---
+
+language: javascript | typescript | javascriptreact | typescriptreact
+-
+import $0 from "$0";
```
afallou, line 8: This was the `import from` statement in @AndreasArvidsson's code; it's now just `snip import`. I kept `snip import from` for cases where the imported symbol name differs from the library name.

core/snippets/snippets/javascript.snippet
```diff
@@ -24,12 +24,6 @@ phrase: import star
 import * as $0 from "$0";
```
knausj85, line 24: While we are here, I'd be inclined to move importStarStatement into the new import.snippet file as well since python supports it

`from library_name import *`

knausj85 (changes requested):

core/snippets/snippets/importFrom.snippet
```diff
@@ -0,0 +1,14 @@
...
+---
+
+language: javascript | typescript | javascriptreact | typescriptreact
+-
+import { $0 } from "$1";
+---
+
+language: python
+-
+from $1 import $0
```
nriley, line 13: Please replace the existing voice command to use your snippet — thanks!

https://github.com/talonhub/community/blob/main/lang/python/python.talon#L54

core/snippets/snippets/importFrom.snippet
```diff
@@ -0,0 +1,14 @@
+name: importFromStatement
+phrase: import from
+
+---
+
+language: javascript | typescript | javascriptreact | typescriptreact
+-
+import { $0 } from "$1";
```
nriley, line 8: From the community backlog session we recommended that you remove the braces around `$0`, and swap $0 and $1 (i.e., the token order you have in python is correct)

nriley (changes requested):

core/snippets/snippets/importFrom.snippet
afallou: Done

core/snippets/snippets/importFrom.snippet
afallou: Done

core/snippets/snippets/javascript.snippet
afallou: Done

core/snippets/snippets/importFrom.snippet
nriley: It looks like you just removed the voice command rather than changing it to use your snippet?

nriley (dismissed):

core/snippets/snippets/importFrom.snippet
afallou: My bad, I misunderstood your comment; I thought that the snippet would effectively replace the voice command.

afallou: Done, for real this time.

core/snippets/snippets/import.snippet
```diff
@@ -0,0 +1,29 @@
...
+---
+
+language: python
+-
+import $0
+---
+
```
knausj85, line 15: I'm not sure why, but this is failing to parse with the addition of the python snippet

```
2025-03-07 22:46:14.509    IO ERROR | import.snippet:15 | Missing body in snippet document '
name: importStarStatement
phrase: import star

'
```

AndreasArvidsson: You can only have one header per file.

knausj85: Ah, so we need to split the import star into its own file.

AndreasArvidsson: That or don't use a header for the file at all. It's perfectly valid to have multiple different snippets in one file, but then I wouldn't use a header because they don't have enough in common. In this particular case I would recommend a separate file.

knausj85 (approved):

PR #1749, opened 2025-02-11
Make the word wheel optional for stop continuous scrolling commands

FireChickenProductivity:
Allows stopping continuous scrolling with "stop" and "stop here". 

Something worth discussing: should we indicate the grammar change in the continuous scrolling graphical display and if so how? Should we just replace "wheel stop" in the display with "stop" or should we indicate that the word wheel can optionally be used?

### DIFF
--- a/plugin/mouse/continuous_scrolling.talon
+++ b/plugin/mouse/continuous_scrolling.talon
@@ -2,7 +2,7 @@ tag: user.continuous_scrolling
 -
 <number_small>: user.mouse_scroll_set_speed(number_small)
 
-wheel stop: user.mouse_scroll_stop()
-wheel stop here:
+[wheel] stop: user.mouse_scroll_stop()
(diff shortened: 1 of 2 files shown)


FireChickenProductivity:
> I would suggest updating the graphical display to "[wheel] stop" to indicate that wheel is optional. This would be consistent with talon files.

The current text for the graphics window is "Wheel Stop [stop scrolling]". The stop scrolling part does not correspond to anything in the command grammar. I think it would be confusing to put the word wheel in square brackets without changing the already bracketed part. What do you think about replacing those brackets with a colon?
[wheel] stop: stop scrolling

nriley (changes requested):
On the community backlog session we discussed that it may make sense to have the button in the imgui match the voice command, `[wheel] stop`.

nriley (approved):

PR #1750, opened 2025-02-12
Add snippets for python comprehensions

afallou:
## Why?
Those are somewhat hard to build by voice without a snippet, and are used very often in Python.

## What?
Add snippet for python list, dictionary and set comprehensions.


core/snippets/snippets/pythonComprehension.snippet
```diff
@@ -0,0 +1,23 @@
...
+insertionScope: statement
+-
+{$0 for $0 in $1}
+---
+
+name: dictComprehension
+phrase: dict comp
+insertionScope: statement
+-
+{$0: $2 for $2 in $1}
```
afallou, line 22: In all three cases the first element to fill out in the snippet is the thing being iterated on, as it enables the IDE to help in filling out the other parts through suggestions.

knausj85:
Since these seem pretty python specific, I would be inclined to move these new snippets into python.snippet. 

Are there any other languages that would benefit from these?

nriley (changes requested):
Comments from the community backlog session:

- We felt that the main use of [x for x in y] where both x's are the same is for filtering yet you have not included an "if"; potentially there is a benefit to having multiple snippets per comprehension
- It's worth also including a comprehension that produces a generator (with parentheses as delimiters)
- We weren't aware of any other supported languages to have something equivalent to comprehensions, however it may make sense to not include the word python in the file name of the snippets file.

afallou:
Those are purely python specific so I move them to `python.snippet`; I added a generator expression, and an if clause in the list comprehension.

NonsenseRemedy:
I'm excited to see these added!

There are few other comprehensions that I often use:

[f(x) for x in xs]
[f(x) if foo(x) else g(x) for x in xs]

Maybe it would make sense to call these `list comp`, `list else comp`, with the existing one called `list if comp`?

AndreasArvidsson (approved):

nriley (approved):

PR #1752, opened 2025-02-13
Add full BSPWM support

BlueDrink9:
I'm not entirely sure that this fits in community rather than being its own separate plugin. In particular BWPWM is different enough ( and has its own clearly  defined grammar anyway) that I haven't included compatibility with the existing community window management commands. However, it's always helpful to just be able to clone community and have it available for whoever does use it. 

Regardless, I would love to get some feedback on this and hear suggestions if there are ways to do things better,  particularly in terms of implementing a grammar with talon syntax.

### DIFF
--- a/apps/bspwm/README.md
+++ b/apps/bspwm/README.md
@@ -49,11 +49,14 @@ direction, for example "node left" or "screen right". This is enabled as a speci
 case in the talon file. You can also swap the verb and subject for simple actions,
 for example "focus workspace previous" or "go window south".
 
+To quickly change state you may omit the word "state", e.g. "node fullscreen".
+
+
 If you wish to customise these, you will have to redeclare the whole lists or rules
 in a python file. For example:
 ```python
 from talon import Context
-ctx = Context()
+ctx = Context(")"
 # Ensure your file takes prescedence over the default one
 ctx.matches = """
 os: linux
@@ -62,6 +65,6 @@ language: en
 
 # Name is the name in bspwm.py, prefixed with "user.bspwm_"
 ctx.lists["user.bspwm_to_desktop"] = {
-  "meander in the direction of space": "--to-desktop",
+  ""meander" in the direction of space": "--to-desktop",
 }
 ```
--- a/apps/bspwm/README.md
+++ b/apps/bspwm/README.md
@@ -4,6 +4,7 @@ This integrates the bspc control grammar for the [bspwm window
 manager](https://github.come/baskerville/bspwm) with talon.
 
 Enable it by adding the `tag(): user.bspwm` to one of your .talon files, e.g.
+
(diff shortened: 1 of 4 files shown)


apps/bspwm/README.md
```diff
@@ -0,0 +1,72 @@
+## BSPWM talon
+
+This integrates the bspc control grammar for the [bspwm window
+manager](https://github.come/baskerville/bspwm) with talon.
```
nriley, line 4:
```suggestion
manager](https://github.com/baskerville/bspwm) with talon.
```

apps/bspwm/README.md
```diff
@@ -0,0 +1,72 @@
...
+for example "focus workspace previous" or "go window south".
+
+To quickly change state you may omit the word "state", e.g. "node fullscreen".
+
+If you wish to customise these, you will have to redeclare the whole lists or rules
+in a python file. For example:
+
+```python
+from talon import Context
+ctx = Context(")"
```
nriley, line 61:
```suggestion
ctx = Context()
```

apps/bspwm/README.md
```diff
@@ -0,0 +1,72 @@
...
+# Ensure your file takes prescedence over the default one
+ctx.matches = """
+os: linux
+language: en
+"""
+
+# Name is the name in bspwm.py, prefixed with "user.bspwm_"
+ctx.lists["user.bspwm_to_desktop"] = {
+  ""meander" in the direction of space": "--to-desktop",
```
nriley, line 70:
```suggestion
  '"meander" in the direction of space': "--to-desktop",
```

nriley (commented):
From the community backlog session — this looks like interesting work but we are worried that this is a relatively obscure window manager on a platform that none of the community maintainers use regularly. I added a couple typo fixes to your documentation.

Our preference would be that you add this to your own repository or community fork (the latter would be indexed by search.talonvoice.com).

It looks like you are doing some pretty dynamic things with lists and creating a DSL — these may get easier on the beta or the upcoming release of Talon if it exposes dynamic lists. At first glance, this dynamism makes it a little hard to review.

BlueDrink9:
Thanks for the review, I'll keep an eye out for dynamic lists. I read about them on the docs but couldn't immediately see the documentation.

Is there somewhere I could read more about the indexing of forks?

nriley:
@BlueDrink9 yup! If you search on search.talonvoice.com for dynamic_list you can see some examples of users who've incorporated it into their repositories. Depending on what you return from a dynamic list function it is either treated like a regular Talon list (dict) or a "selection list" (string) where you can say any substring.

I think any fork of talonhub/community gets indexed by search.talonvoice.com. I'm not sure how often the indexer runs though!

PR #1753, opened 2025-02-14
Add Mojeek to search engines

SuperMeepBoy:
A small change because I personally use Mojeek a lot.


knausj85:
I think this is a better fit as a local modification to your configuration. Essentially, I don't think we should add every obscure search engine to community. 

We also migrated this to .talon-list to make such local customizations much easier :)

PR #1758, opened 2025-02-16
Migrate elixir operators

FireChickenProductivity:
This migrates the existing operators and adds some of the missing ones.

The existing grammar is a bit confusing to me because it uses strict boolean operators (the language has non strict boolean operators as well) but non strict equality and inequality. It might be worth discussing what to do with strict and non strict operator variants.


AndreasArvidsson:
Feedback from the community session. Especially for javascript it makes more sense for strict to be the default and have a separate weak comparison.

FireChickenProductivity:
> Feedback from the community session. Especially for javascript it makes more sense for strict to be the default and have a separate weak comparison.

I added weak operator definitions with implementations for Elixir. Let me know if there is anything you would recommend changing for the code organization or naming.

nriley:
From the community backlog session — we don't feel comfortable merging this as none of us have any experience with Elixir. @jaresty — since you originally contributed this could you take a quick look and ensure it doesn't cause any regressions?

jaresty:
This looks fine to me-thanks for putting it together!

nriley (approved):

PR #1760, opened 2025-02-16
Update Visual Studio Code README.

### LINKED ISSUE #1723: Be consistent about how we recommend terminal tags be referenced
In `terminal.talon` we say that tags should be activated per-terminal-app:

https://github.com/talonhub/community/blob/16adf1c2e8103d8eae74629b4ee5b57e85af49bb/tags/terminal/terminal.talon#L3

and this is done by activating `user.generic_unix_shell` (and sometimes other tags) in individual terminals throughout community, for example:

https://github.com/talonhub/community/blob/16adf1c2e8103d8eae74629b4ee5b57e85af49bb/apps/kde_konsole/kde_konsole.talon#L8

But in the VSCode README we recommend activating tags directly on the `terminal` tag (with an example):

https://github.com/talonhub/community/blob/16adf1c2e8103d8eae74629b4ee5b57e85af49bb/apps/vscode/README.md#L21

These recommendations are at least somewhat inconsistent with one another. I can see it'd be annoying if you used multiple terminals, terminals on multiple platforms, etc.

nriley:
- Update links
- Add reference to Andreas Talon (IMO, everyone should be using this one)
- Link directly to settings (friendlier than editing settings.json)
- Don't recommend activating tags on the terminal tag alone (fixes #1723)
- Add option to open integrated terminals as editors

### DIFF
--- a/apps/vscode/README.md
+++ b/apps/vscode/README.md
@@ -20,7 +20,7 @@ Note that the full window title may not be displayed at the top of VS Code windo
 
 Change the [`window.title`](vscode://settings/window.title) setting to:
 
-```json
+```
 ${activeEditorShort}${separator}${rootName}${separator}${profileName}${separator}focus:[${focusedView}]
 ```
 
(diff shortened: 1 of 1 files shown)


knausj85 (approved):
Looks good, thanks!

Issue #780, opened 2022-02-20
create_spoken_forms does not handle apostrophes well


nriley:
We (https://github.com/phillco/talon_accessibility) are starting to use this for UI elements, where "don’t" is pretty commonly used on macOS.

```
>>> actions.user.create_spoken_forms_from_list(["Don’t Save", "Don't Save"])
{
    'don t save': 'Don’t Save',
    'save': 'Don’t Save',
    'don t': 'Don’t Save',
    'don': 'Don’t Save',
    'don apostrophe t save': "Don't Save"
}
```

PR #1763, opened 2025-02-17
Move symbols into their own python file

knausj85:
#1561

### DIFF
--- a/core/create_spoken_forms.py
+++ b/core/create_spoken_forms.py
@@ -6,16 +6,16 @@ from typing import Any, List, Mapping, Optional
 
 from talon import Module, actions
 
-from .keys.keys import symbol_key_words
 from .numbers.numbers import digits_map, scales, teens, tens
 from .user_settings import track_csv_list
+from .keys.symbols import symbols_for_create_spoken_forms
 
 mod = Module()
 
 DEFAULT_MINIMUM_TERM_LENGTH = 2
 EXPLODE_MAX_LEN = 3
 FANCY_REGULAR_EXPRESSION = r"[A-Z]?[a-z]+|[A-Z]+(?![a-z])|[0-9]+"
-SYMBOLS_REGEX = "|".join(re.escape(symbol) for symbol in set(symbol_key_words.values()))
+SYMBOLS_REGEX = "|".join(re.escape(symbol) for symbol in set(symbols_for_create_spoken_forms.values()))
 FILE_EXTENSIONS_REGEX = r"^\b$"
 file_extensions = {}
 
@@ -61,7 +61,7 @@ def on_abbreviations(values):
 
 REVERSE_PRONUNCIATION_MAP = {
     **{str(value): key for key, value in digits_map.items()},
-    **{value: key for key, value in symbol_key_words.items()},
+    **{value: key for key, value in symbols_for_create_spoken_forms.items()},
 }
 
 # begin: create the lists etc necessary for create_spoken_word_for_number
@@ -398,7 +398,8 @@ def create_spoken_forms_from_regex(source: str, pattern: re.Pattern):
     For numeric pieces detected by the regex, generates both digit-wise and full
     spoken forms for the numbers where appropriate.
     """
-    pieces = list(pattern.finditer(source))
+    source_without_apostrophes = source.replace("'", "")
+    pieces = list(pattern.finditer(source_without_apostrophes))
     spoken_forms = list(map(lambda x: x.group(0), pieces))
 
     # NOTE: Order is sometimes important
--- a/core/keys/keys.py
+++ b/core/keys/keys.py
@@ -1,6 +1,15 @@
 from talon import Context, Module, actions, app
+from .symbols import symbols_command_and_dictation, symbols_command, dragon_symbols_command_and_dictation
 
 mod = Module()
+ctx = Context()
+
+ctx_dragon = Context()
+ctx_dragon = Context()
+ctx_dragon.matches = r"""
+speech.engine: dragon
+"""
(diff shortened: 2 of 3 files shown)


core/keys/keys.py
```diff
@@ -103,101 +117,13 @@ def letters(m) -> str:
     return "".join(m.letter_list)
 
 
-ctx = Context()
+@mod.action_class
+class Actions:
+    def get_punctuation_words():
```
knausj85, line 122: Added for gaze OCR to allow customization

core/create_spoken_forms.py
```diff
@@ -398,7 +400,8 @@ def create_spoken_forms_from_regex(source: str, pattern: re.Pattern):
     For numeric pieces detected by the regex, generates both digit-wise and full
     spoken forms for the numbers where appropriate.
     """
-    pieces = list(pattern.finditer(source))
+    source_without_apostrophes = source.replace("'", "")
```
knausj85, line 403: Addresses https://github.com/talonhub/community/issues/780

core/keys/keys.py
```diff
@@ -1,6 +1,20 @@
...
+    dragon_symbols_command_and_dictation,
+    symbols_command,
+    symbols_command_and_dictation,
+)
+
 mod = Module()
+ctx = Context()
+
+ctx_dragon = Context()
+ctx_dragon = Context()
```
AndreasArvidsson, line 13:
```suggestion
ctx_dragon = Context()
```

core/keys/keys.py
```diff
@@ -103,101 +117,13 @@ def letters(m) -> str:
...
-    # Currencies
-    "dollar": "$",
-    "pound": "£",
-}
-
-# make punctuation words also included in {user.symbol_keys}
-symbol_key_words.update(punctuation_words)
-ctx.lists["self.punctuation"] = punctuation_words
-ctx.lists["self.symbol_key"] = symbol_key_words
+ctx.lists["self.punctuation"] = symbols_command_and_dictation
```
AndreasArvidsson, line 127: Feedback from the backlog session. Is probably best if the variables are named the same as the lists.

AndreasArvidsson, line 127: We would probably use `user.` instead of self

core/keys/symbols.py
```diff
@@ -0,0 +1,122 @@
+# define the spoken forms for symbols in command and dictation mode
```
AndreasArvidsson, line 1: This file now became quite complicated. We're leaning towards just having one list with the constants/definitions and then a separate file with all the logic.

core/keys/symbols.py
```diff
@@ -0,0 +1,122 @@
...
+                [command_forms] if isinstance(command_forms, str) else command_forms
+            )
+
+
+currency_symbols = [
+    Symbol("$", ["dollar sign"], ["dollar"]),
+    Symbol("£", ["pound sign"], ["pound"]),
+]
+
+symbols = [
```
AndreasArvidsson, line 55: We're not sure that the data class where we combined punctuation and symbols is easy to read. Also we really need to read through this to make sure we don't have any regression from the current setup.

core/keys/symbols.py
```diff
@@ -0,0 +1,122 @@
...
+    "at": "@",
+    "dot": ".",
+    # for application names like "notepad++"
+    "plus": "+",
+}
+
+
+class Symbol:
+    character: str
+    command_forms: list[str] = None
```
AndreasArvidsson, line 29: Please have these in the same order as in the constructor. We read the order of the fields and were a bit confused about the definitions below.

AndreasArvidsson (changes requested):

core/keys/symbols.py
```diff
@@ -0,0 +1,122 @@
...
+            "brack",
+            "L brack",
+            "bracket",
+            "L bracket",
+            "left bracket",
+            "square",
+            "L square",
+            "left square",
+        ],
+    ),
```
nriley, line 88: From the community backlog session — we did not have an issue with including this in a single list but this formatting is unfortunate.

nriley (changes requested):

core/keys/symbols.py
knausj85: We agreed in the subsequent session that this is probably the least-worst option.
https://github.com/talonhub/community/pull/1763#discussion_r1986123977

core/keys/keys.py
knausj85: I settled on the same names with _dict to avoid name collisions with various captures.

core/keys/symbols.py
knausj85: I have small preference for keeping this as-is, so everything can be customized in one place. We can discuss further in the next session.

.pre-commit-config.yaml
```diff
@@ -38,6 +38,7 @@ repos:
     rev: 25.1.0
     hooks:
       - id: black
+        exclude: core/keys/symbols.py$
```
knausj85, line 41: After a bit of research, this seems like the least-worst approach to addressing the formatting.

AndreasArvidsson: Couldn't we achieve this with `# fmt: off` in the file instead?

knausj85: Done.

AndreasArvidsson (approved):

AndreasArvidsson:
@FireChickenProductivity

knausj85:
"right brace" and "r curly bracket" have been fixed. Thanks!

nriley (approved):

PR #1766, opened 2025-02-18
Fix behavior of unprefixed numbers with mouse grid

andyschweig:
Fix so that in mouse grid mode you can say a string of numbers together even with unprefixed numbers active. Without this change, those number keys are sent to the system instead of narrowing the mouse grid.


knausj85 (approved):

knausj85:
Thanks!

PR #1768, opened 2025-02-18
Migrate lua operators

FireChickenProductivity:
I added the missing exponentiation operator and added a new operator to the original operators definition for floor division.


knausj85:
This looks good to me. Unrelated to your changes, I noticed that lua.talon currently opts into `tag: user.code_operators_pointer` but I see no implementation thereof. Should we remove this while we're here?

knausj85 (approved):

PR #1776, opened 2025-02-22
Call "floor division" integer division

FireChickenProductivity:


auscompgeek (commented):
Renaming it for the sake of most programming languages sounds reasonable, although Python would probably deserve a "floor divide" alias.

nriley (approved):

PR #1777, opened 2025-02-23
vscode: copy, paste, redo and undo through command palette (2nd attempt)

BlueDrink9:
Draft PR to implement https://github.com/talonhub/community/pull/1677#issuecomment-2600944142 properly, such that it only applies in the main editor window (or at least, not in modal/system windows).

TODO:

- [x] Figure out how to parse that info from the window title
- [x] Also add find

### DIFF
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -125,7 +125,9 @@ class EditActions:
 
     def find(text: str = None):
         if text:
-            actions.user.run_rpc_command("editor.actions.findWithArgs", {"searchString": text})
+            actions.user.run_rpc_command(
+                "editor.actions.findWithArgs", {"searchString": text}
+            )
         else:
             actions.user.vscode("actions.find")
 
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -120,9 +120,6 @@ class EditActions:
     def paste():
         actions.user.vscode("editor.action.clipboardPasteAction")
 
-    def save():
-        actions.user.vscode("workbench.action.files.save")
-
     def find(text: str = None):
         if text:
             actions.user.run_rpc_command(
@@ -143,6 +140,9 @@ class EditActions:
 
     def save_all():
         actions.user.vscode("workbench.action.files.saveAll")
+
+    def save():
+        actions.user.vscode("workbench.action.files.save")
 
     def find_next():
         actions.user.vscode("editor.action.nextMatchFindAction")
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -3,6 +3,7 @@ from talon import Context, Module, actions, app
 is_mac = app.platform == "mac"
 
 ctx = Context()
+ctx_editor = Context()
 mac_ctx = Context()
 mod = Module()
 # com.todesktop.230313mzl4w4u92 is for Cursor - https://www.cursor.com/
@@ -60,6 +61,10 @@ and app.exe: /^cursor\.exe$/i
 ctx.matches = r"""
 app: vscode
 """
+ctx_editor.matches = r"""
+app: vscode
+and win.title: /focus:\[Text Editor\]/
+"""
 mac_ctx.matches = r"""
 os: mac
 app: vscode
@@ -98,7 +103,10 @@ class CodeActions:
         actions.user.vscode("editor.action.commentLine")
 
 
-@ctx.action_class("edit")
+# In the editor, use RPC commands to avoid conflicting with the editor's keybindings.
+# Only do this for editor, so that e.g. modal windows can still be pasted into with
+# ctrl-v.
+@ctx_editor.action_class("edit")
 class EditActions:
     def undo():
         actions.user.vscode("undo")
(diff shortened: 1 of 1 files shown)


BlueDrink9:
That seems to be working. Regular paste in the modal windows, but rpc within the editor! Might still not work in situations like copilot chat, where vim extensions are still active but the focus might not technically be a text editor, but for now this is good.

Are we happy with only these particular RPC commands being in the separate context? Or should I move all of them / the majority?

BlueDrink9:
Forgot that this requires tweaking vscode to put the focused view in the window title (I had assumed the vscode extension tweaked that for us), so I think leaving just these ones in the extra context is the right move.

apps/vscode/vscode.py
```diff
@@ -60,6 +61,10 @@
 ctx.matches = r"""
 app: vscode
 """
+ctx_editor.matches = r"""
+app: vscode
+and win.title: /focus:\[Text Editor\]/
```
AndreasArvidsson, line 66: Unfortunately this is not default vscode and will not work for most users

BlueDrink9: I'm aware it isn't default, but it also won't do any harm if they don't have it set up. If it doesn't match the context, it falls back to regular undo etc. 

Adding that to window title is part of the recommended settings, so I thought it would be fine as an optional extra

BlueDrink9: (unsure whether your comment was intended as a "this is a blocker to the PR" comment or whether it was simply an observation)

apps/vscode/vscode.py
```diff
@@ -98,6 +103,33 @@ def toggle_comment():
...
+    def redo():
+        actions.user.vscode("redo")
+
+    def copy():
+        actions.user.vscode("editor.action.clipboardCopyAction")
+
+    def paste():
+        actions.user.vscode("editor.action.clipboardPasteAction")
+
+    def save():
```
nriley, line 123: From the community backlog session — seems like this may be a regression with the save action if this context never activates itself unless you have the window title configured as is recommended?

nriley (changes requested):

apps/vscode/vscode.py
```diff
@@ -98,6 +103,33 @@ def toggle_comment():
...
+        actions.user.vscode("workbench.action.files.save")
+
+    def find(text: str = None):
```
nriley, line 126: From the community backlog session — this doesn't seem a substantial improvement over using a keyboard shortcut. There is an alternative action called "editor.actions.findWithArgs" that would accept the search string as an argument rather than having to type it or paste it as below.

nriley (changes requested):
From the community backlog session — we had a few comments on a couple of the specific actions you have implemented but overall want to understand the advantage of doing this for basic editing commands over using keyboard shortcuts that would typically work regardless of context within Visual Studio Code.

apps/vscode/vscode.py
BlueDrink9: Oh yes, that is a mistake. Thanks

apps/vscode/vscode.py
BlueDrink9: That would be better, I'll switch it. Re: keyboard shortcuts, the argument I made on the other post still stands: not everyone uses the same keyboard shortcuts, e.g. in vim mode, so bypassing them where possible increases reliability.

BlueDrink9:
Thanks all for the review, I'll address those. @nriley may I tag you once those are addressed?

apps/vscode/vscode.py
nriley: Gotcha, sorry it's been a while and I missed this…

nriley:
> Thanks all for the review, I'll address those. @nriley may I tag you once those are addressed?

Of course!

apps/vscode/vscode.py
BlueDrink9: Actually on second thoughts this is not a regression. Previously the save action used the VSCode command only because I had submitted that pr previously. With this pr it goes back to using the normal save action, which is control s and will work for anyone who doesn't have the extra context enabled, and only uses the vs code command if it knows that it is in the editor action. 

(In reality using the save command either inside this context or in all VSCode contexts doesn't really matter too much because you don't ever need to use save in a modal dialogue anyway, but just in case, it's better in the extra context section.)

apps/vscode/vscode.py
BlueDrink9: I've updated this. Assuming my other comment about the save action is acceptable, this should be ready to merge @nriley

BlueDrink9:
@nriley ready for review (pinging again now it's the weekend)

BlueDrink9:
@nriley tagging to get this merged in the upcoming community session

apps/vscode/vscode.py
nriley: From the community backlog session including @AndreasArvidsson — add a comment that links to the [VSCode README](https://github.com/talonhub/community/blob/main/apps/vscode/README.md#option-1-add-focused-view-to-window-title) and I think we are good.

apps/vscode/vscode.py
nriley: From the community backlog session — we discovered that saying "file save" on Windows with the terminal focused will still save the frontmost editor, and if this change goes into place you will no longer be able to save the file without focusing it first, so it is worth undoing this change.

nriley (changes requested):

nriley (approved):

PR #1780, opened 2025-02-25
add assert and pass to python's user.code_keyword

NonsenseRemedy:
I'm not sure if this is where these belong,  but I figured I'd ask via merge request rather than  via a post to slack  :-)

### DIFF
--- a/lang/python/python.py
+++ b/lang/python/python.py
@@ -62,7 +62,7 @@ ctx.lists["user.code_type"] = {
 }
 
 ctx.lists["user.code_keyword"] = {
-    "assert": "assert",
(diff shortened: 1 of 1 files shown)


lang/python/python.py
```diff
@@ -62,13 +62,15 @@
 }
 
 ctx.lists["user.code_keyword"] = {
+    "assert": "assert ",
```
NonsenseRemedy, line 65: I often use assert like `assert foo is not None`,  but could also see an argument for putting this in functions

nriley (approved):
Seems like a no brainer to me. Thanks!

PR #1782, opened 2025-02-26
Python's not operator should have space before it

NonsenseRemedy:


NonsenseRemedy:
I'm not sure about this one --  I see an argument that the way `not` is used should always have generated a space before saying `op not` (e.g. `snip if op not word foo`), but I think I prefer to be sure that the operator is always inserted separated by space and let my formatter clean that up as necessary.

knausj85:
I agree with this.

knausj85 (approved):

PR #1790, opened 2025-03-01
Migrate rust operators

FireChickenProductivity:
Migrate the existing rust operators and add pointer operators. 

I think that the original
```python
def code_operator_structure_dereference():
        actions.auto_insert("*")
```
may be mistaken. I migrated that operator as is because I do not know much about rust.

### DIFF
--- a/lang/rust/rust.py
+++ b/lang/rust/rust.py
@@ -263,7 +263,6 @@ operators = Operators(
     # code_operators_pointer
     POINTER_INDIRECTION="*",
     POINTER_ADDRESS_OF="&",
-    POINTER_STRUCTURE_DEREFERENCE="*",
 )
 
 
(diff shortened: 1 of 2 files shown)


lang/rust/rust.py
```diff
@@ -221,9 +223,55 @@ def code_type(m) -> str:
...
+    MATH_GREATER_THAN_OR_EQUAL=" >= ",
+    MATH_LESS_THAN=" < ",
+    MATH_LESS_THAN_OR_EQUAL=" <= ",
+    MATH_AND=" && ",
+    MATH_OR=" || ",
+    ASSIGNMENT_INCREMENT=" += 1",
+    # code_operators_pointer
+    POINTER_INDIRECTION="*",
+    POINTER_ADDRESS_OF="&",
+    POINTER_STRUCTURE_DEREFERENCE="*",
```
nriley, line 266:
```suggestion
```

nriley (approved):

PR #1797, opened 2025-03-04
Migrate scala operators (from the ones defined as actions)

FireChickenProductivity:
This does not migrate the actions in the talon file. 

I am not sure what to do with these

```
op right arrow: " -> "
op left arrow: " <- "
op plus plus: " ++ "
op subtype: " <: "
```

I did add some missing operators.

### DIFF
--- a/lang/scala/scala.py
+++ b/lang/scala/scala.py
@@ -1,5 +1,7 @@
 from talon import Context, Module, actions, settings
 
+from ..tags.operators import Operators
+
 ctx = Context()
 mod = Module()
 ctx.matches = r"""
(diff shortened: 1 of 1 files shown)


lang/scala/scala.py
```diff
@@ -83,102 +85,54 @@
...
+    # code_operators_array
+    SUBSCRIPT=lambda: actions.user.insert_between("(", ")"),
+    # code_operators_assignment
+    ASSIGNMENT=" = ",
+    ASSIGNMENT_ADDITION=" += ",
+    ASSIGNMENT_SUBTRACTION=" -= ",
+    ASSIGNMENT_MULTIPLICATION=" *= ",
+    ASSIGNMENT_DIVISION=" /= ",
+    ASSIGNMENT_MODULO=" %= ",
+    ASSIGNMENT_INCREMENT="++",
```
nriley, line 98: From the community backlog session — according to the scala language specification it appears that `++` is not a standard operator either when used infix or postfix. Therefore if we are not adding any additional operators, it's fine to preserve the existing behavior.

nriley (approved):
None of us know scala well enough to weigh in on this but @knausj85 did test this and ensure that all the operators continue to work.

PR #1798, opened 2025-03-05
Migrate javascript operators

FireChickenProductivity:
I believe this was my first time deprecating something. Please let me know if I should handle that differently. 

Strict equality is now default. 

This migrates the javascript operators to the new format and adds the increment operator.


lang/javascript/javascript.py
```diff
@@ -73,9 +75,56 @@
...
+    MATH_EXPONENT=" ** ",
+    MATH_EQUAL=" === ",
+    MATH_NOT_EQUAL=" !== ",
+    MATH_OR=" || ",
+    MATH_AND=" && ",
+    MATH_GREATER_THAN=" > ",
+    MATH_LESS_THAN=" < ",
+    MATH_GREATER_THAN_OR_EQUAL=" >= ",
+    MATH_LESS_THAN_OR_EQUAL=" <= ",
+    MATH_WEAK_EQUAL=" == ",
```
knausj85, line 118: Should these two need to be defined in operators.py? 

It's also probably worth adding some javascript specific commands to actually support these? I don't know how common these are in other languages

FireChickenProductivity: Which two? Are you referring to weak comparison? Several languages support those operators but the grammar is currently inconsistent in community.

nriley:
We agreed on the community backlog session for @FireChickenProductivity to ask on the slack about whether the current plan for deprecation makes sense to other users as none of the contributors were heavy users of javascript and community.

lang/javascript/javascript.py
nriley: From the community backlog session — these are already implemented in operators.py for other languages. So I think we are good here.

nriley (approved):

PR #1806, opened 2025-03-17
Add more features to discord app

ProfDoof:
Went through the Discord help list and tried to add any feature that wasn't already supported. I haven't gotten the chance to completely test everything yet as I am not on Windows. I wasn't sure how to approach handling editing messages though.

The message editing commands are currently sitting as a set of comments in the talon file. I thought about possibly having a context flag that would set a tag for messages being focused. Something like "focus messages" that tabs once, and then they could navigate up and down. Then, they would be able to use an edit command, reply command, etc. However, I wanted to get other people's thoughts on that before doing something more complex like that.

I looked at the contribution guidelines and they say to prefer object verb but the structure is already verb object so I'm not sure whether it should be redesigned or left in that form.

### DIFF
--- a/apps/discord/discord.py
+++ b/apps/discord/discord.py
@@ -68,6 +68,9 @@ class discord_actions:
     def discord_decline_call():
         """Decline incoming call"""
 
+    def discord_start_call():
+        """Start a call in a private DM or group"""
+
     def discord_quick_switcher(dest_type: str, dest_search: str):
         """Open up the quick switcher, optionally specifying a type of destination"""
 
@@ -77,9 +80,37 @@ class discord_actions:
     def discord_toggle_dms():
         """Toggle between dms and your most recent server"""
 
+    def discord_create_join_server():
+        """Create or join a new server"""
+
+    def discord_toggle_soundboard():
+        """Toggle the soundboard while in a call"""
+
+    def discord_create_private_group():
+        """Create private group"""
+
+    def discord_channel_oldest_unread():
+        """Go to oldest unread message in channel"""
+
+    def discord_open_support():
+        """Open the Discord support page in browser"""
+
+    def discord_start_lofi():
+        """Listen to Lofi beats with Wumpus""" 
+
 
 @ctx.action_class("user")
 class UserActions:
     # Navigation: Channels
     def messaging_open_channel_picker():
         actions.user.discord_quick_switcher("#", "")
+
+
+@ctx.action_class("edit")
+class EditActions:
+    def find_next():
+        pass
+
+    def find_previous():
+        pass
+
--- a/apps/discord/discord.talon
+++ b/apps/discord/discord.talon
@@ -2,6 +2,22 @@ app: discord
 -
 tag(): user.messaging
 tag(): user.emoji
+tag(): user.navigation
+tag(): user.find
+
+# Messages
+# Edit Message: E
+# Delete Message: Backspace
+# Pin Message: P
+# Add Reaction: +
+# Reply: R
+# Copy Text: CMD+C or Ctrl+C
+# Mark Unread: OPT+ENTER
+# Focus Text Area: ESC
+# Forward Message: F
+
+# Navigation: Messages
+go oldest unread: user.discord_channel_oldest_unread()
 
 # Navigation: QuickSwitcher
 {user.discord_destination} [<user.text>]:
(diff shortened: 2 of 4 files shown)


apps/discord/discord.py
```diff
@@ -77,9 +80,36 @@ def discord_go_current_call():
...
 class UserActions:
     # Navigation: Channels
     def messaging_open_channel_picker():
         actions.user.discord_quick_switcher("#", "")
+
+
+@ctx.action_class("edit")
+class EditActions:
+    def find_next():
+        pass
```
auscompgeek, line 112: Action implementations within a context shouldn't do nothing. Unless there's a good reason to do this, in which case there should be a comment here?

auscompgeek (changes requested):

apps/discord/discord.py
ProfDoof: I provided a comment explaining my logic. I can remove this section though if that logic is incorrect for some reason.

ProfDoof: @auscompgeek does the provided logic make sense?

nriley: On the community backlog session — this makes sense as there is no find next command in Discord.

apps/discord/discord.talon
```diff
@@ -18,7 +34,9 @@ current call: user.discord_go_current_call()
 toggle pins: user.discord_toggle_pins()
 toggle inbox: user.discord_toggle_inbox()
 toggle (members | member list): user.discord_toggle_members()
+toggle (vee see text | voice text): user.discord_toggle_members()
```
nriley, line 37: Have you tried this with `v c text` instead? This usually works fine with conformer and is a bit more intention revealing.

apps/discord/discord.talon
```diff
@@ -32,3 +50,8 @@ mark inbox channel read: user.discord_mark_inbox_read()
 [toggle] (deafen | undeafen): user.discord_deafen()
 answer call: user.discord_answer_call()
 decline call: user.discord_decline_call()
+start call: user.discord_start_call()
+(create | join) server: user.discord_create_join_server()
+create private group: user.discord_create_private_group()
+open support: user.discord_open_support()
+start low fie: user.discord_start_lofi()
```
nriley, line 57: From the community backlog session — I realize we do not have a lot of consistency here (issue on this is incoming) already but with new voice commands we tend to prefer a noun verb syntax.

apps/discord/discord.talon
```diff
@@ -2,6 +2,22 @@ app: discord
...
+# Pin Message: P
+# Add Reaction: +
+# Reply: R
+# Copy Text: CMD+C or Ctrl+C
+# Mark Unread: OPT+ENTER
+# Focus Text Area: ESC
+# Forward Message: F
+
+# Navigation: Messages
+go oldest unread: user.discord_channel_oldest_unread()
```
nriley, line 20: From the community backlog session — this should probably go into `messaging.talon` and have a syntax more similar to the existing commands for navigating around unread channels.

apps/discord/discord_win.py
```diff
@@ -90,8 +90,29 @@ def discord_answer_call():
...
+    def discord_toggle_soundboard():
+        actions.key("ctrl-shift-b")
+
+    def discord_create_private_group():
+        actions.key("ctrl-shift-t")
+
+    def discord_channel_oldest_unread():
+        actions.key("shift-pageup")
+
+    def discord_open_support():
```
nriley, line 114: Wondering if this should be a more general command to open help rather than something discord specific.

nriley (changes requested):

ProfDoof:
I'll try to get back to this in the next couple of days.

apps/discord/discord.talon
ProfDoof: Do you guys have a standard method of deprecating commands? If so, I could deprecate the prior syntax and convert to the new syntax.

apps/discord/discord_win.py
ProfDoof: Personally, I'm for it being a general command. However, I'm unsure as to what a clean syntax for that would look like. Help is already taken. It could be `app help` maybe?

Is there a standard way of handling common functionalities between applications that need to be differentiated from talon ones?

nriley: Sure, you could consider defining a tag for it, or (probably better in this case) implementing it for an entire platform at a time via an action.

apps/discord/discord.talon
nriley: Yup! See core/deprecations.py and BREAKING_CHANGES.txt.

PR #1807, opened 2025-03-18
Add deeper sleep support

FireChickenProductivity:
Basic functionality is working, but I need to test that the centralized sleep and wake up actions are behaving as intended in every context. I also need to figure out what to do with the "welcome back" command.


FireChickenProductivity:
#1807

FireChickenProductivity:
I am not sure where to put the following comment now that sleep_mode.talon has no more commands left (because I moved welcome back)

#================================================================================
# Commands to wake Talon
#================================================================================

# Note: these have repeaters on them (+) to work around an issue where, in sleep mode,
# you can get into a situation where these commands are difficult to trigger.

# These commands are fully anchored (^ and $), which means that there must be
# silence before and after saying them in order for them to recognize (this reduces
# false positives during normal sleep mode, normally a good thing).

# However, ignored background speech during sleep mode also counts as an utterance.

# Thus, if you say "blah blah blah talon wake", these won't trigger, because "blah
# blah blah" was part of the same utterance. You have to say "blah blah blah" <pause,
# wait for speech timeout>, "talon wake" <pause, wait for speech timeout>.

# Sometimes people would forget the second pause, notice things weren't working, and
# say "talon wake" over and over again before the speech timeout ever gets hit, which
# means that these won't recognize. The (+) handles this case, so if you say
# <pause> "talon wake talon wake" <pause>, it'll still work.

AndreasArvidsson:
Feedback from the community session. Is probably simpler to us disable the normal wake up command when the tag is active and instead activate a separate command. This separate command could be the same phrase multiple times or could just be a more verbose wake up command. I don't think we need a user interface for this.

FireChickenProductivity:
I decided it would be easiest to close this and make a new PR

PR #1811, opened 2025-03-21
Migrate php operators

FireChickenProductivity:
Migrate php operators. I believe this is the final operators migration. I added some missing operators. There are some further operators that might be worth adding, but that might be best left to pull requests from people who understand the language. 

I wonder if adding "is loosely equal" and "is loosely not equal" as aliases might be better than deprecation.


AndreasArvidsson (approved):

AndreasArvidsson:
Please give this a second test and then merge it

FireChickenProductivity:
I have tested it again and found no issues, but I do not think I have permission to do the merge.

PR #1813, opened 2025-03-23
Add deeper sleep support

FireChickenProductivity:
Add a tag for enabling sleep with a longer wakeup command


core/modes/sleep_mode_not_deep_sleep_or_dragon.talon
```diff
@@ -0,0 +1,15 @@
+mode: command
```
AndreasArvidsson, line 1: sleep_mode_not_dragon

core/modes/sleep_mode_not_deep_sleep.talon
```diff
@@ -0,0 +1,8 @@
+mode: sleep
```
AndreasArvidsson, line 1: sleep_mode_dragon

plugin/deep_sleep/deep_sleep.talon
```diff
@@ -0,0 +1,5 @@
+mode: sleep
```
AndreasArvidsson, line 1: sleep_mode_deep
move to modes directory

plugin/deep_sleep/deep_sleep.py
```diff
@@ -0,0 +1,22 @@
+from talon import Context, Module, actions
```
AndreasArvidsson, line 1: move to modes directory

AndreasArvidsson (dismissed):
See existing comments

core/modes/sleep_mode_not_deep_sleep.talon
FireChickenProductivity: Given that the file is for wake up commands available with and without dragon, I renamed this to sleep_mode_wakeup.talon for now. Is that acceptable? I renamed the file dealing with dragon sleep commands "sleep_mode_dragon.talon".

FireChickenProductivity:
I have tested this including with dragon. I recommend someone else test this before merge simply because breaking sleep would not be fun.

nriley:
From the community backlog session — we are looking for some additional testers. @knausj85 and I may have some time next weekend but would appreciate if anyone else could give this a look as sleep is one of those things we do not want to break!

core/modes/deep_sleep.py
```diff
@@ -0,0 +1,22 @@
+from talon import Context, Module, actions
+
+mod = Module()
+mod.tag(
```
davidsgrogan, line 4: Do you mind adding a short comment somewhere that tells why this is useful? Took me a while to figure out that deep sleep is _just_ to better prevent accidental wakeups; it doesn't actually disable more stuff than regular sleep. (right?) And, there's no provided command that calls deep_sleep_enable(), it's up to the user to define their own, right?

something like:
\# If you want to decrease the chance of accidental wakeups, add a command that calls deep_sleep_enable(). Then talon will have to detect "wake up and listen" instead of the normal "wake up".

FireChickenProductivity: I am not sure where the best place to document this would be. I can ask in the next backlog session. 

There is indeed no built in command for calling deep_sleep_enable(). It does not disable more stuff than regular sleep.

The main way I would expect people to activate this would actually be with the tag rather than defining a command. The tag can be activated with apps like meeting apps where it would be convenient to have regular sleep commands put talon in a state that is harder to wake up on accident.

AndreasArvidsson:
Please add a documentation string describing this feature. What it does and why it's useful.

FireChickenProductivity:
> Please add a documentation string describing this feature. What it does and why it's useful.

I added a documentation string to the deep sleep enabling action. (As a reminder, we are still waiting on further testing, so please do not merge this early!)

core/modes/deep_sleep.py
davidsgrogan: Your below comment looks great.

core/modes/deep_sleep.py
```diff
@@ -0,0 +1,27 @@
...
+ctx = Context()
+
+
+@mod.action_class
+class Actions:
+    def deep_sleep_enable():
+        """Enable deep sleep.
+        Deep sleep has the advantage of requiring a longer wakeup command to exit sleep mode, which can help prevent unintended wakeups from conversations, meetings, listening to videos, etc.
+        Community does not come with a built in command for activating deep sleep. The expected way to use it is enabling the "user.deep_sleep" tag in applications where unwanted wakeups are more likely or problematic like meeting apps. Users can still enable the action with a command by creating a custom command using this action.
+        Note that when activating the tag on a per application basis, you must remember to explicitly match sleep mode as command mode is implicitly matched by default.
```
nriley:
```suggestion
        Note that when activating the tag on a per application basis, you must remember to explicitly match sleep mode, as command mode is implicitly matched by default.
```

core/modes/deep_sleep.py
```diff
@@ -0,0 +1,27 @@
...
+
+ctx = Context()
+
+
+@mod.action_class
+class Actions:
+    def deep_sleep_enable():
+        """Enable deep sleep.
+        Deep sleep has the advantage of requiring a longer wakeup command to exit sleep mode, which can help prevent unintended wakeups from conversations, meetings, listening to videos, etc.
+        Community does not come with a built in command for activating deep sleep. The expected way to use it is enabling the "user.deep_sleep" tag in applications where unwanted wakeups are more likely or problematic like meeting apps. Users can still enable the action with a command by creating a custom command using this action.
```
nriley:
```suggestion
        Deep sleep requires a longer wakeup command to exit sleep mode (defined in `sleep_mode_deep.talon`), which can help prevent unintended wakeups from conversations, meetings, listening to videos, etc. Instead of invoking this action directly, consider enabling the `user.deep_sleep` tag in applications where unwanted wakeups are more likely or problematic, such as meeting apps. With this tag active, any sleep command triggers deep sleep.
        Users can also manually activate deep sleep by defining a custom voice command using this action.
```

core/modes/deep_sleep.py
```diff
@@ -0,0 +1,45 @@
...
+        Instead of invoking this action directly, consider enabling
+        the `user.deep_sleep` tag in applications where unwanted
+        wakeups are more likely or problematic, such as meeting
+        apps. With this tag active, any sleep command triggers deep
+        sleep.
+
+        When activating the tag on a per application basis, you must
+        remember to explicitly match sleep mode, as command mode is
+        implicitly matched by default.
+
```
nriley: I'm not sure if this is needed. I didn't have any trouble just declaring the tag and it stayed active when needed despite being in sleep mode.

image

AndreasArvidsson: It's only commands that implicitly match command mode. tags, settings etc are enabled for all modes if no mode is specified.

nriley:
```suggestion
```

nriley (approved):

PR #1814, opened 2025-03-25
change "homophones hide"=>"homophones exit"

NonsenseRemedy:
I consistently get misrecognitions when trying to hide the homophones picker menu, since "phones hide" sounds a lot like "phone side", which opens another homophones picker menu.

if other people have this problem too, it might make sense to change the defaults. (and unlike most cases of changing spoken commands, I had to dig a bit to get the gui text to match the .talon file)

I also thought of using "phones close", but that's a homophone too 😆

### DIFF
--- a/core/homophones/homophones.py
+++ b/core/homophones/homophones.py
@@ -162,7 +162,7 @@ def gui(gui: imgui.GUI):
                 actions.user.homophones_hide()
             index = index + 1
 
-        if gui.button("Phones exit"):
+        if gui.button("Phones (hide | exit)"):
             actions.user.homophones_hide()
 
 
--- a/core/homophones/homophones.py
+++ b/core/homophones/homophones.py
@@ -162,7 +162,7 @@ def gui(gui: imgui.GUI):
                 actions.user.homophones_hide()
             index = index + 1
 
-        if gui.button("Phones hide"):
+        if gui.button("Phones exit"):
             actions.user.homophones_hide()
 
 
--- a/core/homophones/homophones.talon
+++ b/core/homophones/homophones.talon
@@ -3,7 +3,7 @@ phones that: user.homophones_show_auto()
 phones force <user.homophones_canonical>:
     user.homophones_force_show(homophones_canonical)
 phones force: user.homophones_force_show_selection()
-phones hide: user.homophones_hide()
+phones exit: user.homophones_hide()
 phones word:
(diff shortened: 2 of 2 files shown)


core/homophones/homophones.talon
```diff
@@ -3,7 +3,7 @@ phones that: user.homophones_show_auto()
 phones force <user.homophones_canonical>:
     user.homophones_force_show(homophones_canonical)
 phones force: user.homophones_force_show_selection()
-phones hide: user.homophones_hide()
+phones exit: user.homophones_hide()
```
AndreasArvidsson, line 6: Instead of changing it we could have hide or exit. eg `hide | exit`

AndreasArvidsson (dismissed):

AndreasArvidsson (changes requested):
See existing comment

core/homophones/homophones.talon
nriley: On the community backlog session we discussed that theres some inconsistently between closing various popup windows in community but that hide is at least somewhat common, so we should preserve this functionality.

NonsenseRemedy: duh. Of course it's better to preserve the existing functionality. Updated.

core/homophones/homophones.talon
```diff
@@ -3,7 +3,7 @@ phones that: user.homophones_show_auto()
 phones force <user.homophones_canonical>:
     user.homophones_force_show(homophones_canonical)
 phones force: user.homophones_force_show_selection()
-phones hide: user.homophones_hide()
+phones [hide | exit]: user.homophones_hide()
```
nriley, line 6:
```suggestion
phones (hide | exit): user.homophones_hide()
```

core/homophones/homophones.py
```diff
@@ -162,7 +162,7 @@ def gui(gui: imgui.GUI):
                 actions.user.homophones_hide()
             index = index + 1
 
-        if gui.button("Phones hide"):
+        if gui.button("Phones exit"):
```
nriley, line 165:
```suggestion
        if gui.button("Phones (hide | exit)"):
```

nriley (changes requested):

nriley (approved):

AndreasArvidsson (approved):

PR #1820, opened 2025-04-13
C++ support

chdoc:
I promised a long time ago that I would set out to unify support for C and C++. I've been tinkering with this for quite a while now, and I finally decided to wrap what I had up into a commit and put it here for inspection. 

Isn't really cleaned up yet, but I would nevertheless welcome feedback. In particular, I myself do not actually program in C. So any feedback from someone who actually uses the current C support would be valuable.

While certainly far from perfect, it has absolutely improved my experience in writing C++ code.


lang/c/c.py
```diff
@@ -250,16 +303,22 @@ def code_insert_is_not_null():
         actions.auto_insert(" != NULL")
 
     def code_state_if():
-        actions.insert("if () {\n}\n")
-        actions.key("up:2 left:3")
+        actions.insert("if (")
```
nriley, line 306: I'm not sure we documented this anywhere but on the community backlog sessions we had decided to just directly use snippets here.

lang/c/c.talon
```diff
@@ -7,7 +8,7 @@ tag(): user.code_comment_line
 tag(): user.code_comment_block_c_like
 tag(): user.code_data_bool
 tag(): user.code_data_null
-tag(): user.code_functions
+# tag(): user.code_functions
```
nriley, line 11: Could you explain why you have disabled this?

lang/c/c.talon
```diff
@@ -25,27 +26,32 @@ settings():
...
+include <user.code_libraries>:
+    user.code_insert_library(code_libraries, "")
+
 state type deaf: insert("typedef ")
 state type deaf struct:
     insert("typedef struct")
     insert("{\n\n}")
     edit.up()
-    key('tab')
+    key('tab')preprocessor
```
nriley, line 40:
```suggestion
    key('tab')
```

nriley: This may be that you have not finished migration yet but it makes sense for anything where you are doing text navigation in the midst of insertion to migrate to a snippet as this is less likely to fight with builtin editor commands.

lang/c/c.talon
```diff
@@ -58,35 +64,45 @@ state default: "default:\nbreak;"
...
+
+push (semi|semicolon):
+     edit.line_end()
+     insert(";")
+     key(enter)
 
 # Declare variables or structs etc.
 # Ex. * int myList
-<user.c_variable> <phrase>:
+(var|variable) <user.c_variable> <phrase>:
```
nriley, line 84: Would recommend keeping track of any places where you are changing existing functionality such it can be added to the breaking changes file.

lang/c/cpp.py
```diff
@@ -0,0 +1,80 @@
...
+code.language: cpp
+language: en
+"""
+
+mod.list("cpp_standard_functions", desc="Functions in the std namespace")
+mod.list("cpp_standard_constants", desc="Constants in the std namespace")
+mod.list("cpp_standard_types", desc="Types in the std namespace")
+mod.list("cpp_namespace", desc="Commonly used namespaces")
+
+ctx.lists["user.code_libraries"] = {
```
nriley, line 20: Potentially consider using Talon lists for these

nriley (commented):
From the community backlog session — thank you so much for all this work. Just a few minor suggestions and happy to keep reviewing or answering questions as you have them.

lang/c/c.py
chdoc: Admittedly, I didn't quite know what to do with these. They existed in the original C-support, and all I did was comment out the part that breaks if you use an editor that adds closing braces/parentheses. Personally, I stopped using the "state <structure>" commands. I'm totally fine with removing them.

lang/c/c.talon
chdoc: That is a good question. I remember being upset that `funky` didn't have any means to sneak in a return type for the function you're defining. Started experimenting with a more specific `funky` command, but from the looks of it I never finished that...
```
[<user.c_qualifier_list>] <user.c_type> funky <user.text>:
    qualifiers = c_qualifier_list or ""
    insert("{qualifiers} {text}")
```
This has been alleviated for me by defining the function definition snippet to use the `auto` return type and simply using `snip funky` for everything. Maybe I should just reinstate that tab.

lang/c/c.talon
chdoc: I thought about something like this, but then I basically completely overhauled the mode. I really wish I knew someone who was actually using the current C mode (or was confident that there actually isn't anyone).

lang/c/c.talon
chdoc: Apart from the accidental insert of the word "preprocessor", I didn't touch this at all. The use of `typedef` is discouraged in modern C++.

nriley: Gotcha, OK. If you don't want to touch the existing C support that's fine; I had assumed that your accidental insertion meant you were trying to edit this code.

lang/c/c.talon
nriley: There are definitely people using the existing C mode. If you add things that are C++ only, no need to document that, only if you change them. This is the standard we used for community maintenance.

nriley:
As above, we (primarily @FireChickenProductivity) are starting to merge pieces of this PR into community. We'll close this PR when we get everything incorporated. Please let us know if you have comments in the meantime!

chdoc:
@FireChickenProductivity If you get around to adding parts of this as separate PRs, please make sure to reference this so that I can stay in the loop. I currently have a few cycles to spare and I would like to further the cause of C++ support in talon.

PR #1827, opened 2025-04-28
Various list extensions

Enyium:
- The lines with `#` on its own were necessary to prevent Black from destroying the grouping.
- Are the special keys "num lock", "caps lock", and "scroll lock" correct on Mac?


nriley:
From the community backlog session — apologies for overlooking this PR for several months.

Could you please split your changes into a single PR for each file you are changing? Thanks!

nriley (changes requested):
Please see the comment above.

PR #1836, opened 2025-04-29
Tab handling improvements/extensions

Enyium:
See commits.

If you're asking yourself what apps support a tab switcher that can be sufficiently controlled, Directory Opus does, for which I wrote a voice command set. I know that Firefox also has a tab switcher, but - to my knowledge - doesn't allow control about it like persistently showing it instead of only while holding a modifier key down. But I guess there's a chance that such apps will make persistent tab switcher dialogs available in the future, if their developers learn that it may be useful.

I saw Talon comes with an `app.tab_detach()` function slot, so I also made a phrase for it. It wasn't used before.

### DIFF
--- a/core/windows_and_tabs/tabs.talon
+++ b/core/windows_and_tabs/tabs.talon
@@ -1,10 +1,11 @@
 tag: user.tabs
 -
 tab (open | new): app.tab_open()
+tab (reopen | restore): app.tab_reopen()
+tab (duplicate | clone): user.tab_duplicate()
+tab close: user.tab_close_wrapper()
+
 tab (last | previous): app.tab_previous()
 tab next: app.tab_next()
-tab close: user.tab_close_wrapper()
-tab (reopen | restore): app.tab_reopen()
 go tab <number>: user.tab_jump(number)
 go tab final: user.tab_final()
-tab (duplicate | clone): user.tab_duplicate()
--- a/core/windows_and_tabs/tabs.talon
+++ b/core/windows_and_tabs/tabs.talon
@@ -5,7 +5,11 @@ tab (reopen | restore): app.tab_reopen()
 tab (duplicate | clone): user.tab_duplicate()
 tab close: user.tab_close_wrapper()
 
-tab (last | previous): app.tab_previous()
-tab next: app.tab_next()
+tab (last | previous | left | up): app.tab_previous()
+tab (next | right | down): app.tab_next()
 go tab <number>: user.tab_jump(number)
 go tab final: user.tab_final()
+
+tab move (left | up): user.tab_move_left()
+tab move (right | down): user.tab_move_right()
+tab detach: app.tab_detach()
--- a/core/windows_and_tabs/tabs.talon
+++ b/core/windows_and_tabs/tabs.talon
@@ -10,6 +10,10 @@ tab (next | right | down): app.tab_next()
 go tab <number>: user.tab_jump(number)
 go tab final: user.tab_final()
 
+tab focus <user.text>: user.tab_switcher_focus(text)
+tab focus$: user.tab_switcher_menu()
+tab focus last: user.tab_switcher_focus_last()
+
 tab move (left | up): user.tab_move_left()
 tab move (right | down): user.tab_move_right()
 tab detach: app.tab_detach()
--- a/core/windows_and_tabs/tabs_win.py
+++ b/core/windows_and_tabs/tabs_win.py
@@ -1,0 +1,15 @@
+from talon import Context, actions
+
+ctx = Context()
+ctx.matches = r"""
+os: windows
+"""
+
+
+@ctx.action_class("user")
+class UserActions:
+    def tab_move_left():
+        actions.key("ctrl-shift-pageup")
+
+    def tab_move_right():
+        actions.key("ctrl-shift-pagedown")
--- a/core/windows_and_tabs/tabs.py
+++ b/core/windows_and_tabs/tabs.py
@@ -12,10 +12,10 @@ class tab_actions:
         actions.app.tab_close()
 
     def tab_duplicate():
-        """Duplicates the current tab."""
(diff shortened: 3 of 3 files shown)


core/windows_and_tabs/tabs.talon
```diff
@@ -1,10 +1,19 @@
...
+
+tab (last | previous | left | up): app.tab_previous()
+tab (next | right | down): app.tab_next()
 go tab <number>: user.tab_jump(number)
 go tab final: user.tab_final()
-tab (duplicate | clone): user.tab_duplicate()
+
+tab focus <user.text>: user.tab_switcher_focus(text)
+tab focus$: user.tab_switcher_menu()
+tab focus last: user.tab_switcher_focus_last()
```
nriley, line 15: From the community backlog session — we did not think this was a great name for this command, specifically because "tab focus last" and "tab last" do two different things, but also couldn't think of a better one. Since you do not have an implementation for this we would recommend that you just remove it from the PR and potentially suggest some other names.

core/windows_and_tabs/tabs_win.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+
+
+@ctx.action_class("user")
+class UserActions:
+    def tab_move_left():
+        actions.key("ctrl-shift-pageup")
+
+    def tab_move_right():
+        actions.key("ctrl-shift-pagedown")
```
nriley, line 15: These commands at least work in the Mac version of Microsoft Edge so would recommend you implement this cross platform.

nriley (changes requested):

core/windows_and_tabs/tabs.talon
```diff
@@ -1,10 +1,19 @@
...
 tab (open | new): app.tab_open()
-tab (last | previous): app.tab_previous()
-tab next: app.tab_next()
-tab close: user.tab_close_wrapper()
 tab (reopen | restore): app.tab_reopen()
+tab (duplicate | clone): user.tab_duplicate()
+tab close: user.tab_close_wrapper()
+
+tab (last | previous | left | up): app.tab_previous()
+tab (next | right | down): app.tab_next()
```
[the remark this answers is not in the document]
Enyium: - It would be inconsistent to have "tab move left/right", but only "tab last/next". If you'd want to change the former, what would the phrases be?
- I find the relative "tab last/previous/next" to be not intuitive enough. Tab sequences don't have a strict reading direction like text. The user - like me - may have evolved various personal standards of ordering tabs for different situations that don't always harmonize with the meaning of "last"/"previous" and "next". Other phrases like for moving the cursor, and snapping windows also have directional phrases.

See also the second next conversation block.

core/windows_and_tabs/tabs.talon
```diff
@@ -1,10 +1,19 @@
...
 go tab <number>: user.tab_jump(number)
 go tab final: user.tab_final()
-tab (duplicate | clone): user.tab_duplicate()
+
+tab focus <user.text>: user.tab_switcher_focus(text)
+tab focus$: user.tab_switcher_menu()
```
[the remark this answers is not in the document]
Enyium: I'd summarize the current uses of "hunt" as being related to either text searches (which includes symbol hunts in VS Code), or to searches for items out in the world or the periphery, i.e., not yet brought into your workspace, e.g., "file hunt", and web searches.

Neither of those two commands, which are derived from the respective window-related commands, do interactive filtering. The one capturing text immediately switches to the first tab matching the text (just like the window-related phrase); the one not capturing text simply shows *all* tabs, just like Windows' Alt+Tab dialog shows all windows (with rare exceptions of certain kinds of dialogs).

You could *additionally* have "tab hunt", which could replace the existing "tab search" that in fact is interactive (at least in Firefox), currently implemented for Chrome and Firefox.

Note that, as I wrote in #1840, I'd actually prefer "switch" to "focus" (but not introduced by this PR, of course).

core/windows_and_tabs/tabs.talon
Enyium: (You seem to only refer to the last code line with `tab focus last`.)

If `tab focus <user.text>` and `tab focus$` stay (see misunderstanding above), it would only be consistent with the window-related phrases to not deviate from them regarding *this* phrase.

"last" is ambiguous (temporal vs. spacial meaning), and inconsistently used in repo. It should actually be standardized - to the temporal meaning, in my opinion. This would mean deprecating "tab last", or making it an alias of "tab focus last".

One could provide a simple implementation for some app where pressing Ctrl+Tab always switches to the last active tab, and not just to the tab to the right of the active one. That'd be a start. It could be Notepad++, but it has a setting to change it, whose default value I don't yet know.

core/windows_and_tabs/tabs.talon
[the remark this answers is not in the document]
nriley: Until dynamic list support gets released (which makes such commands work great; I've built several of them now) I'm not sure what benefit a *non*-searching `<user.text>` capture has, other than potentially confusing users. Since there was no implementation in your PR it was hard to tell what it did.

core/windows_and_tabs/tabs.talon
nriley: Absolutely — "last" is extra confusing in whether it means "item selected before the current item" (temporal) or "item before the current item" or "final item" (spatial).

One benefit is that it's quick to say when compared with some of the other options (e.g. "previous"), and it's *pretty* consistently indicating "item before the currently item" in community. Breaking the speech patterns of everyone who uses community is not something we do lightly.

I just thought of `tab most recent`, which is comparatively unambiguous. Rango uses `tab back` (https://rango.click/#/?id=focus-tabs) and also supports `tab hunt`. As we build out more sophisticated tab commands we could consider migrating some of the actions from Rango into community to avoid conflicts as there have been between Cursorless and community.

PR #1837, opened 2025-04-29
Added "talon copy win class name" phrase.

Enyium:
Analogous to "talon copy title".

### DIFF
--- a/plugin/talon_helpers/talon_helpers.talon
+++ b/plugin/talon_helpers/talon_helpers.talon
@@ -18,6 +18,9 @@ talon copy bundle:
 talon copy title:
     title = win.title()
     clip.set_text(title)
+talon copy class:
+    class_name = user.talon_get_active_window_class_name()
+    clip.set_text(class_name)
 talon dump version:
     result = user.talon_version_info()
     print(result)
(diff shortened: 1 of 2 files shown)


plugin/talon_helpers/talon_helpers.talon
```diff
@@ -15,9 +15,12 @@ talon copy executable:
 talon copy bundle:
     bundle = app.bundle()
     clip.set_text(bundle)
-talon copy title:
+talon copy [win] title:
     title = win.title()
     clip.set_text(title)
+talon copy [win] class [name]:
+    class_name = user.talon_get_active_window_class_name()
+    clip.set_text(class_name)
```
nriley, line 23:
```suggestion
talon copy title:
    title = win.title()
    clip.set_text(title)
talon copy class:
    class_name = user.talon_get_active_window_class_name()
    clip.set_text(class_name)
```
From the community backlog session — we did not think it was necessary to add a bunch of optional ways to say these things to the global command namespace.

nriley (changes requested):
Definitely something I would use — thanks for the contribution!

plugin/talon_helpers/talon_helpers.talon
Enyium: Technically, it's not the class, but the class name, which is why I also provided the "...class name" variant. But I reduced it to "talon copy class" now.

nriley (approved):

PR #1846, opened 2025-05-12
Commented out EditActions that don't work with the default iterm2 settings

caroby:
Now it defaults to ctrl-a and ctrl-e in readline.py which always works.

### DIFF
--- a/apps/iterm/iterm.py
+++ b/apps/iterm/iterm.py
@@ -15,13 +15,13 @@ directories_to_remap = {}
 directories_to_exclude = {}
 
 
-@ctx.action_class("edit")
-class EditActions:
-    def line_start():
-        actions.key("home")
+# @ctx.action_class("edit")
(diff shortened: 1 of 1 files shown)


BlueDrink9:
I know iterm is clever, but I would never have expected the original to work. Strange. That sort of thing is handled at the shell level, not the terminal level. I think it would be less confusing to delete them rather than comment them though.

This probably will need chiming in from at least one other person that uses iterm. What's your shell?

AndreasArvidsson:
I don't use Iterm, but we should definitely remove them and not just leave them commented if we don't want this functionality.

caroby:
I deleted them instead. Without this fix, line_start() and line_end() do not work on iTerm with the out-of-the-box default settings on mac.

BlueDrink9 (approved):

caroby:
how do I get this approved?

PR #1860, opened 2025-05-20
Start migrating stata to snippets

FireChickenProductivity:
There is probably something better we can do with the for loop, but I will leave that to someone more familiar with the language.

### DIFF
--- a/core/snippets/snippets/genericForLoopStatement.snippet
+++ b/core/snippets/snippets/genericForLoopStatement.snippet
@@ -1,0 +1,10 @@
+name: genericForLoopStatement
+phrase: for
+insertionScope: statement
+---
+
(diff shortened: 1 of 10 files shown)


core/snippets/snippets/genericForLoopStatement.snippet
```diff
@@ -0,0 +1,10 @@
+name: genericForLoopStatement
+phrase: for
+insertionScope: statement
+---
+
+language: stata
+-
+forval $1 {
```
AndreasArvidsson, line 8: Looks like forval is a ranged loop. This should probably just be moved to the other for snippet

AndreasArvidsson (approved):

PR #1868, opened 2025-05-22
Start migrating scala to use snippets

FireChickenProductivity:
I am not used to languages with this kind of paradigm, so I would prefer if someone familiar with the language would review this.

### DIFF
--- a/core/snippets/snippets/switchStatement.snippet
+++ b/core/snippets/snippets/switchStatement.snippet
@@ -15,3 +15,10 @@ language: python
 match $1:
     $0
 ---
+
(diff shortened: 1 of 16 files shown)


core/snippets/snippets/switchStatement.snippet
```diff
@@ -15,3 +15,10 @@ language: python
 match $1:
     $0
 ---
+
+language: scala
+-
+match {
```
AndreasArvidsson, line 21:
```suggestion
$1 match {
```

nriley (approved):

PR #1880, opened 2025-05-24
Start migrating elixir to snippets

FireChickenProductivity:
There does not seem to be an else if statement in the language, so I did not try to migrate that action. 
I was not sure what to do about the fact that a switch statement is called a case statement in the language, so changing the code state case action to do what that action usually does would not reflect what things are named in this language. 
There is an issue with actions conflicting with commands with the elixir setup, so I migrated the actions but left the commands alone. 
I followed the structure in the existing talon file of explicitly defining state commands. 
I am not familiar enough with the language to know how to properly handle the import actions, which are currently unusable anyways because they do not correspond to action definitions.

### DIFF
--- a/core/snippets/snippets/genericForLoopStatement.snippet
+++ b/core/snippets/snippets/genericForLoopStatement.snippet
@@ -1,0 +1,11 @@
+name: genericForLoopStatement
+phrase: for
+insertionScope: statement
+---
+
+language: elixir
+-
+for $1 do
+	$0
+end
+---
--- a/lang/elixir/elixir.talon
+++ b/lang/elixir/elixir.talon
@@ -31,8 +31,6 @@ state try: "try do"
 state rescue: "rescue"
 state after: "after"
 state end: "end"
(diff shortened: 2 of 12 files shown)


core/snippets/snippets/genericForLoopStatement.snippet
```diff
@@ -0,0 +1,11 @@
+name: genericForLoopStatement
```
AndreasArvidsson, line 1: Not sure why this isn't just a for loop snippet. Do we really need a new generic for loop snippet?

lang/elixir/elixir.talon
```diff
@@ -31,6 +31,8 @@ state try: "try do"
 state rescue: "rescue"
 state after: "after"
 state end: "end"
+state while: user.insert_snippet_by_name("whileLoopStatement")
+state for: user.insert_snippet_by_name("genericForLoopStatement")
```
AndreasArvidsson, line 35:
```suggestion
```

AndreasArvidsson (approved):

PR #1885, opened 2025-05-26
Refactor edit commands and create readline for vi mode

BlueDrink9:
Generally speaking the refactor allows you to override how edit actions are applied, which I uses to make sure that the modifier/selection applies by creating a vim motion instead of trying to select terminal text. The refactors to the edit commands are fairly minimal, it's just converting the functions into user actions and refactoring out actions to access the internal dictionaries.

In draft mode while I dog food it for a week or so to ensure it is ready

### DIFF
--- a/core/edit/edit_command_actions.py
+++ b/core/edit/edit_command_actions.py
@@ -98,7 +98,7 @@ class Actions:
         action_type = action.type
 
         if action_type in simple_action_callbacks:
-            callback = simple_action_callbacks[action_type]
+            callback = actions.user.get_simple_action_callback(action_type)
             callback()
             return
 
@@ -117,3 +117,8 @@ class Actions:
 
             case _:
                 raise ValueError(f"Unknown edit action: {action_type}")
+
+    def get_simple_action_callback(action_type: str):
+        """Convert a edit action type created from a string into its associated Callback"""
+        assert action_type in simple_action_callbacks
+        return simple_action_callbacks[action_type]
--- a/core/edit/edit_command_actions.py
+++ b/core/edit/edit_command_actions.py
@@ -88,26 +88,32 @@ simple_action_callbacks: dict[str, Callable] = {
 }
 
 
-def run_action_callback(action: EditAction):
-    action_type = action.type
+@mod.action_class
+class Actions:
+    def run_action_callback(action: EditAction):
+        """
+        Run a callback that applies an edit action to the selected
+        Intended for internal use and overwriting
+        """
+        action_type = action.type
 
-    if action_type in simple_action_callbacks:
-        callback = simple_action_callbacks[action_type]
-        callback()
-        return
+        if action_type in simple_action_callbacks:
+            callback = simple_action_callbacks[action_type]
+            callback()
+            return
 
-    match action_type:
-        case "insert":
-            assert isinstance(action, EditInsertAction)
-            actions.insert(action.text)
+        match action_type:
+            case "insert":
+                assert isinstance(action, EditInsertAction)
+                actions.insert(action.text)
 
-        case "wrapWithDelimiterPair":
-            assert isinstance(action, EditWrapAction)
-            return lambda: actions.user.delimiter_pair_wrap_selection(action.pair)
+            case "wrapWithDelimiterPair":
+                assert isinstance(action, EditWrapAction)
+                return lambda: actions.user.delimiter_pair_wrap_selection(action.pair)
 
-        case "applyFormatter":
-            assert isinstance(action, EditFormatAction)
-            actions.user.formatters_reformat_selection(action.formatters)
+            case "applyFormatter":
+                assert isinstance(action, EditFormatAction)
+                actions.user.formatters_reformat_selection(action.formatters)
 
-        case _:
-            raise ValueError(f"Unknown edit action: {action_type}")
+            case _:
+                raise ValueError(f"Unknown edit action: {action_type}")
--- a/core/edit/edit_command_modifiers.py
+++ b/core/edit/edit_command_modifiers.py
@@ -71,11 +71,14 @@ class Actions:
         Run a callback that selects or prepares text ready to apply and edit action.
         Intended for internal use and overwriting
         """
+        count = modifier.count
+        modifier = actions.user.get_modifier_callback(EditModifier)
+        for i in range(1, count + 1):
+            modifier.callback()
+
+    def get_modifier_callback(modifier: EditModifier):
+        """Convert an edit modifier created from a string into its associated EditModifierCallback"""
         modifier_type = modifier.type
         if modifier_type not in modifier_dictionary:
             raise ValueError(f"Unknown edit modifier: {modifier_type}")
-
-        count = modifier.count
-        modifier = modifier_dictionary[modifier_type]
-        for i in range(1, count + 1):
-            modifier.callback()
+        return modifier_dictionary[modifier_type]
--- a/core/edit/edit_command_modifiers.py
+++ b/core/edit/edit_command_modifiers.py
@@ -64,12 +64,18 @@ modifier_dictionary: dict[str, EditModifierCallback] = {
 }
 
 
-def run_modifier_callback(modifier: EditModifier):
-    modifier_type = modifier.type
-    if modifier_type not in modifier_dictionary:
-        raise ValueError(f"Unknown edit modifier: {modifier_type}")
+@mod.action_class
+class Actions:
+    def run_modifier_callback(modifier: EditModifier):
+        """
+        Run a callback that selects or prepares text ready to apply and edit action.
+        Intended for internal use and overwriting
+        """
+        modifier_type = modifier.type
+        if modifier_type not in modifier_dictionary:
+            raise ValueError(f"Unknown edit modifier: {modifier_type}")
 
-    count = modifier.count
-    modifier = modifier_dictionary[modifier_type]
-    for i in range(1, count + 1):
-        modifier.callback()
+        count = modifier.count
+        modifier = modifier_dictionary[modifier_type]
+        for i in range(1, count + 1):
+            modifier.callback()
--- a/core/edit/edit_command_modifiers.py
+++ b/core/edit/edit_command_modifiers.py
@@ -72,7 +72,7 @@ class Actions:
         Intended for internal use and overwriting
         """
         count = modifier.count
-        modifier = actions.user.get_modifier_callback(EditModifier)
+        modifier = actions.user.get_modifier_callback(modifier)
         for i in range(1, count + 1):
             modifier.callback()
 
--- a/tags/terminal/readline_vi.py
+++ b/tags/terminal/readline_vi.py
@@ -9,8 +9,9 @@ ctx.matches = """
 tag: user.readline_vi
 """
 
+
 @dataclass
-class PendingSelection():
+class PendingSelection:
     motion: str
     endAction: str
     count: int = 1
@@ -20,15 +21,16 @@ pendingSelection: PendingSelection | None = None
 
 
 edit_action_vi_keys: dict[str, str] = {
-        "cut": "c",
-        "delete": "c",
-        "goBefore": "",
-        "goAfter": "",
-        # We can't actually use the clipboard for these operations
-        "copyToClipboard": "y",
-        "cutToClipboard": "c",
-        "pasteFromClipboard": "y",
+    "cut": "c",
+    "delete": "c",
+    "goBefore": "",
+    "goAfter": "",
+    # We can't actually use the clipboard for these operations
+    "copyToClipboard": "y",
+    "cutToClipboard": "c",
+    "pasteFromClipboard": "y",
 }
+
 
 def normal_cmd(keys):
     actions.key("escape")
@@ -87,10 +89,8 @@ class EditActions:
         pendingSelection = PendingSelection("w", "i")
 
 
-
 @ctx.action_class("user")
 class Actions:
-
     def run_action_callback(action):
         """
         Run a callback that applies an edit action to the selected
--- a/tags/terminal/readline_vi.py
+++ b/tags/terminal/readline_vi.py
@@ -20,14 +20,15 @@ class PendingSelection:
 pendingSelection: PendingSelection | None = None
 
 
+# using d instead of c for removal so that we remain in normal mode and the character sent to exit normal mode doesn't get inserted
 edit_action_vi_keys: dict[str, str] = {
-    "cut": "c",
-    "delete": "c",
+    "cut": "d",
+    "delete": "d",
     "goBefore": "",
     "goAfter": "",
     # We can't actually use the clipboard for these operations
     "copyToClipboard": "y",
-    "cutToClipboard": "c",
+    "cutToClipboard": "d",
     "pasteFromClipboard": "y",
 }
 
@@ -37,6 +38,14 @@ def normal_cmd(keys):
     # sleep to avoid interpreting as an escape sequence
     sleep(0.2)
     actions.insert(keys)
+
+
+def add_pending(motion, end):
+    global pendingSelection
+    if pendingSelection and pendingSelection.motion == motion:
+        pendingSelection.count += 1
+    else:
+        pendingSelection = PendingSelection(motion, end)
 
 
 # This operates on a paradigm of staying in insert mode, but at least allows standard community text editing commands if the user has set their shell to vi mode in read line.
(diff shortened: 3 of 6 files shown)


BlueDrink9:
I haven't noticed any issues with this while dogfooding it for the last two weeks, including both inside and outside of terminal, so it looks like I haven't broken the regular commands either.

core/edit/edit_command_actions.py
```diff
@@ -88,26 +88,37 @@ def edit_action(m) -> EditAction:
...
-
-        case "applyFormatter":
-            assert isinstance(action, EditFormatAction)
-            actions.user.formatters_reformat_selection(action.formatters)
-
-        case _:
-            raise ValueError(f"Unknown edit action: {action_type}")
+@mod.action_class
+class Actions:
+    def run_action_callback(action: EditAction):
```
AndreasArvidsson, line 93:
```suggestion
    def run_action_callback(action: EditAction) -> Callable :
```

core/edit/edit_command_actions.py
```diff
@@ -88,26 +88,37 @@ def edit_action(m) -> EditAction:
...
+                return lambda: actions.user.delimiter_pair_wrap_selection(action.pair)
+
+            case "applyFormatter":
+                assert isinstance(action, EditFormatAction)
+                actions.user.formatters_reformat_selection(action.formatters)
+
+            case _:
+                raise ValueError(f"Unknown edit action: {action_type}")
+
+    def get_simple_action_callback(action_type: str):
```
AndreasArvidsson:
```suggestion
    def get_simple_action_callback(action_type: str) -> Callable | None:
```

core/edit/edit_command_actions.py
```diff
@@ -88,26 +88,37 @@ def edit_action(m) -> EditAction:
...
+@mod.action_class
+class Actions:
+    def run_action_callback(action: EditAction):
+        """
+        Run a callback that applies an edit action to the selected
+        Intended for internal use and overwriting
+        """
+        action_type = action.type
+
+        if action_type in simple_action_callbacks:
```
AndreasArvidsson: This if statements means that you can't add more callbacks to a context implementation of `get_simple_action_callback()`. 
`get_simple_action_callback()` could return None to indicate that there is no simple action call back for this action type.

core/edit/edit_command_modifiers.py
```diff
@@ -64,12 +64,21 @@ def edit_modifier(m) -> EditModifier:
...
+class Actions:
+    def run_modifier_callback(modifier: EditModifier):
+        """
+        Run a callback that selects or prepares text ready to apply and edit action.
+        Intended for internal use and overwriting
+        """
+        count = modifier.count
+        modifier = actions.user.get_modifier_callback(modifier)
+        for i in range(1, count + 1):
+            modifier.callback()
```
AndreasArvidsson: May be better to just have `get_modifier_callback()` return a callable? Otherwise any context specific implementation of this needs to construct the dataclass.

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,148 @@
...
+    def extend_word_left():
+        add_pending("b", "i")
+
+    def extend_word_right():
+        add_pending("w", "i")
+
+
+@ctx.action_class("user")
+class Actions:
+    def run_action_callback(action):
```
AndreasArvidsson: This doesn't give you an error in the Talon log since you are omitting the argument type?

BlueDrink9: no, it takes it from the module version's signature presumably

core/edit/edit_command_actions.py
BlueDrink9: That's a good catch, I have altered it. I don't think there is any way to call the original mod action function once you have created a context override, correct? So while your modification would let someone entirely replace the callbacks, it would not let them extend it. Any ideas for how to do this? It will actually have an immediate benefit, because I will be able to map ctrl+w to clear word left, which will be faster and cleaner than the current implementation of having to press escape

AndreasArvidsson: You can call `actions.next()` to call that next most specific implementation.

core/edit/edit_command_actions.py
```diff
@@ -88,26 +88,39 @@ def edit_action(m) -> EditAction:
...
+
+            case _:
+                raise ValueError(f"Unknown edit action: {action_type}")
+
+    def get_simple_action_callback(action_type: str) -> Callable | None:
+        """Convert a edit action type created from a string into its associated Callback"""
+        if action_type in simple_action_callbacks:
+            return simple_action_callbacks[action_type]
+        else:
+            return None
```
auscompgeek:
```suggestion
        return simple_action_callbacks.get(action_type)
```

core/edit/edit_command_actions.py
```diff
@@ -88,26 +88,36 @@ def edit_action(m) -> EditAction:
...
+            return
+
+        match action_type:
+            case "insert":
+                assert isinstance(action, EditInsertAction)
+                actions.insert(action.text)
+
+            case "wrapWithDelimiterPair":
+                assert isinstance(action, EditWrapAction)
+                return lambda: actions.user.delimiter_pair_wrap_selection(action.pair)
```
nriley: This was not your code, but it looks like there is an error here in which a lambda is returned rather than just running the action. Could you fix this as part of your PR?

BlueDrink9:
Do we know how many vi testers we will have? If the concern is testing that nothing breaks outside of this mode, could we just have a few community maintainers use or merge this fork for themselves for a few weeks?

That bug was unfortunate, but I caught it fairly soon after the first two weeks --- I just didn't have time to fix it for a while. Otherwise I have been using this code consistently for several months and even outside of terminal, I haven't noticed any further issues. (At one point I thought I had a broken "clear line left" until I discovered that actually that was just not the correct command)

core/edit/edit_command_actions.py
BlueDrink9: Fixed, and also removed the reliance on string literals for the match

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,186 @@
...
+    # "copyToClipboard": "y",
+    # "cutToClipboard": "d",
+    # "pasteFromClipboard": "y",
+}
+
+
+def normal_cmd(keys):
+    actions.key("escape")
+    # sleep to avoid interpreting as an escape sequence
+    sleep(0.2)
```
auscompgeek, line 40: Is there a reason for using `time.sleep` instead of `actions.sleep`?

BlueDrink9: oh yeah that's a mistake, thanks

Arrow2718:
## kitty

user and edit actions added by BlueDrinks  PR #1885
Tested on Endeavour OS KDE Plasma

| command word             | method/actions             |                           |
| ------------------------ | -------------------------- | ------------------------- |
| **we delete line**       | *edit.delete_line()*       | works                     |
| **we word left**         | *edit.word_left()*         | works                     |
| **we word right**        | *edit.word_right()*        | works                     |
| **we line end**          | *edit.line_end()*          | works                     |
| **we line start**        | *edit.line_start()*        | works                     |
| **we undo**              | *edit.undo()*              | works                     |
| **we redo**              | *edit.redo()*              | triggers reverse-i-search |
| **we extend line end**   | *edit.extend_line_end()*   | ???                       |
| **we extend line start** | *edit.extend_line_start()* | ???                       |
| **wee extend word left** | *edit.extend_word_left()*  | ???                       |
| **we extend word right** | *edit.extend_word_right()* | ???                       |
| **we cut line**          | *user.cut_line()*          | works                     |
| **we cut word left**     | *user.cut_word_left()*     | doesn't go to clipboard   |
| **we copy word left**    | *user.copy_word_left()*    | doesn't go to clipboard   |
| **we copy word right**   | *user.copy_word_right()*   | doesn't go to clipboard   |

Arrow2718:
I'll post these as comments in the code if required.

BlueDrink9:
extend doesn't make sense in a terminal/bash, since there's no no selection. I think redo doesn't have an action mapped by default if ctrl+r goes to rev search… but I'm not sure that's default, have you customised your shell at all?

cut and copy use the vi clipboard rather than system, no way around that. I'm not sure default readline bindings have a way to do it either; And regardless, I think you would prefer to have, cut and copy and paste all operate on the terminal selection rather than read line (i.e so you can copy text out of your terminal to the system clipboard)

nriley:
> extend doesn't make sense in a terminal/bash, since there's no no selection. I think redo doesn't have an action mapped by default if ctrl+r goes to rev search… but I'm not sure that's default, have you customised your shell at all?

```
$ bind -m vi-command -P | grep redo
vi-redo can be found on ".".
$ bind -m vi-insert -P | grep redo
vi-redo is not bound to any keys
```

Looks like it is only bound in command mode in bash by default. Undo is not bound in zsh's insert mode at all.

> cut and copy use the vi clipboard rather than system, no way around that. I'm not sure default readline bindings have a way to do it either; And regardless, I think you would prefer to have, cut and copy and paste all operate on the terminal selection rather than read line (i.e so you can copy text out of your terminal to the system clipboard)

Some shells do indeed support a selection, but I agree it makes sense to use the system clipboard.

core/edit/edit_command_actions.py
nriley: Can you add the word edit into the name of this action somewhere? It is a little too generic otherwise.

core/edit/edit_command_actions.py
```diff
@@ -88,26 +88,37 @@ def edit_action(m) -> EditAction:
...
+                actions.user.delimiter_pair_wrap_selection(action.pair)
+
+            case EditFormatAction.type:
+                assert isinstance(action, EditFormatAction)
+                actions.user.formatters_reformat_selection(action.formatters)
+
+            case _:
+                raise ValueError(f"Unknown edit action: {action_type}")
+
+    def get_simple_action_callback(action_type: str) -> Callable | None:
```
nriley: Please add edit to this one too.

core/edit/edit_command_modifiers.py
```diff
@@ -64,12 +64,21 @@ def edit_modifier(m) -> EditModifier:
...
-
-    count = modifier.count
-    modifier = modifier_dictionary[modifier_type]
-    for i in range(1, count + 1):
-        modifier.callback()
+@mod.action_class
+class Actions:
+    def run_modifier_callback(modifier: EditModifier):
+        """
+        Run a callback that selects or prepares text ready to apply and edit action.
```
nriley:
```suggestion
        Run a callback that selects or prepares text ready to apply an edit action.
```
Please add edit to this one as well.

core/edit/edit_command_modifiers.py
```diff
@@ -64,12 +64,21 @@ def edit_modifier(m) -> EditModifier:
...
+        Intended for internal use and overwriting
+        """
+        count = modifier.count
+        modifier_callback = actions.user.get_modifier_callback(modifier)
+        for i in range(1, count + 1):
+            modifier_callback()
+
+    def get_modifier_callback(modifier: EditModifier):
```
nriley: … And again.

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def line_end():
+        normal_cmd("A")
+
+    def line_start():
+        normal_cmd("I")
+
+    def undo():
+        # Technically control underscore works in vi readline mode as well, but this also works in zsh
+        normal_cmd("ua")
```
nriley: Particularly in zsh which does not coalesce single character deletion, so you will end up needing to undo multiple times, this causes unwanted cursor movement. Think that "ui" might work better.

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,187 @@
...
+    if pendingSelection and pendingSelection.motion == motion:
+        pendingSelection.count += 1
+    else:
+        pendingSelection = PendingSelection(motion, end)
+
+
+def delete_word_right():
+    # need a custom function, because otherwise the right key that moves the cursor will interfere with the pending action
+    actions.key("escape")
+    sleep(0.1)
```
nriley: Can you explain why you have a different sleep here? If there is some benefit to having different sleep values, potentially it would make sense to have a second argument to `normal_cmd` (assuming you also update it to be able to handle arbitrary key presses and not just insert)?

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,187 @@
...
+@ctx.action_class("edit")
+class EditActions:
+    def delete_line():
+        normal_cmd("cc")
+
+    def word_left():
+        normal_cmd("bi")
+
+    def word_right():
+        actions.key("escape")
```
nriley: Same thing here — if you have a good reason to have a different sleep value and not use `normal_cmd` please mention it!

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def line_start():
+        normal_cmd("I")
+
+    def undo():
+        # Technically control underscore works in vi readline mode as well, but this also works in zsh
+        normal_cmd("ua")
+
+    def redo():
+        actions.key("escape")
```
nriley: Would it make more sense to use `normal_cmd(".i")` here? . seems to be bound as redo both in readline/bash and zsh.

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@ctx.action_class("user")
+class Actions:
+    def run_action_callback(action):
+        """
+        Run a callback that applies an edit action to the selected
+        Intended for internal use and overwriting
+        """
+        global pendingSelection
```
nriley: `pending_selection`?

tags/terminal/readline_vi.py
```diff
@@ -0,0 +1,187 @@
...
+    # "cutToClipboard": "d",
+    # "pasteFromClipboard": "y",
+}
+
+
+def normal_cmd(keys):
+    actions.key("escape")
+    # sleep to avoid interpreting as an escape sequence
+    sleep(0.2)
+    actions.insert(keys)
```
nriley: Given that you have several places where you need to duplicate this logic, consider using `actions.key` here instead.

nriley (changes requested):

BlueDrink9:
> Looks like it is only bound in command mode in bash by default. Undo is not bound in zsh's insert mode at all.

vi-redo is not the same as a traditional redo, it's actually a repeat command. So no, there is no way to redo

tags/terminal/readline_vi.py
BlueDrink9: agree

tags/terminal/readline_vi.py
BlueDrink9: That's not redo in the sense that we mean it, it is repeat. readline doesn't have redo; however I still left this here so that in zsh (or other vi emulation) it does work

tags/terminal/readline_vi.py
BlueDrink9: The sleeps should probably be the same. Do the existing comments not explain why this can't just go through normal cmd?

tags/terminal/readline_vi.py
BlueDrink9: I am not at all clear what you are asking,  or what the benefits of actions . key is, Could you please clarify?

tags/terminal/readline_vi.py
BlueDrink9: Oh, I see I was actually wrong about it not needing to go through normal!

tags/terminal/readline_vi.py
BlueDrink9: Oh, now I think I understand. I didn't realize that keys could take multiple keys to insert

nriley:
> > Looks like it is only bound in command mode in bash by default. Undo is not bound in zsh's insert mode at all.
> 
> vi-redo is not the same as a traditional redo, it's actually a repeat command. So no, there is no way to redo

It does work as redo in zsh's vi command mode though.

BlueDrink9:
> > > Looks like it is only bound in command mode in bash by default. Undo is not bound in zsh's insert mode at all.
> > 
> > 
> > vi-redo is not the same as a traditional redo, it's actually a repeat command. So no, there is no way to redo
> 
> It does work as redo in zsh's vi command mode though.

 "ctrl-r" does zsh redo, which it is what we have currently. "." does zsh repeat

PR #1894, opened 2025-05-29
Migrate for loop actions to use snippets

FireChickenProductivity:
This migrates all the for loop actions not handled in a separate pull request. If we are going to migrate the previous state for command, I recommend having a separate python snippet for "for in" without the colon for more complex comprehensions. I have no idea what to call it though.


AndreasArvidsson (approved):

lang/go/go.py
```diff
@@ -134,7 +134,7 @@ def code_insert_is_not_null():
         actions.insert(" != nil")
 
     def code_state_for():
-        actions.user.insert_snippet_by_name("whileLoopStatement")
+        actions.user.insert_snippet_by_name("forLoopStatement")
```
AndreasArvidsson: All of these are now defined on the module so you can remove them for the specific languages

AndreasArvidsson (changes requested):
Remove superfluous context actions.

AndreasArvidsson (approved):

PR #1897, opened 2025-05-30
Migrate more javascript to use snippets

FireChickenProductivity:


FireChickenProductivity:
We decided to use a for each snippet inserting an actual for each statement instead of the .forEach method. We already have the "dot for each" command for inserting the .forEach method.

Issue #1562, opened 2024-09-10
Help Search Command


Mark-Phillipson:
Sometimes when using the help search command the word I'm searching for does not get recognized.

As a workaround I would suggest a command like the following:

```talon
help search clipboard:
    text = clip.text()
    user.help_search(text)
```
This would enable us to put the word in the clipboard first then there would be no confusion as to what to search for.

PR #1902, opened 2025-05-30
Add clipboard help search command

FireChickenProductivity:
Do we want to address this issue https://github.com/talonhub/community/issues/1562? If so, here is an implementation. Other commands, in particular the screenshot ones, use clip to refer to the clipboard, so I used that here. I put the word clip before the word search in case someone wanted to do a help search on the word clip itself.

### DIFF
--- a/apps/vscode/vscode.py
+++ b/apps/vscode/vscode.py
@@ -15,6 +15,8 @@ and app.bundle: com.microsoft.VSCodeInsiders
 os: mac
 and app.bundle: com.vscodium
 os: mac
+and app.bundle: co.posit.positron
+os: mac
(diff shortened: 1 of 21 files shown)


FireChickenProductivity:
We agreed in the backlog session to discuss this later. It might be better, for instance, to allow searching based on the current selection instead of requiring the person to copy text first and then use a separate command to search based on the clipboard text. We also want to make sure the grammar is consistent.

AndreasArvidsson:
I think this command looks fine. Personally I would probably have done something like `"help search clip clip"` but I think the current command works as well. 
I would recommend doing two commands one for the clipboard content and one for the current selection because I think they're both useful. eg `"help this search"`?

Issue #1889, opened 2025-05-27
Allow snippets in unsupported editors to insert the tab character with a setting


FireChickenProductivity:
I believe that some people using snippets in unsupported editors are going to want them to press the tab key instead of spaces for indentation. I recommend we support this through a setting.

PR #1907, opened 2025-06-02
Support customizing indentation with raw text snippet insert

FireChickenProductivity:
Now that we are migrating existing functionality to snippets, I want to provide settings that can allow customizing how snippets are inserted without editor support to improve the experience for people using editors other than VSCode. As the first pull request towards this goal, I added a setting for deciding if tabs should be converted to spaces and a setting for determining the number of spaces that should be used per tab when converting tabs to spaces.

### DIFF
--- a/settings.talon
+++ b/settings.talon
@@ -78,7 +78,7 @@ settings():
     # .snippet files. Changing this setting requires a restart of Talon.
     # user.snippets_dir = "snippets"
 
-     # Set to the number of spaces to use for each tab when inserting snippets as raw text (without editor support). Set to -1 to insert tabs as tabs, such as in code editors that can expand tabs in pasted or typed text. This setting is provided for applications like web browsers and chat apps that do not understand code formatting.
+    # Set to the number of spaces to use for each tab when inserting snippets as raw text (without editor support). Set to -1 to insert tabs as tabs, such as in code editors that can expand tabs in pasted or typed text. This setting is provided for applications like web browsers and chat apps that do not understand code formatting.
     user.snippet_raw_text_spaces_per_tab = 4
 
     # Uncomment to insert text longer than 10 characters (customizable) by pasting from


FireChickenProductivity:
Closes #1889

core/snippets/snippets_insert_raw_text.py
```diff
@@ -27,9 +42,14 @@ def insert_snippet_raw_text(body: str):
...
 
+def compute_indentation_as_spaces():
+    return " " * settings.get("user.snippets_raw_text_spaces_per_tab")
+
+
 def parse_snippet(body: str):
     # Some IM services will send the message on a tab
-    body = re.sub(r"\t", INDENTATION, body)
+    if settings.get("user.snippets_raw_text_convert_indentation"):
+        body = re.sub(r"\t", compute_indentation_as_spaces(), body)
```
AndreasArvidsson: Have you verified that if you don't replace tabs with spaces the calculation further down of columns still work?

core/snippets/snippets_insert_raw_text.py
```diff
@@ -1,9 +1,24 @@
 import re
 from dataclasses import dataclass
 
-from talon import actions
+from talon import Module, actions, settings
+
+mod = Module()
+
+mod.setting(
```
AndreasArvidsson: You could probably turn this into a single setting. Just have it default to minus one.

FireChickenProductivity: I converted it to use a single setting.

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: Based on some manual testing with a print statement, calculating the column number for the stops seems to work fine without replacing tabs with spaces.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -1,9 +1,18 @@
...
-from talon import actions
+from talon import Module, actions, settings
+
+mod = Module()
+
+
+mod.setting(
+    "snippets_raw_text_spaces_per_tab",
+    type=int,
+    default=-1,
```
AndreasArvidsson:
```suggestion
    default=4,
```

core/snippets/snippets_insert_raw_text.py
```diff
@@ -28,9 +37,14 @@ def insert_snippet_raw_text(body: str):
         left(stop.columns_left)
 
 
+def compute_indentation_as_spaces():
+    return " " * settings.get("user.snippets_raw_text_spaces_per_tab")
```
AndreasArvidsson: How about doing a replace tabs function instead? I'm not a big fan of reading the setting twice.

```suggestion
def format_tabs(text: str) -> str:
    """Possibly replaces tabs with spaces in the given text."""
    spaces_per_tab: int = settings.get("user.snippet_raw_text_spaces_per_tab")
    if spaces_per_tab < 0:
        return text
    return re.sub(r"\t", " " * spaces_per_tab, text)
```

core/snippets/snippets_insert_raw_text.py
```diff
@@ -28,9 +37,14 @@ def insert_snippet_raw_text(body: str):
...
+
+
 def parse_snippet(body: str):
     # Some IM services will send the message on a tab
-    body = re.sub(r"\t", INDENTATION, body)
+    if settings.get("user.snippets_raw_text_spaces_per_tab") >= 0:
+        body = re.sub(r"\t", compute_indentation_as_spaces(), body)
```
AndreasArvidsson:
```suggestion
    body = format_tabs(body)
```

core/snippets/snippets_insert_raw_text.py
```diff
@@ -1,9 +1,18 @@
...
 from dataclasses import dataclass
 
-from talon import actions
+from talon import Module, actions, settings
+
+mod = Module()
+
+
+mod.setting(
+    "snippets_raw_text_spaces_per_tab",
```
AndreasArvidsson:
```suggestion
    "snippet_raw_text_spaces_per_tab",
```

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: Because it does not always replace the tabs, I would prefer to give this a different name like format_tabs. Otherwise, I agree with extracting that logic into a separate function to keep the main function simple.

AndreasArvidsson: Sounds good

AndreasArvidsson (approved):

nriley:
From the community backlog session — please explain a little bit about the fact that many/most code editors will automatically convert tabs to spaces when you paste, so this is really only necessary when you are inserting snippets into contexts that generally don't understand code formatting, such as web browsers or chat sessions. Also, please add the setting to settings.talon.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -5,14 +5,21 @@
 
 mod = Module()
 
+
+mod.setting(
+    "snippet_raw_text_spaces_per_tab",
+    type=int,
+    default=4,
+    desc="""The number of spaces to use for each tab in snippets inserted as raw text when converting tabs to spaces. A negative value prevents tabs from getting converted to spaces. This should usually be negative for actual code editors because they can usually handle tabs correctly. This setting is provided applications like web browsers and chat apps.""",
```
nriley:
```suggestion
    desc="""The number of spaces per tab when inserting snippets as raw text. Set to -1 to insert tabs as tabs, such as in code editors that can expand tabs in pasted or typed text. This setting is provided for applications like web browsers and chat apps.""",
```

core/snippets/snippets_insert_raw_text.py
```diff
@@ -40,9 +47,17 @@ def insert_snippet_raw_text(body: str):
         left(stop.columns_left)
 
 
+def format_tabs(text: str) -> str:
+    """Possibly replaces tabs with spaces in the given text."""
+    spaces_per_tab: int = settings.get("user.snippet_raw_text_spaces_per_tab")
+    if spaces_per_tab < 0:
+        return text
+    return re.sub(r"\t", " " * spaces_per_tab, text)
```
nriley, line 55: There's no need to use a regular expression here to replace a single character when `str.replace` will do.

settings.talon
```diff
@@ -78,6 +78,9 @@ settings():
     # .snippet files. Changing this setting requires a restart of Talon.
     # user.snippets_dir = "snippets"
 
+    # Set the number of spaces to convert each tab to for inserting snippets without editor support. A negative number means that tabs will not be converted to spaces. This setting should usually be set to a negative number for actual code editors because they can usually format tabs correctly. This setting is available for contexts like web browsers and chat applications that do not understand code formatting.
```
```suggestion
     # Set to the number of spaces to use for each tab when inserting snippets as raw text (without editor support). Set to -1 to insert tabs as tabs, such as in code editors that can expand tabs in pasted or typed text. This setting is provided for applications like web browsers and chat apps that do not understand code formatting.
```

nriley (changes requested):

nriley (approved):

PR #1909, opened 2025-06-02
Have raw snippet insert reach the first stop from the end of the line

FireChickenProductivity:
The raw snippet insert currently reaches the first stop by moving to the correct column, going to the beginning of the line, and pressing right. This can fail when there is indentation at the start of the line by putting the cursor in the wrong position. This pull request addresses this by moving the cursor to the end of the line and then moving left the required number of times.

### DIFF
--- a/core/snippets/snippets_insert_raw_text.py
+++ b/core/snippets/snippets_insert_raw_text.py
@@ -17,14 +17,15 @@ class Stop:
 
 def insert_snippet_raw_text(body: str):
     """Insert snippet as raw text without editor support"""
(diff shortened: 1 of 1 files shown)


AndreasArvidsson:
I would just update the stop dataclass and add  `col_left` since we already have `row_up`.

FireChickenProductivity:
> I would just update the stop dataclass and add `col_left` since we already have `row_up`.

Good idea. I have that implemented now.

AndreasArvidsson (approved):

PR #1910, opened 2025-06-02
Add snip next without editor support

FireChickenProductivity:
This offers snip next without editor support assuming that the user has not moved the cursor. This will probably create merge conflicts with the rest of my snippet system pull requests.


FireChickenProductivity:
I would make this a draft again if I could. Is not going to work until the columns left field gets fixed.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -16,9 +16,59 @@ class Stop:
...
+
+def go_to_next_stop():
+    """Goes to the next snippet stop if it exists"""
+    global stop_stack
+
+    if stop_stack:
+        current_stop = stop_stack.pop()
+        next_stop = stop_stack[-1]
+        if current_stop.row == next_stop.row:
+            move_to_correct_column(current_stop.col, next_stop.col)
```
AndreasArvidsson: Doesn't this assume that the user hasn't typed anything into the previous stop? I would always go to the start or the end of the line and find the column from there.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -16,9 +16,59 @@ class Stop:
...
+stop_stack: list[Stop] | None = None
+
+
+def update_stop_information(stops: list[Stop]):
+    global stop_stack
+    if len(stops) > 1:
+        stop_stack = stops
+        stop_stack.reverse()
+    else:
+        stop_stack = None
```
AndreasArvidsson: Is there reason you can't just assign an empty list?

core/snippets/snippets_insert_raw_text.py
```diff
@@ -16,9 +16,59 @@ class Stop:
...
+
+
+def go_to_next_stop():
+    """Goes to the next snippet stop if it exists"""
+    global stop_stack
+
+    if stop_stack:
+        current_stop = stop_stack.pop()
+        next_stop = stop_stack[-1]
+        if current_stop.row == next_stop.row:
```
AndreasArvidsson: next_stop might be None here

core/snippets/snippets_insert_raw_text.py
```diff
@@ -16,9 +16,59 @@ class Stop:
     col: int
 
 
+stop_stack: list[Stop] | None = None
+
+
+def update_stop_information(stops: list[Stop]):
```
AndreasArvidsson: I would move all these help/private functions below insert_snippet_raw_text

core/snippets/snippets_insert.py
```diff
@@ -3,7 +3,7 @@
 from talon import Module, actions
 
 from .snippet_types import Snippet
-from .snippets_insert_raw_text import insert_snippet_raw_text
+from .snippets_insert_raw_text import go_to_next_stop, insert_snippet_raw_text
```
AndreasArvidsson: `go_to_next_stop_raw` to be consistent

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: Good idea. I updated it to always move the cursor to the column position from the end of the line.

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: I decided to refactor the code to always have the stop stack be a list to simplify the logic.

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: I believe I avoided that issue, but I refactored the code to make it easier to read.

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: I moved the helper functions below insert_snippet_raw_text to improve organization.

core/snippets/snippets_insert.py
FireChickenProductivity: I agree that helps with readability and made the change.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -99,7 +188,7 @@ def left(n: int):
 
 def key(stop: Stop):
```
nriley: I realize that you didn't write this code originally — this is confusing both in its naming and that it is not a method on `Stop`.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -36,8 +58,75 @@ def insert_snippet_raw_text(body: str):
...
+    for line in lines:
+        # Lines with only one stop are always in order
+        if len(line) > 1:
+            stop = line[0]
+            stop_key = key(stop)
+            for next_stop in line[1:]:
+                next_key = key(next_stop)
+                # If the ordering between the keys and columns are inconsistent,
+                # the stops on this line go from right to left
+                if next_key < stop_key != stop.col < next_stop.col:
```
nriley: Please add some parentheses to this to match the operator precedence.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -36,8 +58,75 @@ def insert_snippet_raw_text(body: str):
...
+                    return True
+                stop_key = next_key
+                stop = next_stop
+    return False
+
+
+def move_to_correct_column(stop: Stop):
+    actions.edit.line_end()
+    left(stop.columns_left)
```
nriley: This function name no verb.

nriley (commented):
Looks good aside from some naming issues. Will leave this to @AndreasArvidsson for a final decision.

core/snippets/snippets_insert_raw_text.py
AndreasArvidsson: I agree that this should be moved into the stop class.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -32,9 +34,29 @@ class Stop:
...
+stop_stack: list[Stop] = []
+
+
+def go_to_next_stop_raw():
+    """Goes to the next snippet stop if it exists"""
+    global stop_stack
+    if len(stop_stack) > 1:
+        current_stop = stop_stack.pop()
+        next_stop = stop_stack[-1]
+        if current_stop.row != next_stop.row:
```
AndreasArvidsson: I would move this logic into move_to_correct_row

FireChickenProductivity: I am confused. Could you please elaborate? 
I would prefer to avoid hiding the side effect of popping the current stop off of the stack behind a helper function like move_to_correct_row.

AndreasArvidsson: I meant only the line `if current_stop.row != next_stop.row:`

FireChickenProductivity: That makes sense. I will make that change.

FireChickenProductivity: I made the requested change.

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: This function name now yes verb.

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: I agree that improves readability. I added the requested parentheses.

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: I moved the function inside the stop class.

AndreasArvidsson (approved):

jaresty:
Wow, this is fancy.

PR #1911, opened 2025-06-02
Add setting for raw snippet insertion through pasting

FireChickenProductivity:


AndreasArvidsson:
Personally I don't think `user.paste` should ever be used. It's much better to context define `insert()` to do the preferred thing in all scenarios. We already have a setting for this in community.

FireChickenProductivity:
> Personally I don't think `user.paste` should ever be used. It's much better to context define `insert()` to do the preferred thing in all scenarios. We already have a setting for this in community.



> Personally I don't think `user.paste` should ever be used. It's much better to context define `insert()` to do the preferred thing in all scenarios. We already have a setting for this in community.

I do not use paste to insert because code editors that I have used treat pasting symbols differently from typing them and using that setting would break many of my existing commands. If I needed to use an editor other than VSCode or was not able to use the command server, I would prefer the ability to have snippets pasted to bypass the need to take into account numerous factors like automatic changes to the indentation level and automatic completion, which has behavior that not only differs on a per editor level but also on a per language level. That does not mean I am going to want everything else inserted through pasting. Even if I was using paste to insert with a certain character limit designed to optimize performance, I might want snippets pasted consistently instead of having to choose between the optimum for regular insertion and improving snippet behavior.

core/snippets/snippets_insert_raw_text.py
```diff
@@ -1,7 +1,16 @@
 import re
 from dataclasses import dataclass
 
-from talon import actions
+from talon import Module, actions, settings
+
+mod = Module()
+
+mod.setting(
+    "snippets_raw_text_always_paste",
```
AndreasArvidsson:
```suggestion
    "snippet_raw_text_paste",
```

core/snippets/snippets_insert_raw_text.py
```diff
@@ -19,7 +28,10 @@ def insert_snippet_raw_text(body: str):
     """Insert snippet as raw text without editor support"""
     updated_snippet, stop = parse_snippet(body)
 
-    actions.insert(updated_snippet)
+    if settings.get("user.snippets_raw_text_always_paste"):
```
AndreasArvidsson:
```suggestion
    if settings.get("user.snippet_raw_text_paste"):
```

AndreasArvidsson (approved):

Issue #1817, opened 2025-04-03
Use an alternate test character for context-sensitive dictation in Notion


pokey:
We insert a space to allow us to do our peek operation for context-sensitive dictation, but in Notion, pressing spacebar at beginning of line triggers AI assistant 😅

https://github.com/talonhub/community/blob/f325de37d9695b14d90ba8e6a35055b3c47c6280/core/text/text_and_dictation.py#L516

We should either use a different char always, or at least override it in Notion

nriley:
From the community backlog session — none of us use Notion so would appreciate some assistance from someone who does!

PR #1915, opened 2025-06-03
Add setting for determining context sensitive dictation peek character

FireChickenProductivity:
In response to this issue: https://github.com/talonhub/community/issues/1817, this pull request adds a setting for determining  the character inserted during context sensitive dictation without application support to handle the case of nothing before or after the cursor. This would make changing the character on a per application basis easier and let someone pick a better default for their use case.

### DIFF
--- a/core/text/text_and_dictation.py
+++ b/core/text/text_and_dictation.py
@@ -15,6 +15,13 @@ mod.setting(
     desc="Look at surrounding text to improve auto-capitalization/spacing in dictation mode. By default, this works by selecting that text & copying it to the clipboard, so it may be slow or fail in some applications.",
 )
 
+mod.setting(
+    "context_sensitive_dictation_peek_character",
+    type=str,
+    default=" ",
+    desc="This is the character inserted during dictation_peek to ensure that some text is selected even if the cursor is at the start or end of the document.",
+)
+
 mod.list("prose_modifiers", desc="Modifiers that can be used within prose")
 mod.list("prose_snippets", desc="Snippets that can be used within prose")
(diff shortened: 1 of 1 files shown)


core/text/text_and_dictation.py
```diff
@@ -15,6 +15,13 @@
     desc="Look at surrounding text to improve auto-capitalization/spacing in dictation mode. By default, this works by selecting that text & copying it to the clipboard, so it may be slow or fail in some applications.",
 )
 
+mod.setting(
+    "context_sensitive_dictation_peek_character",
+    type=str,
+    default=" ",
+    desc="This is the character inserted during dictation_peek to ensure that some text is selected even if the cursor is at the start or end of the document.",
```
AndreasArvidsson: Add note about the the setting needs to be a single character.

core/text/text_and_dictation.py
```diff
@@ -547,5 +554,5 @@ def dictation_peek(left: bool, right: bool) -> tuple[Optional[str], Optional[str
             actions.edit.extend_word_right()
             after = actions.edit.selected_text()[1:]
             actions.edit.left()
-            actions.key("delete")  # remove space
+            actions.key("delete")  # remove peek character
```
AndreasArvidsson:
```suggestion
            actions.user.delete_right() # remove peek character
```

core/text/text_and_dictation.py
FireChickenProductivity: I added the note.

AndreasArvidsson (approved):

Issue #793, opened 2022-03-02
Subtitles and command history don't match; "wake up" doesn't go into command history


rntz:
It would be good if the command history and subtitles agreed by default. However, the subtitles include wake up commands, but the command history does not.

PR #1916, opened 2025-06-03
Include wakeup commands in command history

FireChickenProductivity:
This addresses issue: https://github.com/talonhub/community/issues/793.

Including wakeup commands in the history it is useful if a user woke up talon without knowing what command did that and wants to know what happened. This also improves consistency between the command history and the talon subtitles.

I did a little refactoring to simplify the code and add more type annotations to the variables.

### DIFF
--- a/plugin/command_history/command_history.py
+++ b/plugin/command_history/command_history.py
@@ -8,20 +8,28 @@ mod = Module()
 mod.setting("command_history_size", type=int, default=50)
 mod.setting("command_history_display", type=int, default=10)
 
-hist_more = False
-history = []
+hist_more: bool = False
+history: list[str] = []
+was_asleep: bool = False
 
-
-def on_phrase(j):
+def handle_phrase(j):
     global history
-
     words = j.get("text")
-
     text = actions.user.history_transform_phrase_text(words)
-
     if text is not None:
         history.append(text)
         history = history[-settings.get("user.command_history_size") :]
+
+def on_phrase(j):
+    global was_asleep
+    was_asleep = not actions.speech.enabled()
+    if actions.speech.enabled():
+        handle_phrase(j)
+
+def after_phrase(j):
+    """This handles a situation where the user used a command that woke up talon"""
+    if was_asleep and actions.speech.enabled():
+        handle_phrase(j)
 
 
 # todo: dynamic rect?
@@ -44,6 +52,7 @@ def gui(gui: imgui.GUI):
 
 
 speech_system.register("phrase", on_phrase)
+speech_system.register("post:phrase", after_phrase)
 
 
 @mod.action_class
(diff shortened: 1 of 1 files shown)


AndreasArvidsson:
1. If this is consistency with subtitles we should probably use the same logic as the subtitles. https://github.com/talonhub/community/blob/365d8cfe8c9d6eaf5d644f20640be96b0ab0a7ca/plugin/subtitles/on_phrase.py
2. Please properly link the issue (under development) to the pull request so it will be closed with it.

FireChickenProductivity:
> 1. If this is consistency with subtitles we should probably use the same logic as the subtitles. https://github.com/talonhub/community/blob/365d8cfe8c9d6eaf5d644f20640be96b0ab0a7ca/plugin/subtitles/on_phrase.py
> 2. Please properly link the issue (under development) to the pull request so it will be closed with it.

2. I do not have sufficient permissions to do that, which is probably because I do not have permission to close someone else's issue.

1. For some reason I forgot about the community subtitles when I made this pull request. I will make that change.

plugin/subtitles/on_phrase.py
```diff
@@ -26,4 +25,10 @@ def skip_phrase_in_sleep(phrase: Phrase) -> bool:
     )
 
 
+def parse_phrase(phrase: Phrase) -> str:
+    words = phrase["phrase"]
+    current_phrase = " ".join(words)
+    return current_phrase
```
AndreasArvidsson:
```suggestion
    return " ".join(words)
```

plugin/command_history/command_history.py
```diff
@@ -82,11 +82,3 @@ def history_get(number: int) -> str:
         """returns the history entry at the specified index"""
         num = (0 - number) - 1
         return history[num]
-
-    def history_transform_phrase_text(words: list[str]) -> Optional[str]:
```
AndreasArvidsson, line 86: This was available to be context redefined. We should probably have a discussion on the community meetup if we want to keep it as an action.

AndreasArvidsson: Feedback from community session. Leave this as an action.

FireChickenProductivity: I put the action back.

plugin/subtitles/on_phrase.py
```diff
@@ -8,8 +8,7 @@ def on_pre_phrase(phrase: Phrase):
     if skip_phrase(phrase):
         return
 
-    words = phrase["phrase"]
-    current_phrase = " ".join(words)
+    current_phrase = parse_phrase(phrase)
```
AndreasArvidsson:
```suggestion
    words = phrase["phrase"]
    current_phrase = " ".join(words)
```

plugin/subtitles/on_phrase.py
```diff
@@ -26,4 +25,9 @@ def skip_phrase_in_sleep(phrase: Phrase) -> bool:
     )
 
 
+def parse_phrase(phrase: Phrase) -> str:
+    words = phrase["phrase"]
+    return " ".join(words)
+
+
```
AndreasArvidsson:
```suggestion
```

PR #1918, opened 2025-06-05
Migrate continue to snippets

FireChickenProductivity:
Ideally, for languages where the continue statement can take an argument, we would have a way to insert a space first but otherwise have no space without an argument.


core/snippets/snippets/continue.snippet
```diff
@@ -0,0 +1,29 @@
...
+---
+
+language: c | csharp | java
+-
+continue;
+---
+
+language: javascript | typescript | javascriptreact | typescriptreact | php | rust
+-
+continue$0;
```
AndreasArvidsson: I would remove `$0` and merge this with the above definition.

FireChickenProductivity: I would not. In these languages, the continue statement can take an argument.

AndreasArvidsson: Well yes, it's just that it's seldom used and if you expect another argument is should really be a whitespace after `continue`. Let's talk about it on the community meetup.

FireChickenProductivity:
> Feedback from community session.We will split this into two separate snippets. continue and continue with.

I implemented the requested change.

PR #1926, opened 2025-06-20
Migrate break statement to snippets

FireChickenProductivity:


nriley (approved):

nriley:
Note that the $0 for break in Lua is just there such that the inserted text will end with a space — it is common to have break on the same line as other keywords in Lua.

PR #1927, opened 2025-06-23
Migrate block to snippets

FireChickenProductivity:
This raises the question of what to do with the code_block_c_like tag.

I think we can leave it as users might have made use of it in their custom repositories, but I think it should be moved out of imperative.

### DIFF
--- a/core/snippets/snippets/codeBlock.snippet
+++ b/core/snippets/snippets/codeBlock.snippet
@@ -1,0 +1,23 @@
+name: codeBlock
+phrase: block
+---
+
+language: c | cpp | csharp | java | javascript | typescript | javascriptreact | typescriptreact | php | scala | kotlin | r | rust | go | css | terraform | protobuf
+-
+{
+    $0
+}
+---
+
+language: python
+-
+:
+    $0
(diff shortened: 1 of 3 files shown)


AndreasArvidsson (changes requested):
Please remove the tag and add noting breaking changes

FireChickenProductivity:
I removed the tag and put a note in breaking changes. Is my description of the change adequate?

AndreasArvidsson (approved):

BREAKING_CHANGES.txt
```diff
@@ -7,6 +7,7 @@ and when the change was applied given the delay between changes being
 submitted and the time they were reviewed and merged.
 
 ---
+* 2025-06-28 Remove user.code_block_c_like tag in favor of using the code block snippet to handle code blocks.
```
AndreasArvidsson:
```suggestion
* 2025-06-28 Remove user.code_block_c_like tag in favor of using `codeBlock.snippet` to handle code blocks.
```

AndreasArvidsson (approved):

PR #1928, opened 2025-06-25
Start deprecating imperative actions

FireChickenProductivity:


AndreasArvidsson (changes requested):
Please do a workspace wide search for all action names to make sure they are not used anywhere else.

FireChickenProductivity:
There is a complication of what to do with ruby because state do is not used in that language for the do while loop.

FireChickenProductivity:
I have found no other remaining issues with this.

lang/rust/rust.talon
```diff
@@ -61,7 +61,6 @@ state (mod | module): "mod "
 state ref (mute | mutable): "ref mut "
 state ref: "ref "
 state trait: "trait "
-state match: user.code_state_switch()
```
AndreasArvidsson, line 64: We probably need to deprecate these voice commands and not just remove them.

FireChickenProductivity: No. They already exist as part of the standard imperative grammar. Those were duplicates.

FireChickenProductivity: The standard imperative command for switch statements is state (switch | match): user.insert_snippet_by_name("switchStatement"). 
Talon was not even using the language specific ones for me in the first place.

AndreasArvidsson (approved):

FireChickenProductivity:
This is not ready until we address the ruby issue.

AndreasArvidsson:
Ah I completely misinterpreted `I have found no other remaining issues with this.`. Sorry.

AndreasArvidsson:
Isn't it enough to use a snippet for ruby?

AndreasArvidsson:
I would vote for either adding a ruby snippet for that if one is available or just removing it then.

PR #1929, opened 2025-06-25
Deprecate talon wake/sleep without dragon

FireChickenProductivity:


AndreasArvidsson (dismissed):
Please update the wiki to stop recommending these commands
https://talon.wiki/basic%20usage/basic_usage/#mode-switching

nriley (approved):

nriley:
There is another mention of `Talon wake` on the troubleshooting page of the wiki [here](https://talon.wiki/resource%20hub/speech%20recognition/troubleshooting/#check-that-talon-is-awake).

PR #1930, opened 2025-06-26
Expand markdown snippets

BlueDrink9:
I'm hoping this doesn't end up stepping on the toes of efforts to migrate language specific things to snippets. These are some snippets that I expected to be there and were not


FireChickenProductivity:
These were not planned as part of the migration itself, so no toe stepping.

core/snippets/snippets/markdown.snippet
```diff
@@ -4,5 +4,48 @@ language: markdown
 name: link
 phrase: link
 -
-[$1]($0)
+[$0]($1)
```
FireChickenProductivity: You seem to have inverted the order of the snippet holes for the link snippet. Was there a reason for this?

core/snippets/snippets/markdown.snippet
```diff
@@ -4,5 +4,48 @@ language: markdown
...
+---
+
+name: image
+phrase: image
+-
+![$0]($1)
```
FireChickenProductivity: Because $0 is for the last snippet hole, this goes from right to left. Is there a reason to do that with this snippet?

lang/markdown/markdown.py
```diff
@@ -16,4 +16,6 @@
     "shell": "shell",
     "bash": "bash",
     "json": "json",
+    "are": "r",
```
FireChickenProductivity, line 19: This looks like a separate change from the snippets. Based on my experience attending backlog sessions, the maintainers might ask you to make that a separate pull request.

core/snippets/snippets/markdown.snippet
```diff
@@ -4,5 +4,48 @@ language: markdown
...
+name: quote
+phrase: quote
+-
+> $0
+---
+
+name: comment
+phrase: comment
+-
+<!-- $0 -->
```
FireChickenProductivity: I am not familiar with markdown comment syntax, but we have snippet files for line comments and block comments, so I would recommend putting this in the appropriate file if this is a line or block comment to improve organization. If this is a special markdown thing, then it makes sense to leave it in the markdown snippet file.

BlueDrink9: Good point, thank you. It's technically an HTML comment, but it's fairly common for it to be used in Markdown as well. I'll migrate it

core/snippets/snippets/markdown.snippet
BlueDrink9: Oops, that was an accident on my part. That being said, do you think it would be open to keeping it like this?

I would want link and image to be consistent. It doesn't make much sense to me to have the URL, which is much more static and more likely to be pasted, as the final point when most people would be more likely to want to free-form edit the description.
I would even potentially be tempted to have the final stop outside of the link entirely so that the person can continue with whatever they were typing.

core/snippets/snippets/markdown.snippet
BlueDrink9: Ah ha, now I understand the original logic

core/snippets/snippets/markdown.snippet
BlueDrink9: Now I have seen the other comment. I think the latter is the better solution, so I have updated it to that. However, this does change the stops slightly for that snippet. Is that likely to be a problem, do you think?

lang/markdown/markdown.py
BlueDrink9: Thanks for the heads up. It is a fairly minor change and functions very similarly to a snippet "python block" rather than "snip code block"), so I'll leave it and hope.

core/snippets/snippets/markdown.snippet
FireChickenProductivity: I think the argument that it would be convenient to have a quick way to exit the snippet structure could be applied to many snippets. I personally think it would be better to have a command that works in the general case for that instead of modifying individual snippets. 
I can discuss this with whoever shows up at the next backlog session (tomorrow).
I doubt the maintainers would want to change the existing link snippet given that it was part of the migration, and we do not want to create breaking changes too frequently. I would leave it for now though and wait for their input.

BlueDrink9: That makes sense and would make it inconsistent with all the other snippets. I will revert it

AndreasArvidsson: Feedback from community session. We agree that we want to keep the current snippet behavior. If you feel strongly that you'll want a way to go to the end of snippets that is definitely something we can have a look at. Please file an issue.

BlueDrink9:
oops, had forgotten to address issues. Will finalise today!

core/snippets/snippets/markdown.snippet
```diff
@@ -6,3 +6,40 @@ phrase: link
...
+---
+
+
+name: checkbox
+phrase: to do | check box | tick box
+-
+* [ ] $0
+---
+
+name: bold
```
nriley, line 23: Markdown syntax is not defined as bold/italic/bold italic but just as increasingly strong forms of emphasis; perhaps something like emph[asis]/strong/very strong? Even newer definitions like GitHub Flavored Markdown don't include any definitions of how the various emphasis formats should be rendered (see https://github.github.com/gfm/#emphasis-and-strong-emphasis)

nriley (commented):
Just one comment, otherwise looks good. I don't have any problem with you putting the "r" thing in here

core/snippets/snippets/markdown.snippet
BlueDrink9: Hmm, that's an interesting note. Because intuitively, and for the purpose of saying things, I feel like most people would still refer to it as bold/italic, since that is the most common rendering fashioned by far. However, it is technically incorrect… Perhaps we just include both as options?

nriley: I'm OK with that if the rendering is so consistent across different Markdown flavors/renderers.

BlueDrink9: Added both in case anyone uses the other as a phrase. Other than that, ready to merge?

nriley: Does that actually work with a portion of a word in `[]`? I've never tried that.

AndreasArvidsson: It does not. Only `a-z` and `|` is supported.

BlueDrink9: Sorry, it was less than responsible of me to not test that. Fixed now, and tested

nriley (approved):

PR #1931, opened 2025-06-28
Quick fix: Fix snip next without editor support comparison bug

FireChickenProductivity:
Adding parentheses to the if statement caused it to misbehave for some reason. This caused an unnecessary announcement that the snippet was converted from right to left to go from left to right. I fixed it by extracting the boolean values into their own variables.


core/snippets/snippets_insert_raw_text.py
```diff
@@ -122,7 +122,9 @@ def is_any_line_from_right_to_left(lines) -> bool:
                 next_key = next_stop.compute_sorting_key()
                 # If the ordering between the keys and columns are inconsistent,
                 # the stops on this line go from right to left
-                if (next_key < stop_key) != (stop.col < next_stop.col):
+                is_next_key_smaller = next_key < stop_key
+                is_next_col_smaller = next_stop.col < stop.col
+                if is_next_key_smaller != is_next_col_smaller:
```
nriley:
```suggestion
                if (next_key < stop_key) != (next_stop.col < stop.col):
```
The logic was just backwards.

nriley (changes requested):

core/snippets/snippets_insert_raw_text.py
FireChickenProductivity: Weird. It works without the parentheses.

nriley: Yes, because Python's operator precedence isn't what you think it is; [all comparison operators are at the same precedence level](https://docs.python.org/3/reference/expressions.html#comparisons).

nriley (approved):

PR #1933, opened 2025-06-28
Fix ruby state do

FireChickenProductivity:
This is a temporary fix until we can decide on something better.

Ruby used code state do for something other than do while loops, so it did not get covered in the snippet migration.

I am not sure why my migration pull request got merged before we figured out what to do with this. I guess I can try to be more clear with my comments going forward.

### DIFF
--- a/lang/ruby/ruby.talon
+++ b/lang/ruby/ruby.talon
@@ -32,6 +32,7 @@ state end: "end"
 state begin: "begin"
 state rescue: "rescue "
 state module: "module "
+state do: user.code_state_do()
 
 ^instance <user.text>$:
     insert("@")


AndreasArvidsson:
Will this even run? Or will the default snippet implementation be used? Trying to overload spoken forms like this is problematic.

FireChickenProductivity:
I tested this before making the pull request and it runs. 
We could do that, but that is not what the do while loop looks like in ruby.

> Will this even run? Or will the default snippet implementation be used? Trying to overload spoken forms like this is problematic.
> 
> Also I would vote for adding a ruby do while snippet since that is what the spoken form does in all other languages.

I tested this before making the pull request and it runs. 
We could do that, but that is not what the do while loop looks like in ruby. Ruby has a more general do keyword that state do was triggering for better or for worse. I wanted to preserve the preexisting behavior until we had a chance to think this one through.

FireChickenProductivity:
I made the change based on research. Someone familiar with ruby should tell us if I got the do while snippet correct or not before we merge this.

nriley (approved):

PR #1934, opened 2025-07-04
Add snippet stop at the end

FireChickenProductivity:
Add a snippet stop at the end of every snippet that does not already have one there to allow exiting the snippet with "snip next". 

I replaced .* with .*? in the regular expression to match the smallest possible match to better handle multiple instances of the same snippet stop with default values.

Someone should probably test this other than me as breaking snippets would not be fun.


core/snippets/snippets_parser.py
```diff
@@ -110,8 +136,14 @@ def validate_snippet(document: SnippetDocument, snippet: Snippet) -> bool:
...
 def is_variable_in_body(variable_name: str, body: str) -> bool:
+    return (
+        re.search(create_variable_regular_expression(variable_name), body) is not None
+    )
+
+
+def create_variable_regular_expression(variable_name: str) -> str:
     # $value or ${value} or ${value:default}
-    return re.search(rf"\${variable_name}|\${{{variable_name}.*}}", body) is not None
+    return rf"\${variable_name}|\${{{variable_name}.*?}}"
```
AndreasArvidsson, line 127: Please add a comment why we need the ? a the end

AndreasArvidsson:
Please test what happens if you don't have a $zero stop

AndreasArvidsson:
Feed back from the session. Instead of having a fixed replacement name we should find the highest numerical value and pick something that's larger than that. 


Which a probably unit test this

core/snippets/snippets_parser.py
nriley: "this"?

core/snippets/snippets_parser.py
```diff
@@ -110,8 +115,16 @@ def validate_snippet(document: SnippetDocument, snippet: Snippet) -> bool:
...
+    return (
+        re.search(create_variable_regular_expression(variable_name), body) is not None
+    )
+
+
+def create_variable_regular_expression(variable_name: str) -> str:
     # $value or ${value} or ${value:default}
-    return re.search(rf"\${variable_name}|\${{{variable_name}.*}}", body) is not None
+    # *? is used to find the smallest possible match.
+    # We need this here to avoid treating multiple stops as a single one
```
nriley:
```suggestion
    # This stops multiple stops from being treated as a single stop.
```

PR #1936, opened 2025-07-09
Add more complete Python keyword support

FireChickenProductivity:
I used the keywords from here: https://docs.python.org/3/reference/lexical_analysis.html#identifiers

This was a draft prototype of more general keyword support. Now that we are migrating a lot to snippets, I still want users to be able to use keywords directly.

Short keywords require the `keyword` or `put` prefix. Long keywords do not.


FireChickenProductivity:
I found this set of google search suggestions while I was doing research for this amusing.
image

lang/python/code_keyword_long.talon-list
```diff
@@ -0,0 +1,7 @@
+list: user.code_keyword_long
+code.language: python
+-
+
+finally: "finally"
```
auscompgeek: I don't think there's a use case for inserting `finally` without also starting a block, is there?

lang/python/code_keyword.talon-list
```diff
@@ -0,0 +1,41 @@
...
+in: " in "
+while: "while "
+with: "with "
+try: "try"
+raise: "raise "
+except: "except "
+and: " and "
+not: "not "
+or: " or "
+else: "else"
```
auscompgeek: Since this would be most useful for writing a ternary expression, maybe inserting a space here as well would be good? If people accidentally use this to start an `else:` block then an autoformatter should clean it up.
```suggestion
else: "else "
```

lang/python/code_keyword.talon-list
```diff
@@ -0,0 +1,41 @@
+list: user.code_keyword
+code.language: python
+-
+
```
nriley, line 4: It would be great if you could sort this list.

lang/python/code_keyword_long.talon-list
```diff
@@ -0,0 +1,7 @@
+list: user.code_keyword_long
```
nriley: I would recommend you name this something like `code_keyword_bare`.

lang/tags/keywords.py
```diff
@@ -5,9 +5,13 @@
 mod.tag("code_keywords", desc="Tag for enabling commands for keywords")
 
 mod.list("code_keyword", desc="List of keywords for active language")
+mod.list(
+    "code_keyword_long",
+    desc="List of long keywords (at least 3 syllables) for active language",
```
nriley: Update the description so it reflects that these keywords are available alone as well as with the words keyword or put.

auscompgeek (commented):
A couple of thoughts, but by no means consider this a blocking review.

lang/python/code_keyword_long.talon-list
nriley: From the community backlog session — there may be a case when you want to insert finally without necessarily inserting a finally block, such as when editing an existing control structure. We did not want to be too opinionated about supporting keywords only when there is a specific purpose for them as we felt that being a complete representation of the language syntax was more important. We were also trying to avoid being too "magical" for new users of Talon so we are only adding spacing.

lang/tags/keywords.py
FireChickenProductivity: Is this a good enough description? I updated it to the following:
"List of keywords for active language that can be dictated by name alone or the put (name) command"

nriley (approved):

PR #1940, opened 2025-07-13
Add shortcat integration

NLCas8:
Integrates [Shortcat](https://shortcat.app/) with Talon. 

This allows you to say for example `sot click [BUTTON LABEL]` where `BUTTON LABEL` is the label of a button anywhere on the screen (it works in almost any application, so not only the browser). I have done the same for right click, hover, and double click.

See screenshot where I said `sot click landscape`. It also works with button labels that are not visible on the screen, but only appear when hovering over the button.
image


jaresty:
I didn't know about shortcat. Cool!

auscompgeek:
FWIW I think a bunch of folks moved on from Shortcat to [Homerow](https://homerow.app).

apps/shortcat/shortcat.py
```diff
@@ -0,0 +1,40 @@
+from talon import Module, actions
+
+mod = Module()
+
+# Hotkey that is configured to activate Shortcat
+SHORTCAT_HOTKEY = "cmd-shift-space"
```
auscompgeek: This should be a setting so it can be changed by users without modifying the code.

NLCas8:
> FWIW I think a bunch of folks moved on from Shortcat to [Homerow](https://homerow.app).

I didn't know about Homerow. I will look into this, thanks!

apps/shortcat/shortcat.py
NLCas8: I replaced both constants with settings

apps/shortcat/shortcat.py
```diff
@@ -0,0 +1,48 @@
...
+    def shortcat_hover(text: str):
+        "Hover over a button using shortcat"
+        actions.key(settings.get("user.shortcat_hotkey"))
+        actions.sleep(settings.get("user.shortcat_click_delay"))
+        actions.insert(text)
+
+    def shortcat_click(text: str, click_delay: str = "0ms"):
+        "Click a button using shortcat"
+        actions.user.shortcat_hover(text)
+        actions.sleep(click_delay)
```
nriley, line 30: From the community backlog session, we had a couple comments on this behavior.

If the delay is to give you time to cancel if the wrong button is about to be clicked, then using `actions.sleep` will block Talon and not allow you to cancel using a voice command or other Talon action. A couple options:

- Having a separate confirmation action. This is how my [Homerow support works](https://github.com/nriley/talon_community/blob/f887406dc5d5a4f1364407617fb23290afa61eae/apps/homerow/homerow_search.talon#L3)
- Using `cron` such that Talon is active while it is waiting.

Our preference is for the first option as we feel that any substantial delay long enough to give you a chance to cancel would be frustrating to use in practice.

apps/shortcat/shortcat.py
```diff
@@ -0,0 +1,48 @@
...
+        actions.mouse_click(0)
+
+    def shortcat_double_click(text: str, click_delay: str = "0ms"):
+        "Double click a button using shortcat"
+        actions.user.shortcat_hover(text)
+        actions.sleep(click_delay)
+        actions.key("enter")
+        actions.sleep("100ms")
+        actions.mouse_click(0)
+        actions.mouse_click(0)
```
nriley, line 42: Consider using a single action for single and multiple clicks.

apps/shortcat/shortcat.talon
```diff
@@ -0,0 +1,16 @@
+os: mac
+-
+
+^(shot | sot) [<user.text>]: user.shortcat_hover(text or "")
+
+^(shot | sot) click [<user.text>]:
+    # Add some delay to cancel click if needed
+    user.shortcat_click(text or "", "1500ms")
```
nriley, line 8: I think that this delay would be incredibly frustrating in practice, not to mention it is likely to trigger Talon's watchdog timer and spam your log. I would recommend that you have this as a configurable setting, if you decide to use the cancel delay rather than a confirmation action.

nriley (changes requested):
Overall, we appreciate your contribution, and it has certainly caused me to take a second look at Shortcat — it has grown a lot of features since the last time I checked!

Since not everybody will be using Shortcat on their Mac, you should be able to scope your changes to Shortcat appearing in `user.running`.

PR #1943, opened 2025-07-21
Improve C keywords support

FireChickenProductivity:
I used the following link for understanding the language keywords: https://en.cppreference.com/w/c/keyword.html

I decided to leave defining the bare list and figuring out what to do with the _ keywords to someone more familiar with the language.


lang/c/code_keyword.talon-list
```diff
@@ -0,0 +1,50 @@
...
+-
+
+align as: "alignas"
+align of: "alignof"
+auto: "auto "
+bool: "bool"
+break: "break"
+case: "case "
+char: "char"
+const exp: "constexpr "
```
nriley:
```suggestion
const expr: "constexpr "
```

lang/c/code_keyword.talon-list
```diff
@@ -0,0 +1,50 @@
...
+else: "else"
+enum: "enum "
+extern: "extern "
+false: "false"
+float: "float"
+for: "for"
+go to: "goto "
+if: "if"
+in line: "inline "
+int | integer: "int"
```
nriley: Double-check that the or works here.

FireChickenProductivity: It did not. Not sure what I was thinking there. Regardless, I fixed it.

chdoc:
Looking at this, I have several questions:
- Given that the `code_keywords` tag already has commands for inserting keywords both with and without a prefix, shouldn't this PR also remove the command for inserting *all* keywords without any prefix? If the intention is to allow inserting all these keywords without any prefix, which I doubt is a good idea, I suppose this should instead define the list `code_keyword_unprefixed`, no?
- What is the reason for including the control flow keywords? Normally, these would be inserted through snippets, no?

chdoc:
Isn't the behavior for unprefixed keywords (almost) exactly the same as it was before? I don't think this is a breaking change at all.

Which exact keywords to put into the prefixed and unprefixed lists may require some fine tuning by users. Personally, I would put `const`, `constexpr`, and possibly also `return` into the unprefixed list. Same for `nullptr`, `true`, and `false` for which I find the `state` prefix rather counterintuitive. But that's theory without taking misrecognitions into account.

Issue #1942, opened 2025-07-19
Be smarter about inserting one or multiple keywords into code


nriley:
The current `user.code_keyword` action has several deficiencies.

* It does not attempt to insert or remove spacing between consecutive keywords
* it calls the `insert` action multiple times instead of joining together the keywords
* it does not attempt to examine the document context to understand if a leading or trailing space should be removed or added
* it has a name which is not intention revealing

@AndreasArvidsson has solutions to some of these issues in his own repository.

https://github.com/AndreasArvidsson/andreas-talon/blob/4327ba6d4f0420c7a8deb8148c74e832b6537f93/languages/tags/code_keywords.py#L10-L13

https://github.com/AndreasArvidsson/andreas-talon/blob/4327ba6d4f0420c7a8deb8148c74e832b6537f93/core/edit/edit.py#L124-L138

PR #1946, opened 2025-07-25
Better handle keyword spaces

FireChickenProductivity:
Better handle keyword spaces with the keyword grammar. Addresses part of issue #1942.

### DIFF
--- a/lang/tags/keywords.py
+++ b/lang/tags/keywords.py
@@ -20,5 +20,13 @@ def code_keyword(m) -> str:
 class Actions:
     def code_keyword(keywords: list[str]):
         """Adds keywords"""
-        for keyword in keywords:
-            actions.insert(keyword)
+        if len(keywords) == 1:
+            actions.insert(keywords[0])
+        else:
+            # every keyword is separated by a space
+            # the spacing before the first keyword and after the last keyword is kept
+            combined_text: str = keywords[0].rstrip() + " "
+            for i in range(1, len(keywords) - 1):
+                combined_text += keywords[i].strip() + " "
+            combined_text += keywords[-1].lstrip()
+            actions.insert(combined_text)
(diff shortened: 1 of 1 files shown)


lang/tags/keywords.py
```diff
@@ -20,5 +20,13 @@ def code_keyword(m) -> str:
...
-        for keyword in keywords:
-            actions.insert(keyword)
+        if len(keywords) == 1:
+            actions.insert(keywords[0])
+        else:
+            # every keyword is separated by a space
+            # the spacing before the first keyword and after the last keyword is kept
+            combined_text: str = keywords[0].rstrip() + " "
+            for i in range(1, len(keywords) - 1):
+                combined_text += keywords[i].strip() + " "
```
auscompgeek: String concatenation in a loop like this is O(n²). This really should use `" ".join`.

FireChickenProductivity: I am aware that string concatenation in a loop like that is quadratic in python. I was not concerned with the performance given how small the input sizes are going to be in real use cases, but I can make the change if the maintainers want.

FireChickenProductivity: No one else showed up at today's backlog session to weigh in, so I decided to make the change.

lang/tags/keywords.py
```diff
@@ -20,5 +20,31 @@ def code_keyword(m) -> str:
...
-            actions.insert(keyword)
+        if len(keywords) == 1:
+            actions.insert(keywords[0])
+        else:
+            # every keyword is separated by a space
+            # the spacing before the first keyword and after the last keyword is kept
+            num_leading_spaces = compute_number_of_leading_spaces(keywords[0])
+            num_trailing_spaces = compute_number_of_trailing_spaces(keywords[-1])
+            stripped_words = [word.strip() for word in keywords]
+            combined_text = (
```
nriley: Just go ahead and strip the first and last words.

nriley (changes requested):

lang/tags/keywords.py
FireChickenProductivity: I made the change.

AndreasArvidsson (approved):

PR #1955, opened 2025-08-01
Migrate javascript common functions to talon list

FireChickenProductivity:
To make it easier to match javascript variants, I created a javascript tag. It may be worth discussing alternative approaches.


AndreasArvidsson (changes requested):
Please go back to be more explicit without a javascript tag

AndreasArvidsson (approved):

PR #1960, opened 2025-08-04
fixed microphone letters in mode_indicator for windows

DonnerWolfBach:
At least on my win 11 system, I always got "Mi" as letters because the strings seem to always start with "Microphone (...".


plugin/mode_indicator/mode_indicator.py
```diff
@@ -129,7 +129,13 @@ def on_draw(c: SkiaCanvas):
             Point2d(x, y), radius, [color_text, color_text]
         )
 
-        text = current_microphone[:2]
+        text = current_microphone
+        # on windows the current_microphone string always starts with "microphone (..."
```
AndreasArvidsson, line 143: This isn't actually true. In windows you can rename your audio devices to whatever you want. Even without renaming them there exists microphones out there that have names that doesn't start with microphone.

We think it's probably easier for you to just rename your microphones so they each start with a unique letter.
We should probably add a comment write above the setting to enable the microphone name in the mode indicator telling people to rename their microphone in this case.
Does that sound reasonable?

AndreasArvidsson (changes requested):

plugin/mode_indicator/mode_indicator.py
DonnerWolfBach: Oh, that is interesting, did not know that. 

Indeed one of my headset calls itself "headset microphone (...)".

For me renaming microphones would be a bit of a chore since I have some more and switch computers quite often, but I think the possibility should be definitely mentioned.

I will most definitely keep my little dirty fix on my fork, maybe it makes sense to hide it behind another option for the community? Might provide a nicer out of the box experience if instead of renaming your microphone (clean but a bit more effort in setup - everytime you connect a new mic) you could also activate this little hack (not guaranteed to work always but a bit easier).

But I dont have strong opinions on that.

AndreasArvidsson: I think a setting and then actually check if the name contains parenthesis would be a decent solution.
Another solution would be to do a map the same way we have with overriding application names, but for microphones. That way you only have to change a csv file and would work for all platforms.

PR #1963, opened 2025-08-04
Add wait option to `insert_between` action

BlueDrink9:
Allows users to fix issues like paired delimiter insertion always failing in VSCode, without having to have an incredibly long insert or key wait

### DIFF
--- a/core/edit/insert_between.py
+++ b/core/edit/insert_between.py
@@ -1,12 +1,20 @@
-from talon import Module, actions
+from time import sleep
+from talon import Module, actions, settings
 
 mod = Module()
 
+mod.setting(
+    "insert_between_wait",
+    type=int,
+    default=0,
+    desc="Time in milliseconds to sleep after inserting text with `insert_between` (e.g. when using paired delimiters like 'box' or 'round'), before moving the cursor back. Useful to set on a per-application basis, to prevent moving the moving the cursor before text is inserted",
+)
 
 @mod.action_class
 class module_actions:
     def insert_between(before: str, after: str):
         """Insert `before + after`, leaving cursor between `before` and `after`. Not entirely reliable if `after` contains newlines."""
         actions.insert(f"{before}{after}")
+        sleep(int(settings.get("user.insert_between_wait", 0)) / 1000)
         for _ in after:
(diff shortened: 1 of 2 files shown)


FireChickenProductivity:
I recommend using the sleep action over time.sleep.

settings.talon
```diff
@@ -1,4 +1,4 @@
-settings():
+:
```
FireChickenProductivity: You have removed `settings()` from the start of the settings block. Can you put it back?

BlueDrink9: oops, thanks and sorry

core/edit/insert_between.py
```diff
@@ -1,12 +1,22 @@
...
+    desc="Time in milliseconds to sleep after inserting text with `insert_between` (e.g. when using paired delimiters like 'box' or 'round'), before moving the cursor back. Useful to set on a per-application basis, to prevent moving the moving the cursor before text is inserted",
+)
+
 
 @mod.action_class
 class module_actions:
     def insert_between(before: str, after: str):
         """Insert `before + after`, leaving cursor between `before` and `after`. Not entirely reliable if `after` contains newlines."""
         actions.insert(f"{before}{after}")
+        sleep(int(settings.get("user.insert_between_wait", 0)) / 1000)
```
AndreasArvidsson, line 20: Please use the Talon action for this instead.

AndreasArvidsson (changes requested):

AndreasArvidsson:
Are you using the clipboard/paste to do insertions? 
https://github.com/talonhub/community/blob/6e09d96d3e9afe56439a3da6724d8eab0f6a65f2/settings.talon#L86

core/edit/insert_between.py
BlueDrink9: action for what? Sleep??

BlueDrink9:
might try that, although imo talon pollutes my clipboard history enough as-is… For the time being, this has been pretty good so I think it's worth having!

core/edit/insert_between.py
BlueDrink9: ah, I see, didn't realise that existed

PR #1964, opened 2025-08-07
Use described functions for operators

FireChickenProductivity:
The git dif on this was not pretty. I needed to use app.register("ready", on_ready) to define operators to allow working with an action. 
The lua actions could probably be better described.


nriley:
Sorry, not only did I accidentally close this, we seem to have gotten confused along the way. The problem we were trying to solve was not the inability to access a docstring, i.e. the `__doc__` attribute (without which Python's help system would not work), but the inability to access the *code* of a function or lambda (`__code__`). So just use `__doc__`.

core/described_functions.py
```diff
@@ -0,0 +1,21 @@
+# This file provides support for defining functions that have an associated description
+# Talon's version of python does not allow getting the code of a function through reflection, so we are instead allowing associating doc strings with lambdas and functions as a way to apply descriptions to the community help system
```
nriley:
```suggestion
# Talon's version of Python does not allow getting the code of a function through reflection, so we are instead allowing associating doc strings with lambdas and functions as a way to apply descriptions to the community help system
```

core/described_functions.py
```diff
@@ -0,0 +1,21 @@
...
+    """Creates a function with an associated doc string. Primarily intended to be used with lambdas"""
+    function.__doc__ = description
+    return function
+
+
+def create_described_insert_between(before: str, after: str) -> Callable:
+    """Creates a described function for calling actions.user.insert_between"""
+    return create_described_function(
+        lambda: actions.user.insert_between(before, after),
+        f"actions.user.insert_between('{before}', '{after}')",
```
nriley:
```suggestion
        f"wraps the cursor with {before} and {after}",
```

lang/lua/lua.py
```diff
@@ -85,15 +89,28 @@ def code_operator_bitwise_right_shift():
...
     # code_operators_assignment
     ASSIGNMENT=" = ",
     # code_operators_bitwise
-    BITWISE_AND=code_operator_bitwise_and,
-    BITWISE_OR=code_operator_bitwise_or,
-    BITWISE_EXCLUSIVE_OR=code_operator_bitwise_exclusive_or,
-    BITWISE_LEFT_SHIFT=code_operator_bitwise_left_shift,
-    BITWISE_RIGHT_SHIFT=code_operator_bitwise_right_shift,
+    BITWISE_AND=create_described_function(
+        code_operator_bitwise_and, "insert bitwise and based on lua version"
```
nriley: Include these specific symbols that you use depending on the Lua version (and capitalize Lua), and mention the name of the setting that you use to define your Lua version.

FireChickenProductivity: Apparently it relies on a library call with older Lua versions. There is a todo comment to add support for other libraries. Given that it may depend on more than the version in future updates, do you still want me to describe exactly what gets inserted? I could instead do something like "insert (symbol) or library call based on user.lua_version"

nriley: Sounds reasonable to me.

FireChickenProductivity: I made the change and updated help operators to use the documentation string when available.

core/help/help.py
```diff
@@ -123,6 +123,8 @@ def update_operators_text():
                     operator = operators.get(operator_text)
                     if type(operator) == str:
                         text = ": " + operator
+                    elif hasattr(operator, "__doc__") and operator.__doc__:
+                        text = ": " + operator.__doc__
```
nriley: Try something like `elif doc := getattr(f, '__doc__', None):`

core/help/help.py
```diff
@@ -123,6 +123,8 @@ def update_operators_text():
                     operator = operators.get(operator_text)
                     if type(operator) == str:
```
nriley: Please use `isinstance` instead of `type`.

lang/lua/lua.py
```diff
@@ -83,17 +87,34 @@ def code_operator_bitwise_right_shift():
         actions.insert(" bit.rshift() ")
 
 
+def compute_bitwise_operator_description(operator_text: str) -> str:
+    return f"Insert {operator_text} or library call based on user.lua_version"
+
+
```
nriley: Since we completely control the implementations here please just add a docstring to each. Thanks!

nriley (approved):
Cool. Well, thank you again

core/help/help.py
FireChickenProductivity: Done

lang/lua/lua.py
FireChickenProductivity: Done

nriley (approved):

PR #1968, opened 2025-08-09
Improve list help display

nriley:
- Display "List: " at the beginning and pagination in parentheses to match other help displays
- Display the list description (wrapped to 70 characters if needed)
- Display "empty" instead of "1/0" if the list is empty

Before:
image

After:
image

### DIFF
--- a/core/help/help.py
+++ b/core/help/help.py
@@ -663,6 +663,7 @@ def gui_list_help(gui: imgui.GUI):
 
     gui.text(f"List: {selected_list} ({page_info})")
 
+    # Extract description from list declaration, i.e. mod.list(..., desc=...))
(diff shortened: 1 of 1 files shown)


FireChickenProductivity (approved):
If I did not know what this was doing already, it would not have been clear to me that this was getting the description from the list description string. I recommend adding a comment. 

I tested this and found no problems. This is a significant improvement to the help list system!

nriley:
> If I did not know what this was doing already, it would not have been clear to me that this was getting the description from the list description string. I recommend adding a comment.
> 
> I tested this and found no problems. This is a significant improvement to the help list system!

Makes sense. I have added a comment.

AndreasArvidsson (approved):

PR #1978, opened 2025-08-12
Add c++ specific snippets

FireChickenProductivity:
This defines some c++ specific snippets based on my own setup and chdoc's pull request.

### DIFF
--- a/core/snippets/snippets/c++.snippet
+++ b/core/snippets/snippets/c++.snippet
@@ -1,0 +1,53 @@
+language: cpp
+---
+
+name: templateDeclaration
+phrase: template
+-
+template <$0>
+---
+
+name: publicSectionDeclaration
+phrase: public
+-
+public:
+    $0
+---
+
+name: protectedSectionDeclaration
+phrase: protected
+-
+protected:
+    $0
+---
+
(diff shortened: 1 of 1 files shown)


core/snippets/snippets/c++.snippet
```diff
@@ -0,0 +1,53 @@
+language: cpp
+---
+
+name: templateDeclaration
+phrase: template
+-
+template <$0>
+---
+
+name: publicSectionDeclaration
```
nriley: Could you please check the C++ language specification and use names that reflect official terminology? Thanks!

nriley (changes requested):

core/snippets/snippets/c++.snippet
FireChickenProductivity: The language specification seems to be behind a paywall. Would it be sufficient to use terminology consistent with https://en.cppreference.com/w/cpp/language/access.html?

nriley: If you are pretty confident that this is a high quality reference material then I am not asking you to dig through or pay for the the specification! Looks good to me.

nriley (approved):

PR #1982, opened 2025-08-14
Improve c++ for each statements and add mutable for each variant

FireChickenProductivity:


nriley (approved):

nriley:
We did do a bit today!

PR #1985, opened 2025-08-14
Add education website extension

FireChickenProductivity:


auscompgeek:
I feel like we should stop shoving TLDs into a list called "file extensions"…

PR #1991, opened 2025-08-15
Adding Positron and Quarto files support 

EduardAnsaldo:
Adding Positron (A fork of VSCode used for data science) and Quarto (.qmd) files support for both positron and vscode.


apps/positron/positron_vscode_qmd.talon
```diff
@@ -0,0 +1,23 @@
+# Adding Quarto Markdown Support For Positron and VSCode
+app: vscode
+app: positron
+win.file_ext: .qmd
```
BlueDrink9, line 4: should probably use "code.language: quarto" instead - that way it can be forced, either by the editor (eg if it isn't a file yet) or by talon. And add it to https://github.com/talonhub/community/blob/030505955d83a687b3449da862b57e98e09fe9ca/core/modes/code_languages.py#L49

apps/positron/positron.talon
```diff
@@ -0,0 +1,24 @@
+#Adding Positron commands, building on Nicolas Riley's personal repo
+app: positron
+-
+notebook new: user.vscode("quarto.newNotebook")
```
AndreasArvidsson, line 4: We should probably use `run_rpc_command` here instead of `vscode`

apps/positron/positron.talon
```diff
@@ -0,0 +1,24 @@
+#Adding Positron commands, building on Nicolas Riley's personal repo
```
nriley:
```suggestion
#Adding Positron commands, building on Nicholas Riley's personal repo
```

nriley (approved):
Apologies for taking so long to review this — the plot commands look good.

PR #1997, opened 2025-08-20
Add python common member functions

FireChickenProductivity:
Expand the javascript common member functions into a general grammar and add python support. I also added a help command for the common member functions.

### DIFF
--- a/lang/tags/object_oriented.py
+++ b/lang/tags/object_oriented.py
@@ -8,7 +8,7 @@ mod.tag(
     desc="Tag for enabling basic object oriented programming commands (objects, classes, etc)",
 )
 
-mod.list("code_common_method", desc="Function to use in a dotted chain, eg .foo()")
+mod.list("code_common_method", desc="Commonly invoked method, e.g. 'foo' in '.foo()'")
 
 
 @mod.action_class


lang/tags/common_member_function.talon
```diff
@@ -0,0 +1,5 @@
+tag: user.code_common_member_function
+-
+
+dot {user.code_common_member_function}:
```
auscompgeek: I've been meaning to update #1254 to make changing `dot` in this command possible. I'll see if I can find some time to rebase that PR sometime soon...

lang/tags/common_member_function.py
```diff
@@ -0,0 +1,19 @@
+from talon import Context, Module, actions
+
+ctx = Context()
+mod = Module()
+
+mod.tag(
+    "code_common_member_function",
```
auscompgeek: FWIW I don't think it makes sense for this to have its own tag. It should probably just be folded into the object oriented tag.

FireChickenProductivity: We already have enough tags to clutter the help system, so I agree.

lang/tags/object_oriented.py
```diff
@@ -20,3 +22,7 @@ def code_self():
     def code_define_class():
         """Starts a class definition (e.g., Java's "class" keyword)"""
         actions.user.insert_snippet_by_name("classDeclaration")
+
+    def code_method(name: str):
+        """Inserts a member function call"""
```
nriley:
```suggestion
        """Inserts a method invocation"""
```

lang/tags/object_oriented.py
```diff
@@ -8,6 +8,8 @@
     desc="Tag for enabling basic object oriented programming commands (objects, classes, etc)",
 )
 
+mod.list("code_common_method", desc="Function to use in a dotted chain, eg .foo()")
```
nriley:
```suggestion
mod.list("code_common_method", desc="Commonly invoked method, e.g. 'foo' in '.foo()'")
```

PR #1998, opened 2025-08-20
Mouse wake enables controls mouse and no zoom

mpinkert:
This behavior was extremely confusing, where telling talon to `sleep all` would shut everything off but having it wake up from that with `welcome_back` would only activate zoom mode on the mouse. It took me several minutes to realize this unusual behavior.

I think most users are going to want to default to waking up to having control over the mouse and not having zoom mode enabled; while zoom mode is useful, I think most users would use the control setting most of the time.


nriley:
From the community backlog session — neither of us on the call this week used eye tracking frequently. Could someone else take a look at this?

nriley:
From the community backlog session — there are now so many different control mouse and Zoom mouse options available that there is no "one size fits all" option.

We would recommend you declare a variable to save the eye tracking modes enabled (control mouse, zoom mouse, etc., returned by the `tracking.*_enabled` actions) at the time you put Talon to sleep so they can be restored on wake.

nriley (changes requested):
Please see above request.

nriley:
This PR is now redundant and was implemented in #2082 as @FireChickenProductivity mentioned.

PR #1999, opened 2025-08-20
Significantly expand formatters and formatter help commands

mpinkert:
This change includes many new formatters for the word and prose
categories the play around with spaces, capitalization, commas, 
and periods. In addition I have introduced new help commands to bring up the individual lists for 
code, prose, word formatters since it has become bloated to just see all of the code and prose together.

I acknowledge that this is a rather significant expansion, so if there is
disagreement about the new formatters I think the help commands
are still useful on their own and can be split off into a separate request.

The purpose of these formatters is to that the user doesn't need to
break their flow to make these additional specifications.

I have found in my own personal workflow that I was often wishing for
these commands, and now that I am used to them it has significantly
sped up common tasks. Talon also often misheard space, so that
solves that issue.

Wherever possible I tried to map word formatters to prose formatters.
E.g., `walk` and `hike` are trailing space formatters for word and prose
respectively.

I want to call out a few duplicate formatters.

`trot` = `throw`, `gallop` = `hurl`

I thought that `throw` and `hurl` can better match the motion
naming whereas `trot` and `gallop` randomly introduced horse words.
Nonetheless I did not want to remove `trot` for current users,
so I also added `gallop` for an equivalence.

I also introduced `begin` as an alternative to `sentence`
as talon frequently misheard sentence.


core/help/help.py
```diff
@@ -657,17 +657,7 @@ def gui_list_help(gui: imgui.GUI):
...
 
-    if total_page_count == 0:
-        page_info = "empty"
-    else:
-        page_info = f"{current_list_page}/{total_page_count}"
-
-    gui.text(f"List: {selected_list} ({page_info})")
-
-    # Extract description from list declaration, i.e. mod.list(..., desc=...))
-    if (desc := registry.decls.lists[selected_list].desc) is not None:
```
FireChickenProductivity, line 672: Your pull request seems to have removed the help code that gets the list description from the registry and displays it. Can you put that back?

BlueDrink9:
Personally, I am not  a fan of the new formatters - at least, not mixing them in with the existing ones. Just looking at them, I found it difficult to distinguish between their subtle differences, and there are a lot of them. The cost of adding a bunch of new formatters is not actually zero, because it significantly adds to the sense of overwhelm that (especially newer) users get when they look at the available formatter list. There are already a decent number to learn.

Possibly more importantly, it doesn't look like any of them save more than 1-3 'keystrokes', which is in contrast to formatters which save 1-2 _per word_. This is especially the case for the trailing ones, which would be better served as custom phrase enders. I don't think it is that much more difficult to say "say xyz spacer" instead of "gallop xyz", or "space say xyz" rather than "hike xyz", so IMO it isn't worth the cost
 
 Perhaps this could be added as a separate file with an optional tag to enable it? At least for the ones that actually save significant time

davidsgrogan:
Counterpoint: my version of "hike" broke a few months ago and I've been procrastinating putting it back. I'd love to see it in upstream. I find "space say xyz" noticeably more cognitively burdensome than "hike xyz".

core/help/help.py
FireChickenProductivity: On second thought, you might want to wait until you get feedback from the maintainers before making additional changes.

FireChickenProductivity:
We did not get to this pull request in today's backlog session as we largely focused on the wiki.

I am not sure if having so many formatters in community would make sense or not, but I find myself gradually expanding my personal set of formatter commands for the sake of convenience. Making common tasks one less word adds up over time. I think having an extended set available would help more advanced users.

I am curious. What is the advantage for you for having these as part of the formatter system instead of as separate commands?

mpinkert:
Talon has an accepted way of formatting strings in a particular manner. You can always make a new place to do the same thing but it is cognitively easier for a person to understand and find where a thing is if it is all in one place. While I was learning I frequently used the help command to list out the different formatters and even still do so on occasion. Perhaps that is a sign that this is a little bit too much cognitive overhead at startup, but I have found it helping out significantly over the long run.

core/formatters/formatters.py
```diff
@@ -445,35 +441,70 @@ def formatters_reformat_selection(formatters: str):
...
             name = formatters[phrase]
             demo = format_text_without_adding_to_history("one two three", name)
-            if phrase in prose_formatter_names:
-                phrase += " *"
+
             formatters_help_demo[phrase] = demo
         return formatters_help_demo
 
+    def get_prose_formatter_words() -> dict:
+        """Returns words currently used as prose, and a demonstration string using those formatters"""
```
nriley, line 457:
```suggestion
        """Returns words currently used as prose formatters, and a demonstration string using those formatters"""
```

core/formatters/formatters.py
```diff
@@ -445,35 +441,70 @@ def formatters_reformat_selection(formatters: str):
...
+        prose_formatters = get_prose_formatters()
+        for phrase in prose_formatters:
+            name = prose_formatters[phrase]
+            demo = format_text_without_adding_to_history("one two three", name)
+
+            formatters_help_demo[phrase] = demo
+        return formatters_help_demo
+
+    def get_word_formatter_words() -> dict:
+        """In returns words currently used as word formatters."""
```
nriley, line 468:
```suggestion
        """Returns words currently used as word formatters."""
```

core/help/help.py
nriley: Please put this back.

core/help/help.py
```diff
@@ -122,11 +121,8 @@ def update_operators_text():
                     # If the operator is implemented as text insertion,
                     # display the operator text
                     operator = operators.get(operator_text)
-                    if isinstance(operator, str):
+                    if type(operator) == str:
                         text = ": " + operator
-                    # If a documentation string is available, use that
-                    elif doc_string := getattr(operator, "__doc__", None):
-                        text = ": " + doc_string
                     # Otherwise display the operator name from list
```
nriley, line 126: Please revert this change, thanks.

nriley (changes requested):
From the community backlog session — thank you for splitting out the infrastructural changes from the individual formatters. This is looking pretty good if you could address the issues where you may have inadvertently reverted some changes that had happened to the formatter help code since you started working on this.

PR #2006, opened 2025-08-26
Fix snippet substitution issue

FireChickenProductivity:


dsheyp:
I briefly tested this PR and found that it fixed [cursorless issue #3075](https://github.com/cursorless-dev/cursorless/issues/3075)

FireChickenProductivity:
Testing Summary:
I have tested my changes to:
sql.py
rust.py
r.py
lua.py
css.py

Stuff I have tested with snippets (using Java):
Wrapping: 
"try wrap line"
"if wrap line" (tests $0 with wrapping)
"if cond wrap line" (tests $1 with wrapping)

Modified snippet insertion:
"snip if"
"snip else"
"snip return"
"snip ternary" (corner case: $0 already at the end)
"snip class"
"snip class test name"
"snip break" (corner case: no snippet number present)
"snip ternary" with the body I modified to "$1 ? $2 : $3" (corner case: no $0 with snippet numbers present)

Original snipped insertion actions:
"snip if"
"snip else"
"snip return"
"snip ternary"
"snip class"
"snip class test name"
"snip break"

nriley (changes requested):
If you could just add a quick comment in BREAKING_CHANGES to mention the change in default behavior and how you can return to it either for a single snippet or for all snippets as desired comment that would be great. Thanks.

core/snippets/snippets_parser.py
```diff
@@ -151,30 +150,38 @@ def combine_variables(
     return list(variables.values())
 
 
-def add_final_stop_to_snippet_body(
-    body: str, variables: list[SnippetVariable]
-) -> tuple[str, list[SnippetVariable]]:
+def add_final_stop_to_snippet_body(body: str) -> str:
     """Make the snippet body end with stop $0 to allow exiting the snippet with `snip next`.
     If the snippet has a stop named `0`, it will get replaced with the largest number of a snippet variable name
     plus 1 with the original variable metadata for stop `0` now associated with the replacement.
```
nriley: Please clarify this comment. Thanks!

nriley (approved):

PR #2007, opened 2025-08-29
Add initial cplusplus type support

FireChickenProductivity:
Thank you to chdoc for the cpp_standard_type list!

This makes c_types defined with cplusplus and add support for types from the standard library.


lang/cpp/cpp.py
```diff
@@ -3,14 +3,29 @@
...
 from ..tags.operators import Operators
 
+mod = Module()
 ctx = Context()
 
 ctx.matches = r"""
 code.language: cpp
 """
 
+mod.list("cpp_standard_type", desc="Types from the cplusplus standard library")
```
nriley:
```suggestion
mod.list("cpp_standard_type", desc="Types from the C++ standard library")
```

nriley (approved):

PR #2012, opened 2025-09-06
cplusplus add null data support

FireChickenProductivity:
Given that nullptr is the standard way to deal with the concept of "null" in modern c++, I am using that for the standard null actions. The current intent is to have separate commands for dealing with NULL.

### DIFF
--- a/lang/cpp/cpp.py
+++ b/lang/cpp/cpp.py
@@ -34,6 +34,12 @@ class UserActions:
     def code_insert_is_not_null():
         actions.auto_insert(" != nullptr")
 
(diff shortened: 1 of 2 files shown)


auscompgeek:
Hm, I wanted to turn this into a list, see #1255

FireChickenProductivity:
> Hm, I wanted to turn this into a list, see #1255

I think this should be a list for languages that have multiple concepts of null. I plan on discussing that next backlog session when we have time.

nriley (approved):

PR #2013, opened 2025-09-06
cplusplus add c NULL commands

FireChickenProductivity:
These commands add support for working with the old C way of dealing with null. There is a separate pull request for supporting the modern way of doing this in C++. #2012 

What do you think of this implementation @chdoc?

I would use "c" instead of "see", but I expect this to generally get better recognition.


chdoc:
Instead of separate commands, I would favor a toggle changing the meaning of the regular commands.

FireChickenProductivity:
> Instead of separate commands, I would favor a toggle changing the meaning of the regular commands.

I have not had to deal with NULL. Would you prefer this in a static setting or the ability to change behavior dynamically with a command?

Do you think having separate commands in addition to that would be problematic? I would personally prefer to have a consistent mapping between what I say and what gets inserted for a given language.

chdoc:
Coming to think of it, I do not anticipate to ever use this in C++, where `nullptr` is the preferred way since C++11. So I suppose this is "good enough". 

The situation is very different for C, where `nullptr` was only introduced with C23. Given that a lot of C/C++ code is developed in settings that do not allow language features newer than a decade, for C a verbal toggle with a persistent setting for the default might be the way to go. Then C++ could reuse this with the default being `nullptr` instead of `NULL`. But this is a discussion for another day.

nriley:
From the community backlog session — the existing nullptr support should work for C++11 and people are using things like ancient embedded compilers then we could potentially turn null into a list.

PR #2014, opened 2025-09-06
cpp draft object oriented support

FireChickenProductivity:
This is a little complicated by accessing self in C++ being `this->` or `*(this).` instead of `this.`, but the standard command is `self dot`. I made questionable choices to accommodate the existing spoken form and the different things you do with `this` in C++.

I made `self dot` do `*(this).` to be consistent with the name of the command while actually accessing the object properly. I made `self dot` now use an overwritable action that defaults to the prior behavior. 

I added custom actions for C++ based on common ways to interact with the `this` keyword. 

```talon
self arrow: insert("this->")
(star | dereference) self: insert("*this")
```

@chdoc do you think this is a reasonable way to handle this? Do you have alternative suggestions?

### DIFF
--- a/lang/cpp/cpp.py
+++ b/lang/cpp/cpp.py
@@ -1,4 +1,4 @@
-from talon import Context, Module
+from talon import Context, Module, actions
 
 from ..c.c import operators
 from ..tags.operators import Operators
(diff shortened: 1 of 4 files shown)


chdoc:
I'd prefer "self dot" yielding `this->`. It is basically unavoidable that uniform commands yield things like this. A bit like "op equal" yielding `<-` in R. After all "this" and "that" are already taken.

I don't envision to ever want to enter `(*this).` , so I would prefer to not override the standard command with that. 

I'm okay with the "dereference self".

nriley (approved):

PR #2016, opened 2025-09-07
Escape/unescape $ in snippet substitutions (fixes #2015)

### LINKED ISSUE #2015: Support snippet escaping $ outside VSCode
\$ works in VSCode for escaping $ but does not work with raw snippet insertion.

nriley:
Fixes #2015


AndreasArvidsson:
I assume you have tested that the dollar escapement works find both in vscode and outside.

nriley:
Yes! It works great in both. Here's the snippet I used to test: https://github.com/nriley/talon_community/blob/1b339fa2540355bd109e6c38e6633c502408b059/core/snippets/snippets/switchStatement.snippet#L76

PR #2017, opened 2025-09-07
Properly position the cursor at the first stop in some cases when inserting snippets without editor support

nriley:
Applies when:

- there is text on the line after the inserted snippet
- the first stop in the snippet is on the last (or only) line

Also documents an assumption in `move_to_correct_column` that caused the incorrect behavior.

Per discussion on Slack, we may want to change the behavior of `move_to_correct_column` to not go to the end of the line, but this involves different tradeoffs that should be discussed further.

### DIFF
--- a/core/snippets/snippets_insert_raw_text.py
+++ b/core/snippets/snippets_insert_raw_text.py
@@ -21,7 +21,7 @@ mod.setting(
     desc="""If true, inserting snippets as raw text will always be done through pasting""",
 )
 
-RE_STOP = re.compile(r"\$(\d+|\w+)|\$\{(\d+|\w+)\}|\$\{(\d+|\w+):(.+)\}")
+RE_STOP = re.compile(r"\\?(?:\$(\d+|\w+)|\$\{(\d+|\w+)\}|\$\{(\d+|\w+):(.+)\})")
 LAST_SNIPPET_HOLE_KEY_VALUE = 1000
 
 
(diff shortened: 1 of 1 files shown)


core/snippets/snippets_insert_raw_text.py
```diff
@@ -21,7 +21,7 @@
     desc="""If true, inserting snippets as raw text will always be done through pasting""",
 )
 
-RE_STOP = re.compile(r"\$(\d+|\w+)|\$\{(\d+|\w+)\}|\$\{(\d+|\w+):(.+)\}")
+RE_STOP = re.compile(r"\\?(?:\$(\d+|\w+)|\$\{(\d+|\w+)\}|\$\{(\d+|\w+):(.+)\})")
```
AndreasArvidsson: This is now complicated enough that it probably warrants a comment breaking down the different parts.

AndreasArvidsson (approved):

core/snippets/snippets_insert_raw_text.py
nriley: Adding comments also made me realize I messed this up. Not sure how it worked in my testing! Updated version incoming...

nriley: Agreed! Commenting the regex made me realize I was doing unnecessary work when matching a `\$`.

PR #2020, opened 2025-09-08
Windows Explorer Integration: Added support for other languages & tabs (Win 11) + code refactoring

xavier-ateo:
The file picker GUI for Explorer did not work for me on Windows 11 24H2. I have addressed several issues:

**Multiple language support**
Previously, the Windows Explorer integration was only targeted towards English as the operating system language. It now also supports German, and other languages can be easily added.

**Windows 11 tabs**
Path detection was broken when multiple tabs were open in one Explorer window. This is now fixed for English and German.

**Logging and code refactoring**
Logging can be easily enabled to debug issues and figure out which format a specific language uses when you want to add support for it.

**Todo**
- I took over the previously existing logic for determining OneDrive user folders. However I did not test if it still works as intended.
- I also did not test it under Windows 10.


apps/windows_explorer/windows_explorer.py
```diff
@@ -1,7 +1,51 @@
...
+import re
 
 from talon import Context, Module, actions, app, ui
 
+# Configuration
+DEBUG_LOGGING = False
+
+# Configure logging
+if DEBUG_LOGGING:
+    logging.basicConfig(level=logging.DEBUG, format="%(levelname)s: %(message)s")
```
lunixbochs: remove this please, talon already enables logging

apps/windows_explorer/windows_explorer.py
```diff
@@ -29,125 +73,326 @@
...
+        },
+        "de": {
+            "suffix_patterns": [
+                r" und [0-9]+ weitere Registerkarten – Explorer$",
+                r" und 1 weitere Registerkarte – Explorer$",
+                r" – Datei-Explorer$",
+            ],
+            "folder_names": {  # Right side is the localized name
+                "Desktop": "Desktop",
+                "Documents": "Dokumente",
```
lunixbochs: surely there's a win32 api to get the names for these folders rather than hardcoding per language

apps/windows_explorer/windows_explorer.py
```diff
@@ -1,7 +1,51 @@
...
+"""
+
+import logging
 import os
+import re
 
 from talon import Context, Module, actions, app, ui
 
+# Configuration
+DEBUG_LOGGING = False
```
nriley: In general, we tend to use Talon settings rather than constants in the code to configure logging verbosity.

apps/windows_explorer/windows_explorer.py
nriley: If you want to declare a logger that is private to your module you can use something like https://github.com/talonhub/community/blob/14e89d69cfb8b59a0e7f403838c352249fa0c9af/core/command_client/rpc_client/rpc_client.py#L13. Then generally you can just set the log level based on the Talon setting as above and the overhead from constructing log messages will be negligible when compared with the verbosity repeated checks add to your code.

apps/windows_explorer/windows_explorer.py
```diff
@@ -29,125 +73,326 @@
...
+                logging.debug(f"User mapping: {user_display_name} -> {user_path}")
+
+        return mappings
+
+    # Initialize directory mappings and exclusions
+    directories_to_remap = build_directory_mappings(
+        current_language_mapping["folder_names"], get_folder_path, user_display_name
+    )
+
+    directories_to_exclude = [""] + current_language_mapping["excludes"]
```
nriley: Can you explain why you have an empty string in this list?

apps/windows_explorer/windows_explorer.py
```diff
@@ -29,125 +73,326 @@
...
+        actions.user.file_manager_hide_pickers()
+        return ""
+    return path
+
 
 @ctx.action_class("user")
 class UserActions:
 
     def file_manager_open_parent():
+        """Navigate to parent directory."""
```
nriley: In general, we do not put docstrings in a context implementation, just in the module definition of an action.

nriley (changes requested):

apps/windows_explorer/windows_explorer.py
xavier-ateo: It was part of the original implementation, therefore I kept it. I don't know if it is needed or not.

```
    directories_to_exclude = [
        "",
        "Run",
        "Task Switching",
        "Task View",
        "This PC",
        "File Explorer",
        "Program Manager",
```

apps/windows_explorer/windows_explorer.py
xavier-ateo: The original implementation also had docstrings. Should I remove all of them?

apps/windows_explorer/windows_explorer.py
xavier-ateo: You are right, there is a win32 API. I will use that instead of the hardcoded values for the user folder names.

apps/windows_explorer/windows_explorer.py
xavier-ateo: I'll do that, thanks

apps/windows_explorer/windows_explorer.py
nriley: Please, as long as there is nothing missing from the module docstring.

nriley: Please, as long as there is nothing missing from the module docstring.

apps/windows_explorer/windows_explorer.py
```diff
@@ -29,59 +91,380 @@
...
+            ("Data2", ctypes.c_ushort),
+            ("Data3", ctypes.c_ushort),
+            ("Data4", ctypes.c_ubyte * 8),
+        ]
+
+    guid_string = guid_string.strip("{}").replace("-", "")
+    guid = GUID()
+    guid.Data1 = int(guid_string[0:8], 16)
+    guid.Data2 = int(guid_string[8:12], 16)
+    guid.Data3 = int(guid_string[12:16], 16)
```
nriley, line 165: The Python `uuid.UUID` object can give you these with `.time_low`, `.time_mid` and `.time_hi_version`. I don't see a way to get 8 individual bytes so you'll have to do that yourself!

nriley (changes requested):
From the community backlog session — apologies for the delay in review, as we did not realize that you had made some additional changes. This is looking really good! Do you want to have us consider merging it soon and make the other changes in a future PR?

apps/windows_explorer/windows_explorer.py
```diff
@@ -92,14 +475,16 @@ def file_manager_open_parent():
 
     def file_manager_current_path():
         path = ui.active_window().title
```
knausj85, line 477: For what it's worth, I switched my implementation of this function to use the below, as it provides the actual path. It also resolves https://github.com/talonhub/community/pull/1386 without using accessibility


```

def get_active_explorer_path():
    shell = win32com.client.Dispatch("Shell.Application")
    active_window = ui.active_window()
    for window in shell.Windows():
        try:
            if window.HWND == active_window.id:
                path = window.Document.Folder.Self.Path
                print(f"{path}")
                return path
        except Exception:
            continue

    return None
```

I wonder if (a) it works properly for other languages and (b) might be more robust and simplify the overall logic here. For e.g. the Desktop folder, it provides the full path. For "special" folders like "This PC," it returns a UUID.

The caveat is that I currently use the [dynamic list](https://github.com/talonhub/community/pull/1621) implementation, so I don't know how well this would play with the title change approach used by the current implementation. There could be timing issues.

nriley:
From the community backlog session — labeling this as unfortunately it contains dependencies that are not currently compatible with beta Talon using Python 3.14t. Our tracking bug for this is #2190.

PR #2021, opened 2025-09-09
C++ support: base types, standard functions and heater files, namespaces

chdoc:
This is mostly the result of salvaging functionality from #1820 . I didn't want to break things up into a flurry of individual pull-request, but I did break up the individual features into individual commits.

This shouldn't change any existing functionality except of course he existing "type" command for C++, which becomes a lot more powerful. So I would hope that this can get merged with minimal fuss.

CC: @FireChickenProductivity

### DIFF
--- a/lang/c/c.py
+++ b/lang/c/c.py
@@ -1,3 +1,4 @@
+from contextlib import suppress
 from talon import Context, Module, actions, settings
 
 from ...core.described_functions import create_described_insert_between
@@ -27,6 +28,13 @@ c_and_cpp_ctx.lists["self.stdint_signed"] = {
     "you": "u",
 }
 
+c_and_cpp_ctx.lists["self.c_type_bit_width"] = {
+    "eight": "8",
+    "sixteen": "16",
+    "thirty two": "32",
+    "sixty four": "64",
+}
+
 c_and_cpp_ctx.lists["self.c_signed"] = {
     "signed": "signed",
     "unsigned": "unsigned",
@@ -54,15 +62,18 @@ c_and_cpp_ctx.lists["self.c_types"] = {
     "char": "char",
     "short": "short",
     "long": "long",
+    "long long": "long long",
     "int": "int",
     "integer": "int",
     "void": "void",
     "double": "double",
+    "long double": "long double",
     "struct": "struct",
     "struck": "struct",
     "num": "enum",
     "union": "union",
     "float": "float",
+    "size tea": "size_t",
 }
 
 ctx.lists["user.code_libraries"] = {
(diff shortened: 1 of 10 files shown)


lang/cpp/cpp.py
```diff
@@ -20,12 +34,60 @@ def cpp_standard_type(m) -> str:
...
 
+@mod.capture(rule="{user.cpp_standard_prefix} {user.cpp_standard_function}")
+def cpp_standard_function(m) -> str:
+    # Discard the prefix word and insert properly prefixed function call
+    return "std::" + m[1]
+
+
+@mod.capture(rule="{user.cpp_standard_prefix} {user.cpp_standard_constant}")
+def cpp_standard_constant(m) -> str:
+    # Discard the prefix word and insert properly prefixed function call
```
FireChickenProductivity: This comment says this is inserting a properly prefixed function call even though it is for a constant

lang/cpp/cpp.py
```diff
@@ -20,12 +34,60 @@ def cpp_standard_type(m) -> str:
...
+
+
+@mod.capture(
+    rule="([<user.c_signed>] <user.c_types>) | <user.c_fixed_integer> | <user.cpp_standard_type>"
+)
+def code_type_raw(m) -> str:
+    return " ".join(list(m))
+
+
+@mod.capture(rule="<user.code_type_raw> [{user.cpp_pointers}+] <user.text>")
```
FireChickenProductivity: We typically put something like formatting a variable in a command rather than a capture.

FireChickenProductivity:
I think this looks great. I do not see anything wrong with the actual functionality. Given your limited availability, I think this can be merged as is and I could make a followup pull request if needed to get the implementation more aligned with typical community patterns.

lang/c/c.py
```diff
@@ -27,6 +29,13 @@
     "you": "u",
 }
 
+c_and_cpp_ctx.lists["self.c_type_bit_width"] = {
```
FireChickenProductivity, line 32: The use of `self.` is an outdated pattern that we are trying not to use in newer code. You should be able to use `user.` instead.

lang/cpp/cpp.py
chdoc: Well, what can I say, copy paste error.

lang/cpp/cpp.py
chdoc: I was hoping to factor out the common logic of this capture and the `code_type` capture. However, naively passing the parts doesn't work because of the use of `supress`. Left this for later cleanup. If one capture is inlined into a command, the two will never be unified.

lang/c/c.py
chdoc: Noted. But, for now, I think it lives in good company with other old uses of `self.`. Changing all in one go is probably preferable.

lang/c/c.py
```diff
@@ -90,6 +102,17 @@
...
+
+
+# capture explicitly referenced from the C++ files
+@mod.capture(rule="(fix|fixed) [{self.stdint_signed}] [int] {self.c_type_bit_width}")
+def c_fixed_integer(m) -> str:
+    """fixed-width integer types (e.g. "uint32_t")"""
+    prefix = ""
+    with suppress(AttributeError):
+        prefix = m.stdint_signed
+    return prefix + "int" + m.c_type_bit_width + "_t"
```
nriley:
```suggestion
    return f"{prefix}int{m.c_type_bit_width}_t"
```

lang/cpp/cpp.py
```diff
@@ -10,22 +12,82 @@
...
+def variable_declaration(m) -> str:
+    suffix = ""
+    with suppress(AttributeError):
+        suffix = "".join(m.cpp_pointers_list)
+    return (
+        m.code_type_raw
+        + " "
+        + suffix
+        + actions.user.formatted_text(m.text, "SNAKE_CASE")
```
nriley: Please use the formatter specified in `user.code_private_variable_formatter`.

nriley (changes requested):

lang/cpp/cpp.py
```diff
@@ -10,22 +12,82 @@
...
+    )
```
nriley: Please move the formatting of the variable name itself into a .talon file and take it out of the capture. We may eventually want to use a setting for SNAKE_CASE but for the moment you can include it just as the `var` command in PHP does.

nriley (approved):

PR #2022, opened 2025-09-13
Add c++ keywords

FireChickenProductivity:


chdoc:
> If you have time, could you let me know what you think of the spacing and what should go in the unprefixed list?

I don't really have any strong opinions on the spacing. As it comes to the list of prefixed keywords, I would include those keywords that are verbally distinctive, commonly used, don't have any other commands inserting them yet, and are not accompanied by braces or other structure elements. For me, I suppose the most important ones would be `static`, `constexpr`, and `const`. `namespace` should probably be a snippet. The word that I am I'm sure about is `auto`. Personally, I have added
```
var auto <user.text>: 
    insert("auto ")
    user.code_private_variable_formatter(text)
    insert(" = ")
```
for auto-typed variable declarations, so I wouldn't include it in the unprefixed list.

FireChickenProductivity:
> > If you have time, could you let me know what you think of the spacing and what should go in the unprefixed list?
> 
> I don't really have any strong opinions on the spacing. As it comes to the list of prefixed keywords, I would include those keywords that are verbally distinctive, commonly used, don't have any other commands inserting them yet, and are not accompanied by braces or other structure elements. For me, I suppose the most important ones would be `static`, `constexpr`, and `const`. `namespace` should probably be a snippet. The word that I am I'm sure about is `auto`. Personally, I have added
> 
> ```
> var auto <user.text>: 
>     insert("auto ")
>     user.code_private_variable_formatter(text)
>     insert(" = ")
> ```
> 
> for auto-typed variable declarations, so I wouldn't include it in the unprefixed list.

Thank you for your feedback. I added the keywords you suggested and `explicit` and `virtual`.

lang/cpp/code_keyword.talon-list
```diff
@@ -0,0 +1,109 @@
...
+auto: "auto "
+bit and: " bitand "
+bit or: " bitor "
+bool
+break
+case: "case "
+catch: "catch "
+char
+char eight tea: "char8_t"
+char sixteen tea: "char16_t"
```
AndreasArvidsson:
```suggestion
char eight tea: char8_t
char sixteen tea: char16_t
```

lang/cpp/code_keyword.talon-list
```diff
@@ -0,0 +1,109 @@
+list: user.code_keyword
+code.language: cpp
+-
+asm: asm
+aye es em: asm
+align as: "alignas"
```
AndreasArvidsson:
```suggestion
align as: alignas
```

lang/cpp/code_keyword.talon-list
```diff
@@ -0,0 +1,109 @@
+list: user.code_keyword
+code.language: cpp
+-
+asm: asm
+aye es em: asm
+align as: alignas
+a line as: "alignas"
+align of: "alignof"
+a line of: "alignof"
```
AndreasArvidsson:
```suggestion
align of: alignof
a line of: alignof
```

lang/cpp/code_keyword.talon-list
```diff
@@ -0,0 +1,109 @@
+list: user.code_keyword
+code.language: cpp
+-
+asm: asm
+aye es em: asm
+align as: alignas
+a line as: "alignas"
```
AndreasArvidsson:
```suggestion
a line as: alignas
```

AndreasArvidsson (approved):

PR #2026, opened 2025-09-14
Fix Lua's tryCatchStatement so it can be used to wrap arbitrary code

nriley:
Move it into tryCatchStatement.snippet with other implementations.

Uses the example from https://www.lua.org/pil/8.4.html.

Otherwise, try wrap in VSCode breaks due to the snippet not having $1.

Also add better information about *which* snippet is breaking wrapping, as it took me a while to realize that another language implementation was causing the problem.y

### DIFF
--- a/core/snippets/snippet_types.py
+++ b/core/snippets/snippet_types.py
@@ -47,7 +47,7 @@ class Snippet:
     def get_variable_strict(self, name: str):
         variable = self.get_variable(name)
         if variable is None:
-            raise ValueError(f"Snippet '{self.name}' has no variable '{name}'")
+            raise ValueError(f"No variable '{name}' in {self}")
         return variable
 
 
--- a/core/snippets/snippets/lua.snippet
+++ b/core/snippets/snippets/lua.snippet
@@ -4,6 +4,7 @@ language: lua
 name: forInIPairs
 phrase: for eye pairs
 insertionScope: statement
+
 $1.insertionFormatter: SNAKE_CASE
 -
 for _, $1 in ipairs($2) do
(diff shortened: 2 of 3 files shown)


core/snippets/snippet_types.py
```diff
@@ -47,7 +47,7 @@ def get_variable(self, name: str):
     def get_variable_strict(self, name: str):
         variable = self.get_variable(name)
         if variable is None:
-            raise ValueError(f"Snippet '{self.name}' has no variable '{name}'")
+            raise ValueError(f"No variable '{name}' in {self}")
```
AndreasArvidsson, line 50: Will `{self}` give a good serialization? It's an class instance.

nriley: It's a dataclass, so yes — it looks quite readable. It looks like this:

```py
Snippet(name='tryCatchStatement', body='if pcall(function () $1 end) then\n\t$2\nelse\n\t$0\nend', description=None, phrases=['try catch'], insertion_scopes=['statement'], languages=['lua'], variables=[SnippetVariable(name='1', insertion_formatters=None, wrapper_phrases=['try'], wrapper_scope='statement'), SnippetVariable(name='0', insertion_formatters=None, wrapper_phrases=['catch'], wrapper_scope='statement')])
```

AndreasArvidsson: And you think that is an improvement compared to just the name? We could always add the language if that's what you needed?

nriley: It being just the name was the issue I was trying to fix — there was a broken lua tryCatch implementation, in a different file from all the other ones, that was preventing me from wrapping my Python code in a snippet, and the error message was so unhelpful it took me 10 minutes tracing through code to figure out what was wrong.

Ideally we'd include the filename of the snippet file, but I figured having extra info would be useful so you could easily locate it.

AndreasArvidsson: Ah of course we load all of them now.
Filename sounds reasonable but not data we have right now. In fine with this.

AndreasArvidsson (approved):

PR #2028, opened 2025-09-16
Cleanup c.py file

FireChickenProductivity:
Stop using `self.` and `"Returns a string"`. 
Use snippet system for importing standard header files and inserting function calls instead of user.paste.


lang/c/c.py
```diff
@@ -115,53 +115,53 @@ def c_fixed_integer(m) -> str:
...
+    """A C pointer"""
     return m.c_pointers
 
 
 # capture explicitly referenced from the C++ files
-@mod.capture(rule="{self.c_signed}")
+@mod.capture(rule="{user.c_signed}")
 def c_signed(m) -> str:
-    "Returns a string"
+    """Used for prefixing a signed or unsigned type"""
```
chdoc, line 127: I'm not a fan of the "used for". Unless there is already and agreed-upon pattern I would favor just briefly describing the output, e.g. "signed or unsigned prefix for types". But then, anything is better than "Returns a string". 😅

AndreasArvidsson (changes requested):
Waiting for update off doc strings

FireChickenProductivity:
> Waiting for update off doc strings

I updated the documentation strings.

AndreasArvidsson (approved):

PR #2031, opened 2025-09-16
Migrate c to talon list files

FireChickenProductivity:


lang/c/c_pointers.talon-list
```diff
@@ -0,0 +1,6 @@
+list: user.c_pointers
+code.language: c
+code.language: cpp
```
chdoc, line 3: Please double check, but I do not think that C++ actually makes use of this list. C++ has its own list that includes references and the likes.

lang/c/stdint_types.talon-list
```diff
@@ -0,0 +1,18 @@
+list: user.stdint_types
+code.language: c
+code.language: cpp
```
chdoc, line 3: Please double check, but I do not think that C++ actually makes use of this list.

AndreasArvidsson:
```suggestion
```

lang/c/c_pointers.talon-list
AndreasArvidsson:
```suggestion
```

AndreasArvidsson (approved):

PR #2038, opened 2025-09-23
Fixed the .target DeprecationWarning in help.py by changing it to .script

caroby:


FireChickenProductivity:
Last I checked, the plan was not to remove the deprecated talon.skia imports until the non-deprecated import got added to non beta talon. Trying to import skia does not work for me without beta.

caroby:
I took out the second commit that touches the talon.skia imports, but kept the val.script.code. Does that work with non-beta? @FireChickenProductivity

PR #2039, opened 2025-09-24
Add more contributing guidelines

FireChickenProductivity:


CONTRIBUTING.md
```diff
@@ -11,3 +11,6 @@ Each of the principles is numbered for easy referencing. The body is formatted a
...
+- P07 - Make atomic pull requests unless you have a very good reason not to. Separating code that does not depend on eachother into separate pull requests makes peer review easier and can lead to changes getting merged faster. Large pull requests trying to do too much at once often lose momentum and become abandoned in peer review because there are so many pieces that need to be properly reviewed before the entire thing can go in. A pull request could get delayed because no one who made it to that week's backlog session has some relevant experience or knowledge needed to properly review a small part of it. A large pull request can consequently get delayed for a while even if most of it is fine.
```
chdoc:
```suggestion
- P07 - Try to create pull requests that focus on a single feature at a time. Separating pieces of code that do not depend on each other into separate pull requests makes peer review easier and will usually lead to changes getting merged faster. Large or unfocused pull requests trying to do too much at once often lose momentum and become abandoned in peer review. The reason for this is usually that there are too many pieces that need to be properly reviewed before the entire thing can be merged. Smaller, more focused pull requests allow uncontroversial changes to merged quickly while facilitating more focused discussions for those changes that do merit it. 
```
I tried to rephrase this in terms of potential positives instead of potential negatives.

chdoc: TIL: the suggest changes feature of GitHub does not deal well with lines that encompass an entire paragraph 🙄

FireChickenProductivity: I think this is a better way of wording this.

CONTRIBUTING.md
```diff
@@ -11,3 +11,6 @@ Each of the principles is numbered for easy referencing. The body is formatted a
 
 - P03 - Use the `app.bundle` matcher for apps on OSX. This is the least ambiguous way of referring to a particular program.
 - P04 - Use both `app.name` and `app.exe` matchers for apps on Windows. That is the context should OR together one matcher of each type. Apparently the [MUICache](https://www.magnetforensics.com/blog/forensic-analysis-of-muicache-files-in-windows/) can break, perhaps making one of these matchers stop working.
+- P05 - Use `.talon-list` files for talon lists unless there is a good reason not to. `.talon-list` files are easier for users to edit — especially users who are not programmers. If a list needs to be constructed or referenced through Python code, it may make sense to instead leave it in Python.
```
nriley:
```suggestion
- P05 - Use `.talon-list` files for Talon lists unless there is a good reason not to. `.talon-list` files are easier for users to edit — especially users who are not programmers. If a list needs to be constructed or referenced through Python code, it may make sense to instead leave it in Python.
```

PR #2040, opened 2025-09-27
Improve java generics

FireChickenProductivity:
Make it easier to insert java generic types.
e.g.:
built in types: "list of string"→`List<String> `
user type with the type prefix: "type user type of string"→`UserType<String> `
multiple type parameters: "map of string and integer"→`Map<String, Integer> `
user type as a parameter does not require the type prefix: "list of user type"→`List<UserType> `

After we decide on how we want the functionality, I should add a comment to the Talon file explaining this.


lang/java/java.py
```diff
@@ -131,6 +133,36 @@
 )
 
 
+def public_camel_case_format_variable(variable: str):
+    return actions.user.formatted_text(variable, "PUBLIC_CAMEL_CASE")
+
+
+@mod.capture(rule="{user.java_boxed_type} | <user.text>")
+def java_type_parameter_argument(m) -> str:
+    """A java type parameter for a generic data structure"""
```
nriley:
```suggestion
    """A Java type parameter for a generic data structure"""
```

lang/java/java.py
```diff
@@ -131,6 +133,36 @@
...
+    with suppress(AttributeError):
+        return m.java_boxed_type
+    return public_camel_case_format_variable(m.text)
+
+
+@mod.capture(rule="[type] {user.java_generic_data_structure} | type <user.text>")
+def java_generic_data_structure(m) -> str:
+    """A java generic data structure that takes tight parameter arguments"""
```
nriley:
```suggestion
    """A Java generic data structure that takes tight parameter arguments"""
```

nriley:
Waiting for a little feedback from someone who has used Java in the past ten years :)

FireChickenProductivity:
> My main problem with this is that it's a lot of code that is Java specific. I would much rather that the grammar and captures were generic and there was only a single action per language to actually format it correctly. We have finally begun to move things away from the language python files and I'm hesitant to start making the language implementations more complex again.

My current plan is for us to have support for several languages before trying to abstract out the general grammar. There is considerable variation between languages here. In some languages,  (including rust and c++) you want to reference name spaces in some of your types. c++ type parameters can include constants and not just subtypes. Are you fine with this being language specific for the short term?

lang/java/java.py
```diff
@@ -131,6 +133,44 @@
...
+def java_type_parameter_argument(m) -> str:
+    """A Java type parameter for a generic data structure"""
+    with suppress(AttributeError):
+        return m.java_boxed_type
+    return public_camel_case_format_variable(m.text)
+
+
+@mod.capture(rule="[type] {user.java_generic_data_structure} | type <user.text>")
+def java_generic_data_structure(m) -> str:
+    """A Java generic data structure that takes tight parameter arguments"""
```
nriley:
```suggestion
    """A Java generic data structure that takes type parameter arguments"""
```

PR #2045, opened 2025-10-01
Added spelling prose capture

AndreasArvidsson:


nriley:
From the community backlog session (with @AndreasArvidsson this time) — we discussed a plan to expand such that we create a capture for spelling, replace commands such as in keys.talon, and add comments so it is still easy to understand what is going on. We also talked about potentially deprecating/removing multiple ways of dictating uppercase letters — uppercase, over and lowercase.

core/text/text_and_dictation.py
```diff
@@ -114,6 +114,12 @@ def prose_time(m) -> str:
     return str(m)
 
 
+@mod.capture(rule="spell <user.letters>")
+def prose_spell(m) -> str:
+    """Spell word phoneticly"""
```
nriley:
```suggestion
    """Spell word phonetically"""
```

PR #2046, opened 2025-10-06
Use shorter spoken form for bitwise operators

FireChickenProductivity:
Use `(bit|bitwise)` instead of `op bitwise` as the prefix for bitwise operators. 

I am waiting on advice for how to handle deprecation/notifying users about this change before updating breaking changes.

### DIFF
--- a/lang/tags/operators.talon
+++ b/lang/tags/operators.talon
@@ -2,7 +2,7 @@
 # so changing those commands will make the "help operators" command display the wrong prefixes
 op {user.code_operators_array}: user.code_operator(code_operators_array)
 op {user.code_operators_assignment}: user.code_operator(code_operators_assignment)
-op {user.code_operators_bitwise}: user.code_operator(code_operators_bitwise)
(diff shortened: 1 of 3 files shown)


core/help/help.py
```diff
@@ -103,9 +103,12 @@ def update_operators_text():
         operators = actions.user.code_get_operators()
 
         # Associate the names of the operator lists with the corresponding prefix
-        op_list_names = ["array", "assignment", "bitwise", "lambda", "math", "pointer"]
+        op_list_names = ["array", "assignment", "lambda", "math", "pointer"]
         names_with_prefix = [(name, "op") for name in op_list_names]
-        names_with_prefix.append(("math_comparison", "is"))
+        names_with_prefix.extend(
+            [("math_comparison", "is"), ("bitwise", "(bit | bitwise)")]
+        )
```
nriley:
```suggestion
        names_with_prefix += [("math_comparison", "is"), ("bitwise", "(bit | bitwise)")]
```

PR #2058, opened 2025-11-01
make Emacs filename detection work with the default window title

chdoc:


apps/emacs/emacs.py
```diff
@@ -356,8 +356,11 @@ def language():
 
 @ctx.action_class("win")
 class WinActions:
-    # This assumes the title is/contains the filename.
-    # To do this, put this in init.el:
-    # (setq-default frame-title-format '((:eval (buffer-name (window-buffer (minibuffer-selected-window))))))
+    # This assumes the title either is the buffer name or contains the buffer
+    # name before a space-surrounded hyphen. This is the default for the latest
+    # GNU Emacs, and can be restored by putting the following into your init.el:
+    # (setq frame-title-format '(multiple-frames "%b" ("" "%b - GNU Emacs at " system-name)))
```
nriley: Is there a reason why you switched from using `setq-default` to `setq` here?

nriley:
One more thing I noticed while investigating, might be worth mentioning: `xterm-set-window-title` is available in customize. We could potentially use it to recognize Emacs in a terminal window by this title (that I otherwise hate :-).

apps/emacs/emacs.py
```diff
@@ -356,8 +357,19 @@ def language():
...
+    # name before a space-surrounded hyphen. This is the default for the latest
+    # GNU Emacs. If you are not on macOS and your flavor of Emacs defaults to
+    # something incompatible, you may need to put one of the following two
+    # declarations into your init.el ("%b" being replaced by the buffer name):
+    # (setq frame-title-format "%b")
+    # (setq frame-title-format '(multiple-frames "%b" ("" "%b - Emacs at " system-name)))
     def filename():
-        return actions.win.title()
+        # On macOS, get the filename directly
+        if doc := getattr(ui.active_window(), "doc"):
```
auscompgeek: 2-arg getattr isn't any safer than accessing the attribute directly when the attribute name is statically known.

auscompgeek (dismissed):

apps/emacs/emacs.py
nriley: Sorry about this, my code has been broken for a long time! Will fix to use a 3-arg one.

apps/emacs/emacs.py
```diff
@@ -362,8 +363,19 @@ def language():
...
```
nriley:
```suggestion
        # On macOS, get the filename directly
        if doc := getattr(ui.active_window(), "doc", None):
```

apps/emacs/emacs.py
nriley: From the community backlog session — the difference is immaterial as `frame-title-format` can never be buffer local.

nriley:
> One more thing I noticed while investigating, might be worth mentioning: `xterm-set-window-title` is available in customize. We could potentially use it to recognize Emacs in a terminal window by this title (that I otherwise hate :-).

From the community backlog session — we decided this was out of scope for this PR.

nriley (approved):

PR #2059, opened 2025-11-01
Add Java nested type support

FireChickenProductivity:


lang/java/java.py
```diff
@@ -158,14 +159,68 @@ def java_generic_data_structure(m) -> str:
...
+    return m.java_generic_type_connector_list, m.java_type_parameter_argument_list
+
+
+@mod.capture(
+    rule="<user.java_type_parameter_argument> [<user.java_generic_type_additional_type_parameters>]"
+)
+def java_type_parameter_arguments(m) -> str:
+    """Formatted Java type parameter arguments"""
+    parameters = [m.java_type_parameter_argument]
+    try:
```
nriley: Consider using tuple unpacking and getattr to avoid [0]/[1] and the exception.

lang/java/java.py
```diff
@@ -158,14 +159,68 @@ def java_generic_data_structure(m) -> str:
...
+        parameters.extend(additional_parameters[1])
+    except AttributeError:
+        connectors = []
+    pieces = []
+    nesting: int = 0
+    for i, parameter in enumerate(parameters):
+        pieces.append(parameter)
+        if i < len(connectors):
+            connector = connectors[i]
+            if connector == GenericTypeConnector.AND:
```
nriley: A little clearer if you switch to match.

lang/java/java.py
FireChickenProductivity: Because we need to support multiple consecutive stops, I am going to do some refactoring that will make this irrelevant anyways.

lang/java/java.py
```diff
@@ -158,14 +160,90 @@ def java_generic_data_structure(m) -> str:
...
+)
+def java_type_parameter_arguments(m) -> str:
+    """Formatted Java type parameter arguments"""
+    parameters = [m.java_type_parameter_argument]
+    with suppress(AttributeError):
+        parameters.extend(m.java_generic_type_additional_type_parameters)
+    pieces = []
+    nesting: int = 0
+    for parameter in parameters:
+        if isinstance(parameter, str):
```
nriley, line 221: I think you should be able to write this as `match parameter: case str: ...`. I'll test this and make the changes myself if it works!

nriley (approved):
This looks good as a solution for Java. We are intentionally not documenting this further as we would like to implement generic/nested type support for a couple more languages and settle on a grammar that handles all of them.

If I get some time this weekend, I will submit a second PR with the match change.

PR #2071, opened 2025-12-08
populate rust snippets

BlueDrink9:


lang/rust/rust.talon
```diff
@@ -64,6 +64,7 @@ state (some | sum): "Some"
 state static: "static "
 self taught: "self."
 state use: user.code_import()
+state fish | state turbo fish | state turbo: user.insert_between("::<", ">")
```
chdoc: I'm not a Rust expert, but we usually try to keep the proliferation of alternative pronunciations to a minimum. Would `state [turbo] fish` be acceptable to Rust programmers?

BlueDrink9: That would probably be fine. To be honest, I learned about that term the same day that I wrote this, so I was still undecided about what to call it. (But it is the official term.) I also struggled to decide whether it was technically an operator or something else. Maybe I should see if Java defines something for <T>

nriley: From the Community Backlog session — we felt that since this the feature is relatively unique in Rust among language implementations that it makes sense to simply call it `turbo fish` without `state`. This nickname is even referenced in the Rust programming language book (https://doc.rust-lang.org/book/appendix-02-operators.html).

core/snippets/snippets/forLoopStatement.snippet
```diff
@@ -67,3 +60,10 @@ language: terraform
 -
 for $1 in $2 : $0
 ---
+
+language: rust
+-
+for $1 in $2 {
+    $0
+}
+---
```
nriley: Unsure why you have relocated the snippet without modifying it.

core/snippets/snippets/caseStatement.snippet
```diff
@@ -42,3 +42,9 @@ language: kotlin
 -
 $1 -> $0
 ---
+
+language: rust
+insertionScope: branch
```
nriley:
```suggestion
```
This is not needed as it is declared at the top of the file.

nriley (changes requested):

core/snippets/snippets/forLoopStatement.snippet
BlueDrink9: total accident related to pulling new community, didn't think it was worth fixing unless other issues were found to need addressing

nriley (approved):

Issue #1822, opened 2025-04-15
Add PowerShell Compatible Talon User Dir For Windows


wrchasesims:
Purely a recommendation as pwsh has started to become the standard Win shell.

In step 3 of the Windows section of the [Readme](https://github.com/talonhub/community?tab=readme-ov-file#windows), user is instructed to input the following into *Command Prompt*:

`cd %AppData%\Talon\user`

This works in cmd prompt, but throws an error in PowerShell. couldn't find a syntax that works in both as i'm not super familiar with cmd prompt, but adding the following for pwsh might be nice.

`cd "~/AppData/Roaming/talon/user"`

auscompgeek:
I believe `$env:APPDATA` is the correct syntax for PowerShell

wrchasesims:
> I believe `$env:APPDATA` is the correct syntax for PowerShell

yes, I can confirm that both `cd $env:APPDATA/talon/user` and `cd "~/AppData/Roaming/talon/user"` get pwsh users to the right place. Thanks!

nriley:
From the community backlog session — it looks like on Windows 11 that searching for Command Prompt still opens cmd.exe but there is also an option for "Terminal" that opens PowerShell by default. We would be happy to take a PR that updates the README.

PR #2072, opened 2025-12-11
Updated Readme Windows installation Instruction to use Powershell and reduce no. of clicks

SayakDhole2718:
Corresponding issue : https://github.com/talonhub/community/issues/1822

1.  Changed the terminal used to powershell as it is both supported directly in community and is soon to be a default for windows.
2.  Changed the git installation to use winget instead of manual configuring through GUI Installation to reduce clicks.
3. Updated the path to talon user dir to work with powershell.

I have tested this on a Win11 VM. Willing to provide a video demo since I am recording tutorials anyway.

I'll be responding on slack and the talon community backlog sesh as well as here.


nriley (changes requested):
No additional comments, just marking as changes requested to indicate that this PR is not yet ready to merge.

PR #2073, opened 2025-12-11
[i3wm] Change settings to use full key combinations for key triggered actions

chdoc:


chdoc:
While I still think that what pre-commit does isn't an improvement, at least in this version it doesn't mess up code that is executed.

apps/i3wm/i3wm.talon
```diff
@@ -1,7 +1,12 @@
-# NOTE: If you want to use i3wm you must enable the tag settings.talon. i.e.: `tag(): user.i3wm`
+# NOTE: If you want to use i3wm you must enable the tag in settings.talon. i.e.: `tag(): user.i3wm`
 os: linux
 tag: user.i3wm
 -
+# Change the settings below, if your key combination for these actions differs from the defaults.
+# settings():
```
nriley:
```suggestion
```

PR #2074, opened 2025-12-11
[i3wm] deprecate `port ...` actions in favor of standard `desk ...` actions

chdoc:


apps/i3wm/i3wm.py
```diff
@@ -64,6 +64,17 @@ def switcher_focus(name: str):  # type: ignore
         # focus_on_window_activation is set to "smart" or "urgent"
         actions.user.i3msg('[urgent="latest"] focus')
 
+    # the default implementation considers desktops consecutively numbered
```
auscompgeek, line 67: This should probably do something about the window_move_desktop actions too. See https://github.com/talonhub/community/pull/1167#discussion_r1161266974

chdoc: No, I think that is an entirely separate discussion: the "desk" commands have about the same length as the commands that are used in `i3wm`. As things stand, the main (that is shortest) way to move a window to another desk is `shuffle <number_small>`. I'm certainly not deprecating that in favor as something as verbose as `window move desk <number>`. 

Moreover, `shuffle left` has nothing to do with desktops and swaps the window one slot to the left.

Even worse the shuffle commands are currently way to flexible. I'd be fine with deprecating everything except the "shuffle" variant of 
```
(shuffle | move (win | window) [to] port) <number_small>:
    user.i3msg("move container to workspace number {number_small}")
(shuffle | move (win | window) [to]) last port:
    user.i3msg("move container to workspace back_and_forth")
(shuffle | move) flipper: user.i3msg("move container to workspace back_and_forth")
(shuffle | move (win | window)) {user.arrow_key}: user.i3msg("move {arrow_key}")
```
But again, all of this is only tangentially related to this PR.

PR #2078, opened 2025-12-14
feat: Add 'select using items' functionality for selecting commands

Mark-Phillipson:
This pull request adds new voice commands to the `text_navigation.talon` file that allow users to select multiple items using the "items" keyword. These commands provide an alternative way to select text by specifying a quantity and direction.

New selection commands:

* Added `<number_small> items:` command to select multiple items downward using `shift-down` and repeat for the specified count.
* Added `left <number_small> items:` and `right <number_small> items:` commands to select multiple items to the left or right using `shift-left` and `shift-right`, repeating for the specified count.

### DIFF
--- a/plugin/text_navigation/text_navigation.talon
+++ b/plugin/text_navigation/text_navigation.talon
@@ -85,10 +85,10 @@ big word pre [<number_small>]:
 # Select Using the Items Keyword
 <number_small> items:
     key(shift-down)
-    repeat(number_small-1)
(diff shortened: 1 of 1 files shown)


Mark-Phillipson:
Lines not words is what I found useful

Being able to select multiple lines with a single command.

nriley:
@Mark-Phillipson Try `select down third` or `select three down` for example.

PR #2081, opened 2025-12-17
Add some common talon python functions to python common functions

FireChickenProductivity:
The goal is to make things like "settings.get()" and "actions.clip.text()" a little easier.


nriley (changes requested):
Comments from the community backlog session:
- Use actions instead of corresponding methods where possible
- For actions, enable `user.talon_populate_lists` by default for non-dragon and fix the documentation in the file that declares the actions that reference this.

lang/talon/talon_with_conformer.talon
```diff
@@ -0,0 +1,11 @@
+# enables the user.talon_populate_lists tag with conformer
+# for talon script and talon python
```
nriley:
```suggestion
# for TalonScript and Talon Python
```

nriley (changes requested):

nriley:
From the community backlog session — this is ready to merge, but we want to make sure it does not adversely affect performance on slower computers and the release version of Talon.

This is likely to cause a delay on grammar compilation which typically occurs when you issue your first voice command in a new context.

lang/python/code_common_function.talon-list
```diff
@@ -28,4 +28,16 @@ string: "str"
 sum
 tuple
 type
+update
```
nriley:
```suggestion
```

lang/python/code_common_function.talon-list
```diff
@@ -28,4 +28,16 @@ string: "str"
...
+
+# Talon Python functions
+app notify: actions.app.notify
+clip set text: actions.clip.set_text
+clip text: actions.clip.text
+cron after: cron.after
+cron cancel: cron.cancel
+cron interval: cron.interval
+eff es watch: fs.watch
+fs watch: fs.watch
```
nriley:
```suggestion
F S watch: fs.watch
```

nriley (approved):

nriley:
From the community backlog session — it has been too long with this PR. @FireChickenProductivity will test this with his current setup and we will merge it later today posting on the #community channel for people to watch out in case they're using both the release version of Talon and slow computers.

PR #2082, opened 2025-12-19
Improve mouse wake and sleep

FireChickenProductivity:
Have mouse sleep track eye tracking variables so that mouse wake can set them to the last values.

### DIFF
--- a/plugin/mouse/mouse.py
+++ b/plugin/mouse/mouse.py
@@ -74,7 +74,7 @@ class Actions:
 
     def mouse_wake():
         """Re-enable eye tracking state and disables cursor"""
-        # restore the eye tracking state to what it was before the last user.mouse_sleep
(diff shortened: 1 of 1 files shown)


plugin/mouse/mouse.py
```diff
@@ -45,8 +73,14 @@ def zoom_close():
         actions.tracking.zoom_cancel()
 
     def mouse_wake():
-        """Enable control mouse, zoom mouse, and disables cursor"""
-        actions.tracking.control_zoom_toggle(True)
+        """Re-enable eye tracking state and disables cursor"""
+        # restore the eye tracking state to what it was before the last user.mouse_sleep
```
nriley:
```suggestion
        # restore eye tracking modes enabled as of the last user.mouse_sleep
```

nriley (approved):

PR #2090, opened 2025-12-28
Support dictation mode help

FireChickenProductivity:
Provide "help dictation" command for seeing main dictation mode commands. Allow navigating help menus when shown in dictation mode. Also allow "help active" in dictation mode.

I added support for wildcard matching on selected contexts for matching anything ending with the text after the a wildcard `*`. This was an implementation detail to match any context ending with `.dictation_mode.talon`. This enables showing the right commands even if the user has a dictation_mode.talon of their own. 

On my computer this shows:
image

image

### DIFF
--- a/core/help/help_dictation.talon
+++ b/core/help/help_dictation.talon
@@ -1,4 +1,3 @@
-mode: command
 mode: dictation
 -
 ^help active$: user.help_context_enabled()


core/help/help.py
```diff
@@ -779,6 +798,10 @@ def help_selected_context(m: str):
         register_events(True)
         ctx.tags = ["user.help_open"]
 
+    def help_dictation():
+        """Display primary dictation mode commands"""
+        actions.user.help_selected_context(("*.dictation_mode.talon"))
```
nriley:
```suggestion
        actions.user.help_selected_context("*.dictation_mode.talon")
```

core/help/help_dictation.talon
```diff
@@ -0,0 +1,4 @@
+mode: command
```
AndreasArvidsson:
```suggestion
```

PR #2093, opened 2025-12-30
Provide help commands for the edit lists

FireChickenProductivity:


core/edit/edit_command_actions.py
```diff
@@ -49,7 +49,10 @@ def __str__(self):
 ]
 
 mod = Module()
-mod.list("edit_action", desc="Actions for the edit command")
+mod.list(
+    "edit_action",
+    desc="Actions for the edit command. Dictate an action name followed by an edit modifier name to perform the action on the target referred to by the modifier.",
```
nriley:
```suggestion
    desc="Edit command actions. Follow an action by an edit modifier to perform the action on the modifier's target.",
```

core/edit/edit_command_modifiers.py
```diff
@@ -8,7 +8,7 @@
 mod.list("edit_modifier", desc="Modifiers for the edit command")
 mod.list(
     "edit_modifier_repeatable",
-    desc="Modifiers for the edit command that are repeatable",
+    desc="Modifiers for the edit command that are repeatable. You say a number before the modifier to repeat it that many times.",
```
nriley:
```suggestion
    desc="Edit modifiers that are repeatable. Say a number before the modifier to repeat the action that many times.",
```

PR #2094, opened 2025-12-30
Adds support for xfce terminal

DanielCalvo:
This is nearly identical to the gnome terminal config, minus a shortcut or two.

This is my first PR here -- let me know if anything needs adjusting, I'm happy to follow up.

### DIFF
--- a/core/edit_text_file/edit_text_file_list.talon-list
+++ b/core/edit_text_file/edit_text_file_list.talon-list
@@ -16,3 +16,5 @@ file extensions: settings/file_extensions.csv
 words to replace: settings/words_to_replace.csv
 contacts json: private/contacts.json
 contacts csv: private/contacts.csv
+
+settings: settings.talon


chdoc (approved):
I downloaded the individual files and everything seems to be working as advertised.

apps/xfce4_terminal/xfce4_terminal.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+
+ctx = Context()
+ctx.matches = r"""
+app: xfce4_terminal
+"""
+
+
+@ctx.action_class("user")
+class user_actions:
```
nriley:
```suggestion
class UserActions:
```

apps/xfce4_terminal/xfce4_terminal.py
```diff
@@ -0,0 +1,72 @@
...
+    def tab_jump(number):
+        actions.key(f"alt-{number}")
+
+
+@ctx.action_class("app")
+class app_actions:
```
nriley:
```suggestion
class AppActions:
```

PR #2095, opened 2026-01-01
Added special keys "num lock", "caps lock", and "scroll lock".

Enyium:
Are the special keys "num lock", "caps lock", and "scroll lock" correct on Mac?


nriley (changes requested):
Please see comment.

PR #2098, opened 2026-01-01
Added a few homophones

Enyium:


core/homophones/homophones.csv
```diff
@@ -171,8 +172,10 @@ damn,dam
 Dane,deign
 days,daze
 dear,deer
+defer,differ
```
nriley:
```suggestion
```

A lot of these are accent-specific and may be better configurations by individual users. From the community backlog session, none of us had accents in which these sounded similar.

PR #2099, opened 2026-01-01
More file extensions

Enyium:
The lines with `#` on its own were necessary to prevent Black from destroying the grouping.


FireChickenProductivity:
I cannot get your spoken forms to recognize with the latest ConformerD2, but `dot ps one`, `dot odt`, and `dot ods` work. Could you add those as alternative spoken forms?

I agree that we should add these extensions. I have tested that the reorganization did not lose any of the original extensions. The extensions were highly disorganized before, so I am fine with your reorganization.

AndreasArvidsson (changes requested):
Looks good. We would recommend removing the grouping and just sorting all items alphabetically. If you use cursorless you can just say `"sort every item"`

Also please add those three additional spoken forms mentioned above.

alexander-clarke:
Another useful addition to the list could be `.go`

PR #2100, opened 2026-01-01
Add captures for fractional decimal numbers

Enyium:
Benefits:

- I implemented basic mouse move commands that move the mouse cursor the said amount of centimeters (changeable to inches and other units), which I plan to contribute in the future. Centimeter values like "point five" or "two point five" are a regular occurrence.
- Rango's commands "upper" and "downer" currently only accept integers to scroll whole pages. The only fractional amount is used when omitting the number (0.66 pages). It's very easy to update Rango to simply use a float capture instead to allow for *any* fractional amount like "point five" or "one point five", or even "point one" or "point oh one" (like when you just scrolled, but the mouse cursor needs to get out of the way ever so slightly).
- EDIT: Input float literals into code editors, calculator apps, etc.

### DIFF
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -219,7 +219,10 @@ number_word_leading = f"({'|'.join(leading_words)})"
 
 
 mod.list("number_small", "List of small (0-99) numbers")
-mod.list("decimal_separator", "A decimal separator separating the fractional from the integer part")
+mod.list(
+    "decimal_separator",
+    "A decimal separator separating the fractional from the integer part",
+)
 mod.tag("unprefixed_numbers", desc="Dont require prefix when saying a number")
 ctx.lists["user.number_small"] = get_spoken_form_under_one_hundred(
     0,
@@ -338,7 +341,7 @@ def decimal_string(m) -> str:
     may_omit_int_part_0 = has_decimal_places and m.decimal_separator == "."
 
     string = getattr(m, "number_string", "" if may_omit_int_part_0 else "0")
-    
+
     if has_decimal_places:
         string += m.decimal_separator
         string += m.digit_string
@@ -349,7 +352,7 @@ def decimal_string(m) -> str:
 @mod.capture(rule="<user.decimal_string>")
 def normalized_decimal_string(m) -> str:
     """`user.decimal_string`, normalized in the following manner:
-    
+
     - Integer part is always present
     - Decimal separator, if present, is always `.`
     """
--- a/core/numbers/numbers.py
+++ b/core/numbers/numbers.py
@@ -1,7 +1,7 @@
 import math
 from typing import Iterator, Union, cast
 
-from talon import Context, Module
+from talon import Context, Module, registry
 
 mod = Module()
 ctx = Context()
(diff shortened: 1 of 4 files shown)


core/numbers/numbers.py
```diff
@@ -299,3 +302,16 @@ def number_signed_small(m) -> int:
     """Parses an integer between -99 and 99."""
     number = m[-1]
     return -number if (m[0] in ["negative", "minus"]) else number
+
+
+@mod.capture(rule="[<user.number_string>] point <digit_string> | <user.number_string>")
+def number_float(m) -> float:
```
auscompgeek: Having these without a corresponding string version will be a massive footgun, as people would try to use this to insert strings and end up with rounding errors.

We have a few places already which are intended to insert decimals. We should probably unify them.

(Given this footgun, I wonder if it'd be worth making it explicit in the capture names? Like `number_decimal_string` and `number_decimal_as_float`)

auscompgeek (changes requested):

core/numbers/numbers.py
Enyium: I added string variants. I also allowed for "comma" as the decimal separator, because [many countries use it](https://en.wikipedia.org/wiki/Decimal_separator#Conventions_worldwide), and Talon is also used in English to write for other language contexts.

> We have a few places already which are intended to insert decimals. We should probably unify them.

Where? I could find any.

> (Given this footgun, I wonder if it'd be worth making it explicit in the capture names? Like `number_decimal_string` and `number_decimal_as_float`)

I left the names unchanged for now to wait for other reviewers' opinion.

Enyium: I also added strict variants that only ever output commas normalized to points, which could perhaps be useful for string-based REST APIs and the like for which you don't want rounding errors.

The names are awful. May I break with the previous convention and write `signed_strict_float_string` instead of `number_strict_float_signed_string` etc.?

auscompgeek: > Where? I could find any.

Found these by searching for `number_string point`:

https://github.com/talonhub/community/blob/36e2cd647c3a11b5412a498a7bf43605264d886e/lang/css/css.talon#L29

https://github.com/talonhub/community/blob/36e2cd647c3a11b5412a498a7bf43605264d886e/core/text/text_and_dictation.py#L67-L68

auscompgeek: "float string" seems like an oxymoron since float refers to the floating point data type, which strings aren't.

core/numbers/numbers.py
```diff
@@ -299,3 +302,43 @@ def number_signed_small(m) -> int:
...
     number = m[-1]
     return -number if (m[0] in ["negative", "minus"]) else number
+
+
+@mod.capture(
+    rule="[<user.number_string>] (point | comma) <digit_string> | <user.number_string>"
+)
+def number_float_string(m) -> str:
+    """Parses a phrase for a floating-point number, returning the number as a string with one of different possible decimal separators."""
+    decimal_separator = "." if "point" in m else ","
```
auscompgeek: This should just be a list instead. Would make it easier for users to customise if they so choose.

PR #2101, opened 2026-01-01
More abbreviations

Enyium:
Question: What is `"some": "sum"` good for? Is this on the same level as "luv" for "love"? It's conflicting with `"start of month": "som"`, though, and is also available using the command "phones some". Should this be removed?

---

I took the abbreviated mathematical functions from:

- https://docs.python.org/3/library/functions.html
- https://docs.python.org/3/library/math.html
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math
- https://en.cppreference.com/w/cpp/numeric/math.html
- https://doc.rust-lang.org/std/primitive.f64.html
- https://doc.rust-lang.org/std/primitive.u32.html
- https://en.wikipedia.org/wiki/Trigonometric_functions#Notation

I added abbreviations like `eq` for "equals", or `lt` for "less than", because they can [appear in function names](https://doc.rust-lang.org/std/cmp/trait.Eq.html?search=lt).

---

I deliberately didn't sort alphabetically in the third commit and grouped the new abbreviations. This should make review easier. After that: Should I ruthlessly sort everything alphabetically, even though this scatters month names etc.?


core/abbreviate/abbreviate.py
```diff
@@ -132,7 +132,7 @@
     "double ended queue": "deque",
     "double": "dbl",
     "dupe": "dup",
-    "duplicate": "dup",
+    "duplicate": "dupe",
```
AndreasArvidsson:
```suggestion
    "duplicate": "dup",
```

core/abbreviate/abbreviate.py
```diff
@@ -446,6 +446,186 @@
...
+    "equals": "eq",
+    "not equal to": "ne",
+    "not equals": "ne",
+    "doesnt equal": "ne",
+    "less than": "lt",
+    "less than or equal to": "le",
+    "less equals": "le",
+    "greater than": "gt",
+    "greater than or equal to": "ge",
+    "greater equals": "ge",
```
AndreasArvidsson, line 490: These looks like something that belongs in programming languages. We can't really see anyone using these abbreviations in the prose.

AndreasArvidsson (changes requested):
Looks good. When it comes to sorting we feel that everything that is just prose should be sorted alphabetically. Things that logically belonged together like the names of the month and math concepts could be grouped together separately.

PR #2102, opened 2026-01-02
Added system actions: exit menu etc.

Enyium:
- Implemented them for Windows.
- Transferred one action for Linux from an existing .talon file.

---

See the new file `core/system/system.talon` first for a quick overview.

### DIFF
--- a/core/system/system_linux.py
+++ b/core/system/system_linux.py
@@ -1,0 +1,26 @@
+from talon import Context, actions
+
+i3wm_ctx = Context()
+i3wm_ctx.matches = r"""
+os: linux
+tag: user.i3wm
+"""
+
+
+@i3wm_ctx.action_class("user")
+class I3wmUserActions:
+    def system_switch_screen_power(on: bool):
+        # TODO
+        pass
+
+    def system_show_settings():
+        # TODO
+        pass
+
+    def system_lock():
+        # Adapt key configuration as needed.
+        actions.key("super-shift-x")
+
+    def system_show_exit_menu():
+        # TODO
+        pass
--- a/core/system/system_linux.py
+++ b/core/system/system_linux.py
@@ -17,6 +17,10 @@ class I3wmUserActions:
         # TODO
         pass
 
+    def system_show_task_manager():
+        # TODO
+        pass
+
     def system_lock():
         # Adapt key configuration as needed.
         actions.key("super-shift-x")
--- a/core/system/system_win.py
+++ b/core/system/system_win.py
@@ -2,10 +2,12 @@ import ctypes
 import os
 from typing import TYPE_CHECKING
 
-from talon import Context, app
+from talon import Context, actions, app
 
 if app.platform == "windows" or TYPE_CHECKING:
     from ctypes import wintypes
+
+    import win32con
 
     user32 = ctypes.windll.user32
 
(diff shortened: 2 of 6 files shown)


core/system/system_linux.py
```diff
@@ -0,0 +1,26 @@
...
+        # TODO
+        pass
+
+    def system_show_settings():
+        # TODO
+        pass
+
+    def system_lock():
+        # Adapt key configuration as needed.
+        actions.key("super-shift-x")
```
chdoc, line 26: I think this is very problematic: it moves a setting that is neither a default nor really a widely adopted convention to a file that has nothing to do with i3wm. Thus, this would become very hard to discover. I am not even entirely convinced that there is any configuration for these commands that would work reliably across desktop environments. I'm not aware of any general configuration for the major environments like Gnome and KDE in `community`. Of course, the major desktop environments have recently dropped support for X11 and, by extension, Talon. But that's a different can of worms.

core/system/system_win.py
```diff
@@ -0,0 +1,68 @@
...
+os: windows
+"""
+
+
+@ctx.action_class("user")
+class UserActions:
+    def system_switch_screen_power(on: bool):
+        """Turns all screens off, or, if all are off, turns those on that were on when having turned all off.
+
+        Turning the screens on didn't work for the author on Windows 11 24H2, even though it worked for him on Windows 10 before when the same call to `DefWindowProcW()` was made in a different programming language. Turning the screens on using code like `actions.mouse_move(actions.mouse_x() + 1, actions.mouse_y() + 1)` also didn't work.
```
chdoc: This definitely shouldn't be a documentation string. In fact, I would argue that it doesn't make sense to merge any functionality that only works in Windows 10. After all, Windows 10 has basically reached end of life in October 2025.

core/system/system_linux.py
Enyium: > a setting that is neither a default nor really a widely adopted convention 

I needed to abstract the existing voice command `lock screen`, because it semantically belongs to what the new `core/system/system.py` declares. I can't see how it should just continue to exist in parallel, *if* the platform-independent abstraction is accepted.

> this would become very hard to discover.

`i3wm.talon` could refer to this code in a comment. Do you think the implementation in `core/system` should be moved?!

> I am not even entirely convinced that there is any configuration for these commands that would work reliably across desktop environments.

That's why I defined the `i3wm_ctx` with the respective constraint. The new `core/system/system_linux.py` was intended to be augmented with different Talon `Context`s for different desktop environments.

AI generated code for me for the class `I3wmUserActions` that it thought would implement the actions (same for Mac). I just didn't accept it, because I can't test it, because I'm neither a Linux user nor knowledgable in Linux. (Disclaimer: I wrote the Windows code diligently myself.)

core/system/system_win.py
Enyium: I found a way to support turning the monitor on on Windows. This makes the function much better now.

Enyium:
Maybe I should move `system_switch_screen_power()` into the `core/screens` directory?

AndreasArvidsson (changes requested):
See above

nriley:
Unfortunately, (beta) Talon with Python 3.14t is not currently compatible with PyWin32. Since there has been no response to our comments from January, closing this. Can consider reopening if there is better platform support in future.

PR #2105, opened 2026-01-04
turn dictation helpers into actions to make them configurable

chdoc:
The motivation for this change is my current work on German mode, which implements these new actions as follows:
```python

# characters to replace if ASCII punctuation is desired
_ascii_replace = {'–': '-', '„': '"', '“': '"', "‚": "'", "‘": "'", "’": "'"}
_ascii_trans = str.maketrans(_ascii_replace)

@ctx.action_class("user")
class Actions:
    def omit_space_before(text: str) -> bool:
        return bool(not text or text[0] in _no_space_before)

    def omit_space_after(text: str) -> bool:
        return bool(not text or text[-1] in _no_space_after)
    
    def dictation_replace(text: str) -> str:
        if settings.get("user.german_unicode"):
            return text
        return text.translate(_ascii_trans)
```
Note that the rules regarding spacing around quotes are very different between English and German. In particular, the English opening quote is the German closing quote.

### DIFF
--- a/core/text/text_and_dictation.py
+++ b/core/text/text_and_dictation.py
@@ -493,11 +493,11 @@ class Actions:
         return bool(not text or no_space_before.search(text))
 
     def omit_space_after(text: str) -> bool:
-        """Test if string needs space after"""
+        """Test if dictated text needs space after"""
         return bool(not text or no_space_after.search(text))
 
     def needs_space_between(before: str, after: str) -> bool:
-        """Test if two strings need a space between them"""
+        """Test if two text strings need a space between them"""
         return not (
             actions.user.omit_space_after(before)
             or actions.user.omit_space_before(after)


core/text/text_and_dictation.py
```diff
@@ -498,30 +488,51 @@ def dictation_reformat_no_space():
         """Removes space before the last utterance"""
         reformat_last_utterance(lambda s: s[1:] if s.startswith(" ") else s)
 
+    def omit_space_before(text: str) -> bool:
+        """Test if string needs space before"""
```
nriley:
```suggestion
        """Test if dictated text needs space before"""
```

core/text/text_and_dictation.py
```diff
@@ -498,30 +488,51 @@ def dictation_reformat_no_space():
...
+        return bool(not text or no_space_before.search(text))
+
+    def omit_space_after(text: str) -> bool:
+        """Test if string needs space after"""
```
nriley:
```suggestion
        """Test if dictated text needs space after"""
```

core/text/text_and_dictation.py
```diff
@@ -498,30 +488,51 @@ def dictation_reformat_no_space():
...
+        return bool(not text or no_space_after.search(text))
+
+    def needs_space_between(before: str, after: str) -> bool:
+        """Test if two strings need a space between them"""
```
nriley:
```suggestion
        """Test if two text strings need a space between them"""
```

PR #2113, opened 2026-01-10
Filter applications by .app on MacOS

FireChickenProductivity:
This was recommended by Aegis to prevent matching non application files.

### DIFF
--- a/core/app_switcher/app_switcher.py
+++ b/core/app_switcher/app_switcher.py
@@ -221,23 +221,28 @@ elif app.platform == "linux":
 elif app.platform == "mac":
     mac_application_directories = [
         "/Applications",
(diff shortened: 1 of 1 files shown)


nriley:
I don't use this feature ("launch" controls LaunchBar on Mac and Fluent Search on Windows for me) so I've never really looked at it. Fixed it so it supports subfolders of the standard Applications folders beyond Utilities, as these are quite common. There's also some system apps in `/System/Library/CoreServices` but most are not exposed to the user and not named in a user-friendly fashion, so I deliberately chose to exclude these.

Note that app names can be localized; we don't make any attempt to use the localized names if you are running in a language other than US English.

PR #2115, opened 2026-01-18
Expand r support

BlueDrink9:
Also a tiny correction to the vimscript comment - add the missing space after "

### DIFF
--- a/lang/r/code_keyword.talon-list
+++ b/lang/r/code_keyword.talon-list
@@ -2,8 +2,8 @@ list: user.code_keyword
 code.language: r
 -
 
-N A: "NA"
-N A character: "NA_character_"
-N A real: "NA_real_"
-N A complex: "NA_complex_"
-N A integer: "NA_integer_"
+N A: NA
+N A character: NA_character_
+N A real: NA_real_
+N A complex: NA_complex_
+N A integer: NA_integer_


FireChickenProductivity:
The issue with state is that state was originally intended for referring to statements (such as state while) but unfortunately morphed into the default prefix. If those are in fact keywords, you could move them to a `user.code_keyword` list for r and make r activate the `user.code_keywords` tag.

core/snippets/snippets/commentLine.snippet
```diff
@@ -21,12 +21,12 @@ language: xml | html | markdown
 
 language: vimscript
 -
-"$0
+" $0
```
nriley: Please split this out into a separate PR since it has nothing to do with R.

lang/r/code_common_function.talon-list
```diff
@@ -41,11 +56,16 @@ paste: "paste0"
...
+row names: "rownames"
 scale: "scale"
 sequence along: "seq_along"
 sequence length: "seq_len"
 sequence: "seq"
 set working directory: "setwd"
+set difference: "setdiff"
+set diff: "setdiff"
+sym diff: "symdiff"
+symetric diff: "symdiff"
```
nriley:
```suggestion
symmetric diff: "symdiff"
```

lang/r/r.talon
```diff
@@ -30,6 +30,12 @@ library <user.code_libraries>:
     " %>%"
     key(enter)
 state na: insert("NA")
+state na character: insert("NA_character_")
+state na real: insert("NA_real_")
+state na complex: insert("NA_complex_")
+state na integer: insert("NA_integer_")
```
nriley: From the Community Backlog Session — since all of these are reserved words/keywords in the R language, we should switch these over to keywords. Please add a deprecation for "state na".

core/snippets/snippets/commentLine.snippet
BlueDrink9: that's true, but is it really necessary? This change is tiny

nriley: Yes :) The PR will also be tiny.

lang/r/code_keyword.talon-list
```diff
@@ -0,0 +1,9 @@
+list: user.code_keyword
+code.language: r
+-
+
+N A: "NA"
+N A character: "NA_character_"
+N A real: "NA_real_"
+N A complex: "NA_complex_"
+N A integer: "NA_integer_"
```
AndreasArvidsson:
```suggestion
N A: NA
N A character: NA_character_
N A real: NA_real_
N A complex: NA_complex_
N A integer: NA_integer_
```

PR #2118, opened 2026-01-20
Make Talon icon location description operating system specific

FireChickenProductivity:
Say it is in the system tray on Windows, in the status area on MacOS, and in the system tray or status area on Linux. I tested this on Windows and MacOS but was not sure what to do for Linux.

### DIFF
--- a/plugin/new_user_message/new_user_message.py
+++ b/plugin/new_user_message/new_user_message.py
@@ -48,7 +48,7 @@ def new_user_gui(gui: imgui.GUI):
     gui.line()
 
     gui.text(
-        f"You can see the Talon menu by right clicking the Talon icon in the {talon_menu_location}"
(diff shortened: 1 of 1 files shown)


plugin/new_user_message/new_user_message.py
```diff
@@ -41,7 +48,7 @@ def new_user_gui(gui: imgui.GUI):
     gui.line()
 
     gui.text(
-        "You can see the Talon menu by right clicking the Talon icon in the system tray or status area"
+        f"You can see the Talon menu by right clicking the Talon icon in the {talon_menu_location}"
```
AndreasArvidsson:
```suggestion
        f"You can see the Talon menu by clicking the Talon icon in the {talon_menu_location}"
```

PR #2119, opened 2026-01-21
Fix help active context detection

FireChickenProductivity:
Make active context detection consistent in the help system. This change makes the cached active context list match the commands that `help active` will show. While I was here, I simplified the `refresh_context_command_map` function a little.

### DIFF
--- a/core/edit/edit_command_actions.py
+++ b/core/edit/edit_command_actions.py
@@ -49,7 +49,10 @@ EditAction = Union[
 ]
 
 mod = Module()
-mod.list("edit_action", desc="Actions for the edit command")
+mod.list(
+    "edit_action",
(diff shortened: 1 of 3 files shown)


core/help/help.py
```diff
@@ -538,7 +531,34 @@ def refresh_context_command_map(enabled_only=False):
     rule_word_map = refresh_rule_word_map(local_context_command_map)
 
     ctx.lists["self.help_contexts"] = cached_short_context_names
-    update_active_contexts_cache(active_contexts)
+    update_active_contexts_cache(active_context_cache)
+
+
+def update_short_names(
```
nriley: From the community backlog session — this seems to just generate spoken forms aside from the use of the overrides dictionary, which is empty by default. Perhaps rename it to `compute_spoken_forms` or `update_spoken_forms`?

PR #2127, opened 2026-02-06
Improve golang support, moving lists to talon-list

alexander-clarke:
I seperated the user.code_keyword and the user.code_type lists into talon-list files.
I added a missing keyword (var).
I fixed the types list, which seemed to be built from a copy of the C types, however not all of those types exist in go
I also added the user.code_keywords tag so that the previously defined keywords could actually be used in go


lang/go/code_type.talon-list
```diff
@@ -0,0 +1,14 @@
+list: self.code_type
+code.language: go
+-
+
+boolean: "bool"
+int: "int"
+you int: "uint"
+you int pointer: "uintptr"
+float 32: "float32"
```
auscompgeek: I don't believe this spoken form will be recognised. Probably needs to be written out in words?

alexander-clarke: yep, forgot you had to do that, it works now thanks

lang/go/code_type.talon-list
```diff
@@ -0,0 +1,14 @@
+list: self.code_type
+code.language: go
+-
+
+boolean: "bool"
+int: "int"
+you int: "uint"
+you int pointer: "uintptr"
+float thirty two: "float32"
+float: "float64"
```
nriley: From the Community Backlog session, a couple options here
- include all the data types supported in Go individually (less preferred, but please include all the ones in https://go.dev/tour/basics/11)
- use a capture and one or more lists to generate data types like we do in C/C++ (example: https://github.com/talonhub/community/blob/d1dd588e59baf720331244926318c64202ddfb9c/lang/c/c.py#L31)

nriley: Also, since there is no data type called `float` we do not think it made sense to default to `float64`; of course, you are welcome to do this in your local configuration.

alexander-clarke: >     * use a capture and one or more lists to generate data types like we do in C/C++ (example: https://github.com/talonhub/community/blob/d1dd588e59baf720331244926318c64202ddfb9c/lang/c/c.py#L31
>       )

Should I reuse any applicable lists or create new ones?
Particularly: https://github.com/talonhub/community/blob/d1dd588e59baf720331244926318c64202ddfb9c/lang/c/stdint_signed.talon-list
Which would be identical for this use case

nriley: Good question. I think reuse should be fine.

PR #2136, opened 2026-02-16
Fix bug where mode indicator didn't update for screen changes

AndreasArvidsson:
@nriley This is probably the bug you are running into.


nriley (approved):

nriley:
That'd do it :-) Thanks!

(At the community backlog session I only mentioned the issue with multiple screen size updates within a short period because it affected another app of mine; this is a clear bug!)

PR #2141, opened 2026-02-18
Go keyword and function improvements

alexander-clarke:
Add spacing for keywords as needed  
Add missing keywords based upon this list: https://go101.org/article/keywords-and-identifiers.html
Add some missing functions to the common golang list

### DIFF
--- a/core/snippets/snippets/includeSystemStatement.snippet
+++ b/core/snippets/snippets/includeSystemStatement.snippet
@@ -1,6 +1,8 @@
 name: includeSystemStatement
 phrase: include system | include sys
 insertionScope: statement
+
+$0.insertionFormatter: SNAKE_CASE
 ---
 
 language: c | cpp


lang/go/code_keyword.talon-list
```diff
@@ -4,20 +4,30 @@ code.language: go
 
 break: "break"
 continue: "continue"
-struct: "struct"
-type: "type"
-return: "return"
-package: "package"
-import: "import"
+struct: " struct "
+interface: " interface "
```
chdoc, line 8: I'm a bit surprised to see these surrounded by spaces. As far as the understand, in other languages that is only used to surround keywords that are binary operators. What is the rationale here?

alexander-clarke: In go, struct or interface always comes after some name, and before another symbol
https://go.dev/tour/moretypes/2
https://go.dev/tour/methods/9

and so as it is always in the middle I think should have spaces on both sides.

lang/go/code_keyword.talon-list
```diff
@@ -4,20 +4,30 @@ code.language: go
...
+type: "type "
+channel: "chan "
```
nriley:
```suggestion
chan: "chan "
```

PR #2149, opened 2026-02-28
Add gamepad tester getters for stick and trigger

rokubop:
I've been working on a test suite for my [talon-gamepad-rig](https://github.com/rokubop/talon-gamepad-rig) repository, and found the need for these extra functions when testing for [deadzones](https://github.com/rokubop/talon-gamepad-rig/blob/b0ed11bd78077d6b71b558bd96691288fb58cdeb/tests/test_deadzone.py#L13-L19).


AndreasArvidsson:
I guess you're just polling for values? You could of course register your own gamepad bindings and get on change events instead? In Talon there is no problem registering the same gamepad or hotkey multiple times. 

Could you elaborate a bit on how you're using this?

PR #2150, opened 2026-02-28
Add option to use by_lines for continuous scrolling

BlueDrink9:
As well as adding options, this is vital for working around bugs in certain applications. https://github.com/talonvoice/talon/issues/330

### DIFF
--- a/plugin/mouse/mouse_scroll.py
+++ b/plugin/mouse/mouse_scroll.py
@@ -50,12 +50,12 @@ class Scroller:
         self._is_vertical: bool = False
         self._scroll_dir = 1
 
-    def scroll_in_direction(self, amount: int):
+    def scroll_in_direction(self, amount: int, by_lines: bool):
         scroll_delta = self._scroll_dir * amount
         if self._is_vertical:
-            actions.mouse_scroll(scroll_delta)
+            actions.mouse_scroll(scroll_delta, by_lines=by_lines)
         else:
-            actions.mouse_scroll(0, scroll_delta)
+            actions.mouse_scroll(0, scroll_delta, by_lines=by_lines)
 
     def is_direction_equal_to(self, direction: ScrollingDirection) -> bool:
         return self.direction == direction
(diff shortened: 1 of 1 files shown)


plugin/mouse/mouse_scroll.py
```diff
@@ -94,7 +164,16 @@ def start_continuous_scrolling_job(self):
...
+        by_lines = settings.get("user.continuous_scroll_by_lines")
+        if by_lines:
+            # Set to 2 because usual speeds are too large as a lines value.
+            # Instead, sleep for a proportional time.
+            sleep_time = 1 / speed
+            speed = 2
+
+        self.scroller.scroll_in_direction(speed, by_lines)
+        if by_lines:
+            actions.sleep(sleep_time)
```
AndreasArvidsson, line 176: This function is called via a cron interval. Doing a sleep inside a cron callback is a bad idea because it can affect Talon performance. What we should do is change the cron interval.

plugin/mouse/mouse_scroll.py
```diff
@@ -94,7 +164,16 @@ def start_continuous_scrolling_job(self):
 
     def scroll_continuous_helper(self):
         speed = self.compute_scrolling_speed()
-        self.scroller.scroll_in_direction(speed)
+        by_lines = settings.get("user.continuous_scroll_by_lines")
+        if by_lines:
+            # Set to 2 because usual speeds are too large as a lines value.
+            # Instead, sleep for a proportional time.
+            sleep_time = 1 / speed
+            speed = 2
```
AndreasArvidsson, line 172: Instead of arbitrarily setting the speed to 2 maybe we should just ignore the speed if we are scrolling by lines? Or could you elaborate on why this is a better solution?

AndreasArvidsson (changes requested):

plugin/mouse/mouse_scroll.py
BlueDrink9: I did this by feel, comparing a few different speeds without and without by lines. It would probably actually be better to set speed to one and divide the wait time in half. But that is the one that seems to scroll at the closest rate to the equivalent speed without by lines, in terms of roughly perceived lines/s

plugin/mouse/mouse_scroll.py
BlueDrink9: Could you please give me some hints about how to do that? Currently that is a little bit beyond me.

AndreasArvidsson: If scrolling is by lines you would change the cron interval when the scrolling starts

https://github.com/talonhub/community/blob/c18a84ce598b7a53c400a2ad3bcb3db52feba356/plugin/mouse/mouse_scroll.py#L91
https://github.com/talonhub/community/blob/c18a84ce598b7a53c400a2ad3bcb3db52feba356/plugin/mouse/mouse_scroll.py#L101

PR #2152, opened 2026-03-04
Add compound edit commands for readline

alexander-clarke:
Implements some compound edit commands to readline. ie can do clear word left, clear way left, etc

It also implements copy word left and cut word left but this uses readline's clipboard rather than the desktops. I think this is the only way of doing it but might be unexpected.

I also added readline support to the gnome terminal, to enable these commands in there

### DIFF
--- a/tags/terminal/readline.py
+++ b/tags/terminal/readline.py
@@ -67,13 +67,30 @@ def cut_line_end():
     actions.key("ctrl-k")
 
 
-cut_actions = {
-    "wordLeft": actions.user.cut_word_left,
-    "wordRight": actions.user.cut_word_right,
-    "word": cut_word,
-    "line": actions.user.cut_line,
-    "lineStart": cut_line_start,
-    "lineEnd": cut_line_end,
+compound_actions = {
+    # Delete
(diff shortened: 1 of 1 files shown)


tags/terminal/readline.py
```diff
@@ -53,16 +98,19 @@ def copy_word_right():
...
+            if modifier_name in cut_actions:
+                callback = cut_actions[modifier_name]
+
+                if action_name == "delete":
+                    callback = with_clip_revert(callback)
+                elif action_name == "copyToClipboard":
+                    callback = with_selection_revert(callback)
+
+                return callback
+        return actions.next(pair)
```
chdoc: For the sake of uniformity, I would prefer this to be a simple look up, similar to what is done in `readline_vi.py`. That makes it easier to see and edit what action and modifier combination calls which function.

AndreasArvidsson: Feedback from community session. We agreed that a compound actions dictionary like we have in readline vi is the best solution.
https://github.com/talonhub/community/blob/339e5c4060caa99cf7f7b11b43145a91534eca1b/tags/terminal/readline_vi.py#L62-L74

PR #2155, opened 2026-03-10
Replace pre commit hook for Talon formatter and add snippet formatter

AndreasArvidsson:
We have just set up a Cursorless repository with multiple formatters. The ones that are applicable to community is Talon and snippets. Snippets would just be a brand new pre commit hook, while Talon would be a replacement. We also don't need `remove-tabs` anymore.

The reasoning for switching the Talon formatter is that it's the same one we use for Cursorless and that I actively maintain. It's written in typescript and have good test coverage. It supports `.editorconfig` for easy configuration of both default and custom options.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,24 +22,29 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: mixed-line-ending
+
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: "v4.0.0-alpha.8"
     hooks:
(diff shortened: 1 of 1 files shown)


chdoc:
> Properly implement the specified 88 character line length

I'm a bit on the fence about this one. We often have logical groups of commands, and this might lead to some of them having line breaks while the others don't. If one of them needs to be broken across multiple lines due to the 88 character limit, I would like to preserve the possibility of adding line breaks to the other commands as well. In my view, homogeneous appearance of related commands trumps filling every line to the maximum.

Issue #2146, opened 2026-02-22
Clean up terminal/shell tags


nriley:
There are a number of commands in tags/terminal/terminal.py that are shell rather than terminal commands; these should be separated as there are multiple terminals that support different types of shell (e.g. cmd.exe vs PowerShell vs a Linux shell with Windows Terminal, or even just different shells on Linux or macOS). Consider refactoring these so there doesn't need to be a 1:1 relationship between shell and terminal emulator.

@chdoc wanted me to tag him on this…

PR #2156, opened 2026-03-13
Refactor file_manager tag implementation for shells to use a common base implementation

alexander-clarke:
Currently when we implement file manager for terminals there is a lot of repeated code. I have factored this out so that most of the implementation is inside of terminal.py, and to implement it in a new terminal you just need to set how to get the current path. As an example I have implemented this in gnome. 

I could refactor existing terminals as there is then a lots of code that can be deleted, but that could be a separate pull request.

Could end up relating to https://github.com/talonhub/community/issues/2146 But I'm not sure if there are yet any plans on how things might look different


tags/terminal/terminal.py
```diff
@@ -30,3 +33,36 @@ def terminal_rerun_search(command: str):
 
     def terminal_kill_all():
         """kills the running command"""
+
+
+@ctx.action_class("user")
+class TerminalActions:
+    # user.file_manager
```
AndreasArvidsson, line 52: remove commented code

Issue #2143, opened 2026-02-20
TODO comments


FireChickenProductivity:
Compilation of todo comments. 
- [ ] require tags for https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/1password/1password_global.talon#L1
- [ ] # in german version TODO: differentiate languages, https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/adobe/adobe_acrobat_reader_dc_win.py#L28 zoom_in, zoom_out actions
- [ ] # TODO: mac context and implementation: https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/adobe/adobe_acrobat_reader_dc.py#L15
- [ ] # TODO: fit this to terminal.py https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/apple_terminal/apple_terminal.py#L5
- [ ] # TODO: explain https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/apple_terminal/apple_terminal.talon#L18, above tag(): tags. This might have meant explaining tag(): user.file_manager
- [ ] # TODO: mac implementation, https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/calibre/calibre_viewer.py#L18
- [ ] # TODO: mac context, https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/calibre/calibre.py#L19
- [ ] # todo: these commands should only be available when it's focused: https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/eclipse/eclipse_win.py#L129 before `def find_toggle_match_by_case():`
- [ ] #TODO: should this be added to linecommands? https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/eclipse/eclipse.talon#L142 before the `copy line down` and `copy line up` commands. 
- [ ] # TODO: handle corner-cases like key(" ") and key("ctrl- "), etc. in https://github.com/talonhub/community/blob/3ed8732fa6ae622a9e04b37d4e9191749277fd49/apps/emacs/emacs.py#L64 `def emacs_key(keys: str):`
[...]

PR #2157, opened 2026-03-14
Remove todo comments

FireChickenProductivity:
Removes todo comments, which are now documented here: #2143


apps/windows_terminal/windows_terminal.py
```diff
@@ -151,18 +146,16 @@ def split_clear():
 
     def split_clear_all():
         """Clears all splits"""
-        # TODO: decide whether to implement it at all since it either doesn't makes sense or closes the window/whole tab
+        pass
```
AndreasArvidsson: Please remove the entire function instead

PR #2160, opened 2026-03-17
Use same Python version as public Talon for pre-commit

FireChickenProductivity:


auscompgeek (commented):
Do any of our pre-commit tools depend on their runtime version to determine what version of Python to target? I would've thought the only tool that'd affect is `black`, but only as a fallback when it's not configured, which it is:

https://github.com/talonhub/community/blob/92bc38e155366fd519c6a1cc67b070bf02274fea/pyproject.toml#L1-L2

My concern here (which I think I brought up last time this was proposed) is this would unnecessarily force contributors to have a `python3.11` installed when they otherwise wouldn't.

PR #2166, opened 2026-03-28
Mode indicator show and hide

FireChickenProductivity:
The `mode indicator hide` command forces the indicator to be hidden regardless of the `user.mode_indicator_show` setting. The `mode indicator show` command forces the indicator to be hidden regardless of the `user.mode_indicator_show` setting. If neither command was used this session, the setting decides if the indicator is shown. I implemented this using an enum to keep track of these possible states.


plugin/mode_indicator/mode_indicator.py
```diff
@@ -8,6 +10,18 @@
 canvas: Canvas = None
 current_mode = ""
 current_microphone = ""
+
+
+class VisibilityMode(Enum):
```
AndreasArvidsson: Feedback from session: Which should probably just make an optional boolean.

plugin/mode_indicator/mode_indicator.py
```diff
@@ -222,6 +243,21 @@ def poll_microphone():
         update_indicator()
 
 
+@mod.action_class
+class Actions:
+    def mode_indicator_show():
```
AndreasArvidsson: Replace these functions with a single optional  boolean action

PR #2171, opened 2026-03-29
Migrate block comment commands to actions

FireChickenProductivity:
Migrate block comment commands to use actions. I fixed one of the commands not always making the comment on a single line as a side effect of the snippet migration. 

The differences between these commands are not very clear from the names. There is stuff to improve here, but I would prefer to make this PR just a migration.


lang/tags/comment_block.py
```diff
@@ -14,22 +16,46 @@
...
         """Block comment end syntax"""
 
+    def code_block_comment_line(text: Optional[str] = None):
+        """Wraps current line in block comment markers. Puts `text` between them if given"""
+        actions.edit.line_start()
+        actions.user.code_comment_block_prefix()
+        actions.key("space")
+        if text is not None:
+            actions.insert(text)
+        actions.edit.line_end()
```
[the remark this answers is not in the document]
FireChickenProductivity: This is no longer even relevant

auscompgeek:
Is the plan to convert these to snippets in a follow up PR?

FireChickenProductivity:
> Is the plan to convert these to snippets in a follow up PR?

There are no followup plans yet. Suggestions are welcome. This PR is mostly to satisfy an ancient todo comment.

nriley (changes requested):
From the community backlog session — makes sense as a first step, migrating the implementations from voice commands to actions. Could you please rename all the new actions to start with `code_block_`? Also for the actions that include a new optional argument, can you use snippet insertion or a single `insert` to insert both the comment delimiters and contents in a single action? Thanks!

FireChickenProductivity:
> From the community backlog session — makes sense as a first step, migrating the implementations from voice commands to actions. Could you please rename all the new actions to start with `code_block_`? Also for the actions that include a new optional argument, can you use snippet insertion or a single `insert` to insert both the comment delimiters and contents in a single action? Thanks!

What about instead using `code_comment_block` as the action prefix to be consistent with the original actions?

FireChickenProductivity:
There was no clear way to do the c like block comment insertion in a single insert without changing behavior. We may want to leave updating that one for a snippet migration follow up.

core/snippets/snippets_insert.py
```diff
@@ -70,6 +70,12 @@ def compute_snippet_body_with_substitutions(
     snippet: Snippet, substitutions: dict[str, str]
 ) -> str:
     body = snippet.body
+    return compute_snippet_body_with_substitutions_from_body(body, substitutions)
+
+
+def compute_snippet_body_with_substitutions_from_body(
+    body: str, substitutions: dict[str, str]
```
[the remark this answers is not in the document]
FireChickenProductivity: This is currently completely irrelevant

lang/tags/comment_block.py
```diff
@@ -14,22 +16,54 @@
...
+    def code_comment_block_line():
+        """Wraps current line in block comment markers"""
+        actions.edit.line_start()
+        actions.user.code_comment_block_prefix()
+        actions.key("space")
+        actions.edit.line_end()
+        actions.key("space")
+        actions.user.code_comment_block_suffix()
+
+    def code_comment_block_at_line_start(text: str):
```
nriley: From the Community Backlog session — don't think that `code_comment_block_at_line_start` and `code_comment_block_at_line_end` need to be migrated into individual actions.

lang/tags/comment_block.py
```diff
@@ -14,22 +16,54 @@
...
+        actions.edit.line_start()
+        actions.user.code_block_comment_inline(text)
+
+    def code_comment_block_at_line_end(text: str):
+        """Inserts a block comment at the end of the line"""
+        actions.edit.line_end()
+        actions.user.code_block_comment_inline(text)
+
+    def code_block_comment_inline(text: str):
+        """Inserts an inline block comment at the end of the line with the specified text"""
```
nriley:
```suggestion
        """Inserts an inline block comment at the cursor position with the specified text"""
```

lang/tags/comment_block.py
```diff
@@ -14,22 +16,54 @@
...
+        """Inserts a block comment at the start of the line"""
+        actions.edit.line_start()
+        actions.user.code_block_comment_inline(text)
+
+    def code_comment_block_at_line_end(text: str):
+        """Inserts a block comment at the end of the line"""
+        actions.edit.line_end()
+        actions.user.code_block_comment_inline(text)
+
+    def code_block_comment_inline(text: str):
```
nriley: Please make `text` optional to match `code_comment_block`.

nriley (changes requested):

lang/tags/comment_block.py
FireChickenProductivity: I updated the comment. I do not think we need to spell out that it inserts the comment at the cursor.

lang/tags/comment_block.py
FireChickenProductivity: How would that work? How do we move the cursor in between the comment suffix and prefix if they are inserted through actions?
I could try convoluted workarounds, but we might want to leave that for the snippet migration.

FireChickenProductivity:
This now changes the lua snippet for block comments to be consistent with the rest by creating an inline comment. I should probably update breaking changes before merge but want feedback first. I created a separate snippet for multiline block comments so that lua could override the block comment action to use that instead. Is that an acceptable solution to minimize disruption for lua programmers?
I would appreciate suggestions on what the spoken form should be for the multiline block comment snippet.

FireChickenProductivity:
Are my changes to the snippet code acceptable? (I am waiting for feedback before testing more thoroughly)

nriley (changes requested):
Plan for snippet phrases:
`commentBlock.snippet: inline comment`
`commentMultilineBlock.snippet: block comment`

core/snippets/snippets_insert.py
```diff
@@ -69,17 +74,23 @@ def insert_snippet_with_stop_at_the_end(body):
 def compute_snippet_body_with_substitutions(
     snippet: Snippet, substitutions: dict[str, str]
 ) -> str:
-    body = snippet.body
+    return compute_snippet_text_with_substitutions(snippet.body, substitutions)
+
+
+def compute_snippet_text_with_substitutions(
+    text: str, substitutions: dict[str, str]
```
nriley: Add the snippet name as an optional argument such that it can be returned in error messages instead of the snippet body. Also:

```suggestion
    body: str, substitutions: dict[str, str]
```

nriley (changes requested):

core/snippets/snippets_insert.py
```diff
@@ -45,6 +46,11 @@ def insert_snippet_by_name_with_phrase_and_stop_at_end(name: str, phrase: str):
         body = get_snippet_body_by_name_with_phrase_substitutions(name, phrase)
         insert_snippet_with_stop_at_the_end(body)
 
+    def insert_snippet_with_substitutions(body: str, substitutions: dict[str, str]):
```
nriley:
```suggestion
    def insert_snippet_body_with_substitutions(body: str, substitutions: dict[str, str]):
```

core/snippets/snippets_insert.py
```diff
@@ -67,17 +73,29 @@ def insert_snippet_with_stop_at_the_end(body):
 def compute_snippet_body_with_substitutions(
```
nriley:
```suggestion
def expand_snippet_with_substitutions(
```

core/snippets/snippets_insert.py
```diff
@@ -67,17 +73,29 @@ def insert_snippet_with_stop_at_the_end(body):
...
     snippet: Snippet, substitutions: dict[str, str]
 ) -> str:
-    body = snippet.body
+    return compute_snippet_text_with_substitutions(
+        snippet.body, substitutions, snippet.name
+    )
+
+
+def compute_snippet_text_with_substitutions(
```
nriley:
```suggestion
def expand_snippet_body_with_substitutions(
```

nriley (changes requested):

nriley (approved):

PR #2176, opened 2026-03-30
Proper expanded obsidian support via rpc or palette

BlueDrink9:
This was initially based off someone else's obsidian support that I found online, but I can't remember whose, and I don't 
think much of it is left. I extracted and formatted the command list myself, so that will definitely be accurate, but I used GPT5 to adapt the commands into voice commands, modelling it off VSCode where possible. I have reviewed everything and tested several commands, but it is possible that something erroneous has slipped through in that process. I have tested that both command palate and cli interfaces work.

Happy to take feedback on command words


apps/obsidian/obsidian.py
```diff
@@ -14,3 +13,363 @@
...
+
+mod = Module()
+# Expand app detection similar to VSCode to support all OS variants
+mod.apps.obsidian = """
+os: linux
+and app.name: Obsidian
+"""
+mod.apps.obsidian = """
+os: mac
+and app.bundle: md.obsidian
```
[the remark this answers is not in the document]
BlueDrink9: this is how vscode's example does it, I presume talon does something funky under the hood with that setter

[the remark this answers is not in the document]
chdoc: Indeed, this double assignment is documented in the wiki: https://talon.wiki/Customization/Talon%20Framework/apps/

apps/obsidian/obsidian.py
```diff
@@ -14,3 +13,382 @@
 class CodeActions:
     def language():
         return "markdown"
+
+
+ctx = Context()
+ctx.matches = r"""
+app: Obsidian
```
nriley:
```suggestion
app: obsidian
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +13,382 @@
...
+and app.exe: /^obsidian\.exe$/i
+os: windows
+app.name: Obsidian
+"""
+
+mac_ctx = Context()
+mac_ctx.matches = r"""
+os: mac
+app: vscode
+"""
```
nriley, line 45: Please remove this context and its implementation; it is redundant with vscode.py.

apps/obsidian/obsidian.py
```diff
@@ -14,3 +13,382 @@
...
+        subprocess.run(["obsidian", "command", f"id={command_id}"], timeout=1.0)
+    else:
+        obsidian_palette_command(command_id)
+
+
+def get_command_palette_name(command_id: str) -> str:
+    try:
+        return obsidian_command_names[command_id]
+    except Exception as ex:
+        print(f"ERROR: Could not find obsidian pellet name for command_id {command_id}")
```
nriley:
```suggestion
        print(f"ERROR: Could not find obsidian palette name for command_id {command_id}")
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +13,382 @@
...
+    "workspace:undo-close-pane": "Undo close tab",
+}
+
+
+def command_uri_or_client_fallback(command_id: str):
+    """If advanced-uri is supported, use it to invoke the command.
+    Otherwise, fall back to the client"""
+
+    if settings.get("user.obsidian_use_cli"):
+        subprocess.run(["obsidian", "command", f"id={command_id}"], timeout=1.0)
```
nriley: Consider using `check_call` or `check_output` depending on how/whether the `obsidian` CLI propagates errors.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,131 @@
...
+fold properties toggle: user.obsidian("editor:toggle-fold-properties")
+line move up: user.obsidian("editor:swap-line-up")
+line move down: user.obsidian("editor:swap-line-down")
+
+# display and theme
+theme switch: user.obsidian("theme:switch")
+theme toggle: user.obsidian("theme:toggle-light-dark")
+zoom in: user.obsidian("window:zoom-in")
+zoom out: user.obsidian("window:zoom-out")
+zoom reset: user.obsidian("window:reset-zoom")
```
nriley: Please use existing `user.zoom_*` actions here.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,131 @@
...
+bold toggle: user.obsidian("editor:toggle-bold")
+italic toggle: user.obsidian("editor:toggle-italics")
+code toggle: user.obsidian("editor:toggle-code")
+blockquote toggle: user.obsidian("editor:toggle-blockquote")
+list bullet toggle: user.obsidian("editor:toggle-bullet-list")
+list number toggle: user.obsidian("editor:toggle-numbered-list")
+checkbox toggle: user.obsidian("editor:toggle-checklist-status")
+line numbers toggle: user.obsidian("editor:toggle-line-numbers")
+readable length toggle: user.obsidian("editor:toggle-readable-line-length")
+spellcheck toggle: user.obsidian("editor:toggle-spellcheck")
```
nriley: Would recommend removing toggle from all/most of these. Users are used to the commands working this way and there's no other option. Also consider removing "list" from "list bullet toggle" too — that's getting to be a really long, awkward command.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,131 @@
...
+    sleep(50ms)
+    insert(user.text or "")
+replace here [<user.text>]:
+    user.obsidian("editor:open-search-replace")
+    sleep(50ms)
+    insert(user.text or "")
+command hunt [<user.text>]:
+    user.obsidian("command-palette:open")
+    sleep(50ms)
+    insert(user.text or "")
```
nriley: The existing please command does this.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,131 @@
...
+note daily: user.obsidian("daily-notes")
+note daily previous: user.obsidian("daily-notes:goto-prev")
+note daily next: user.obsidian("daily-notes:goto-next")
+
+# links and navigation
+link follow | follow this: user.obsidian("editor:follow-link")
+link open: user.obsidian("editor:open-link-in-new-leaf")
+link split: user.obsidian("editor:open-link-in-new-split")
+go back: user.obsidian("app:go-back")
+go forward: user.obsidian("app:go-forward")
```
nriley: Please implement the actions in the user.navigation tag instead.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,131 @@
...
+
+# reading/source toggle
+view toggle: user.obsidian("markdown:toggle-preview")
+bar switch: user.obsidian("app:toggle-left-sidebar")
+panel switch: user.obsidian("app:toggle-right-sidebar")
+ribbon switch: user.obsidian("app:toggle-ribbon")
+
+# window and settings
+window reload | reload it: user.obsidian("app:reload")
+window close: user.obsidian("workspace:close-window")
```
nriley: Please use `app.window_close`.

nriley (changes requested):
From the community backlog session — Overall this looks like it's getting there, but there are many cases where you define a voice command where community already has a corresponding voice command and action. Please go through your voice commands, search community for corresponding words and make changes as appropriate. Thanks!

apps/obsidian/obsidian.py
nriley: Our standard is not to use capitalized words in defining generic app names.

apps/obsidian/obsidian.py
BlueDrink9: oops, left that in when copy-pasting, sorry

apps/obsidian/obsidian.talon
BlueDrink9: from memory this doesn't close the os window, I'll double check what it's actually doing and whether community has an alternative. Thanks for the review!

nriley: If it doesn't close the OS window then it shouldn't use the same voice command. If it does and you can do it more reliably than the existing implementation then please override it for Obsidian. Thanks!

apps/obsidian/obsidian.py
BlueDrink9: Updated to obsidian to make sure that the please command works on mac with key fallback

BlueDrink9:
Ready for fresh review

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,432 @@
...
+        print(f"Error running obsidian command via cli: {e}")
+        print("Obsidian might not be on PATH?")
+    except subprocess.CalledProcessError as e:
+        # Note: As of at least obsidian version 1.12, the CLI never emits a failure exit code, so none of these will trigger currently, but a future update to obsidian might fix that.
+        print(f"Error running obsidian command via cli: {e.output}")
+        if "Command line interface is not enabled" in e.output:
+            print(
+                "Talon settings say to use the obsidian CLI for obsidian commands, but CLI interface control is disabled in obsidian settings."
+            )
+            print("Please turn it on in Settings > General >Advanced")
```
nriley: Consider using a notification rather than a print statement as not all Talon users leave the log visible.

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,432 @@
...
+def obsidian_run_cli_command(command_id: str):
+    try:
+        subprocess.check_output(
+            ["obsidian", "command", f"id={command_id}"], timeout=1.0
+        )
+    except OSError as e:
+        print(f"Error running obsidian command via cli: {e}")
+        print("Obsidian might not be on PATH?")
+    except subprocess.CalledProcessError as e:
+        # Note: As of at least obsidian version 1.12, the CLI never emits a failure exit code, so none of these will trigger currently, but a future update to obsidian might fix that.
```
nriley: This comment is a little confusing given you parse the output afterward. Consider clarifying the cases under which the process returns an error code at the moment and may do so in future.

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,432 @@
...
+
+    if settings.get("user.obsidian_use_cli"):
+        obsidian_run_cli_command(command_id)
+    else:
+        obsidian_palette_command(command_id)
+
+
+def get_command_palette_name(command_id: str) -> str:
+    try:
+        return obsidian_command_names[command_id]
```
nriley, line 306: Consider using `obsidian_command_names.get`.

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,432 @@
...
+def command_palette(key_fallback: str):
+    if settings.get("user.obsidian_use_cli"):
+        obsidian_run_cli_command("command-palette:open")
+    else:
+        actions.key(key_fallback)
+
+
+@mod.action_class
+class Actions:
+    def command_palette():
```
nriley: You already implement the existing action `user.command_search` so there is no need for a duplicate action.

nriley (changes requested):
From the Community Backlog Session — please focus on duplication regarding the command palette. You have methods and functions named similarly that do different things, and it's not clear if these have overlapping functionality. Thanks!

apps/obsidian/obsidian.py
BlueDrink9: I'll need some help rewording it then. It's as I say: we never get a failure code. But we might in future, so I decided to leave in the parsing work I had already set up, even if current obsidian versions won't ever trigger it

nriley: If there's information on stdout or stderr (even though the return code is 0) then you can still parse it. This is what the return value of subprocess.check_output is. See the documentation; you can also capture stderr if that's where the Obsidian CLI tool is sending this information. Would recommend trying it with shell pipelines/redirection first so it's a bit easier to iterate.

apps/obsidian/obsidian.py
BlueDrink9: Since I need to notify on error, I decided it's cleaner to leave as a try except

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,448 @@
...
+        # Note: As of at least obsidian version 1.12, the CLI never
+        # emits a failure exit code, so this block won't trigger unless that changes.
+        parse_obsidian_errors(e.output, command_id)
+
+
+def parse_obsidian_errors(message, command_id):
+    if "Command line interface is not enabled" in message:
+        actions.app.notify(
+            """Talon settings say to use the obsidian CLI for obsidian commands, but CLI interface control is disabled in obsidian settings.
+            Please turn it on in Settings > General >Advanced
```
nriley:
```suggestion
            Please turn it on in Settings > General > Advanced
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,448 @@
...
+    except subprocess.CalledProcessError as e:
+        # Note: As of at least obsidian version 1.12, the CLI never
+        # emits a failure exit code, so this block won't trigger unless that changes.
+        parse_obsidian_errors(e.output, command_id)
+
+
+def parse_obsidian_errors(message, command_id):
+    if "Command line interface is not enabled" in message:
+        actions.app.notify(
+            """Talon settings say to use the obsidian CLI for obsidian commands, but CLI interface control is disabled in obsidian settings.
```
nriley:
```suggestion
            """Talon settings say to use the Obsidian CLI for Obsidian commands, but CLI interface control is disabled in Obsidian settings.
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,448 @@
...
+mac_ctx.matches = r"""
+os: mac
+app: obsidian
+"""
+
+mod.setting(
+    "obsidian_use_cli",
+    type=bool,
+    default=False,
+    desc="""Whether to use cli instead of palette to send obsidian commands. Palette (false) will work out of the box but requires the language to be set to english and is slower. cli (true) it requires command line interface to be enabled in the obsidian general options, and requires obsidian to be on the PATH""",
```
nriley:
```suggestion
    desc="""Whether to use the Obsidian CLI instead of palette to send Obsidian commands. Palette (false) will work out of the box but requires the language to be set to English and is slower. CLI (true) requires command line interface to be enabled in the Obsidian general options, and requires Obsidian to be on the PATH""",
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,448 @@
...
+            timeout=1.0,
+            capture_output=True,
+            encoding="utf-8",
+            check=True,
+        )
+        if result.stdout:
+            parse_obsidian_errors(result.stdout, command_id)
+    except OSError as e:
+        actions.app.notify(
+            f"""Error running obsidian command via cli. Obsidian might not be on PATH?
```
nriley:
```suggestion
            f"""Error running Obsidian command via CLI. The obsidian command might not be on your PATH?
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,448 @@
...
+        actions.app.notify(
+            """Talon settings say to use the obsidian CLI for obsidian commands, but CLI interface control is disabled in obsidian settings.
+            Please turn it on in Settings > General >Advanced
+            """
+        )
+        actions.app.notify(f"Falling back to palette command for {command_id}")
+        obsidian_palette_command(command_id, fallback=True)
+    elif re.match('Command "[^"]*"not found', message):
+        actions.app.notify(
+            f"Command '{command_id}' does not seem to be a valid obsidian command"
```
nriley:
```suggestion
            f"Command '{command_id}' does not seem to be a valid Obsidian command."
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,448 @@
...
+    else:
+        obsidian_palette_command(command_id)
+
+
+def get_command_palette_name(command_id: str) -> str:
+    try:
+        return obsidian_command_names[command_id]
+    except KeyError as ex:
+        actions.app.notify(
+            f"ERROR: Could not find obsidian palette name for command_id {command_id}"
```
nriley:
```suggestion
            f"ERROR: Could not find Obsidian palette name for command_id {command_id}"
```

apps/obsidian/obsidian.py
```diff
@@ -14,3 +14,448 @@
...
+
+    def multi_cursor_add_below():
+        actions.user.obsidian("editor:add-cursor-below")
+
+    def multi_cursor_disable():
+        actions.key("escape")
+
+    def multi_cursor_enable():
+        actions.skip()
+
```
nriley:
```suggestion
```
if there is a default action implementation that does not raise an exception, there is no need for this.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,113 @@
...
+file copy full path: user.obsidian("workspace:copy-full-path")
+file copy url: user.obsidian("workspace:copy-url")
+
+# daily notes
+daily open: user.obsidian("daily-notes")
+daily previous: user.obsidian("daily-notes:goto-prev")
+daily next: user.obsidian("daily-notes:goto-next")
+
+# links and navigation
+link follow | follow this: user.obsidian("editor:follow-link")
```
nriley:
```suggestion
follow: user.obsidian("editor:follow-link")
```
other apps just have this as follow.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,113 @@
...
+
+# find and replace helpers (phrase-capable)
+search here [<user.text>]:
+    user.obsidian("editor:open-search")
+    sleep(50ms)
+    insert(user.text or "")
+search all [<user.text>]:
+    user.obsidian("global-search:open")
+    sleep(50ms)
+    insert(user.text or "")
```
nriley:
```suggestion
```
these are duplicates of `hunt this` and `hunt all` for which you have already implemented the corresponding actions. However, there is an inconsistency between your edit.find implementation and `search here` — one contains a sleep and the other one doesn't. Please update your action implementation if the sleep is actually necessary in order to prevent issues with inserting the search text.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,113 @@
...
+replace here [<user.text>]:
+    user.obsidian("editor:open-search-replace")
+    sleep(50ms)
+    insert(user.text or "")
```
nriley: Same as above but with `user.replace`.

apps/obsidian/obsidian.talon
```diff
@@ -1,4 +1,113 @@
...
+
+# folding and movement
+fold all: user.obsidian("editor:fold-all")
+unfold all: user.obsidian("editor:unfold-all")
+fold more: user.obsidian("editor:fold-more")
+fold less: user.obsidian("editor:fold-less")
+fold toggle: user.obsidian("editor:toggle-fold")
+fold properties toggle: user.obsidian("editor:toggle-fold-properties")
+line move up: user.obsidian("editor:swap-line-up")
+line move down: user.obsidian("editor:swap-line-down")
```
nriley: Please implement instead `edit.line_swap_up / edit.line_swap_down`.

nriley (dismissed):
From the Community Backlog Session — we fixed a few things but there's a few more to do. Getting very close. Thanks.

BlueDrink9:
addressed

AndreasArvidsson (approved):

PR #2177, opened 2026-03-31
Document handling of actions irrelevant to a context

FireChickenProductivity:
This describes the design patterns we decided on in a Community backlog session, which are described [here](https://github.com/talonhub/community/issues/2143#issuecomment-4103783099).


CONTRIBUTING.md
```diff
@@ -15,3 +15,5 @@ Each of the principles is numbered for easy referencing. The body is formatted a
...
 - P08 - Use the stored state management actions defined in [storage_state.py](core/stored_state_management/stored_state.py) to store state on disk. This keeps stored state within the stored_state directory, which makes it easier for users to keep track of and git ignore.
+- P09 - When the default behavior of an action is undesirable in a context, override it to throw an exception. This has the advantage of stopping execution, which prevents the calling code from continuing with incorrect assumptions. You may additionally provide a notification.
+- P10 - Do not implement an action in a context where it is irrelevant if it has no default implementation because Talon handles this case for us.
```
nriley:
```suggestion
- P09 - If there is no reasonable implementation for an action in a context, it should raise an exception. This has the advantage of stopping execution, preventing calling code from continuing with incorrect assumptions.
    * If there is an (inappropriate) implementation for the action active in the context, then define an implementation in the specific context that raises an exception. You may additionally display a notification if it is helpful to let the user know that the action did not complete successfully.
    * If there is *no* implementation for the action active in the context, don't implement it. Talon will appropriately raise an exception when executing the action in the context.
```

nriley (approved):

PR #2182, opened 2026-04-14
Add support for the dolphin file manager

BlueDrink9:


apps/dolphin/dolphin.py
```diff
@@ -0,0 +1,49 @@
+from talon import Context, Module, actions, clip, ui
+
+# App definition. Dolphin is the default file manager for KDE plasma, one of
+# the two biggest linux desktop environments.
+mod = Module()
+mod.apps.dolphin = """
+os: linux
+and app: dolphin
```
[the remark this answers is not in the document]
BlueDrink9: App has different meanings in that context

FireChickenProductivity:
Just a note: `(refresh | reload) it: browser.reload()` is already a command in Community. I wonder if it might be worth copying that spoken form if you are using `reload it` and if we should at some point abstract out reloading from just being a browser thing.

FireChickenProductivity:
Another note: `file rename` is defined in several contexts. We might want to put that behind a tag so we can use action overriding. 
(These notes are mostly intended for the next backlog session)

apps/dolphin/dolphin.py
```diff
@@ -0,0 +1,49 @@
+from talon import Context, Module, actions, clip, ui
```
FireChickenProductivity: Can you remove the unused imports?

apps/dolphin/dolphin.talon
```diff
@@ -0,0 +1,8 @@
+app: dolphin
+-
+
+tag(): user.tabs
+tag(): user.file_manager
```
FireChickenProductivity, line 5: Should you also activate the `user.navigation` tag?

BlueDrink9:
I was surprised to find that file rename was not already in the file manager tag.

I find myself creating a reload or refresh command for quite a lot of applications, comma it definitely seems generally useful outside of just browsers. That being said, it does feel like a nuisance to have a tag for just a single command, but I suppose that's better than making users redefine the command for every application.

apps/dolphin/dolphin.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def file_manager_show_properties():
+        actions.key("alt-enter")
+
+    def file_manager_open_directory(path: str):
+        actions.key("ctrl-l")
+        actions.insert(path)
+        actions.key("enter")
+
+    def file_manager_new_folder(name: str = None):
```
FireChickenProductivity: Note for backlog session: In some places we do `: str = None` in Community actions. Should we instead do `: Optional[str] = None`?

nriley: From the community backlog session — we agree this makes sense, but should be in a separate PR.

apps/dolphin/dolphin.py
nriley: This looks like it could be recursive as you are defining `app: dolphin` by matching `app: dolphin`. Whether or not it works, it is confusing to read as in general, we only use `app:` to refer to apps that are defined with `mod.apps`. Consider using `app.name`.

nriley (changes requested):
From the Community Backlog session — please address all of @FireChickenProductivity 's issues. Thanks!

nriley (approved):

StructSeeker:
You seems to omit a `from typing import Optional` that cause error message in windows?
@BlueDrink9

nriley:
Yup, sorry we did not test it! The needed changes are in another PR currently; see https://github.com/talonhub/community/pull/2191/changes/fab4300408539ede78c4bdb8e977faaaf988ec1c. @BlueDrink9 if you want to submit as a separate PR then happy to merge ASAP.

FireChickenProductivity:
Adding the optional import actually causes another error because it makes the type signature inconsistent with the base action.

FireChickenProductivity:
I submitted the quickest fix: https://github.com/talonhub/community/pull/2193

BlueDrink9:
Oops, thank you! I'm on beta so for whatever reason didn't get the error; or maybe just didn't notice it in the log. Thanks FireChicken, and sorry!

PR #2184, opened 2026-04-25
pre-commit updates with Ruff

nriley:
From the Community Backlog session — we are using a combination of pre-commit hooks which are no longer supported, no longer relevant, slow or have newer/better alternatives.

```
  - repo: https://github.com/pre-commit/mirrors-prettier
    rev: "v4.0.0-alpha.8"
    hooks:
      - id: prettier
```
this is in use for JavaScript, CSS, JSON, etc. — probably for JSON/YAML/Markdown. The repository above is archived and we should find an alternative. @AndreasArvidsson recommends we give [oxc](https://oxc.rs) a look.

```
  - repo: https://github.com/ikamensh/flynt
    rev: "1.0.6"
    hooks:
      - id: flynt
```
I think we can remove this entirely as it's there to support migration to Python 3.6 string formatting.

```
  - repo: https://github.com/pycqa/isort
    rev: 8.0.1
    hooks:
      - id: isort
  - repo: https://github.com/psf/black-pre-commit-mirror
    rev: 26.3.1
    hooks:
      - id: black
```
these two can be replaced by [Ruff](https://docs.astral.sh/ruff/), as in this PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-minimum_pre_commit_version: "3.2.0"
+minimum_pre_commit_version: "4.5.0"
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v6.0.0
@@ -13,7 +13,7 @@ repos:
       - id: detect-private-key
       - id: fix-byte-order-marker
       # NB. To avoid sometimes needing multiple runs, we need:
-      # - trailing-whitespace BEFORE end-of-file-fixer,
+      # - trailing-whitespace BEFORE end-of-sfile-fixer,
       #   otherwise trailing newline followed by whitespace, "\n ",
       #   will need multiple runs.
       # - end-of-file-fixer   BEFORE mixed-line-ending,
@@ -34,7 +34,8 @@ repos:
     hooks:
       # Run the linter.
       - id: ruff-check
-        types_or: [python, pyproject]
+        # Want to use pyproject instead of toml, but pre-commit.ci doesn't yet support it.
+        types_or: [python, toml]
         args: [--fix]
       # Run the formatter.
       - id: ruff-format
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,20 +28,16 @@ repos:
     hooks:
       - id: prettier
 
-  - repo: https://github.com/ikamensh/flynt
-    rev: "1.0.6"
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
+    rev: v0.15.12
     hooks:
-      - id: flynt
-
-  - repo: https://github.com/pycqa/isort
-    rev: 8.0.1
-    hooks:
-      - id: isort
-
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 26.3.1
-    hooks:
-      - id: black
+      # Run the linter.
+      - id: ruff-check
+        types_or: [python, pyproject]
+        args: [--fix]
+      # Run the formatter.
+      - id: ruff-format
 
   - repo: https://github.com/cursorless-dev/talon-tools
     rev: v0.10.2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
       - id: detect-private-key
       - id: fix-byte-order-marker
       # NB. To avoid sometimes needing multiple runs, we need:
-      # - trailing-whitespace BEFORE end-of-sfile-fixer,
+      # - trailing-whitespace BEFORE end-of-file-fixer,
       #   otherwise trailing newline followed by whitespace, "\n ",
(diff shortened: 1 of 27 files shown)


apps/mintty/mintty_win.py
```diff
@@ -67,8 +67,8 @@ def delete_line():
 @ctx.action_class("user")
 class UserActions:
     def file_manager_current_path():
-        path = ui.active_window().title
-        path = get_win_path(path)
+        title = ui.active_window().title
+        path = get_win_path(title)
 
         if path in directories_to_remap:
             path = directories_to_remap[title]
```
AndreasArvidsson:
```suggestion
            path = directories_to_remap[path]
```

AndreasArvidsson (commented):
Looks good. Just had one small change I think we need to make.

.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
       - id: detect-private-key
       - id: fix-byte-order-marker
       # NB. To avoid sometimes needing multiple runs, we need:
-      # - trailing-whitespace BEFORE end-of-file-fixer,
+      # - trailing-whitespace BEFORE end-of-sfile-fixer,
```
AndreasArvidsson:
```suggestion
      # - trailing-whitespace BEFORE end-of-file-fixer,
```

nriley:
CI is happy now. I ran into an issue with `pre-commit.ci` versions which I [filed](https://github.com/pre-commit-ci/issues/issues/265) and worked around.

.pre-commit-config.yaml
```diff
@@ -28,20 +28,17 @@ repos:
     hooks:
       - id: prettier
 
-  - repo: https://github.com/ikamensh/flynt
-    rev: "1.0.6"
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
```
AndreasArvidsson: I would probably remove these three comments. Anyone familiar with pre commit will know what they do

.pre-commit-config.yaml
```diff
@@ -28,20 +28,17 @@ repos:
...
-  - repo: https://github.com/pycqa/isort
-    rev: 8.0.1
-    hooks:
-      - id: isort
-
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 26.3.1
-    hooks:
-      - id: black
+      # Run the linter.
```
AndreasArvidsson: 2

.pre-commit-config.yaml
```diff
@@ -28,20 +28,17 @@ repos:
...
+      - id: ruff-check
+        # Want to use pyproject instead of toml, but pre-commit.ci doesn't yet support it.
+        types_or: [python, toml]
+        args: [--fix]
+      # Run the formatter.
```
AndreasArvidsson: 3

AndreasArvidsson (approved):

.pre-commit-config.yaml
```diff
@@ -29,20 +29,14 @@ repos:
...
-    hooks:
-      - id: isort
-
-  - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 26.3.1
-    hooks:
-      - id: black
+      - id: ruff-check
+        # Want to use pyproject instead of toml, but pre-commit.ci doesn't yet support it.
+        types_or: [python, toml]
```
auscompgeek, line 37: FWIW pre-commit.ci always runs `pre-commit run -a` so the hook would always run in CI anyway.

nriley: Sorry, not sure how that addresses my concern. The issue is if I use `types_or: [python, pyproject]` to limit the types of files that Ruff evaluates, `pre-commit.ci` barfs because it does not use a new enough version of `identify` — see https://github.com/pre-commit-ci/issues/issues/265. I'll try to get a PR submitted today to get this fixed as requested (update: [done](https://github.com/pre-commit-ci/runner-image/pull/329) and merged; I'll try this again next week as not sure how long it takes for the production service to be updated).

auscompgeek: If you don't specify this at all ruff would still check (by necessity) the pyproject.toml, wouldn't it?

AndreasArvidsson: The Ruff pre commit hook specifies by default `types_or: [python, pyi, jupyter]` so I don't think so.
https://github.com/astral-sh/ruff-pre-commit/blob/6fec9b7edb08fd9989088709d864a7826dc74e80/.pre-commit-hooks.yaml#L6

auscompgeek: The `types_or` tells `pre-commit` what files to collect and pass to the hook. But ruff _needs_ to read `pyproject.toml` since that's how you configure ruff in the first place. So it should report issues with parsing its own config regardless?

Sure, adding it here would mean that a local `pre-commit run` (or `prek run`) will rerun ruff if you change the pyproject.toml (without touching any Python files). But that will never be an issue for CI.

PR #2185, opened 2026-04-25
pre-commit updates with Oxfmt

AndreasArvidsson:


.pre-commit-config.yaml
```diff
@@ -23,10 +23,12 @@ repos:
...
       - id: mixed-line-ending
 
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v4.0.0-alpha.8"
+  - repo: https://github.com/oxc-project/mirrors-oxfmt
+    rev: "v0.45.0"
     hooks:
-      - id: prettier
+      - id: oxfmt
+        types_or: [markdown, yaml, toml, json]
```
nriley, line 30: This seems to do a different set of TOML checks to Ruff. OK with both of these?

nriley (approved):
I'm OK either way…

.pre-commit-config.yaml
AndreasArvidsson: We should probably pick one of them. I have never used ruff on tom files so just for familiarity I would probably use  oxfmt.
From your pull request it doesn't look like ruff formated `pythonpath`.

nriley: Looks like they do different things; see https://docs.astral.sh/ruff/rules/invalid-pyproject-toml/. So it doesn't check formatting but content; seems like we can leave both enabled.

AndreasArvidsson: Sounds good!

Issue #2189, opened 2026-05-09
Decide on default Ruff configuration


nriley:
Diff of current pyproject.toml producing the below:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 5c0ef2c00..de98ee1b6 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,9 +6,21 @@ target-version = "py310"
 
 [tool.ruff.lint]
 select = [
+    "E", # pycodestyle
+    "F", # Pyflakes
     "I", # isort
+    "UP", # pyupgrade
+    "B", # flake8-bugbear
+    "SIM", # flake8-simplify
 ]
-ignore = ["E722", "E741"]
+ignore = [
+    "E501", # line-too-long (until we decide)
+    "E722", # except: (consider replacing with catching Exception)
+    "E741", # variable names l, O, or I
+    "UP007", # Union[X, Y] -> X | Y not supported in Talon 0.4 actions
+    "UP045", # Optional[X] -> X | None not supported in Talon 0.4 actions
+]
+
```

Unique issues below:
```py
B006 Do not use mutable data structures for argument defaults
B007 Loop control variable `i` not used within loop body
B007 Loop control variable `line` not used within loop body
B028 No explicit `stacklevel` keyword argument found
B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
B905 `zip()` without an explicit `strict=` parameter
F402 Import `app` from line 1 shadowed by loop variable
F811 Redefinition of unused `Actions` from line 327
F821 Undefined name `talon`
F841 [*] Local variable `ex` is assigned to but never used
SIM102 Use a single `if` statement instead of nested `if` statements
SIM108 Use ternary operator `mapped_source = source if source.isupper() else source.lower()` instead of `if`-`else`-block
SIM108 Use ternary operator `mode = "mixed" if "command" in modes else "dictation"` instead of `if`-`else`-block
SIM108 Use ternary operator `result = title.split(" — ")[0] if is_mac else title.split(" - ")[0]` instead of `if`-`else`-block
SIM108 Use ternary operator `text = text + f"({selection})" if selection else text + "()"` instead of `if`-`else`-block
SIM108 Use ternary operator `value = m.number if hasattr(m, "number") else m[0]` instead of `if`-`else`-block
[...]

PR #2191, opened 2026-05-10
Ruff updates

nriley:
- **Ruff configuration updates and automatic Ruff fixes.**
- **Manual Ruff fixes and "noqa" ignores.**

See #2189 for more.


core/vocabulary/vocabulary.py
```diff
@@ -218,7 +219,7 @@ def append_to_vocabulary(rows: dict[str, str]):
     vocabulary_file_path = actions.user.get_vocabulary_file_path()
     with open(str(vocabulary_file_path)) as file:
         line = None
-        for line in file:
+        for line in file:  # noqa: B007
```
AndreasArvidsson: Please add a comment explaining why we are doing the for loop with no op

core/deprecations.py
```diff
@@ -142,7 +142,7 @@ def deprecate_command(time_deprecated: str, name: str, replacement: str):
             f' Instead, say: "{replacement}".'
             f" See {os.path.join(REPO_DIR, 'BREAKING_CHANGES.txt')}"
         )
-        warnings.warn(msg, DeprecationWarning)
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
AndreasArvidsson: Please add a comment explaining stack level and why we need that

core/user_settings.py
```diff
@@ -98,7 +99,7 @@ def append_to_csv(filename: str, rows: dict[str, str], private: bool = False):
 
     with open(str(path)) as file:
         line = None
-        for line in file:
+        for line in file:  # noqa: B007
```
AndreasArvidsson: Same here please add a comment

AndreasArvidsson (dismissed):

core/deprecations.py
nriley: Turns out this didn't work anyway; I misread and didn't realize that a voice command was triggering the action. So any Talon stack would just be internals of command recognition and not helpful!

plugin/microphone_selection/microphone_selection.py
```diff
@@ -57,7 +57,7 @@ def microphone_selection_hide():
 
     def microphone_select(index: int):
         """Selects a micropohone"""
-        if 1 <= index and index <= len(microphone_device_list):
+        if index >= 1 and index <= len(microphone_device_list):
```
auscompgeek: I'd make this a chained comparison
```suggestion
        if 1 <= index <= len(microphone_device_list):
```

AndreasArvidsson: Personally I think chained comparisons are harder to read

core/homophones/homophones.py
```diff
@@ -222,7 +222,7 @@ def homophones_select(number: int) -> str:
         app.notify(error)
         raise error
 
-    def homophones_get(word: str) -> Optional[[str]]:
+    def homophones_get(word: str) -> Optional[str]:
```
AndreasArvidsson: This should probably be `Optional[list[str]]`?

core/user_settings.py
```diff
@@ -92,15 +93,19 @@ def on_update(f):
     return decorator
 
 
+def needs_final_newline(path: Union[Path, str]) -> bool:
+    with open(path) as file:
+        line = None
+        for line in file:  # noqa: B007
+            pass  # iterate through each line in file
+    return line is not None and not line.endswith("\n")
```
AndreasArvidsson, line 101: I think this logic is easier to follow

```suggestion
    with open(path) as file:
        last_line = None
        for line in file:  # noqa: B007
            last_line = line
    return last_line is not None and not last_line.endswith("\n")
```

core/homophones/homophones.py
nriley: Yes, sorry! Will fix.

core/user_settings.py
nriley: I'm pretty sure it's less efficient that way though, and not *that* much easier to read. Can talk next time you're on the backlog session, no rush!

AndreasArvidsson (approved):

PR #2192, opened 2026-05-10
Split Warp config: Linux/Windows shortcuts in warp_win_linux.py, Mac in warp_mac.py

flomillot:
## Summary

Adds support for Warp on Linux and Windows. Previously `apps/warp/warp.py` only declared `mod.apps.warp` for macOS and held Mac-specific action overrides under an unscoped `app: warp` matcher.

- Renames `apps/warp/warp.py` → `apps/warp/warp_mac.py` and scopes its context with `os: mac` so its Mac-specific actions (`cmd-{n}` tab jumps, `alt-arrow` word navigation) only apply on Mac.
- Adds `apps/warp/warp_win_linux.py` (named for consistency with `firefox_win_linux.py` and `opera_win_linux.py`) declaring `mod.apps.warp` for Linux (`app.name: warp`) and Windows (`app.name: Warp` + `app.exe: warp.exe`, per P04). Its context is scoped `(os: linux OR os: windows) AND app: warp` so its overrides do not leak onto macOS.
- Overrides matching Warp's Linux/Windows defaults:
  - `app.tab_open` / `app.tab_close` → `ctrl-shift-t` / `ctrl-shift-w`
  - `user.tab_jump` / `user.tab_final` → `ctrl-{n}`
  - `edit.word_left` / `edit.word_right` → `ctrl-left` / `ctrl-right`


flomillot:
@david-tejada for https://github.com/talonhub/community/pull/1738

I don't find this condition convenient when we have more than nine tab. 
```
@ctx.action_class("user")
class UserActions:
    def tab_jump(number: int):
        if number < 9:
            actions.key(f"cmd-{number}")

    def tab_final():
        actions.key("cmd-9")
```
Can we change this ?

david-tejada:
> @david-tejada for #1738
> 
> I don't find this condition convenient when we have more than nine tab.
> 
> ```
> @ctx.action_class("user")
> class UserActions:
>     def tab_jump(number: int):
>         if number < 9:
>             actions.key(f"cmd-{number}")
> 
>     def tab_final():
>         actions.key("cmd-9")
> ```
> 
> Can we change this ?

What do you mean that you don't find this condition convenient? This is the way it works, at least on Mac. If you press `cmd-9`, you go to the last tab, no matter how many tabs there are.

apps/warp/warp_win_linux.py
```diff
@@ -0,0 +1,48 @@
+from talon import Context, Module, actions
+
+mod = Module()
+
+mod.apps.warp = """
+os: linux
+and app.name: warp
+"""
+mod.apps.warp = """
```
auscompgeek, line 9: An application's matchers should all be in a single file.

flomillot: I did the same as for Firefox or Opera. Do you mean I shouldn't have ?

flomillot:
> > @david-tejada for #1738
> > I don't find this condition convenient when we have more than nine tab.
> > ```
> > @ctx.action_class("user")
> > class UserActions:
> >     def tab_jump(number: int):
> >         if number < 9:
> >             actions.key(f"cmd-{number}")
> > 
> >     def tab_final():
> >         actions.key("cmd-9")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Can we change this ?
> 
> What do you mean that you don't find this condition convenient? This is the way it works, at least on Mac. If you press `cmd-9`, you go to the last tab, no matter how many tabs there are.

My bad. I didn't think about it, and didn't check the keyboard shortcuts either. Indeed, we can only choose tabs between one and eight anyway.

apps/warp/warp_win_linux.py
flomillot: I didn't want to duplicate the shortcuts on two files. How exactly I should have done it ?

apps/warp/warp_win_linux.py
```diff
@@ -0,0 +1,48 @@
...
+os: windows
+and app.name: Warp
```
AndreasArvidsson:
```suggestion
```

AndreasArvidsson:
```suggestion
```

AndreasArvidsson (approved):

PR #2200, opened 2026-05-21
Insert between wait

BlueDrink9:


FireChickenProductivity:
In Community we use actions.sleep instead of time.sleep.

BlueDrink9:
oops good point

nriley:
Is there a reason you can't use the other/built-in wait settings in this case?

BlueDrink9:
The built-in settings are too universal, in that they apply to all keystrokes entered, whereas I often only have issues specifically when mixing left and right with inserting text, leading to the order getting messed up. This is a targeted fix for a flaky input

nriley:
Can you give an example of an app/platform that has this specific problem? Are you using paste to insert in it?

BlueDrink9:
Most of the things I use insert between for are not long enough to use paste to insert

BlueDrink9:
I mainly added this change to my fork for VS Code before insert between switched to using the snippet engine, but I occasionally also get issues with Kate, the editor on Linux, and occasionally in browser windows.

core/edit/insert_between.py
```diff
@@ -1,12 +1,22 @@
...
+    desc="Time in milliseconds to sleep after inserting text with `insert_between` (e.g. when using paired delimiters like 'box' or 'round'), before moving the cursor back. Useful to set on a per-application basis, to prevent moving the moving the cursor before text is inserted",
+)
+
 
 @mod.action_class
 class module_actions:
     def insert_between(before: str, after: str):
         """Insert `before + after`, leaving cursor between `before` and `after`. Not entirely reliable if `after` contains newlines."""
         actions.insert(f"{before}{after}")
+        sleep(int(settings.get("user.insert_between_wait", 0)) / 1000)
```
nriley:
```suggestion
        actions.sleep(settings.get("user.insert_between_wait", 0))
```

core/edit/insert_between.py
```diff
@@ -1,12 +1,22 @@
-from talon import Module, actions
+from time import sleep
+
```
nriley:
```suggestion
```

settings.talon
```diff
@@ -109,6 +109,9 @@ settings():
     # Time in seconds to wait for the clipboard to change when trying to get selected text
     # user.selected_text_timeout = 0.25
 
+    # Time in milliseconds to sleep after inserting text with `insert_between` (e.g. when using paired delimiters like 'box' or 'round'), before moving the cursor back. Useful to set on a per-application basis, to prevent moving the moving the cursor before text is inserted.
```
```suggestion
    # Time in seconds to sleep after inserting text with `insert_between` (e.g. when using paired delimiters like 'box' or 'round'), before moving the cursor back. Useful to set on a per-application basis, to prevent moving the moving the cursor before text is inserted.
```

core/edit/insert_between.py
```diff
@@ -1,12 +1,22 @@
...
+from talon import Module, actions, settings
 
 mod = Module()
 
+mod.setting(
+    "insert_between_wait",
+    type=int,
```
nriley:
```suggestion
    type=float,
```

core/edit/insert_between.py
```diff
@@ -1,12 +1,20 @@
-from talon import Module, actions
+from talon import Module, actions, settings
 
 mod = Module()
 
+mod.setting(
+    "insert_between_wait",
+    type=float,
+    default=0,
+    desc="Time in milliseconds to sleep after inserting text with `insert_between` (e.g. when using paired delimiters like 'box' or 'round'), before moving the cursor back. Useful to set on a per-application basis, to prevent moving the moving the cursor before text is inserted",
```
nriley:
```suggestion
    desc="Time in seconds to sleep after inserting text with `insert_between` (e.g. when using paired delimiters like 'box' or 'round'), before moving the cursor back. Useful to set on a per-application basis, to prevent moving the moving the cursor before text is inserted",
```

nriley (changes requested):

nriley (approved):

PR #2201, opened 2026-05-23
use tag(): user.code_keywords, deprecate "state <keyword>" for non-statements

chdoc:
I only deprecated the commands for those keywords that should be entered using snippets and the likes.


chdoc:
pre-commit.ci autofix

BREAKING_CHANGES.txt
```diff
@@ -7,6 +7,7 @@ and when the change was applied given the delay between changes being
 submitted and the time they were reviewed and merged.
 
 ---
+* 2026-05-17 Deprecate `state local`, `state end`, etc. for Lua in favour of using `code_keyword`.
```
FireChickenProductivity: I recommend mentioning that user.code_keyword is a capture, using the `user.` prefix, mentioning put here and maybe linking to the wiki.

nriley:
```suggestion
* 2026-05-17 Deprecate `state local`, `state end`, etc. for Lua in favour of using `put` via the list/capture/tag `user.code_keyword`.
```

nriley (approved):

nriley (approved):

PR #2218, opened 2026-06-03
Make context sensitive dictation work for Atlassian web apps

FireChickenProductivity:
Using space as the peek character for context dictation is not working, so this uses a different character.

### DIFF
--- a/apps/atlassian/atlassian.talon
+++ b/apps/atlassian/atlassian.talon
@@ -1,0 +1,6 @@
+tag: browser
+browser.host: /atlassian\.(com|net|cloud)/
+-
+
+settings():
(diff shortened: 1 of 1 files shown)


apps/atlassian/atlassian.talon
```diff
@@ -0,0 +1,6 @@
+tag: browser
+browser.host: /atlassian\.(com|net|cloud)/
```
nriley:
```suggestion
browser.host: /atlassian\.(com|net|cloud)$/
```

nriley (changes requested):

nriley (approved):

PR #2223, opened 2026-06-15
Support escaping snippet spaces

FireChickenProductivity:


core/snippets/snippets_parser.py
```diff
@@ -280,6 +281,15 @@ def normalize_snippet_body_tabs(body: str | None) -> str:
     return "\n".join(normalized_lines)
 
 
+def escape_spaces(body: str) -> str:
+    # treat double backslash as escaped backslash
+    # treat backslash space as space
+    matches = [m for m in re.finditer(ESCAPED_SPACE_EXPRESSION, body)]
+    for match in reversed(matches):
+        body = body[: match.end() - 1] + body[match.end() :]
+    return body.replace("\\\\", "\\")
```
nriley: From the community backlog session — consider replacing this with a couple string replacements that replace the escaped backslash with a rarely used/non-printable Unicode character. If this character is in use in the snippet, it is fine to reject it.

nriley (changes requested):

core/snippets/snippets_parser.py
```diff
@@ -447,13 +468,29 @@ def parse_body(text: str) -> Union[str, None]:
     if match_leading is None:
         return None
 
-    body = text[match_leading.start() :].rstrip()
+    body = rstrip_except_escaped_space(text[match_leading.start() :])
```
FireChickenProductivity: Instead of my current special replacement of the rstrip, it might be better to just move the stripping logic to a better place.

core/snippets/README.md
```diff
@@ -13,12 +13,14 @@ Custom format to represent snippets.
...
 
 - A `.snippet` file can contain multiple snippet documents separated by `---`.
-- If you want a snippet to insert `---`, use `\---` to escape it.
 - Each snippet document has a context and body separated by `-`.
 - Optionally a file can have a single context at the top with no body. This is not a snippet in itself, but default values to be inherited by the other snippet documents in the same file.
-- Some context keys supports multiple values. These values are separated by `|`.
+- Some context keys support multiple values. These values are separated by `|`.
   - For most keys like `language` or `phrase` multiple values means _or_. You can use phrase _1_ or phrase _2_. The snippet is active in language _A_ or language _B_.
   - For `insertionFormatter` multiple values means that the formatters will be applied in sequence.
+- Snippet bodies without tab characters use leading white space as indentation. In this case, the smallest leading sequence of whitespace characters from a snippet body's lines represents a single level of indentation. You must therefore escape leading spaces to treat them as literal spaces in a snippet body that has no tab characters.
```
nriley:
```suggestion
- Snippet bodies without tab characters use leading space character(s) as indentation. In this case, the shortest sequence of space character(s) at the beginning of one of a snippet body's lines represents a single level of indentation. You must therefore escape leading spaces to treat them as literal spaces in a snippet body that has no tab characters.
```

core/snippets/snippets_parser.py
```diff
@@ -89,6 +92,14 @@ def validate_snippet(document: SnippetDocument, snippet: Snippet) -> bool:
         error(document.file, document.line_doc, "Missing snippet variables")
         return False
 
+    if document.body and UNICODE_ESCAPED_BACKSLASH_PLACEHOLDER in document.body:
+        error(
+            document.file,
+            document.line_doc,
+            f"Snippet body {document.body} has unicode character \\u0000, which is not allowed",
```
nriley:
```suggestion
            f"Snippet body {document.body} has Unicode character \\u0000, which is not allowed",
```

nriley (changes requested):
From the community backlog session, things overall look good. A couple points:

- I agree the escaping/stripping logic should be consolidated.
- Consider displaying errors in snippets with notifications and not just logging them as many users do not tend to leave the Talon log open. @AndreasArvidsson does this sound OK if we were to change the behavior of the [error function](https://github.com/FireChickenProductivity/knausj_talon/blob/9eb9d01fe8c52158ec7c44bcfce0baeadba095ae/core/snippets/snippets_parser.py#L498)?

AndreasArvidsson:
> From the community backlog session, things overall look good. A couple points:
> 
>     * I agree the escaping/stripping logic should be consolidated.
> 
>     * Consider displaying errors in snippets with notifications and not just logging them as many users do not tend to leave the Talon log open. @AndreasArvidsson does this sound OK if we were to change the behavior of the [error function](https://github.com/FireChickenProductivity/knausj_talon/blob/9eb9d01fe8c52158ec7c44bcfce0baeadba095ae/core/snippets/snippets_parser.py#L498)?

Sounds good

apps/vscode/vscode.py
```diff
@@ -444,6 +444,8 @@ def select_next_occurrence(text: str):
         actions.key("esc")
 
     def insert_snippet(body: str):
+        # Avoid triggering VSCode's escaping mechanisms on accident
+        body = body.replace("\\", "\\\\")
```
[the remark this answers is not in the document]
FireChickenProductivity: We should try to handle escaping logic ourselves so that snippets work across editors. I can fix it so that VSCode handles escaped stops properly.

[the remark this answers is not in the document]
FireChickenProductivity: I reverted the change for now and will leave it to a followup.

core/snippets/snippets_parser.py
```diff
@@ -280,6 +296,36 @@ def normalize_snippet_body_tabs(body: str | None) -> str:
     return "\n".join(normalized_lines)
 
 
+def escape_spaces(body: str) -> str:
+    # treat double backslash as escaped backslash
+    # treat backslash space as space
+    return (
+        body.replace("\\\\", UNICODE_ESCAPED_BACKSLASH_PLACEHOLDER)
```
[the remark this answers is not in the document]
FireChickenProductivity: I do not plan on addressing VSCode specific issues in this PR without maintainer feedback.

FireChickenProductivity:
Feedback from the community backlog session: I should address the VSCode issue by having the initial escaping convert the text into a format optimized for the textmate snippet format and then adjust raw snippet  insertion to behave identical to (or at least close to) an editor using textmate snippets.

FireChickenProductivity:
Question for backlog session: how thoroughly should we test this before merge?

AndreasArvidsson:
> Question for backlog session: how thoroughly should we test this before merge?

Snippets are becoming quite a core feature of our language implementation nowadays. So at least we need the test that it's working for all normal use cases.

FireChickenProductivity:
> > Question for backlog session: how thoroughly should we test this before merge?
> 
> Snippets are becoming quite a core feature of our language implementation nowadays. So at least we need the test that it's working for all normal use cases.

I tested the following in Python in VSCode:
"snip if"
"snip if else"
"snip class this is a test"
"snip funk test function"
"if wrap line"
"if cond wrap line"
"snip return"
"snip import from"
"snip list comp"
"snip module setting"

I tested all those except for the wrapping ones with raw snippet insertion. 

I could drive these changes in my personal fork for a week if you do not consider that to be adequate testing.

I did some earlier testing with backslashes but do not remember exactly what I did.

FireChickenProductivity:
I just additionally tested every Community snippet with a backslash in it with both raw text insertion and VSCode.

core/snippets/README.md
```diff
@@ -13,12 +13,14 @@ Custom format to represent snippets.
...
 - Optionally a file can have a single context at the top with no body. This is not a snippet in itself, but default values to be inherited by the other snippet documents in the same file.
-- Some context keys supports multiple values. These values are separated by `|`.
+- Some context keys support multiple values. These values are separated by `|`.
   - For most keys like `language` or `phrase` multiple values means _or_. You can use phrase _1_ or phrase _2_. The snippet is active in language _A_ or language _B_.
   - For `insertionFormatter` multiple values means that the formatters will be applied in sequence.
+- Snippet bodies without tab characters use leading space character(s) as indentation. In this case, the shortest sequence of space character(s) at the beginning of one of a snippet body's lines represents a single level of indentation. You must therefore escape leading spaces to treat them as literal spaces in a snippet body that has no tab characters.
+- You can escape a space, a line consisting entirely of `---`, or a backslash by putting a backslash before it.
+- A snippet body cannot have the unicode character `"\u0000"`.
```
nriley:
```suggestion
- A snippet body cannot have the Unicode character `"\u0000"`.
```

nriley (dismissed):

AndreasArvidsson (approved):

PR #2224, opened 2026-06-15
Add aspnetcorerazor code language

Mark-Phillipson:
This programming language needs to be in place for the talon_hud code language icon to work in the status bar for *.razor/*.cshtml files.

Note this type of file can have C# Razor, Javascript, html and css all in the same file.


FireChickenProductivity:
The following comment is above the language list.
"Maps code language identifiers, names and file extensions. _Only put languages
here which have a supported language mode; that's why there are so many
commented out entries._"

Do believe that aspnetcorerazor files require a separate language implementation? What features would you want that to have? Another option would be activating the most dominant language in those files by default and having some functionality from the other language implementations available through a tag. Is there a reason that approach would be inadequate?

Unfortunately no maintainers made it to this week's backlog session. Based on my experience attending most backlog sessions, the maintainers will probably ask for more context before deciding if they want to merge this, so answering my questions might help them make a decision faster.

nriley:
From the community backlog session — we agree with @FireChickenProductivity and the comment in code_languages.py. If you plan to do more with this language mode, then please add it to the PR. Thanks!

PR #2231, opened 2026-06-28
Remove inappropriate/unneeded type annotations identified by Talon 1389.

nriley:
Many action parameters are `str = None`, but `None` is not a `str`. Talon 1389 cares about this; for backwards compatibility with Talon 0.4 we use `Optional[str]` rather than `str | None`.

Action implementations (as opposed to declarations) do not require type annotations; they can't differ from the declarations and Talon doesn't type check them anyway, so remove them.


nriley:
Since it initially confused me, and for future reference, you *do* still need to declare default arguments in every action implementation.

auscompgeek (approved):

PR #2242, opened 2026-07-18
never paste into the terminal

chdoc:
Most terminal emulators on Linux have a safeguard that will check pasted text and require user interaction if that pasted text appears to execute code. Disabling the `paste_to_insert_threshold`, even if it is enabled in the global configuration, sidesteps this issue.

### DIFF
--- a/settings.talon
+++ b/settings.talon
@@ -86,6 +86,7 @@ settings():
 
     # Uncomment to insert text longer than 10 characters (customizable) by pasting from
     # the clipboard. This is often faster than typing.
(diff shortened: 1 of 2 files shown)


FireChickenProductivity:
This makes sense to me. I am concerned that this will confuse some people who do want insert to paste in some terminal applications. What do you think of having a note in settings.talon? To avoid duplicating knowledge, it could just tell users that this setting is set by default in some applications, so that way they would know to search where else the setting is set if they encountered unexpected behavior.

AndreasArvidsson (approved):

Issue #2234, opened 2026-06-30
Support declaring PHP functions using "funky"


ZachWatkins:
The PHP language files should support declaring a function like `function processData()` using the `funky` phrase. Possible solutions:

1. Use a code function modifier like `global funky process data` to print `function processData()`.
2. Change default behavior for the `funky process data` phrase in PHP to match JavaScript's default behavior for the same phrase, which prints `function processData()`.

PHP is most commonly used in an object-oriented way, so it may be more useful for the default behavior for `funky process data` to remain `public function processData()`. PHP developers may write something like `public function processData()` more often day-to-day than `function processData()`.

One downside to keeping this current default behavior is that it is not the default behavior for JavaScript, so users may not be expecting different default behavior between two languages. Another reason to go with option 2 is that there is already support for the `public` modifier using `pub funky`.

I am open to suggestions. I've been writing PHP for 13 years and I think Option 2 is the more intuitive choice, so I will submit a pull requested for this issue next month that implements Option 2.

nriley:
From the community backlog session — we agree that it makes sense to implement `code_default_function` such that it does not insert a modifier before the `function` keyword. This is also consistent with how the `functionDeclaration` snippet works.

ZachWatkins:
Closing issue since PR was merged. Thank you!

PR #2252, opened 2026-07-28
Remove modifier before PHP's default function command

ZachWatkins:
This pull request addresses issue #2234.


ZachWatkins:
I'm not sure why the pre-commit checks "Format Talon files" and "Format snippet files" failed. Does someone have suggestions for changes I can make? The new lines are practically identical to the other similar functions declared in this file.

nriley (approved):
From the community backlog session, we do have a long-term plan to migrate these to snippets, but as none of the other function insertion for PHP is defined with snippets, this makes sense for consistency's sake.

Issue #2233, opened 2026-06-30
Add common functions for PHP


ZachWatkins:
The PHP language files should support common built-in functions like `fopen` and `file_exists`. Commonly used functions which deal with strings, arrays, and files should be included, and others as well when it makes sense. Here is a list of all functions (also class methods) included in PHP's online documentation: https://www.php.net/manual/en/indexes.functions.php. There are too many to support all of them.

Many functions are only available when certain PHP extensions are installed and activated in the runtime, so they are not as high of a priority as the standard built-in functions.

I will open a pull request for this issue some time next month.

Edit: I am working on this now (August 1st) and will reference a list of the top 100 most used PHP functions according to an analysis performed by Exakat [The Top 100 PHP functions in 2024](https://www.exakat.io/the-top-100-php-functions-in-2024/).

ZachWatkins:
Closing my own issue since the PR was merged. Thank you!

PR #2260, opened 2026-08-01
Add common functions for PHP

ZachWatkins:
This addresses issue #2233. ~~I am unable to invoke these functions at this time using "funk <user.code_common_function>" because Talon keeps recognizing what I say as "funky" instead of "funk".~~ Thanks to @FireChickenProductivity I've updated the PR to resolve those issues.

I've spoken all of the phrases to produce these functions using the Conformer D speech model and believe the recognition is sufficient. As I use them in my work, I will see if they are reliable enough to resolve as expected during more frequent use.

I will submit a separate issue to add more keywords and types for PHP.


lang/php/code_common_function.talon-list
```diff
@@ -0,0 +1,161 @@
...
+sprint F: sprintf
+str ends with: str_ends_with
+str r pos: strrpos
+str to time: strtotime
+stripos
+strlen
+strpad
+strpos
+strrepeat
+strreplace
```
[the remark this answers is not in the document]
ZachWatkins: Yes this is true, this was a mistake and will fix it soon.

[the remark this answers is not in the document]
ZachWatkins: I've updated these functions.

AndreasArvidsson (changes requested):
Feedback from backlog session.

* Split separate spoken words
* `str` we already use `stir` in other places
* Where we have a separate letter to be spelled just use that letter as is and in upper case for readability.

eg:
`strtolower` -> `stir to lower: strtolower`
`you` -> `U`

lang/php/code_common_function.talon-list
```diff
@@ -0,0 +1,161 @@
...
+time
+trim
+ucwords
+unlink
+url decode: urldecode
+urlencode: urlencode
+var dump: var_dump
+var export: var_export
+You A sort: uasort
+You C first: ucfirst
```
AndreasArvidsson:
```suggestion
U C first: ucfirst
```

lang/php/code_common_function.talon-list
```diff
@@ -0,0 +1,161 @@
...
+You sort: usort
```
AndreasArvidsson:
```suggestion
U sort: usort
```

lang/php/code_common_function.talon-list
```diff
@@ -0,0 +1,161 @@
...
+tempnam
+time
+trim
+ucwords
+unlink
+url decode: urldecode
+urlencode: urlencode
+var dump: var_dump
+var export: var_export
+You A sort: uasort
```
AndreasArvidsson:
```suggestion
U A sort: uasort
```

lang/php/code_common_function.talon-list
```diff
@@ -0,0 +1,161 @@
...
+stir tr: strtr
+stir val: strval
+sub stir: substr
+tempnam
+time
+trim
+ucwords
+unlink
+url decode: urldecode
+urlencode: urlencode
```
AndreasArvidsson:
```suggestion
url encode: urlencode
```

AndreasArvidsson (approved):

Issue #2262, opened 2026-08-02
Add keywords and more standard types for PHP


ZachWatkins:
The PHP language includes [keywords](https://www.php.net/manual/en/reserved.keywords.php) that I would like to add to the language files. Examples include:

- and
- as
- break
- case
- catch
- clone
- const
- finally
- fn
- function
- global
- in
- include
- include_once
- instanceof
- match
- require
- require_once
- return
- try
- yield

It also includes standard [types](https://www.php.net/manual/en/language.types.type-system.php) not present in the current PHP type file that I will add:

- callable
- iterable
- mixed
- never
- object
- parent
- resource
- self
- static

ZachWatkins:
Closing this issue since my pull requests were merged. Thanks!

PR #2263, opened 2026-08-02
Add more standard types for PHP

ZachWatkins:
This pull request addresses issue #2262.


lang/php/php.talon
```diff
@@ -24,7 +25,16 @@ settings():
 
 state try: user.insert_snippet_by_name("tryStatement")
 state catch: user.insert_snippet_by_name("catchStatement")
+type union [<user.code_type>]: "|{code_type}"
```
FireChickenProductivity: Is this not the same amount of words as saying "pipe type <user.code_type>"? Is having an extra command for this worth it?

lang/php/php.talon
```diff
@@ -24,7 +25,16 @@ settings():
...
 
 var <phrase> [over]:
     insert("$")
     insert(user.formatted_text(phrase, "PRIVATE_CAMEL_CASE"))
+
+returns class <phrase>:
```
FireChickenProductivity: I recommend using <user.text> instead of `<phrase>`, which is what Community uses for similar things elsewhere.

lang/php/code_keyword_unprefixed.talon-list
```diff
@@ -0,0 +1,6 @@
+list: user.code_keyword_unprefixed
+code.language: php
+-
+
+continue: "continue"
```
FireChickenProductivity: I do not think the right hand side is needed here.

lang/php/php.talon
ZachWatkins: That seems fine, I will remove it.

lang/php/code_keyword_unprefixed.talon-list
ZachWatkins: FYI it's written this way in the lua and python `code_keyword_unprefixed.talon-list` files.

lang/php/php.py
```diff
@@ -153,5 +153,8 @@ def code_public_static_function(text: str):
         actions.user.paste(result)
         actions.edit.left()
 
+    def code_insert_type_annotation(type: str):
```
FireChickenProductivity: We do not usually implement this for languages where the type goes before the variable. `is type string` is no shorter than `type string space`

lang/php/php.py
```diff
@@ -153,5 +153,8 @@ def code_public_static_function(text: str):
...
+        actions.insert(f"{type} ")
+
```
AndreasArvidsson:
```suggestion
```

lang/php/php.py
AndreasArvidsson: We discussed this at a backlog session and we are in agreement that `"type string space"` is preferred.

AndreasArvidsson (approved):

PR #2268, opened 2026-08-05
Add keywords for PHP

ZachWatkins:
This pull request addresses a task in issue #2262.


lang/php/code_keyword.talon-list
```diff
@@ -0,0 +1,61 @@
...
+and: " and "
+as: " as "
+break: "break"
+case: "case "
+catch: " catch "
+class: "class "
+clone: "clone "
+const: "const "
+continue: "continue"
+default: "default:"
```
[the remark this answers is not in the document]
FireChickenProductivity: We only use the keyword support for keywords and spacing. I recommend submitting a separate command for "default:" if that would be useful in PHP.

lang/php/code_keyword_unprefixed.talon-list
```diff
@@ -0,0 +1,6 @@
+list: user.code_keyword_unprefixed
+code.language: php
+-
+
+continue: "continue"
```
FireChickenProductivity:
```suggestion
continue
```

lang/php/code_keyword.talon-list
FireChickenProductivity:
```suggestion
default
```

AndreasArvidsson (approved):

Issue #2249, opened 2026-07-25
Make symbols easier to customize


FireChickenProductivity:
I think symbols should be customizable outside of python at least through a configuration file.

AndreasArvidsson:
Feedback from community session. The best solution is probably to replace the python list with a csv file. Something like this:
```csv
symbol, mode, spoken forms
?, both, question mark, question symbol
?, command, question
```

FireChickenProductivity:
I am starting on this.

PR #2269, opened 2026-08-06
Support customizing symbols from csv

FireChickenProductivity:
This will address #2249


core/keys/symbols.py
```diff
@@ -92,15 +94,62 @@ def __init__(
...
+    # for dragon, we add a couple of mappings that don't work for conformer
+    # i.e. dragon supports some actual symbols as the spoken form
+    dragon_punctuation_dict = {
+        "`": "`",
+        ",": ",",
+    }
+    for i, row in enumerate(values):
+        # tolerate a blank line
+        if (len(row) == 1 and (not row[0] or row[0].isspace())) \
+             or len(row) == 0:
```
AndreasArvidsson: I think we can simplify this a bit.

```suggestion
        if (len(row) == 0 or (len(row) == 1 and row[0].isspace()):

**Review by AndreasArvidsson — APPROVED** [review]

File: core/keys/symbols.py [hunk 2]
```diff
@@ -92,15 +94,61 @@ def __init__(
...
+    punctuation_dict.clear()
+    # for dragon, we add a couple of mappings that don't work for conformer
+    # i.e. dragon supports some actual symbols as the spoken form
+    dragon_punctuation_dict = {
+        "`": "`",
+        ",": ",",
+    }
+    for i, row in enumerate(values):
+        # tolerate a blank line
+        if (len(row) == 0 or (len(row) == 1 and row[0].isspace()):
```
AndreasArvidsson:
```suggestion
        if len(row) == 0 or (len(row) == 1 and row[0].isspace()):
```

Issue #2266, opened 2026-08-05
Add return type command for class names to language features


ZachWatkins:
It was suggested in pull request #2263 that I should see if the maintainers want a return type command I made for PHP to be available in the common language files (likely after modifications).

(Edit: Since this issue didn't receive a comment after the last maintainers meeting, I will assume it would be easier to discuss this if I submitted a PR. I will submit 2 separate PRs: (1) for the "returns class" command to be added to `./lang/tags/functions.talon`, and (2) for the "returns global class" command to be added to `./lang/tags/functions.talon` (since "returns global class" may not be necessary for other languages (see below)). I am hoping this makes it easier to discuss and review the different solutions that would add "returns class <user.text>" to the project. If one or both PRs are declined, then I will submit a third to add the declined change(s) to the PHP language file.)

Here is the code I submitted with recommended corrections:

```
returns class <user.text>
    formatted = user.formatted_text(text, "PUBLIC_CAMEL_CASE")
    insert(": {formatted}")

returns global class <user.text>:
    formatted = user.formatted_text(text, "PUBLIC_CAMEL_CASE")
    insert(": \\{formatted}")
```

~~The code above is fine for PHP, but aren't there languages which do not require camel case for their class names? If so, additional language-specific logic is required and possibly as a user setting as well.~~ Edit: Since ./core/snippets/snippets/classDeclaration.snippet uses PUBLIC_CAMEL_CASE as the only insertion formatter, I will assume it is fine to use for these commands.

Edit:

About the "returns global class" command, other languages may not need this command. PHP uses [namespaces](https://www.php.net/manual/en/language.namespaces.definition.php) for class, interface, function, and constant definitions. Namespaces use the `\` character to declare subnamespaces, and a global namespace reference can be made by prefixing it with `\`. When a namespace is declared, namespace resolution first checks whether proceeding classes and functions are registered in that namespace. Prefixing them with "\" explicitly declares they use the global space. Simple example:

```php
<?php
namespace App;
function currentDate(): \DateTime {
    return new \DateTime();
}
```

PR #2271, opened 2026-08-09
Add returns class command to common language features

ZachWatkins:
This pull request is related to issue #2266. It implements the "returns class <user.text>" command in `./lang/tags/functions.talon`.

Since `./core/snippets/snippets/classDeclaration.snippet` uses PUBLIC_CAMEL_CASE as the only insertion formatter, I will assume it is fine to use for this new command.

I will submit a separate PR for the "returns global class <user.text>" command, since I am not sure whether it would be useful for languages other than PHP.


AndreasArvidsson (changes requested):
See above

lang/tags/object_oriented.py
```diff
@@ -10,6 +10,28 @@
 
 mod.list("code_common_method", desc="Commonly invoked method, e.g. 'foo' in '.foo()'")
 
+mod.setting(
```
AndreasArvidsson: I know that all of these settings are probably in the wrong file but I would prefer to have them at the same place

https://github.com/talonhub/community/blob/8fcef5449642eab6392882a28a4e2bbf5f4d73a8/lang/tags/functions.py#L39

AndreasArvidsson (changes requested):

lang/tags/object_oriented.py
```diff
@@ -10,6 +10,28 @@
...
+    "code_class_formatter",
+    type=str,
+    default="PUBLIC_CAMEL_CASE",
+    desc="Formatter used when a class/type name is dictated as free text (e.g. 'returns' followed by a spoken class name)",
+)
+
+ctx.matches = r"""
+tag: user.code_object_oriented
+"""
```
AndreasArvidsson: Please move this further up right after the context.

AndreasArvidsson (dismissed):

nriley (approved):

PR #2277, opened 2026-08-12
Fix help search ignoring help_max_command_lines_per_page setting 

ashbostock:
"help search \<phrase\>" results could overflow a single page when many matching commands came from one context file. The pagination code treated each context's entire result group as an indivisible unit, so groups larger than the page limit were placed on an oversized page unchanged.

Fix: flatten grouped search results into individual commands before paginating, matching the approach already used by the single-context help view. Context headers are accounted for by adding their 3-line overhead to each group's first command.

### DIFF
--- a/core/help/help.py
+++ b/core/help/help.py
@@ -386,22 +386,30 @@ def draw_search_commands(gui: imgui.GUI):
         key=lambda item: context_map[item[0]] not in cached_active_contexts,
     )
 
-    pages = get_pages(
-        [
-            sum(get_command_line_count(command) for command in commands) + 3
-            for _, commands in sorted_commands_grouped
-        ]
-    )
+    # flatten to individual commands so get_pages applies the limit per-command, not per-context group
+    flat_commands = []
+    item_line_counts = []
+    for context, commands in sorted_commands_grouped:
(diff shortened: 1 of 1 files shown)


core/help/help.py
```diff
@@ -386,22 +386,32 @@ def draw_search_commands(gui: imgui.GUI):
...
+            continue
+        if context != current_context:
+            if current_context is not None:
+                gui.spacer()
             gui.text(format_context_title(context))
             gui.line()
-            draw_commands(gui, commands)
-            gui.spacer()
+            current_context = context
+        draw_commands(gui, [command])
```
AndreasArvidsson: Passing a list of 1 is a bit weird. Please convert `draw_commands` into `draw_command` instead.

AndreasArvidsson (changes requested):

core/help/help.py
ashbostock: Done

FireChickenProductivity (approved):
I tested this on the most recent beta and public. It seems to work. I confirmed the presence of the original issue.

AndreasArvidsson (approved):

PR #2278, opened 2026-08-13
Enabled additional ruff rules

AndreasArvidsson:


pyproject.toml
AndreasArvidsson: Can we update this? What python version is Talon public using?

nriley: 3.11.4. Good catch!

apps/emacs/emacs_commands.py
```diff
@@ -44,7 +44,7 @@ def load_commands(f):
         if len(row) > 4:
             print(
                 f"emacs_commands.csv: More than four values in row: {row}. "
-                + " Ignoring the extras"
+                " Ignoring the extras"
```
auscompgeek: Do we actually want to enforce explicit string concat instead? Implicit string concat can be a footgun if you accidentally forget a comma.

pyproject.toml
AndreasArvidsson: Thank you

apps/emacs/emacs_commands.py
AndreasArvidsson: Hmm. Which comma are you referring to?

auscompgeek: Imagine something like this:

```python
argv = [
    "foo"   # oops, is this meant to be a separate string in this list or joined?
    "bar",
    "baz",
]
```

The autoformatter might make it more obvious but for longer strings they may not get rewrapped.

AndreasArvidsson: I completely agree. I have now performed this change.

nriley (approved):

PR #2280, opened 2026-08-15
Replace `self.` with `user.`

AndreasArvidsson:
Also did some general cleanup of capture rules


apps/windbg/windbg.py
```diff
@@ -12,14 +12,14 @@
 tag: user.windbg
 """
 
-ctx_windbg_enabled.lists["self.windows_dlls"] = {
+ctx_windbg_enabled.lists["user.windows_dlls"] = {
```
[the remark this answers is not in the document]
AndreasArvidsson: For Talon lists self and user behaves the same, but user is preferred.

core/contacts/contacts.py
```diff
@@ -275,19 +275,17 @@ def prose_contact_snippet(m) -> str:
...
-        "| <user.prose_username> "
-        "| <user.prose_full_name> "
-        "| <user.prose_full_name_possessive> "
-        "| <user.prose_first_name> "
-        "| <user.prose_first_name_possessive> "
-        "| <user.prose_last_name>"
-        "| <user.prose_last_name_possessive>"
-        "| <user.prose_contact_snippet>"
-    ),
+    rule="<user.prose_name>"
```
nriley: I don't see the benefit in doing this — just seems like additional clutter. Is there some style guideline or pending deprecation of which I'm unaware?

AndreasArvidsson: That was intentional: https://github.com/talonhub/community/pull/2278#discussion_r3776603626

But exact same change was already merged in the Ruff pr so it's not relevant for here anymore.

nriley: Yeah I saw that when I was resolving conflicts. Makes sense I guess though I still think it's ugly.

nriley (approved):

core/contacts/contacts.py
AndreasArvidsson: I agree. I have some ideas on how this can be made to look prettier and still avoid the foot gun situation.

Issue #2286, opened 2026-08-16
Add PHP entries to core snippets where missing


ZachWatkins:
16 of the core snippet files in core/snippets/snippets/ are missing entries for the PHP language.

Assuming I should not submit 16 PRs, I will submit one PR for changes to all 16 files.

ZachWatkins:
Closing my own issue since the PR was merged. Thank you!

PR #2287, opened 2026-08-16
Add missing PHP entries to core snippet files

ZachWatkins:
This addresses issue #2286.

I originally meant for this to update one snippet, but then I discovered that 16 files were missing PHP entries where either an existing snippet could apply to PHP or a new entry could be added to an existing snippet file for PHP.

I'm assuming it is not preferred to submit 16 PRs covering each file change individually. If separate PRs are preferred, please let me know and I will do so.


core/snippets/snippets/forEachMutableStatement.snippet
```diff
@@ -23,3 +23,10 @@ for ($1 of $2) {
     $0
 }
 ---
+
+language: php
+-
+foreach ($1 as &$2) {
+    $0
+}
+---
```
chdoc: I think this unnecessary, because it is identical to the definition in `forEachStatement.snipper`. This file is only to offer an alternative if the default introduces an immutable variable.
```suggestion
```

chdoc (changes requested):

core/snippets/snippets/forEachMutableStatement.snippet
```diff
@@ -23,3 +23,4 @@ for ($1 of $2) {
     $0
 }
 ---
+
```
AndreasArvidsson, line 26:
```suggestion
```

AndreasArvidsson (approved):

PR #2290, opened 2026-08-16
Fix tmux split orientation commands

davitenio:
Tmux uses the -h flag for a left/right split, which applications such as VS Code and Obsidian call a vertical split. That meant issuing the voice commands "split horizontal" and "split vertical" would create the opposite of what a user coming from those other apps would expect. This commit aligns the tmux split voice commands with the orientation conventions used by these other applications: "split horizontal" will create a horizontal split (one pane on top of the other) and "split vertical" will create a vertical split (one pane to the left of the other).

### DIFF
--- a/BREAKING_CHANGES.txt
+++ b/BREAKING_CHANGES.txt
@@ -7,7 +7,7 @@ and when the change was applied given the delay between changes being
 submitted and the time they were reviewed and merged.
 
 ---
-* 2026-08-10 Change tmux split commands to match the orientation conventions we use for other applications such as VS Code and Obsidian.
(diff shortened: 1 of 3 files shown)


BREAKING_CHANGES.txt
```diff
@@ -7,6 +7,7 @@ and when the change was applied given the delay between changes being
 submitted and the time they were reviewed and merged.
 
 ---
+* 2026-08-10 Change tmux split commands to match the orientation conventions we use for other applications such as VS Code and Obsidian.
```
AndreasArvidsson:
```suggestion
* 2026-08-10 Change tmux split commands to match the orientation conventions we use for other applications such as VS Code and Obsidian. (PR #2290)
```

AndreasArvidsson (approved):

PR #2293, opened 2026-08-22
Fix tmux shell commands being unavailable outside a running tmux session

davitenio:
Move tmux shell commands into tags/terminal/unix_utilities.talon so they are available outside running tmux sessions and can be imported using the user.unix_utilities tag. Keep tmux keybinding commands in the file specific to running tmux sessions.


nriley (changes requested):
Please move the contents of `tmux_shell_commands.talon` into `tags/terminal/unix_utilities.talon`. Thanks!

nriley (approved):
Apologies, you did what I asked but I did not write what was in my head, which was to add the commands to the .talon-list file! I made the changes myself.

PR #2301, opened 2026-08-23
Remove action command

FireChickenProductivity:
Given that the command no longer does anything and has been deprecated for a over year, I think we should just remove it.

### DIFF
--- a/BREAKING_CHANGES.txt
+++ b/BREAKING_CHANGES.txt
@@ -7,6 +7,7 @@ and when the change was applied given the delay between changes being
 submitted and the time they were reviewed and merged.
 
 ---
+* 2026-08-23 Removed deprecated Jetbrains command `action` in favour of the `please` command. 
 * 2026-08-10 Change tmux split commands to match the orientation conventions we use for other applications such as VS Code and Obsidian. (PR #2290)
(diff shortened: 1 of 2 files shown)


nriley (approved):

BREAKING_CHANGES.txt
```diff
@@ -7,6 +7,7 @@ and when the change was applied given the delay between changes being
 submitted and the time they were reviewed and merged.
 
 ---
+* 2026-08-23 Removed deprecated Jetbrains command `action` in favour of the `please` command.
```
auscompgeek, line 10: Given the command doesn't do anything I don't think this needs a changelog entry.
```suggestion
```

FireChickenProductivity: I did not think it was needed either, but I chose to stick strictly to
```
1. For 6 months from deprecation a deprecated action or command should
   continue working. Put an entry in the BREAKING_CHANGES.txt file in the
   project root to mark the deprecation and potentially explain how users can
   migrate. Use the user.deprecate_command, user.deprecate_action, or
   user.deprecate_capture actions to notify users.
2. After 6 months you can delete the deprecated command, action, or capture.
   Leave the note in BREAKING_CHANGES.txt so people who missed the
   notifications can see what happened.
```
from deprecations.py for sake of consistency. There was no original note from the original deprecation, so I decided I might as well add one.

Issue #2190, opened 2026-05-09
Investigate removing pywin32 dependencies


nriley:
Beta/experimental Talon makes use of a free-threading/no-GIL Python interpreter (3.14t). As of build 1409, pywin32 is removed from Talon.

Currently, pywin32 is not compatible with free-threading, though there is [work in progress](https://github.com/mhammond/pywin32/pull/2767) to remedy this which is unlikely to be ready for Talon 0.5.

We should replace community's uses of pywin32 with ctypes or other APIs for interfacing with Windows APIs, or potentially (optionally) disable the corresponding features. Core Talon has also been adding actions to replace this usage.

Current status:

- [ ] `core/app_switcher/app_switcher.py`: if import fails, use new `apps.list` action (see https://github.com/nriley/talon_community/commit/cf99cc40db0e2c6990b767c5c057c30e689a3d7d and #2304). Currently `apps.running` doesn't suppress background apps, but it should soon, so we can wait to replace community's implementation.
- [ ] `plugin/mouse/mouse_cursor.py`: use new `ctrl.cursor_visible` action. This action exists but doesn't work as of this writing.
- [ ] `apps/wsl/wsl.py`: Try migrating to the `winreg` module for Windows registry access, which we use elsewhere. I don't have a Windows installation capable of running WSL, so would appreciate some help here. In the meantime we may consider just disabling WSL support. This was originally contributed by @codecat555 (#632).

clin1234:
Hi. Author of the free-threaded support for pywin32 here. https://github.com/mhammond/pywin32/pull/2767 has just been merged in, though consider it experimental.

nriley:
@clin1234 Great start, we'll be following your progress on addressing the no-GIL issues and hope it will let us be able to continue to use pywin32 from Talon!

nriley:
As of build 1409, pywin32 is removed from Talon.

Current status:

- [ ] `core/app_switcher/app_switcher.py`: if import fails, use new `apps.list` action (see https://github.com/nriley/talon_community/commit/cf99cc40db0e2c6990b767c5c057c30e689a3d7d). Currently `apps.running` doesn't suppress background apps, but it should soon (and our current app switching code doesn't rely on pywin32). We can decide whether we want to add some additional filtering on top of it.
- [ ] `plugin/mouse/mouse_cursor.py`: use new `ctrl.cursor_visible` action. This action exists but doesn't work as of this writing.
[...]

PR #2304, opened 2026-08-25
Fix windows app switcher on experimental beta branch

FireChickenProductivity:
The diff looks awkward but it just puts the existing windows implementation in a try/except block on ImportError with the exception branch using actions.apps.list() to get the apps.
I tested on windows 10 with public and the latest beta.


nriley (changes requested):
From the community backlog session — please add a comment and update #2190.

Thanks!

nriley (approved):

PR #2309, opened 2026-09-01
Add voice command for inserting markdown table headers

FireChickenProductivity:
This inserts the table header for a markdown table with the given number of columns. It builds a temporary snippet to insert to take advantage of `snip next`. Note that we could not put this in a .snippet file because the number of placeholders depends on the number of columns.


lang/markdown/markdown.py
```diff
@@ -36,3 +36,21 @@ def markdown_wrap_selection_with_link(url: str):
         """Wraps the selection with a markdown link for the specified url"""
         description = actions.edit.selected_text()
         actions.user.markdown_insert_link(url, description)
+
+    def markdown_insert_table_header(num_columns: int):
+        """Creates a table header with the given number of columns"""
+        if num_columns <= 0:
+            return
```
AndreasArvidsson:
```suggestion
            raise ValueError("Can't insert table header with 0 columns.")
```

AndreasArvidsson (approved):

PR #2311, opened 2026-09-02
Added documentation for the draft editor

AndreasArvidsson:


nriley (approved):
Note that there is an unrelated rename in this PR; we had discussed/approved it at a prior community backlog session so I am not bothering to remove it from the PR before merging it.

PR #2312, opened 2026-09-03
fix: pluralize .talon-list filenames that collide with same-named .talon file

wesleyboar:
## Overview

Fixes `user.emoji` and `user.top_level_domain` registering with zero entries, which made their commands unmatchable.

## Changes

- **renamed** each `.talon-list` file whose basename collided with a same-named `.talon` file
    - `emoji.talon-list`
        **→ `emojis.talon-list`**
    - `top_level_domain.talon-list`
        **→ `top_level_domains.talon-list`**

### Top-Level Domains

1. On `main`, run:
    ```
    echo 'from talon import registry; print(len(dict(registry.lists["user.top_level_domain"][0])))' | ~/.talon/bin/repl
    ```
2. **Verify it prints `0`.**
3. Say "dot com".
4. **Verify nothing is typed.**
5. Check out this branch.
6. Restart Talon.
7. Repeat step 1.
8. **Verify it now prints `8`.**
9. Say "dot com".
10. **Verify `.com` is typed.**

### Emojis

1. On `main`, run:
    ```
    echo 'from talon import registry; print(len(dict(registry.lists["user.emoji"][0])))' | ~/.talon/bin/repl
    ```
2. **Verify it prints `0`.**
3. Focus Slack or Discord.
4. Say "emoji angry".
5. **Verify nothing is pasted.**
6. Check out this branch.
7. Restart Talon.
8. Repeat step 1.
9. **Verify it now prints `28`.**
10. Focus Slack or Discord.
11. Say "emoji angry".
12. **Verify 😠 is pasted.**

## Notes

- Talon skips a `.talon-list` whose basename matches a `.talon` file in the same directory, logging `DEBUG [ ]` instead of `DEBUG [+]` — no warning or error.
- These were the only two collisions that AI found among the repo's 96 `.talon-list` files.


wesleyboar (commented):
> [!NOTE]
> **`tags/emoji/insert_emoji.talon`**
> Verifying this half needs Slack or Discord focus: those are the only two contexts that enable `tag: user.emoji`, which gates these rules. `emoticon` and `kaomoji` were never affected, since their list files have no same-stem `.talon` shadowing them.

> [!NOTE]
> **`core/top_level_domain/insert_top_level_domain.talon`**
> The `insert_` prefix follows the existing `core/vocabulary/edit_vocabulary.talon` precedent for naming a command file differently from its sibling list files. Renaming the `.talon-list` instead would fix the shadowing equally well, but would break the convention that a list file is named after the list it defines.

AndreasArvidsson (changes requested):
Feedback from the community session. We would prefer if you pluralize the list file name. eg `emojis.talon-list`.  Don't rename the list itself just the file name.

AndreasArvidsson (approved):

PR #2315, opened 2026-09-10
Improve file manager pagination

ashbostock:
Previously, total_folder_pages was calculated from ctx.lists["user.file_manager_directories"], whose length counts every spoken-form entry. A single directory can have multiple spoken forms, so this could overestimate the number of visible pages.

This change instead counts folder_selections, which is the deduplicated, sorted list actually displayed in the picker.

The change also uncomments the gui buttons which show how to navigate the file manager. As a first time user of the file/folder manager it wasn't obvious how to switch pages. So I think having these enabled by default makes things more consistent with how the "help active" gui also works.


FireChickenProductivity (approved):

tags/file_manager/file_manager.py
```diff
@@ -320,14 +317,16 @@ def gui_files(gui: imgui.GUI):
...
+        if gui.button("File next"):
+            actions.user.file_manager_next_file_page()
+        if gui.button("File last"):
+            actions.user.file_manager_previous_file_page()
 
-    #   if gui.button("Previous..."):
-    #        actions.user.file_manager_previous_file_page()
+    gui.spacer()
+    if gui.button("Manager close"):
+        actions.user.file_manager_hide_pickers()
```
nriley:
```suggestion
```

nriley (approved):

Issue #1721, opened 2025-02-01
Convert `Optional[X]` to `X | None` in action definitions (after v0.5)


phillco:
* `|` isn't supported in action type definitions in v0.4, only 0.5 beta: https://github.com/talonvoice/talon/issues/634
  * fix will be shipped in 0.5 release

PR #2316, opened 2026-09-10
fix incompatible type signature in readline_vi

BlueDrink9:
Turns out the 3.10 deprecation isn't compatible with the old signature now? I'm not sure what happened, there must have been a change to the signature of the defining action because this was working for me previously.

### DIFF
--- a/tags/terminal/readline_vi.py
+++ b/tags/terminal/readline_vi.py
@@ -1,5 +1,6 @@
 from collections.abc import Callable
 from dataclasses import dataclass
+from typing import Optional
 
 from talon import Context, actions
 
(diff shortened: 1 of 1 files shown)


auscompgeek (approved):
Yep, see #1721.

tags/terminal/readline_vi.py
auscompgeek, line 124: Wait, huh, apparently this _should_ work here because this isn't the action declarations.

BlueDrink9: why should it work? The action declaration uses a different signature, which means this implementation of the action is not a compatible signature, causing a talon error

nriley (changes requested):
From the community backlog session, please remove the type declarations from action implementations (`@ctx.action_class`) and fix the type declaration on the action definition (`@mod.action_class`).

nriley (approved):

PR #2317, opened 2026-09-10
[minor] Add antigravity vscode alias

BlueDrink9:


apps/vscode/vscode.py
```diff
@@ -67,6 +69,10 @@
 os: windows
 and app.exe: /^cursor\.exe$/i
 os: windows
+and app.name: Antigravity IDE
+os: windows
+and app.exe: /^antigravity (IDE)?\.exe$/i
```
auscompgeek: This matches `antigravity .exe` with a space. Surely it's not shipped like that?

BlueDrink9: oops thanks, well spotted!

apps/vscode/vscode.py
```diff
@@ -67,6 +69,10 @@
 os: windows
 and app.exe: /^cursor\.exe$/i
 os: windows
+and app.name: Antigravity IDE
+os: windows
+and app.exe: /^antigravity( IDE)?\.exe$/i
```
nriley, line 74: Are there multiple versions of Antigravity that have different .exe filenames? I've rarely seen Windows executables that have a space in their names.

nriley (dismissed):

apps/vscode/vscode.py
BlueDrink9: Couldn't test windows, so was going off what I scraped online. It does seem like old versions had a space

nriley: I don't think it makes sense to support old versions; these kind of apps tend to have auto-updates and it is very rare that people run old versions.

BlueDrink9: In that case, I will need someone else to test this for me. I don't have a Windows machine to find out what the actual executable is, and I'm not sure which of these is the old or different version.

BlueDrink9: Alternatively, I can remove them entirely, but I figure it would be nicer to provide someone with support, even if one of them is redundant.

nriley: On the community backlog session, I installed the current version of the Antigravity IDE and it uses Antigravity.exe. Please just use this version.

BlueDrink9:
```suggestion
and app.exe: antigravity.exe
```

BlueDrink9: done

apps/vscode/vscode.py
```diff
@@ -67,6 +69,10 @@
 os: windows
 and app.exe: /^cursor\.exe$/i
 os: windows
+and app.name: Antigravity IDE
+os: windows
+and app.exe: antigravity.exe
```
nriley:
```suggestion
and app.exe: /^antigravity\.exe$/i
```

nriley: Still needs to be case-insensitive due to compatibility issues between 0.4 and the beta.

nriley (approved):

apps/vscode/vscode.py
BlueDrink9: Sorry, a little bit unclear. Do you still want this change? (Asking because you resolved the comment and approved the rest)

PR #2318, opened 2026-09-10
Fix obsidian tab commands for more consistency with community

BlueDrink9:
Technically a breaking change, but one that brings commands more into line with the rest of community. Obsidian support was only added by me recently, so I didn't feel like this justified a deprecation notice.


apps/obsidian/obsidian.talon
```diff
@@ -46,7 +46,7 @@ file open folder: user.obsidian("open-with-default-app:show")
 file open default app: user.obsidian("open-with-default-app:open")
 file export pdf: user.obsidian("workspace:export-pdf")
 file copy path: user.obsidian("workspace:copy-path")
-file copy full path: user.obsidian("workspace:copy-full-path")
+file copy full [path]: user.obsidian("workspace:copy-full-path")
```
nriley:
```suggestion
file copy full path: user.obsidian("workspace:copy-full-path")
```
This was the unrelated change to which I was referring.

nriley (changes requested):

nriley (approved):

PR #2319, opened 2026-09-11
Permit multiple user snippet directories

BlueDrink9:
This is useful for example if you have both a personal set of snippets, and a project/job/machine specific set

### DIFF
--- a/core/snippets/snippets.py
+++ b/core/snippets/snippets.py
@@ -4,6 +4,7 @@ from typing import Union
 from talon import Context, Module, actions, app, fs, settings
 
 from ..modes.code_languages import code_languages
+from ..user_settings import parse_snippet_dirs
 from .snippet_types import (
     InsertionSnippet,
     Snippet,
@@ -25,7 +26,19 @@ mod.setting(
     "snippets_dir",
     type=str,
     default=None,
-    desc="Directory (relative to Talon user) containing additional snippets",
+    desc="""
+    Directory path(s) containing additional snippets. Can be relative to the Talon user folder or absolute.
+
+    Accepts a single path, or multiple paths separated by pipe (`|`) characters.
+    Spaces inside path names are supported directly and do not need to be escaped.
+    Whitespace around the `|` delimiter is automatically trimmed.
+
+    Examples:
+      - Single relative path: "my_snippets"
+      - Path containing spaces: "my custom snippets/python"
+      - Absolute path: "/Users/name/documents/snippets"
+      - Multiple paths: "my_snippets | /var/snippets | custom snippets/lang"
+    """,
 )
 
 # `_` represents the global context, ie snippets available regardless of language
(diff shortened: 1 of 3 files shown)


nriley (commented):
From the community backlog session, overall, this looks like a reasonable feature to add to community.

Our concerns were around making sure that we have this tested on all three platforms and whether a delimiter of " | " is the best thing to use. We will revisit this in the next few weeks.

chdoc:
> I specifically chose pipe because it's an invalid unix path character, and very rare on windows. I could allow it to be escaped by doubling, if that helps the decision

I'm not entirely sure what you are saying here, because `|` is a perfectly valid path character on both Linux and macOS:
```
$ mkdir "|"
$ ll | grep "|"
drwxr-xr-x 1 chris chris      0 19. Sep 14:39 |
```
That being said, the standard UNIX path separator for environment variables `:` is also a valid path character. As far as I know, the only invalid non-printable character on Linux and macOS is `/` while Windows forbids `< > : " / \ | ? *`.

So I would prefer to use the UNIX path list separator `:` which (unlike Windows own path list separator  `;`) cannot occur in a Windows path. Also, I have never come across a path containing a `:` on Linux, because using them in path lists would be a pain.

nriley:
Python has a constant [os.pathsep](https://docs.python.org/3/library/os.html#os.pathsep) — please use the platform specific separator, thanks!

BlueDrink9:
os.pathsep would make this setting not cross compatible across operating systems, though for people that switch. They would have to duplicate their setting. That's what I avoided that. I'd be fine to pick just `:`. I also haven't implemented escaping it, but could do that (probably by doubling rather than an escape chracter, since `\` is a path seperator).

BlueDrink9:
> > I specifically chose pipe because it's an invalid unix path character, and very rare on windows. I could allow it to be escaped by doubling, if that helps the decision
> 
> I'm not entirely sure what you are saying here, because `|` is a perfectly valid path character on both Linux and macOS:

My bad, I miswrote those. It's invalid on Windows, and I assumed rare on Unix because it has another use,  (would be easy to confuse with a pipe step). In that respect, it's similar to `:`

nriley:
> os.pathsep would make this setting not cross compatible across operating systems, though for people that switch. They would have to duplicate their setting. That's what I avoided that. I'd be fine to pick just `:`. I also haven't implemented escaping it, but could do that (probably by doubling rather than an escape chracter, since `\` is a path seperator).

Yes, but / vs \ and drive letters are already different between operating systems. We'd need it to differ regardless.

BlueDrink9:
> Yes, but / vs \ and drive letters are already different between operating systems. We'd need it to differ regardless.

I have just been looking into this because I wanted to make the change to pathsep. The / path separator is supported in all cases anyway by the PATH parsing. However, lone \ are not, meaning users could not actually put windows parts into this regardless, unless they double up all of their backslashes. (Unlike python files, .talon files don't support r"" raw strings, which would normally be the preferred option).

I think it would be silly to throw out compatibility across multiple OSes, especially for paths within the talon user directory or the user's home directory. Drive letters are only relevant for machine-specific absolute paths, which I imagine would be rarer than sticking the snippet files in either talon home, or somewhere like the user's `~/Documents`

Since windows users would have to modify their path entries in 1 of 2 ways anyway (either double all backslashes, or change them to /), I would prefer to use a single separator across all OSes.

PR #2320, opened 2026-09-12
joplin support via rpc or palette

BlueDrink9:


DonnerWolfBach:
Does the palette here refer to a command palette like vscodes ctrl-shift-p?

BlueDrink9:
yes

BlueDrink9:
Currently in draft because I'm still testing, but any preliminary suggestions are welcome.

I've learned not to sit on PRs for too long, in case someone else comes along wanting to/doing the same thing (oops, tmux)

BlueDrink9:
Tested and ready for feedback

PR #2324, opened 2026-09-19
(Blocked on next Talon release) Address ui deprecation warning

FireChickenProductivity:


nriley:
Can you make this conditional? That's what we've been doing with other things like this that are fixes rather than new features, rather than using the blocked-0.5 tag. (Oops, did not mean to close this!)

Issue #2326, opened 2026-09-19
Migrate skia imports


FireChickenProductivity:
In the next Talon update, skia should be imported directly, `import skia`, instead of importing it from talon.

PR #2329, opened 2026-09-19
(Blocked on next Talon version) Import skia directly for mouse grid

FireChickenProductivity:
Part of #2326


nriley:
Can you make this conditional? That's what we've been doing with other things like this that are fixes rather than new features, rather than using the blocked-0.5 tag.

FireChickenProductivity:
> Can you make this conditional? That's what we've been doing with other things like this that are fixes rather than new features, rather than using the blocked-0.5 tag.

Given that this is just addressing a deprecation warning, I think we can wait for the release on this rather than needing a followup to remove the conditional statement at some point.

Issue #2327, opened 2026-09-19
Avoid importing talon.mac when not on mac


FireChickenProductivity:
Avoid importing talon.mac when not on Mac because on the latest beta this throws an exception. Someone already has a patch for this and is probably going to make the PR soon.

PR #2331, opened 2026-09-19
fix applescript imports

chdoc:
This fixes an issue with the current beta, where the `applescript` import no longer fails silently but with a notification window.


FireChickenProductivity:
closes #2327

nriley (approved):

PR #2332, opened 2026-09-20
mouse_cursor: Replace pywin32 reference used when hiding/showing mouse cursor on Windows.

nriley:
`ctrl.cursor_visible` doesn't work on Windows (yet). The only dependency in the workaround for this deficiency in core Talon is retrieving a constant from the win32con module, so this is easy enough to change.

Resolves one of the items in #2190.


knausj85 (approved):

knausj85:
At the moment, getting an "unable to read response from server" error when attempting to merge....  this is approved and good to go if it goes away.

